PR #2, opened 2024-11-20
Move external agent fixes into neuro_san

d1donlydfink:
A number of changes to external agent code found in the exercising of them in another repo.
This actually makes the external agents work (given the previous PR)!

* If there are no front men found by looking for someone without function parameters,
look instead for the first guy registered. This will allow top-level agents that can be called as functions
to be invoked correctly.
* Langchain needs its tool names to conform to a certain naming convention that our specs for external agents will never conform to. Create a method on ExternalToolAdapter to make such a name.
* Use that safe name in the CallingTool to make a lookup of safe-name (registered with langchain) -> external tool name (with enough information to call out to an external server)
* When looking for responses from the external tool, look for the last thing the assistant said as the actual result.
* Repair a check in LangChainRunContext so external tools can be discovered.
* Shut up the outgoing grpc_client_retry logging that spams the logs.

Testing: I tested hello_world and esp_decision_assistant and tested these changes in the unileaf repo with a new unileaf-specific agent.  All is well.

Worth noting that pylint, flake8, and unit tests are all passing despite what the build message is saying (Donn is bringing that up).

### DIFF
--- a/neuro_san/graph/tools/external_tool.py
+++ b/neuro_san/graph/tools/external_tool.py
@@ -102,6 +102,12 @@ class ExternalTool(CallableTool):
         # Parse the logs response for the last thing the assistant said
         response_split: List[str] = response.split("assistant: ")
         content: str = response_split[-1]
+        if len(response_split) < 2:
+            # The assistant marker not being in the logs is a very uncommon situation,
+            # but still it would be nice to have some idea of the frequency if it
+            # actually does arise.  It's not really an error as far as we know.
+            # Let the upstream agent figure it out for now until we know what to do.
+            print(f"Response from {self.agent_url} contains no indentifying marker")
 
         # Prepare the output
         message: Dict[str, Any] = {


neuro_san/graph/registry/agent_tool_registry.py
```diff
@@ -203,6 +207,10 @@ def find_front_man(self) -> str:
             if function.get("parameters") is None:
                 front_men.append(name)
 
+        if len(front_men) == 0:
+            # The next way to find a front man is to see which agent was registered first
+            front_men.append(self.first_agent)
```
d1donlydfink, line 212: If there are no front men found by looking for someone without function parameters,
look instead for the first guy registered.  This will allow top-level agents that can be called as functions
to be invoked correctly.

neuro_san/graph/tools/calling_tool.py
```diff
@@ -136,6 +137,26 @@ def get_callable_tool_names(agent_tool_spec: Dict[str, Any]) -> List[str]:
...
         return tool_list
 
+    def get_callable_tool_dict(self) -> Dict[str, str]:
+        """
+        :return: A dictionary of name -> name, where the keys are what the tool has been
+                called in the framework (e.g. langchain) and the value is what the tool
+                has been called in the agent_spec itself.  This is often a reflexive
+                mapping, except in the case of external tools where a safer name is
+                needed for internal framework reference.
+        """
```
d1donlydfink, line 147: Langchain needs its tool names to conform to a certain naming convention that our specs for external agents will never conform to. This method helps make a lookup table between that safe_name and the actual external agent name.

neuro_san/graph/tools/calling_tool.py
```diff
@@ -182,15 +203,18 @@ async def make_one_tool_function_call(self, component_tool_call: ToolCall) -> Di
...
         # Create a new instance of a JSON-speced tool using the supplied callable_tool_name.
-        callable_tool_names: List[str] = self.get_callable_tool_names(self.agent_tool_spec)
-        if tool_name not in callable_tool_names:
-            raise ValueError(f"{tool_name} is not in tools {callable_tool_names}")
+        # At this point tool_name might be an internal reference to an external tool,
+        # so we need to check a mapping.
+        callable_tool_dict: Dict[str, str] = self.get_callable_tool_dict()
+        use_tool_name: str = callable_tool_dict.get(tool_name)
+        if use_tool_name is None:
+            raise ValueError(f"{tool_name} is not in tools {list(callable_tool_dict.keys())}")
```
d1donlydfink, line 211: Lookup the tool name to use within the graph system (which could be an external tool name) in the lookup table.
Most of the time it's not an external agent.

neuro_san/graph/tools/calling_tool.py
```diff
@@ -182,15 +203,18 @@ async def make_one_tool_function_call(self, component_tool_call: ToolCall) -> Di
...
 
         # Note: This is not a BaseTool. This is our own construct within graph
         #       that we can build().
         callable_component: CallableTool = \
             self.factory.create_agent_tool(self.run_context, self.logger,
-                                           tool_name, self.sly_data, tool_arguments)
+                                           use_tool_name, self.sly_data, tool_arguments)
```
d1donlydfink, line 217: Use the tool name found in the look up table.

neuro_san/graph/tools/external_tool.py
```diff
@@ -99,11 +99,16 @@ async def build(self) -> List[Any]:
             # not exist on all replications of a single service pod.
             response = logs_response.get("chat_response")
 
+        # Parse the logs response for the last thing the assistant said
+        response_split: List[str] = response.split("assistant: ")
+        content: str = response_split[-1]
```
d1donlydfink, line 104: When looking for responses from the external tool, look for the last thing the assistant said as the actual result.

neuro_san/run_context/langchain/langchain_openai_function_tool.py
```diff
@@ -83,6 +84,11 @@ def from_function_json(function_json: Dict[str, Any],
         # definition.
         tool = LangChainOpenAIFunctionTool(**function_json)
 
+        # Check for external tools.
+        name: str = function_json.get("name")
+        if ExternalToolAdapter.is_external_agent(name):
+            tool.name = ExternalToolAdapter.get_safe_agent_name(name)
```
d1donlydfink, line 90: The framework-internal name for an external tool needs to have a blessed name to be registered as a usable tool within langchain.

neuro_san/run_context/langchain/langchain_run_context.py
```diff
@@ -134,7 +134,7 @@ async def _create_base_tool(self, name: str) -> BaseTool:
         if agent_spec is None:
 
             # See if the agent name given could reference an external agent.
-            if ExternalToolAdapter.is_external_agent(name):
+            if not ExternalToolAdapter.is_external_agent(name):
```
d1donlydfink, line 137: Repair the check so external tools can be found.

neuro_san/run_context/utils/external_tool_adapter.py
```diff
@@ -124,7 +126,27 @@ def is_external_agent(agent_url: str) -> bool:
...
+        is_external: bool = agent_location is not None
+        return is_external
+
+    @staticmethod
+    def get_safe_agent_name(agent_url: str) -> str:
+        """
+        :param agent_url: The URL describing where to find the desired agent.
+        :return: A name that is suitable for using within agent toolkits (like langchain)
+                for internal tool reference.
+        """
```
d1donlydfink, line 138: Create a safe name based on the external tool's path component of its url.
Something that langchain will not find objectionable.

neuro_san/run_context/utils/external_tool_adapter.py
```diff
@@ -145,4 +167,9 @@ def create_session(agent_location: Dict[str, str]) -> AgentSession:
         #   to minimize socket usage, but for now use the ServiceAgentSession
         #   so as to ensure proper logging even on the same server (localhost).
         session = ServiceAgentSession(host, port, agent_name=agent_name)
+
+        # Quiet any logging from leaf-common grpc stuff.
+        quiet_please = logging.getLogger("leaf_common.session.grpc_client_retry")
+        quiet_please.setLevel(logging.WARNING)
```
d1donlydfink, line 173: Shut up the outgoing grpc_client_retry logging that spams the logs.

.flake8
```diff
@@ -0,0 +1,6 @@
+[flake8]
+max-line-length = 120
+exclude =
+    dist,
+    neuro_san/grpc/generated,
+    proto,
```
d1donlydfink, line 6: Base neuro-san flake8 off of what is in unileaf.

tests/neuro_san/agent_cli.py
```diff
@@ -15,7 +15,6 @@
 
 import argparse
 import json
-import os
```
d1donlydfink, line 18: Remove unused import caught by flake8.

neuro_san/graph/tools/external_tool.py
andreidenissov-cog: No chance of getting IndexError exception here?
Can response be empty?

d1donlydfink: Good point I will fix.

d1donlydfink: Actually looking at this, I think it's OK as-is.
We know that response at this point is a non-None string with length > 0, otherwise it would not have gotten out of the loop above.

If "assistant: " is not in the string at all (however, it is indeed 99% likely to be there), it will merely return the entire log, which I think is OK for now - let the upstream agent deal with the edge case.
Maybe I will log the situation since it shouldn't happen.

d1donlydfink: Addressed with a message logged.

neuro_san/graph/tools/calling_tool.py
ofrancon: Does it mean we can start using spaces in our Agent names?
I tried earlier today, because it makes the agent graph more user friendly, but LangChain crashed

PR #7, opened 2024-11-21
Fix Path Traversal and some references to DECISION_ASSISTANT env vars

d1donlydfink:
* Address a Checkmarx PathTraversal problem from service command line args
* Change the default of the agent_cli to use direct connections
* Change DECISION_ASSISTANT_* env var names to be AGENT_*

### DIFF
--- a/neuro_san/client/agent_cli.py
+++ b/neuro_san/client/agent_cli.py
@@ -119,7 +119,7 @@ class AgentCli:
         arg_parser.add_argument("--sly_data", type=str,
                                 help="JSON string containing data that is out-of-band to the chat stream, "
                                      "but is still essential to agent function")
-        arg_parser.add_argument("--connection", default="service",
+        arg_parser.add_argument("--connection", default="direct",
                                 choices=["service", "direct"],
                                 help="""
 The type of connection to initiate. Choices are to connect to:
--- a/neuro_san/service/agent_main_loop.py
+++ b/neuro_san/service/agent_main_loop.py
@@ -16,7 +16,7 @@ from typing import Dict
 
 import logging
 
-from os import environ
+import os
 
 from argparse import ArgumentParser
 from pathlib import Path
@@ -73,8 +73,7 @@ class AgentMainLoop(ServerLoopCallbacks):
                                 default=self.DEFAULT_TOOL_REGISTRY_FILE,
                                 help=".hocon or .json file defining the AgentToolRegistry for the service")
         arg_parser.add_argument("--port", type=int,
-                                default=int(environ.get("DECISION_ASSISTANT_PORT",
-                                                        AgentSession.DEFAULT_PORT)),
+                                default=int(os.environ.get("AGENT_PORT", AgentSession.DEFAULT_PORT)),
                                 help="Port number for the service")
 
         # Actually parse the args into class variables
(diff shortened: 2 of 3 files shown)


neuro_san/client/agent_cli.py
```diff
@@ -119,7 +119,7 @@ def parse_args(self):
         arg_parser.add_argument("--sly_data", type=str,
                                 help="JSON string containing data that is out-of-band to the chat stream, "
                                      "but is still essential to agent function")
-        arg_parser.add_argument("--connection", default="service",
+        arg_parser.add_argument("--connection", default="direct",
```
d1donlydfink, line 122: Change default of CLI to use direct connections.
This is the best way to test simple agents anyway. 
These days we largely only need a server if top-level agents are referencing each other.

neuro_san/service/agent_main_loop.py
```diff
@@ -73,8 +73,7 @@ def parse_args(self):
                                 default=self.DEFAULT_TOOL_REGISTRY_FILE,
                                 help=".hocon or .json file defining the AgentToolRegistry for the service")
         arg_parser.add_argument("--port", type=int,
-                                default=int(environ.get("DECISION_ASSISTANT_PORT",
-                                                        AgentSession.DEFAULT_PORT)),
+                                default=int(os.environ.get("AGENT_PORT", AgentSession.DEFAULT_PORT)),
```
d1donlydfink, line 76: Rename the env var to AGENT_PORT from DECISION_ASSISTANT_PORT.
I believe we are not actually using this env var anywhere currently.

neuro_san/service/agent_main_loop.py
```diff
@@ -88,6 +87,8 @@ def parse_args(self):
         tool_registry: AgentToolRegistry = None
         tool_registry_file: str = args.tool_registry_file
         if tool_registry_file is not None and len(tool_registry_file) > 0:
+            # Check the path to avoid Path Traversal issues from scans.
+            tool_registry_file = self._check_path(tool_registry_file)
```
d1donlydfink, line 91: Previous scans showed a Path Traversal problem with tool_registry_file coming from the command line args.
Here we pass it to a validation method which raises an exception if it's out-of-bounds to satisfy the scanners.

neuro_san/service/agent_main_loop.py
```diff
@@ -144,6 +145,19 @@ def shutdown_callback(self):
         logger.info("Shutdown: shutting down AsyncioExecutor")
         self.asyncio_executor.shutdown()
 
+    def _check_path(self, file_path: str) -> str:
+        """
+        Checks path to be sure its in-bounds for Checkmarx Path Traversal
+        :param file_path: The file path to check
+        :return: True if the file_path is in bounds.  False if not.
+        """
```
d1donlydfink, line 153: Method to check the bounds of the file.

neuro_san/service/agent_server.py
```diff
@@ -66,8 +66,8 @@ def __init__(self, port: int,
         current_dir = os.path.dirname(os.path.abspath(__file__))
 
         setup_logging(SERVER_NAME_FOR_LOGS, current_dir,
-                      'DECISION_ASSISTANT_SERVICE_LOG_JSON',
-                      'DECISION_ASSISTANT_SERVICE_LOG_LEVEL')
+                      'AGENT_SERVICE_LOG_JSON',
+                      'AGENT_SERVICE_LOG_LEVEL')
```
d1donlydfink, line 70: More renames for the env var to AGENT_* from DECISION_ASSISTANT_*.
I believe we are not actually using these env vars anywhere currently.

PR #10, opened 2024-11-22
Try better specs of setuptools for git versioning

d1donlydfink:
Trying to get better versioning of dist-info going.
I think I have more-proper config going now and I can get the better versioning locally,
but I can't get CF to do it.  See comments in PR.

Even so, this work gets me farther.

### DIFF
--- a/build_scripts/codefresh.yml
+++ b/build_scripts/codefresh.yml
@@ -9,13 +9,6 @@ steps:
         git: github
         revision: "${{CF_REVISION}}"
 
-    get-tags:
-        title: Get git tags for proper versioning
-        image: alpine/git:latest
-        commands:
-            - git init
-            - git fetch --tags
-
     prepare_for_build:
         description: "Do any common preparation for tests"
         type: "freestyle"


build_scripts/codefresh.yml
```diff
@@ -199,7 +199,6 @@ steps:
         description: Build wheel file for distribution
         working_directory: ${{APP_SOURCE}}
         commands:
-            - pip install build wheel
```
d1donlydfink, line 202: Don't need this because these are in the requirements-build.txt now along with other build-time goodies

pyproject.toml
```diff
@@ -2,7 +2,7 @@
 # https://packaging.python.org/en/latest/overview/
 
 [build-system]
-requires = ["setuptools>=61.0"]
+requires = ["setuptools>=61.0", "wheel", "setuptools-git-versioning>=2.0,<3"]
```
d1donlydfink, line 5: Add requirements to the build-system. This is per the setuptools-git-versioning docs.

requirements-build.txt
```diff
@@ -14,3 +14,9 @@ pylint==3.3.1
 
 # Special pylint extensions
 pylint-protobuf==0.22.0
+
+# packaging
+build
+setuptools>=61.0
+wheel
+setuptools-git-versioning>=2.0.<3
```
d1donlydfink, line 22: Having these in the requirements-build.txt and part of my local env makes my own local builds do much much better w/rt getting a proper version number (with post and all that biz too) in the wheel's file name.
Here is a sample wheel file name:
neuro_san-0.1.2.post3+git.3f9d32ad-py3-none-any.whl

However, I cannot get this same degree of fanciness happening in CodeFresh.
All it wants to do is produce a wheel file with version 0.0.1  
This will at least be bad for uploads to s3 because the same file will be written over all the time.
It *might* also be bad for GitHub installs in requirements files because the version information that gets
installed into the venv is not completely correct.  (It's not so bad because we go by tag, but might be bad for tar files, for instance. Dunno)

I have tried some stuff from here, but with Fails: https://codefresh.io/docs/docs/example-catalog/ci-examples/git-checkout-custom/

@donn-leaf : I think this is happening because the git-clone step of the codefresh.yaml is only getting git info about the commit, and not any extra info about the pre-existing tags (which is what the new voodoo requires).
I have tried doing a separate step that does a `git fetch --tags` to no avail (it errors out).  Not sure what the blessed way of doing extra git business in CF is.

d1donlydfink: If we can solve this, then we do not have to keep updating the version in the source files before publishing.

PR #13, opened 2024-11-26
UN-2659 Ensure generated gRPC files get packaged up

d1donlydfink:
Generated gRPC files were not getting in the zip or tar.gz files generated by GitHub.
Solve that by actually checking in the generated gRPC files.

This also means they will get picked up in the wheel file too.

### DIFF
--- a/build_scripts/run_pylint.sh
+++ b/build_scripts/run_pylint.sh
@@ -45,7 +45,7 @@ done
 
 echo "Running pylint on directories '${use_dirs}':"
 # shellcheck disable=2086
-pylint --load-plugins=pylint_protobuf -j 6 -rn --rcfile=${RCFILE} ${use_dirs}
+pylint --load-plugins=pylint_protobuf -j 0 -rn --rcfile=${RCFILE} ${use_dirs}
 retval=$?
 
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -26,5 +26,7 @@ __pycache__/
 site/
 
 # Generated files
-**/generated/*
+# Note: We actually want to check-in the generated files so they can be properly
+#       exported via GitHub zip/tar.gz and wheel file mechanisms.
+# **/grpc/generated/*
```
d1donlydfink, line 31: Specifically do not ignore generated grpc files.

neuro_san/grpc/generated/agent_pb2.py
```diff
@@ -0,0 +1,53 @@
+
+
+
+
+
+# -*- coding: utf-8 -*-
+# Generated by the protocol buffer compiler.  DO NOT EDIT!
+# source: agent.proto
+# Protobuf Python Version: 4.25.1
```
d1donlydfink, line 9: This fgenerated grpc file and its friend need to be checked in so they can be packaged up correctly.

PR #20, opened 2024-12-06
Fix case when starting agent_main_loop outside of Dockerfile

d1donlydfink:
* Fix a problem with running agent_main_loop outside the Dockerfile.
* Add a better exception message in a squirrelly agent tool path case.

Feel free to merge/delete/publish upon approval of this one.

### DIFF
--- a/neuro_san/graph/registry/agent_tool_registry.py
+++ b/neuro_san/graph/registry/agent_tool_registry.py
@@ -87,6 +87,9 @@ class AgentToolRegistry(AgentToolFactory):
                         len(resolved_path) > len(best_path):
                     best_path = resolved_path
 
+            if len(best_path) == 0:
+                raise ValueError("No reasonable agent tool path found in PYTHONPATH for {agent_tool_path}")
+
             # Find the path beneath the python path
             resolve_path = agent_tool_path.split(best_path)[1]
 


neuro_san/service/agent_server.py
```diff
@@ -74,6 +75,12 @@ def __init__(self, port: int,
         self.server_loop_callbacks = server_loop_callbacks
         current_dir = os.path.dirname(os.path.abspath(__file__))
 
+        # Make for easy running from the neuro-san repo
+        if os.environ.get("AGENT_SERVICE_LOG_JSON") is None:
+            # Use the log file that is local to the repo
+            file_of_class = FileOfClass(__file__, path_to_basis="../deploy")
+            os.environ["AGENT_SERVICE_LOG_JSON"] = file_of_class.get_file_in_basis("logging.json")
```
d1donlydfink, line 82: Make the default logging.json file the one in the repo if nothing is set.

leaf-server-common expects the logging.json file to be in the same directory as who is calling the logging setup method here.  In neuro-san now that those files have moved to the deploy directory, that is not the case any more.

neuro_san/graph/registry/agent_tool_registry.py
```diff
@@ -87,6 +87,9 @@ def determine_agent_tool_path(self, agent_tool_path: str) -> str:
                         len(resolved_path) > len(best_path):
                     best_path = resolved_path
 
+            if len(best_path) == 0:
+                raise ValueError("No reasonable agent tool path found in PYTHONPATH for {agent_tool_path}")
```
d1donlydfink, line 91: Add a better exception message for a problem I am trying to debug.

neuro_san/service/agent_server.py
ofrancon: Thanks, that should solve my issue

PR #21, opened 2024-12-06
UN-2672 Easier manifests from external repos

d1donlydfink:
Some moves to make it easier to use manifests and individual tools from external repos.

This basically is the attempt to keep the way folks were invoking the service from external repos largely intact.

I've tested a bunch of combinations of direct/service Dockerfile/agent_main_loop here and in an external repo.
All seems to be well AFAICT.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ Option 1: Build and run the docker container for the hosting agent service:
 
 Option 2: Run the service directly for a specific agent network:
 
-    python ./neuro_san/service/agent_main_loop.py --tool_registry_file hello_world
+    python -m neuro_san.service.agent_main_loop --tool_registry_file hello_world
 
 #### Client
 


neuro_san/graph/registry/agent_tool_registry.py
```diff
@@ -79,19 +79,21 @@ def determine_agent_tool_path(self, agent_tool_path: str) -> str:
         if agent_tool_path.find(os.sep) >= 0:
 
             # Find the best of many resolution paths in the PYTHONPATH
+            resolved_tool_path: str = str(Path(agent_tool_path).resolve())
             best_path = ""
             pythonpath_split = os.environ.get("PYTHONPATH").split(":")
             for one_path in pythonpath_split:
                 resolved_path: str = str(Path(one_path).resolve())
-                if agent_tool_path.startswith(resolved_path) and \
+                if resolved_tool_path.startswith(resolved_path) and \
```
d1donlydfink, line 87: Resolve the tool path to get a proper apples-to-apples comparison with how the PYTHONPATH is looked at.

neuro_san/graph/registry/agent_tool_registry.py
```diff
@@ -79,19 +79,21 @@ def determine_agent_tool_path(self, agent_tool_path: str) -> str:
...
                         len(resolved_path) > len(best_path):
                     best_path = resolved_path
 
             if len(best_path) == 0:
-                raise ValueError("No reasonable agent tool path found in PYTHONPATH for {agent_tool_path}")
+                raise ValueError(f"No reasonable agent tool path found in PYTHONPATH for {agent_tool_path}")
```
d1donlydfink, line 92: Make it an f-string to get good values out for agent_tool_path.

neuro_san/graph/registry/agent_tool_registry.py
```diff
@@ -79,19 +79,21 @@ def determine_agent_tool_path(self, agent_tool_path: str) -> str:
...
 
             # Find the path beneath the python path
-            resolve_path = agent_tool_path.split(best_path)[1]
+            path_split = resolved_tool_path.split(best_path)
+            resolve_path = path_split[1]
```
d1donlydfink, line 96: Did this for some debugging. Harmless to leave as-is.

neuro_san/service/agent_main_loop.py
```diff
@@ -83,8 +83,15 @@ def parse_args(self):
         """
         # Set up the CLI parser
         arg_parser = ArgumentParser()
+
+        # If we are using a manifest file, allow all the tools in the manifest
+        default_tool_registry_file: str = ""
+        if os.environ.get("AGENT_MANIFEST_FILE") is None:
+            # If there is no manifest, then fall back to a default
+            default_tool_registry_file = self.DEFAULT_TOOL_REGISTRY_FILE
```
d1donlydfink, line 91: Redo the default for the single registry file case.
If an AGENT_MANIFEST_FILE is provided via env, then do not specify which tool registry file by default.
This will mean that all agents in the manifest.hocon will be started up.

neuro_san/graph/registry/agent_tool_registry_restorer.py
```diff
@@ -60,9 +60,6 @@ def restore(self, file_reference: str = None):
         if file_reference is None or len(file_reference) == 0:
             raise ValueError(f"file_reference {file_reference} cannot be None or empty string")
 
-        if file_reference.startswith("/") and self.registry_dir is not None:
-            raise ValueError(f"file_reference {file_reference} cannot be an absolute path")
```
d1donlydfink, line 64: Remove the absolute path constraint.  Does not seem to be necessary any more and I don't recall why it was needed in the first place.

README.md
```diff
@@ -61,7 +61,7 @@ Option 1: Build and run the docker container for the hosting agent service:
 
 Option 2: Run the service directly for a specific agent network:
 
-    python ./neuro_san/service/agent_main_loop.py --tool_registry_file hello_world
+    python -m neuro_san.service.agent_main_loop --tool_registry_file hello_world
```
d1donlydfink, line 64: Change the README to reflect a more portable usage that uses python module names (findable when pip installed or via the PYTHONPATH) vs a fixed path.

@ofrancon , @babakatwork: This will be important going forward.
The file path will still work from here, but in the external repos the module path will be a requirement (and those READMEs will change soon enough)

PR #23, opened 2024-12-10
Strawman for streaming chat interface

d1donlydfink:
The objective with this PR is to stub out the new StreamingChat API method on the agent service
so that later in subsequent PRs when we make it do something real, what we did to make it real will be more obvious in the diffs in the next PR.

The idea is to work *now* towards solidifying that protocol before we are hemmed in by having to support users external to the group.  While there is still more to do with the protocol (yes, @dsargent like error reporting), this beginning allows me to set a direction so that I can get the basics of streaming gRPC responses working in that subsequent make-it-real PR.

@dsargent and @andreidenissov-cog : For this to be taken up by unileaf, we will need to update the Python/GoLang/UI scripts that take shared .proto files as input for code generation, as there are 2 new files to consider.

### DIFF
--- a/neuro_san/run_context/langchain/llm_factory.py
+++ b/neuro_san/run_context/langchain/llm_factory.py
@@ -315,6 +315,11 @@ class LlmFactory:
                                                                      "OPENAI_PROXY"),
                             openai_api_type=LlmFactory.get_value_or_env(use_config, "openai_api_type",
                                                                         "OPENAI_API_TYPE"),
+                            azure_endpoint=LlmFactory.get_value_or_env(use_config, "azure_endpoint",
+                                                                       "AZURE_OPENAI_ENDPOINT"),
+                            # AD here means "ActiveDirectory"
+                            azure_ad_token=LlmFactory.get_value_or_env(use_config, "azure_ad_token",
+                                                                       "AZURE_OPENAI_AD_TOKEN"),
                             max_tokens=use_max_tokens,
                             model_name=use_model_name)
         elif base_class == ChatAnthropic:


neuro_san/api/grpc/agent.proto
```diff
@@ -37,6 +38,9 @@ service AgentService {
 
     // Resets the chat initiated with Chat() above.
     rpc Reset (ResetRequest) returns (ResetResponse) {}
+
+    // Unidirectional streaming method which would supercede Chat() and Logs() above.
+    rpc StreamingChat(ChatRequest) returns (stream ChatResponse) {}
```
d1donlydfink, line 43: The new API method.
The whole idea here is to *eventually* be able to only really have this method and Function() above and that's it.
The rest go away.

neuro_san/api/grpc/agent.proto
```diff
@@ -110,6 +114,9 @@ message ChatRequest {
     // The intent is for the key references to be passed to tools,
     // which then grab the values by programmatic means.
     google.protobuf.Struct sly_data = 3 [json_name="sly_data"];
+
+    // Supercedes user_input above
+    chat.ChatMessage user_message = 4 [json_name="user_message"];
```
d1donlydfink, line 119: Add a user_message as what is sent via the ChatRequest.
Eventually user_input above will be superceded, as the new ChatMessage allows for sending media content (among other things).  For now we allow both.

neuro_san/api/grpc/agent.proto
```diff
@@ -125,6 +132,18 @@ message ChatResponse {
...
+    // Idea is that all of the above come back immediately to establish
+    // the connection, but likely the field below would not initially be filled in.
+    // Later on, as messages return from agents, later streamed versions of this
+    // message would have response below filled in with single messages from
+    // specific agents any time any one of them has something new to say.
+    // The origin info on the ChatMessage would allow the client to place
+    // just where in the agent hierarchy the message comes from.
+    // (Think like a URI:
+    //  opportunity_finder_pipeline/opportunity_finder_process_manager/synthetic_data_generator)
+    chat.ChatMessage response = 4 [json_name="response"];
```
d1donlydfink, line 146: With the new StreamingChat API, any ChatResponse will have a response ChatMessage (same form as the input) associated with each one that is streamed back to the client.  This allows us (among other things) to stream media as part of the response.

Future:
* While I was doing this, it occurred to me that the clunkily service-stateful notion of "session_id"s could be replaced with the state necessary to reproduce the session.  All this state really is (currently) is chat history for the FrontMan. This of course comes with tradeoffs that are likely worthwhile to make. That is, we trade greater network usage for the ability to take our agent chat history to any replicated pod instead of having to retry to find the right pod with the right session_id history like we do today.  That will be a beneficial change coming soon to a protocol near you.

* That thought led me to another dope-smoking thought about context/chat-history for all the agents and how that could potentially improve reliability. Will have to chat with some others before going there, though.

* @dsargent wants me to investigate breaking error messages out separately from the chat stream so they are easier to parse. Will need to play around with that. At the moment I'm not sure if such a beast goes here or in the ChatMessage.

neuro_san/api/grpc/agent_pb2_grpc.py
```diff
@@ -41,6 +41,11 @@ def __init__(self, channel):
                 request_serializer=neuro__san_dot_api_dot_grpc_dot_agent__pb2.ResetRequest.SerializeToString,
                 response_deserializer=neuro__san_dot_api_dot_grpc_dot_agent__pb2.ResetResponse.FromString,
                 )
+        self.StreamingChat = channel.unary_stream(
+                '/dev.cognizant_ai.neuro_san.api.grpc.agent.AgentService/StreamingChat',
+                request_serializer=neuro__san_dot_api_dot_grpc_dot_agent__pb2.ChatRequest.SerializeToString,
+                response_deserializer=neuro__san_dot_api_dot_grpc_dot_agent__pb2.ChatResponse.FromString,
+                )
```
d1donlydfink, line 48: I will comment here just to say I am not going to comment on the contents of protoc-generated python files like those that appear under api.grpc here.

neuro_san/api/grpc/chat.proto
```diff
@@ -0,0 +1,63 @@
...
+// Go - https://developers.google.com/protocol-buffers/docs/reference/go-generated
+// Python - https://developers.google.com/protocol-buffers/docs/reference/python-generated
+//
+syntax = "proto3";
+
+package dev.cognizant_ai.neuro_san.api.grpc.chat;
+
+option go_package = "github.com/leaf-ai/neuro_san/internal/gen/dev.cognizant_ai/neuro_san/api/grpc/chat/v1;chat";
+
+import "neuro_san/api/grpc/mime_data.proto";
```
d1donlydfink, line 16: This file is largely lifted from unileaf repo with some minor improvements.

neuro_san/api/grpc/chat.proto
```diff
@@ -0,0 +1,63 @@
...
+message ChatMessage {
+
+    // Type of message corresponding to basic langchain message types
+    // that are in common usage.
+    enum ChatMessageType {
+        UNKNOWN = 0;    // Note: UNKNOWN is not a langchain thing.
+        HUMAN = 1;
+        SYSTEM = 2;
+        AI = 3;
+        TOOL = 4;
```
d1donlydfink, line 31: Enums of AI and TOOL are swapped compared to what is in unileaf repo.
This allows a simple inequality to help us through the initial phase of message filtering.

neuro_san/api/grpc/chat.proto
```diff
@@ -0,0 +1,63 @@
...
+
+    // String contents of any chat message
+    string text = 2;
+
+    // Optional bytes for any non-text media referenced by this message.
+    // For some chat sources, the string text field might also be populated as a
+    // reference for how the data was created.  If this happens, then
+    // it should be safe to assume that the text is enough to represent the
+    // message in any history carried forward.
+    mime_data.MimeData mime_data = 3 [json_name="mime_data"];
```
d1donlydfink, line 49: In the unileaf repo, this field was called "image_data".
That's a limiting name.

neuro_san/api/grpc/chat.proto
```diff
@@ -0,0 +1,63 @@
...
+    // Each string is returned in a URI format:
+    //
+    //          [host:]/agent1/agent2/...
+    //
+    // ... where the host is optional and each agent name represents an agent
+    // in a single agent hierarchy.
+    //
+    // For each top-level agent/front-man (perhaps on another server) that is called,
+    // an extra string is added to the list.
+    repeated string origin = 4 [json_name="origin"];
```
d1donlydfink, line 62: New field from unileaf repo that will allow us to report where in the agent graph a message came from.

neuro_san/api/grpc/mime_data.proto
```diff
@@ -0,0 +1,24 @@
...
+
+// A Message identifying image data
+message MimeData {
+
+    // MIME type of the image data
+    string mime_type = 1 [json_name="mime_type"];
+
+    // Raw bytes of the image
+    bytes mime_bytes = 2 [json_name="mime_bytes"];
+}
```
d1donlydfink, line 24: Exactly the same as in unileaf repo

neuro_san/api/scripts/protobuf_manifest.txt
```diff
@@ -1 +1,3 @@
 neuro_san/api/grpc/agent.proto
+neuro_san/api/grpc/chat.proto
+neuro_san/api/grpc/mime_data.proto
```
d1donlydfink, line 3: Need to add the new proto files to the manifest that tells us what needs to be sent through to protoc.

neuro_san/service/agent_service.py
```diff
@@ -252,3 +253,48 @@ def Reset(self, request: service_messages.ResetRequest,
             self.request_logger.finish_request(f"{self.agent_name}.Reset", log_marker, request_log)
 
         return response
+
+    # pylint: disable=no-member
+    def StreamingChat(self, request: service_messages.ChatRequest,
+                      context: grpc.ServicerContext) \
+            -> Iterator[service_messages.ChatResponse]:
```
d1donlydfink, line 260: Stub out the streaming chat service handling method.

neuro_san/service/agent_service.py
```diff
@@ -252,3 +253,48 @@ def Reset(self, request: service_messages.ResetRequest,
...
+                                     metadata=metadata,
+                                     security_cfg=self.security_cfg)
+        response_dict = session.streaming_chat(request_dict)
+        response_dict["request"] = request_dict
+
+        # Convert the response dictionary to a grpc message
+        # DEF - this will get better
+        response_string = json.dumps(response_dict)
+        response = service_messages.ChatResponse()
+        Parse(response_string, response)
```
d1donlydfink, line 295: While the method return type says that we will be returning an Iterator (that is indeed what gRPC types things as),
we are not yet `yield`-ing results to make the stream happen in the right.  That will happen in a subsequent make-it-real PR and you will see a diff here to reflect the right way of doing that.

neuro_san/service/agent_servicer_to_server.py
```diff
@@ -71,6 +72,11 @@ def add_rpc_handlers(self, server: Server):
                     request_deserializer=agent__pb2.ResetRequest.FromString,
                     response_serializer=agent__pb2.ResetResponse.SerializeToString,
             ),
+            'StreamingChat': unary_stream_rpc_method_handler(
+                    self.servicer.StreamingChat,
+                    request_deserializer=agent__pb2.ChatRequest.FromString,
+                    response_serializer=agent__pb2.ChatResponse.SerializeToString,
+            ),
```
d1donlydfink, line 79: Our own version of the Servicer needs to be updated with the new gRPC method.
Recall that this guy is based on the generated code for agent.proto, but we add some mojo so
that a single AgentServer can serve up multiple agent network hocon definitions.  That extra mojo allows the grpc routing to work based on agent name.

neuro_san/session/agent_service_stub.py
```diff
@@ -94,6 +96,11 @@ def __call__(self, channel: Channel):
                 request_serializer=agent__pb2.ResetRequest.SerializeToString,
                 response_deserializer=agent__pb2.ResetResponse.FromString,
                 )
+        self.StreamingChat = channel.unary_stream(
+                f"/{service_name}/StreamingChat",
+                request_serializer=agent__pb2.ChatRequest.SerializeToString,
+                response_deserializer=agent__pb2.ChatResponse.FromString,
+                )
```
d1donlydfink, line 103: Another similar stub update that helps the grpc routing when a single AgentServer serves up multiple agent types.

neuro_san/session/agent_session.py
```diff
@@ -127,3 +128,36 @@ def reset(self, request_dict: Dict[str, Any]) -> Dict[str, Any]:
                                           never return this value.
         """
         raise NotImplementedError
+
+    def streaming_chat(self, request_dict: Dict[str, Any]) -> Generator[Dict[str, Any], None, None]:
```
d1donlydfink, line 132: The new method on the transport-agnostic AgentSession.  We will always be returning a Generator with this method.

Recall that Python Generators are entities that enable iterable results to be used in a for-loop construct, where the "next iteration" is allowed to block.  This is the basic method that underlies streaming of all kinds of results.
The other tricky Python thing to remember is as soon as you have a method that `yield`-s something, your method returns a Generator whether you like it or not.

neuro_san/session/direct_agent_session.py
```diff
@@ -249,6 +250,39 @@ def reset(self, request_dict: Dict[str, Any]) -> Dict[str, Any]:
...
+                                          service instances will not keep track of the same
+                                          chat sessions.
+                              CREATED   - Returned when no session_id was given (initiation
+                                          of new chat by client) and a new chat session is created.
+            "response"      - An optional ChatMessage dictionary.  See chat.proto for details.
+
+            Note that responses to the chat input might be numerous and will come as they
+            are produced until the system decides there are no more messages to be sent.
+        """
+        raise NotImplementedError
```
d1donlydfink, line 284: Will fill in this blank on the make-it-real PR.

neuro_san/session/service_agent_session.py
```diff
@@ -199,6 +200,45 @@ def reset(self, request_dict: Dict[str, Any]) -> Dict[str, Any]:
...
+            Note that responses to the chat input might be numerous and will come as they
+            are produced until the system decides there are no more messages to be sent.
+        """
+        # pylint: disable=no-member
+        return self.call_grpc_method(
+            "streaming_chat",
+            self._streaming_chat_from_stub,
+            request_dict,
+            service_messages.ChatRequest(),
+            stream_response=True)
```
d1donlydfink, line 240: Fill in this blank on the grpc-service-calling version of the AgentSession.
Note that we are already using the new stream_response functionality of the latest leaf-common library.
The idea all along was to make it easy at this level.

neuro_san/session/service_agent_session.py
```diff
@@ -246,3 +286,15 @@ def _reset_from_stub(stub, timeout_in_seconds,
...
+    def _streaming_chat_from_stub(stub, timeout_in_seconds,
+                                  metadata, credentials, *args):
+        """
+        Global method associated with the session that calls StreamingChat
+        given a grpc Stub already set up with a channel (socket) to call with.
+        """
+        response = stub.StreamingChat(*args, timeout=timeout_in_seconds,
+                                      metadata=metadata,
+                                      credentials=credentials)
+        return response
```
d1donlydfink, line 300: Requisite static method to complete the Service*Agent pattern.

requirements-private.txt
```diff
@@ -9,7 +9,7 @@
 # DEF 07/26/2022: Due to bug in codefresh private repo plugin, this leaf-common
 # comes in from top-level requirements.txt file for now.
 # Public repos
-git+https://${LEAF_SOURCE_CREDENTIALS}@github.com/leaf-ai/leaf-common.git@1.2.12
+git+https://${LEAF_SOURCE_CREDENTIALS}@github.com/leaf-ai/leaf-common.git@1.2.14
```
d1donlydfink, line 12: Use the latest leaf-common to get the stream_response functionality out of AbstractServiceSession.

neuro_san/api/grpc/agent.proto
ofrancon: If sending the state over the network each time becomes a problem, another solution is to store the state in a key-value store like DynamoDB or Redis

d1donlydfink: It should not be a problem AFAICT.  IMHO it should be a goal of the design to not complicate things with a database.

PR #24, opened 2024-12-11
UN-2685 Do not depend on private container registry for neuro-san example Dockerfile

d1donlydfink:
Eventually we will have to give out neuro-san to folks outside of our team.
This means that the example Dockerfile for a neuro-san service that we export should not reference our own ECR.

Also remove some unnecessary information transfer when using the run.sh script.

Along these publication lines, there is still another strata of work to do in allowing a Docker/build.sh path which:
1. Allows the leaf-server-common and leaf-common dependencies in as wheel files from a local directory (Filed https://leaf-ai.atlassian.net/browse/UN-2710)
2. Publish these libs on pypi (tho that likely involves legal and some code prep work in all 3 repos)

### DIFF
--- a/neuro_san/deploy/Dockerfile
+++ b/neuro_san/deploy/Dockerfile
@@ -52,9 +52,6 @@ RUN make prefix=/usr/local all && \
 # Reset to the root directory
 WORKDIR /
 
-# Set the shell and options in each FROM section per hadolint recommendations
-SHELL ["/bin/bash", "-o", "pipefail", "-c"]
-
 # App-specific constants
 ENV USERNAME leaf-ai
 ENV APP_HOME /usr/local/${USERNAME}


neuro_san/deploy/Dockerfile
```diff
@@ -12,7 +12,45 @@
...
+    rm -rf /var/lib/apt/lists/*
+
+# Change to the Git source directory to build and install Git
+WORKDIR /tmp/git-${GIT_VERSION}
+
+RUN make prefix=/usr/local all && \
+    make prefix=/usr/local install
+
+# Reset to the root directory
+WORKDIR /
```
d1donlydfink, line 53: This is taken from the source of the python-and-git Dockerfile.
If we are going to hand this out externally we do not want to have externals depend on our own private ECR

neuro_san/deploy/run.sh
```diff
@@ -64,17 +64,8 @@ function run() {
     docker_cmd="docker run --rm -it \
         --name=$SERVICE_NAME \
         --network=$network \
-        -v $HOME/.aws:/usr/local/leaf/.aws \
-        -v $vault_cert_dir:/usr/local/leaf/vault \
-        -e AWS_ACCESS_KEY_ID \
-        -e AWS_SECRET_ACCESS_KEY \
-        -e AWS_DEFAULT_REGION \
```
d1donlydfink, line 71: Get rid of env vars neuro-san does not need to run its container.

neuro_san/deploy/Dockerfile
donn-leaf: FWIW the repo that holds the python-and-git image is public.

neuro_san/deploy/Dockerfile
donn-leaf, line 1: Lines 55 and 56 are not needed. We set the shell earlier in this build stage.

d1donlydfink: Fixed.

neuro_san/deploy/Dockerfile
d1donlydfink: Didn't realize that. Even so. I think this is more straightforward as something to give other people

PR #26, opened 2024-12-12
UN-2708 Fix checkmarx scans

d1donlydfink:
Fix a couple of PathTraversal and TrustBoundary violations found in Checkmarx scans.
These were the only issues found in recent scans.

Also: Have most of the dependencies in requirements.txt float upwards as per recommendations for library code.
This will satisfy @sidcog 's BlackDuck scans which want us to upgrade grpcio.
I have not, however, let the LLM-related stuff float. Those people are animals and I'd be cleaning up their shit daily if I did let that happen.

Tested: container and agent_main_loop + of neuro_san + unileaf

The possibilities of the testing matrix have decent fan-out with client/server/basis repo all contributing to the mix.
Please verify your own use cases before committing to taking this change.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -18,7 +18,6 @@ protobuf>=4.25.3
 pyhocon>=0.3.60
 pyOpenSSL>=24.0.0
 
-# These guys use >= to allow for conflicts between kserve and s3fs
 boto3>=1.34.51
 botocore>=1.34.51
 idna>=3.6
@@ -32,17 +31,19 @@ ruamel.yaml>=0.18.6
 hvac>=1.1.0
 
 # For LLM access
-langchain>=0.3.7
-langchain-anthropic>=0.2.4
-langchain-community>=0.3.5
-langchain-openai>=0.2.5
-openai>=1.54.1
-tiktoken>=0.8.0
-bs4>=0.0.2
-pydantic>=2.9.2
+# Keep all these guys pinned for now, as it is the wild wild west out there
+# and we don't want to be on a dependencies treadmill with timing not of our choosing.
+langchain==0.3.7
+langchain-anthropic==0.2.4
+langchain-community==0.3.5
+langchain-openai==0.2.5
+openai==1.54.1
+tiktoken==0.8.0
+bs4==0.0.2
+pydantic==2.9.2
 
 # httpx pinned only for this reason: https://github.com/openai/openai-python/issues/1903
-httpx>=0.27.2
+httpx==0.27.2
 
 # For input validation
 validators>=0.22.0


neuro_san/client/agent_cli.py
```diff
@@ -119,7 +120,7 @@ def parse_args(self):
         arg_parser.add_argument("--sly_data", type=str,
                                 help="JSON string containing data that is out-of-band to the chat stream, "
                                      "but is still essential to agent function")
-        arg_parser.add_argument("--connection", default="direct",
+        arg_parser.add_argument("--connection", default="direct", type=str,
```
d1donlydfink, line 123: Add a type to the connection argument. This should fix all Trust Boundary Violations found in the scan.

neuro_san/client/agent_cli.py
```diff
@@ -129,6 +130,13 @@ def parse_args(self):
 """)
         self.args = arg_parser.parse_args()
 
+        # Check some arguments to prevent PathTraversal scans lighting up.
+        # Since this is a command-line tool not intended to be used inside a
+        # Dockerfile service, we don't really care where these files come from.
+        # Check anyway to give warm fuzzies to scans.
+        self.args.thinking_file = FileOfClass.check_file(self.args.thinking_file, "/")
+        self.args.first_prompt_file = FileOfClass.check_file(self.args.first_prompt_file, "/")
```
d1donlydfink, line 138: Check some arguments using new FileOfClass.check_file() method.
Idea is to funnel all of these through this method which actually does the checking in a consistent way.

For these 2 args, this is for a command line tool which is never used as part of any service
(it's a shipping part of the library however so users can play). As such, we really don't care where the
input files live, so we pass "/".  They could be in /tmp (as often happens for me) or "~" or wherever.

This is my best effort to get rid of the Path Traversal vulnerability stemming from these 2 guys.
If we get dinged again, we need exceptions with the reasoning that this is a command-line tool
shipped as part of a library. This never gets executed as part of any service for unileaf or anyone else.

neuro_san/service/agent_main_loop.py
```diff
@@ -186,19 +188,6 @@ def shutdown_callback(self):
...
-        Checks path to be sure its in-bounds for Checkmarx Path Traversal
-        :param file_path: The file path to check
-        :return: True if the file_path is in bounds.  False if not.
-        """
-        bounds_dir: str = os.path.expanduser("~")
-        abs_file_path: str = os.path.abspath(file_path)
-        if not abs_file_path.startswith(bounds_dir):
-            # Appeases Checkmarx
-            raise ValueError(f"file_path {file_path} must be under {bounds_dir}")
-        return abs_file_path
```
d1donlydfink, line 200: The meat of this moves to FileOfClass.check_file()

neuro_san/service/agent_main_loop.py
```diff
@@ -125,7 +126,7 @@ def parse_args(self):
         tool_registry_file: str = args.tool_registry_file
         if tool_registry_file is not None and len(tool_registry_file) > 0:
             # Check the path to avoid Path Traversal issues from scans.
-            tool_registry_file = self._check_path(tool_registry_file)
+            tool_registry_file = FileOfClass.check_file(tool_registry_file, basis="~")
```
d1donlydfink, line 129: Use the method to consolidate file path checks.
Here we make sure the basis lives under the home directory, as this guy does actually get invoked from the service/container.

neuro_san/service/agent_main_loop.py
```diff
@@ -136,6 +137,7 @@ def parse_args(self):
             this_file_dir: str = Path(__file__).parent.resolve()
             registry_dir: str = (Path(this_file_dir) / ".." / "registries").resolve()
             registry_restorer = AgentToolRegistryRestorer(registry_dir)
+            tool_registry_file = FileOfClass.check_file(tool_registry_file, basis=registry_dir)
```
d1donlydfink, line 140: Check the path with the new method.

neuro_san/utils/file_of_class.py
```diff
@@ -69,3 +69,27 @@ def get_file_in_basis(self, filename: str) -> str:
...
+            raise ValueError("basis for file checking is None")
+
+        # Resolve full paths
+        test_abs_path: str = Path(filepath).resolve()
+        basis_abs_path: str = Path(basis).resolve()
+
+        if not test_abs_path.startswith(basis_abs_path):
+            raise ValueError("{test_abs_path} must be under {basis_abs_path}")
+
+        return test_abs_path
```
d1donlydfink, line 100: New method coming from AgentMainLoop._check_path()  A little more generalized for use from other places.

If we still have PathTraversal issues that point to inside this method, then we need to ask for another exception because this is the place where we are actually doing the checking and Checkmarx might not be able to see all that.

requirements.txt
```diff
@@ -7,18 +7,17 @@
...
 
 # These are needed for generating code from .proto files
-grpcio==1.62.0
-grpcio-health-checking==1.62.0
-grpcio-reflection==1.62.0
-grpcio-tools==1.62.0
+grpcio>=1.62.0
+grpcio-health-checking>=1.62.0
+grpcio-reflection>=1.62.0
+grpcio-tools>=1.62.0
```
d1donlydfink, line 13: Also update the grpcio versions per @sidcog 's request.
I verified that this set of requirements works in the example Dockerfile build.sh.

requirements.txt
```diff
@@ -28,10 +27,12 @@ idna>=3.6
...
 # These are transient dependencies needed by leaf-common
-ruamel.yaml==0.18.6
-hvac==1.1.0
+ruamel.yaml>=0.18.6
+hvac>=1.1.0
 
 # For LLM access
+# Keep all these guys pinned for now, as it is the wild wild west out there
+# and we don't want to be on a dependencies treadmill with timing not of our choosing.
 langchain==0.3.7
```
d1donlydfink, line 36: Do not let the LLM-related dependencies float, as is recommended for most libraries.
These guys are all animals and we don't want to drop everything to clean up after their shit.

PR #27, opened 2024-12-17
UN-2731 Add comments for latest neuro-san scan exceptions

d1donlydfink:
Basically everything flagged in the last neuro-san scan needs an exception.
They are all false positives. Luckily, there's not a lot.

### DIFF
--- a/neuro_san/service/agent_main_loop.py
+++ b/neuro_san/service/agent_main_loop.py
@@ -113,7 +113,7 @@ class AgentMainLoop(ServerLoopCallbacks):
         # Actually parse the args into class variables
 
         # Incorrectly flagged as Path Traversal 3, 7
-        # See destination below ~ line 139, 154 for explanation. 
+        # See destination below ~ line 139, 154 for explanation.
         args = arg_parser.parse_args()
         self.port = args.port
         self.server_name = args.server_name


neuro_san/client/agent_cli.py
```diff
@@ -52,6 +52,12 @@ def main(self):
         # Get initial user input
         user_input: str = None
         if self.args.first_prompt_file is not None:
+            # Incorrectly flagged as destination of Path Traversal 4
+            #   Reason: first_prompt_file was previously checked with FileOfClass.check_file()
+            #           which actually does the path traversal check. CheckMarx does not
+            #           recognize pathlib as a valid library with which to resolve these kinds
+            #           of issues.  Furthermore, this is a client command line tool that is never
+            #           used inside servers which just happens to be part of a library offering.
             with open(self.args.first_prompt_file, 'r', encoding="utf-8") as prompt_file:
```
d1donlydfink, line 61: Explanations for exceptions are at destinations so we explain why it's safe to use the values we do.

neuro_san/client/agent_cli.py
```diff
@@ -128,6 +134,11 @@ def parse_args(self):
     "direct"    - a session via library.
 All choices require an agent name.
 """)
+
+        # Incorrectly flagged as source of Path Traversal 1, 2, 4, 5, 6
+        # See destination in file_of_class.py for exception explanation.
+        # Incorrectly flagged as source of Trust Boundary Violation 1, 2
+        # See destination in agent_session_Factory.py for exception explanation.
         self.args = arg_parser.parse_args()
```
d1donlydfink, line 142: Also flag the sources, but point to the destinations for explanations.

PR #33, opened 2024-12-21
UN-2494 StreamingChat client + server - 1st light

d1donlydfink:
I sometimes say "This looks like a lot, but it's not really", but that is not the case here.
There is a lot going on in here, so be prepared to hunker down for ~20 minutes.

Fill in the stubbing that was left open in the previous Chat Strawman PR which this guy depends on.
If you have seen that one yet, take a look at that one first:
https://github.com/leaf-ai/neuro-san/pull/23 Strawman for streaming chat interface

Make it work by:
* Modifying the client agent_cli.py to allow for:
    * streaming or polling chat communication with the service (Note streaming does not have full thinking file functionality yet)
    * subclases in other repos to add their own arguments and session factories by inheritance
* Modifying the ChatSession interface and implementation to allow for:
    * polling method chat() to return an iterator so that there is a single point of entry to talk to the Front Man
    * new method streaming_chat() which sets up messages going on to an asynchronous queue
    * new method queue_consumer() which allows that asynchronous queue to be drained by the service
* Modifying the DirectAgentSession to allow for:
    * synchronously Generator-ing streaming results for grpc streaming responses from an asynchronous Generator from queue_consumer() (above)
    * Handle new-style user input from requests
* Modifying the service infrastructure so that we can keep track of request activity in order to better inform yield vs sleep decisions in the ServerLifetime.  This should improve performance for all cases, but especially so for streaming results.
* Getting rid of a show_json() method that was never being used.

Tested:
* streaming and polling with service using locally run esp_decision_assistant via agent_main_loop
* streaming and polling with service using neuro-san Dockerfile via esp_decision_assistant.
* polling with direct connections - all good
* streaming with direct connections - this still needs a little work, but I figure I can make progress on that in separate PR.


neuro_san/chat/chat_session.py
```diff
@@ -28,7 +30,7 @@ async def set_up(self):
         """
         raise NotImplementedError
 
-    async def chat(self, user_input: str, sly_data: Dict[str, Any]):
+    async def chat(self, user_input: str, sly_data: Dict[str, Any]) -> Iterator[Dict[str, Any]]:
```
d1donlydfink, line 33: Change the signature of the chat() method to reflect that it now returns an Iterator over the messages.
The grpc Chat() method itself does not use the return value at all because its clients are directed to use the Logs() method to get their results, but the new StreamingChat() method now ends up calling this guy to set things in motion.

neuro_san/chat/chat_session.py
```diff
@@ -43,6 +45,23 @@ async def chat(self, user_input: str, sly_data: Dict[str, Any]):
...
+    async def streaming_chat(self, user_input: str, sly_data: Dict[str, Any]):
+        """
+        Main entry-point method for accepting new user input
+
+        :param user_input: A string with the user's input
+        :param sly_data: A mapping whose keys might be referenceable by agents, but whose
+                 values should not appear in agent chat text. Can be None.
+        :return: Nothing.  Response values are put on a queue to be read by a call to queue_consumer()
+        """
+        raise NotImplementedError
```
d1donlydfink, line 57: New streaming_chat() method for the AgentSession.streaming_chat() to call.

neuro_san/chat/chat_session.py
```diff
@@ -43,6 +45,23 @@ async def chat(self, user_input: str, sly_data: Dict[str, Any]):
...
+
+    async def queue_consumer(self) -> AsyncIterator[Dict[str, Any]]:
+        """
+        :return: An AsyncIterator over the messages from the agent(s).
+        """
+        raise NotImplementedError
```
d1donlydfink, line 63: This is the buddy of the method above.  The AsyncIterator is really the handle to whatever is yielding the results at a lower level.  

We don't really want streaming_chat() _returning_ that queue because it would then have to be done streaming all the messages before it returns.

You'll see these in motion in DirectAgentSession.

neuro_san/chat/data_driven_chat_session.py
```diff
@@ -54,6 +60,8 @@ def __init__(self, registry: AgentToolRegistry,
         self.latest_response = None
         self.last_input_timestamp = datetime.now()
         self.sly_data: Dict[str, Any] = {}
+        self.queue: Queue[Dict[str, Any]] = Queue()
+        self.last_streamed_index: int = 0
```
d1donlydfink, line 64: In the ChatSession implementation we now have as members the async queue that acts as the buffer for messages streamed.

We also have to store the last_streamed_index. This is basically a pointer into the chat history to tell us between conversational calls to streaming_chat():  "this is where we left off"

neuro_san/chat/data_driven_chat_session.py
```diff
@@ -77,26 +85,25 @@ async def set_up(self):
...
         :param sly_data: A mapping whose keys might be referenceable by agents, but whose
                  values should not appear in agent chat text. Can be None.
+        :return: An Iterator of chat.ChatMessage dictionaries
 
-        Note that nothing is returned immediately as processing
-        the chat happens asynchronously and might take longer
-        than the lifetime of a socket.  Results are polled from
-        get_latest_response() below.
+        Results are polled from get_latest_response() below, as this non-streaming
+        version can take longer than the lifetime of a socket.
```
d1donlydfink, line 98: While we now return an Iterator from chat(), the older polling Chat() service method doesn't really care.

neuro_san/chat/data_driven_chat_session.py
```diff
@@ -109,17 +116,105 @@ async def chat(self, user_input: str, sly_data: Dict[str, Any]):
 
         try:
             self.logger.write("consulting chat agent(s)...")
+
+            # DEF - drill further down for iterator from here to enable getting
+            #       messages from downstream agents.
             raw_messages: List[Any] = await self.front_man.submit_message(user_input)
```
d1donlydfink, line 122: A note to myself for future improvements.
Right now we are getting all the messages that are returned from the front man.
Later PRs will push the reference to the queue down so that elements of the graph can stream their own messages at their own level.

I haven't gone there yet because:
* Enough going on in this PR
* I really just wanted to focus on getting the streaming mechanisms in place to get them working

I believe I will have to go their in order to fully supplant the old Chat()/Logs() polling methods' ability to return chat history from deep down which powers @babakatwork 's planetarium-lookin' demo.

So this is the first place I will say: _While the streaming chat is beginning to work with this PR, it's not there yet for everything we use the Logs() method for just yet._

neuro_san/chat/data_driven_chat_session.py
```diff
@@ -109,17 +116,105 @@ async def chat(self, user_input: str, sly_data: Dict[str, Any]):
...
-            decision_messages = pretty_the_messages(raw_messages)
+
```
d1donlydfink, line 123: This call to pretty_the_messages moves down.

neuro_san/chat/data_driven_chat_session.py
```diff
@@ -109,17 +116,105 @@ async def chat(self, user_input: str, sly_data: Dict[str, Any]):
...
         except BadRequestError:
             # This can happen if the user is trying to send a new message
             # while it is still working on a previous message that has not
             # yet returned.
-            decision_messages = "Patience, please. I'm working on it."
+            raw_messages: List[Any] = [
+                AgentFrameworkMessage(content="Patience, please. I'm working on it.")
+            ]
```
d1donlydfink, line 130: Use the new AgentFrameworkMessage type (understood only at our level)
to communicate the "wait for it ..." message at this level.

neuro_san/chat/data_driven_chat_session.py
```diff
@@ -109,17 +116,105 @@ async def chat(self, user_input: str, sly_data: Dict[str, Any]):
...
             print(traceback.format_exc())
 
-        # Decision has been made. Update the response.
-        self.latest_response = decision_messages
+        # Update the polling response.
+        prettied_messages = pretty_the_messages(raw_messages)
+        self.latest_response = prettied_messages
```
d1donlydfink, line 135: Pretty the messages always now that they all come in the same form.
Tweak the comment to banish the old DecisionAssistant frame-of-mind.

neuro_san/chat/data_driven_chat_session.py
```diff
@@ -109,17 +116,105 @@ async def chat(self, user_input: str, sly_data: Dict[str, Any]):
...
+
+        chat_messages: List[Dict[str, Any]] = []
+        for raw_message in raw_messages:
+            chat_message: Dict[str, Any] = convert_to_chat_message(raw_message)
+            chat_messages.append(chat_message)
+
+        return iter(chat_messages)
```
d1donlydfink, line 142: Prepare the messages and return the iterator from this plain-old synchronous chat() method even though it's ignored by the service call.   This enables streaming_chat() to call it to keep the same logic together.

neuro_san/chat/data_driven_chat_session.py
```diff
@@ -109,17 +116,105 @@ async def chat(self, user_input: str, sly_data: Dict[str, Any]):
...
+        """
+        Main streaming entry-point method for accepting new user input
+
+        :param user_input: A string with the user's input
+        :param sly_data: A mapping whose keys might be referenceable by agents, but whose
+                 values should not appear in agent chat text. Can be None.
+        :return: Nothing.  Response values are put on a queue to be read by a call to queue_consumer()
+        """
+        # Queue Producer from this:
+        #   https://stackoverflow.com/questions/74130544/asyncio-yielding-results-from-multiple-futures-as-they-arrive
```
d1donlydfink, line 154: New streaming_chat() method.
Leave a breadcrumb as to where I riffed the async queue idea from.

neuro_san/chat/data_driven_chat_session.py
```diff
@@ -109,17 +116,105 @@ async def chat(self, user_input: str, sly_data: Dict[str, Any]):
...
+        # DEF - These put()s will eventually be pushed down into the library.
+        chat_messages: Iterator[Dict[str, Any]] = await self.chat(user_input, sly_data)
+        for index, chat_message in enumerate(chat_messages):
```
d1donlydfink, line 157: As advertised above, inside streaming_chat() we call chat() itself to get an iterator over the messages that are returned. Currently these messages are only from the front man, but they do contain the entire chat history with that guy only.

neuro_san/chat/data_driven_chat_session.py
```diff
@@ -109,17 +116,105 @@ async def chat(self, user_input: str, sly_data: Dict[str, Any]):
...
+
+            if self.is_streamable_message(chat_message, index):
+                # The consumer await-s for self.queue.get()
+                await self.queue.put(chat_message)
+                self.last_streamed_index = index
```
d1donlydfink, line 164: For now we do some filtering in the service with is_streamable_message() which really asks the question: should I actually send this message to the client?  A subsequent PR will have this filtering move client side.

Note that we are in an sync call here, so as we are iterating over the returned messages from chat() here (for now), we put() them onto the asynchronous queue so the DirectAgentSession.streaming_chat() method can get() them out the other end in a synchronous method and stream those results to a grpc client.

Having the chat_messages iterator as an middleman is not the ideal, nor is it the most efficient.
Eventually this put() call will be part of the RunContexts so it can be passed around and any agent in the graph can put their own messages on as they become available.  ... subsequent PR tho.

neuro_san/chat/data_driven_chat_session.py
```diff
@@ -109,17 +116,105 @@ async def chat(self, user_input: str, sly_data: Dict[str, Any]):
...
+
+        end_dict = {"end": True}
+        await self.queue.put(end_dict)
```
d1donlydfink, line 169: Since we will have messages from all kinds of agents coming over the stream eventually we need a signal
to the consumer of the queue that we are truly done.  That's what this "end marker" is about.

neuro_san/chat/data_driven_chat_session.py
```diff
@@ -109,17 +116,109 @@ async def chat(self, user_input: str, sly_data: Dict[str, Any]):
...
+
+    def is_streamable_message(self, chat_message: Dict[str, Any], index: int) -> bool:
+        """
+        Filter chat messages from the full logs that are in/appropriate for streaming.
+
+        :param chat_message: A Chat message dictionary of the form in chat.ChatMessage proto
+        :param index: The place in the chat history the message has
+        :return: True if the given message should be streamed back. False if not.
+        """
```
d1donlydfink, line 178: For now we are doing some filtering of what messages we return to the client in the service.
Eventually some of this functionality will be pushed onto the client, but I can foresee some of that
filtering responsibility being shared for ease of use for client code.  At the moment, I don't know what
a filtering spec would look like, but will get a better picture when I push the queue deeper in future PRs.

neuro_san/chat/data_driven_chat_session.py
```diff
@@ -109,17 +116,109 @@ async def chat(self, user_input: str, sly_data: Dict[str, Any]):
...
+        # This comes from definitions in chat.proto
+        ai_message_type: int = 3
+
+        message_type: int = chat_message.get("type")
+
+        # Only report messages that are important enough to send back as part of chat
+        # (for now).  This includes any response for an AI (read: LLM), a tool, or
+        # agent or framework messages.
+        if message_type is None or message_type < ai_message_type:
+            return False
```
d1donlydfink, line 189: I arranged the enums for type defined in chat.proto such that a simple inequality here acts as a basic filter of messages. E.g:  Show we what the LLM and above in the system says, no user messages or system messages.
For now, this makes the chat results with the client side simpler and when you use agent_cli, you get what you'd expect from a chatbot. More on that later.

neuro_san/chat/data_driven_chat_session.py
```diff
@@ -109,17 +116,109 @@ async def chat(self, user_input: str, sly_data: Dict[str, Any]):
...
+
+        if index <= self.last_streamed_index:
+            return False
```
d1donlydfink, line 192: Don't send what we've already sent before. Recall this ChatSession instance is saved within the context of the current session_id.  What comes back from the front man iterator now is complete chat history.

neuro_san/chat/data_driven_chat_session.py
```diff
@@ -109,17 +116,111 @@ async def chat(self, user_input: str, sly_data: Dict[str, Any]):
...
+
+    async def queue_consumer(self) -> AsyncIterator[Dict[str, Any]]:
+        """
+        Queue Consumer from this:
+            https://stackoverflow.com/questions/74130544/asyncio-yielding-results-from-multiple-futures-as-they-arrive
+
+        Loops until either the timeout is met or the end marker is seen.
+
+        :return: An AsyncIterator over the messages from the agent(s).
+        """
```
d1donlydfink, line 204: This queue_consumer() method returns the AsyncIterator spigot that the synchronous DirectAgentSession
loops over to get the messages put on the queue in streaming_chat() above.

neuro_san/chat/data_driven_chat_session.py
```diff
@@ -109,17 +116,111 @@ async def chat(self, user_input: str, sly_data: Dict[str, Any]):
...
+                if not isinstance(message, Dict):
+                    print(f"object on queue is of type {message.__class__.__name__}")
+                    message = {"end": True}
+
+                # Look for a special end-marker dictionary in the queue that signals
+                # there will be no more messages.
+                done = message.get("end") is not None
+                if not done:
+                    # yield all messages except the end marker
+                    yield message
```
d1donlydfink, line 219: Since the queue itself is not an AsyncIterator, here we pull messages off the queue itself and yield them as they come in, giving us the AsyncIterator that we need for streaming results.
We have to filter out the end-marker so we don't send that to the client.

neuro_san/client/agent_cli.py
```diff
@@ -66,26 +69,30 @@ def main(self):
...
-        state: Dict[str, Any] = {
-            "last_logs": [],
-            "last_chat_response": None,
-            "prompt": self.default_prompt,
-            "timeout": self.input_timeout_seconds
-        }
-
-        print("To see the thinking involved with the agent:\n"
-              f"    tail -f {self.args.thinking_file}\n")
-
```
d1donlydfink, line 78: Oh yeah. Had to modify the client too!
In general you will see a few things going on in here:
* Make it so that the synchronous code and asynchronous code largely use the same loop
* Actually implement the streaming side of the client.
* Ready this client code such that other repos can expand upon it as they like.
* Add a couple of command line args.

These move deleted lines simply move down below.

neuro_san/client/agent_cli.py
```diff
@@ -66,26 +69,30 @@ def main(self):
...
+              f"    tail -f {self.args.thinking_file}\n")
+
+        state: Dict[str, Any] = {
+            "last_logs": [],
+            "last_chat_response": None,
+            "prompt": self.default_prompt,
+            "timeout": self.input_timeout_seconds,
+            "user_input": user_input,
+            "sly_data": sly_data,
+        }
```
d1donlydfink, line 89: What's different from above is that user_input and sly_data are both initialized in the dictionary right off the bat.

neuro_san/client/agent_cli.py
```diff
@@ -96,28 +103,47 @@ def main(self):
...
-                self.send_user_input(user_input, sly_data)
-                user_input = None
+            state["user_input"] = user_input
+            if self.args.stream:
+                state = self.stream_once(state)
             else:
-                sleep(self.poll_timeout_seconds)
-
-            state = self.get_responses(state)
+                state = self.poll_once(state)
```
d1donlydfink, line 110: A little broken up here but the idea is that user_input is now part of the state
but we branch off here how we handle it based on wether or not we are streaming results or not.
poll_once() is the old way and the logic that was here before moves into that new method below.  stream_once() is the new way.

neuro_san/client/agent_cli.py
```diff
@@ -96,28 +103,47 @@ def main(self):
...
 
     def parse_args(self):
         """
         Parse command line arguments into member variables
         """
+
         arg_parser = argparse.ArgumentParser()
+
+        self.add_args(arg_parser)
```
d1donlydfink, line 119: Break parse_args() into a phase that adds the args (above) and a phase that checks the args (below).
This will allow for other clients to expand and add their own args (like NeuroAI already does, but currently copy/pastes this code.).

neuro_san/client/agent_cli.py
```diff
@@ -96,28 +103,47 @@ def main(self):
...
+        # Incorrectly flagged as source of Trust Boundary Violation 1, 2
+        # See destination in agent_session_factory.py for exception explanation.
+        self.args = arg_parser.parse_args()
+
+        # Check some arguments to prevent PathTraversal scans lighting up.
+        # Since this is a command-line tool not intended to be used inside a
+        # Dockerfile service, we don't really care where these files come from.
+        # Check anyway to give warm fuzzies to scans.
+        self.args.thinking_file = FileOfClass.check_file(self.args.thinking_file, "/")
+        self.args.first_prompt_file = FileOfClass.check_file(self.args.first_prompt_file, "/")
```
d1donlydfink, line 132: These checks come up from below.

neuro_san/client/agent_cli.py
```diff
@@ -96,28 +103,47 @@ def main(self):
...
+
+    def add_args(self, arg_parser: argparse.ArgumentParser):
+        """
+        Adds arguments.  Allows subclasses a chance to add their own.
+        :param arg_parser: The argparse.ArgumentParser to add.
+        """
```
d1donlydfink, line 138: Make add_args() so that subclasses can augment functionality as they see fit.

neuro_san/client/agent_cli.py
```diff
@@ -126,6 +152,10 @@ def parse_args(self):
         arg_parser.add_argument("--sly_data", type=str,
                                 help="JSON string containing data that is out-of-band to the chat stream, "
                                      "but is still essential to agent function")
+        arg_parser.add_argument("--stream", default=False, action="store_true",
+                                help="Use streaming chat instead of polling")
+        arg_parser.add_argument("--poll", dest="stream", action="store_false",
+                                help="Use polling chat instead of streaming")
```
d1donlydfink, line 158: Add args for stream vs polling.
Right now polling is still the default, but eventually that will change because streaming is actually much much snappier (will be good for demos)

neuro_san/client/agent_cli.py
```diff
@@ -134,19 +164,10 @@ def parse_args(self):
...
-        # Incorrectly flagged as source of Trust Boundary Violation 1, 2
-        # See destination in agent_session_factory.py for exception explanation.
-        self.args = arg_parser.parse_args()
-
-        # Check some arguments to prevent PathTraversal scans lighting up.
-        # Since this is a command-line tool not intended to be used inside a
-        # Dockerfile service, we don't really care where these files come from.
-        # Check anyway to give warm fuzzies to scans.
-        self.args.thinking_file = FileOfClass.check_file(self.args.thinking_file, "/")
-        self.args.first_prompt_file = FileOfClass.check_file(self.args.first_prompt_file, "/")
```
d1donlydfink, line 149: These checks move above.

neuro_san/client/agent_cli.py
```diff
@@ -134,19 +164,10 @@ def parse_args(self):
...
+        arg_parser.add_argument("--service", dest="connection", action="store_const", const="service",
+                                help="Use a service connection")
+        arg_parser.add_argument("--direct", dest="connection", action="store_const", const="direct",
+                                help="Use a direct/library call for the chat")
```
d1donlydfink, line 170: Some new flag args for service vs direct. This ends up being a little easier when typing out command lines that --connection <foo>

neuro_san/client/agent_cli.py
```diff
@@ -156,9 +177,10 @@ def open_session(self):
         if self.args.host is not None or not self.args.local:
             hostname = self.args.host
 
-        # Open a session
-        self.session = AgentSessionFactory.create_session(self.args.connection, self.args.agent,
-                                                          hostname, self.args.port)
+        # Open a session with the factory
+        factory: AgentSessionFactory = self.get_agent_session_factory()
+        self.session = factory.create_session(self.args.connection, self.args.agent,
+                                              hostname, self.args.port)
```
d1donlydfink, line 183: Break out how the AgentSessionFactory is obtained to a method that could be overridden.
This will also enable other repos to do what they want (like NeuroAI does)

neuro_san/client/agent_cli.py
```diff
@@ -171,6 +193,34 @@ def open_session(self):
...
             thinking.write("\n")
 
+    def get_agent_session_factory(self) -> AgentSessionFactory:
+        """
+        This allows subclasses to add different kinds of connections.
+
+        :return: An AgentSessionFactory instance that will allow creation of the
+                 session with the agent network.
+        """
+        return AgentSessionFactory()
```
d1donlydfink, line 203: Breakout method for factory.

neuro_san/client/agent_cli.py
```diff
@@ -171,6 +193,34 @@ def open_session(self):
...
+
+        if user_input is not None and user_input != self.default_input:
+            print(f"Sending user_input {user_input}")
+            self.send_user_input(user_input, sly_data)
+            state["user_input"] = None
+        else:
+            sleep(self.poll_timeout_seconds)
+
+        state: Dict[str, Any] = self.get_responses(state)
+        return state
```
d1donlydfink, line 222: This method breaks out the old logic from main() above for the polling case.
Right now polling is still the most robust because it actually provides the thinking file contents still which 
(I think) powers @babakatwork 's planetary demo. 

 Streaming does not yet provide the thinking file contents because not all the messages are coming over to populate it. Will get there soon.

neuro_san/client/agent_cli.py
```diff
@@ -179,7 +229,10 @@ def send_user_input(self, user_input: str, sly_data: Dict[str, Any]):
         if user_input != self.default_input:
             chat_request = {
                 "session_id": self.session_id,
-                "user_input": user_input
+                "user_message": {
+                    "type": 1,      # HUMAN from chat.proto
+                    "test": user_input
+                }
```
d1donlydfink, line 235: Send the new-style request with user_message carrying the goods.

neuro_san/client/agent_cli.py
```diff
@@ -243,6 +296,61 @@ def get_responses(self, state: Dict[str, Any]) -> Dict[str, Any]:
         }
         return return_state
 
+    def stream_once(self, state: Dict[str, Any]) -> Dict[str, Any]:
+        """
+        Use polling strategy to communicate with agent.
+        :param user_input: The initial (string) user input to the loop.
+        :param sly_data: The initial dictionary of private sly_data to send.
+        """
```
d1donlydfink, line 304: New streaming method.
Again, we are not really doing the thinking file correctly, so for now this is only for my own proof-of-concept
for making sure the streaming is doing something stream-y.

neuro_san/client/agent_cli.py
```diff
@@ -243,6 +296,61 @@ def get_responses(self, state: Dict[str, Any]) -> Dict[str, Any]:
...
+            return state
+
+        print(f"Sending user_input {user_input}")
+        chat_request = {
+            "session_id": self.session_id,
+            "user_message": {
+                "type": 1,      # HUMAN from chat.proto
+                "text": user_input
+            }
+        }
```
d1donlydfink, line 320: Use the new-style request.

neuro_san/client/agent_cli.py
```diff
@@ -243,6 +296,61 @@ def get_responses(self, state: Dict[str, Any]) -> Dict[str, Any]:
...
+                last_chat_response = text
+
+            return_state: Dict[str, Any] = {
+                "last_logs": last_logs,
+                "last_chat_response": last_chat_response,
+                "prompt": prompt,
+                "timeout": self.input_timeout_seconds
+            }
+
+        return return_state
```
d1donlydfink, line 352: The nice thing about the streaming client is that it doesn't actually return control to the input loop until the service has told you it's done sending messages.  That is much clunkier in the polling world, where you could send new messages while it's still working (and then the service gives you the hand).

neuro_san/client/direct_agent_session_factory.py
```diff
@@ -11,7 +11,7 @@
 # END COPYRIGHT
 from typing import Dict
 
-from leaf_server_common.utils.asyncio_executor import AsyncioExecutor
+from leaf_server_common.asyncio.asyncio_executor import AsyncioExecutor
```
d1donlydfink, line 14: We are using the latest leaf-server-common where this guy moved.

neuro_san/graph/tools/branch_tool.py
```diff
@@ -194,12 +194,8 @@ async def integrate_callable_response(self, run: Run, messages: List[Any]) -> Li
         while run.requires_action():
             # The tool we just called requires more information
             new_run = await self.make_tool_function_calls(run)
-
             new_run = await self.run_context.wait_on_run(new_run, self.logger)
-            self.logger.show_json(run)
```
d1donlydfink, line 199: An early rabbit hole I went down had to do with the logger.
It turns out we call this show_json() method in a number of places but it no longer ever does anything.
So kill it.

neuro_san/graph/tools/external_tool.py
```diff
@@ -130,7 +130,10 @@ async def send_input(self, agent_input: str, sly_data: Dict[str, Any]):
         # Set up a request
         chat_request = {
             "session_id": self.session_id,
-            "user_input": agent_input
+            "user_message": {
+                "type": 1,          # HUMAN from chat.proto
+                "text": agent_input
+            }
```
d1donlydfink, line 136: Use the new request for calling external tools for agent webs.

neuro_san/service/agent_main_loop.py
```diff
@@ -174,7 +177,7 @@ def main_loop(self):
         # Start up the background thread which will process asyncio stuff
         self.asyncio_executor.start()
 
-        grpc_server = AgentServer(self.port,
+        self.server = AgentServer(self.port,
```
d1donlydfink, line 180: Keep track of the server we create. More on this below.

neuro_san/service/agent_main_loop.py
```diff
@@ -183,14 +186,24 @@ def main_loop(self):
...
 
+        # Report back on service activity so the ServerLifetime that calls
+        # this method can properly yield/sleep depending on how many requests
+        # are in motion.
+        agent_services: List[AgentService] = self.server.get_services()
+        for agent_service in agent_services:
+            if agent_service.get_request_count() > 0:
+                return True
+
+        return False
```
d1donlydfink, line 205: OK while bringing up the streaming for the first time I noticed that the streaming was way slower than I expected it to be.  It turned out to be an interaction between the asynchronous event loop and the loop that calls this method, which had been sleep()-ing for 5 seconds periodically and thus, throttling the streamed output to once every 5 seconds.

To address that, I added this reporting mechanism which basically tells that loop "Yield the CPU but don't wait for 5 seconds when you've got something to do" (That is the True return value here).  When there's nothing to do, it goes back to sleeping every 5 seconds.

The result is that all interactions (save the sleeping inherent in the polling) should be snappier now
and the streaming is especially nice and fluid.

neuro_san/service/agent_server.py
```diff
@@ -96,6 +97,13 @@ def __init__(self, port: int,
         self.server_name_for_logs: str = server_name_for_logs
         self.request_limit: int = request_limit
         self.service_prefix: str = service_prefix
+        self.services: List[AgentService] = []
+
+    def get_services(self) -> List[AgentService]:
+        """
+        :return: A list of the AgentServices being served up by this instance
+        """
+        return self.services
```
d1donlydfink, line 106: Some plumbing laid so that the AgentMainLoop can report back about request activity.

neuro_san/service/agent_server.py
```diff
@@ -123,6 +131,8 @@ def serve(self):
                                    self.asyncio_executor,
                                    agent_name,
                                    tool_registry)
+            self.services.append(service)
```
d1donlydfink, line 134: Keep track of all the Agent service handlers we spin up.
(Recall each agent served by the single AgentServer has its own grpc routing address - that is what is going on here).

neuro_san/service/agent_service.py
```diff
@@ -80,6 +82,13 @@ def __init__(self,
         self.asyncio_executor: AsyncioExecutor = asyncio_executor
         self.tool_registry: AgentToolRegistry = tool_registry
         self.agent_name: str = agent_name
+        self.request_counter = AtomicCounter()
```
d1donlydfink, line 85: Use an AtomicCounter (new to leaf-server-common) to keep track of how many requests are in motion.
This helps the sleep vs. yield debate inside ServerLifetime and gives us better overall performance.

neuro_san/service/agent_service.py
```diff
@@ -93,6 +102,7 @@ def Function(self, request: service_messages.FunctionRequest,
         :param context: a grpc.ServicerContext
         :return: a FunctionResponse
         """
+        self.request_counter.increment()
```
d1donlydfink, line 105: For every service method, increment the counter upon entry ...

neuro_san/service/agent_service.py
```diff
@@ -121,6 +131,7 @@ def Function(self, request: service_messages.FunctionRequest,
         if request_log is not None:
             self.request_logger.finish_request(f"{self.agent_name}.Function", log_marker, request_log)
 
+        self.request_counter.decrement()
```
d1donlydfink, line 134: ... and decrement it on exit.

neuro_san/service/agent_service.py
```diff
@@ -285,16 +302,25 @@ def StreamingChat(self, request: service_messages.ChatRequest,
                                      asyncio_executor=self.asyncio_executor,
                                      metadata=metadata,
                                      security_cfg=self.security_cfg)
-        response_dict = session.streaming_chat(request_dict)
-        response_dict["request"] = request_dict
+        response_dict_iterator: Iterator[Dict[str, Any]] = session.streaming_chat(request_dict)
```
d1donlydfink, line 305: OK this is where the rubber hits the road for streaming results.
We get back a synchronous Iterator from the DirectAgentSession.streaming_chat() method,
which yields us the chat messages as they come in...

neuro_san/service/agent_service.py
```diff
@@ -285,16 +302,25 @@ def StreamingChat(self, request: service_messages.ChatRequest,
...
 
-        # Convert the response dictionary to a grpc message
-        # DEF - this will get better
-        response_string = json.dumps(response_dict)
-        response = service_messages.ChatResponse()
-        Parse(response_string, response)
+        for response_dict in response_dict_iterator:
```
d1donlydfink, line 307: Use the classic for-loop construction with the iterator above, but note that when it goes to get its next() response_dict here, it can synchronously block.

neuro_san/service/agent_service.py
```diff
@@ -285,16 +302,25 @@ def StreamingChat(self, request: service_messages.ChatRequest,
...
+            # Convert the response dictionary to a grpc message
+            response_string = json.dumps(response_dict)
+            response = service_messages.ChatResponse()
+            Parse(response_string, response)
+
+            # Yield-ing a single response allows one response to be returned
+            # over the connection while keeping it open to wait for more.
+            # Grpc client code handling response streaming knows to construct an
+            # iterator on its side to do said waiting over there.
+            yield response
```
d1donlydfink, line 319: So because we are at grpc-level here we have to go through a layer of conversion to protobuf messages.
But we still want to yield our results for grpc streaming.  To do that, we simply do our conversion and yield that results, and voila, this method ends up a returning a Generator which is what grpc streaming is based on anyway.

neuro_san/service/agent_service.py
```diff
@@ -285,16 +302,25 @@ def StreamingChat(self, request: service_messages.ChatRequest,
...
+
+        # Iterator has finally signaled that there are no more responses to be had.
+        # Log that we are done.
         if request_log is not None:
             self.request_logger.finish_request(f"{self.agent_name}.StreamingChat", log_marker, request_log)
 
-        return response
+        self.request_counter.decrement()
```
d1donlydfink, line 326: Worth pointing out that even though the yield above is "returning" a single result via a generator,
the flow of control has still not officially exited this method.  Lucky for us because we still have
some business to finish up before we get out of here on the service side.

When the streaming ends here on the service-side, grpc magic also makes sure the iterator for the stream on the client side is done, and the client-side for-loop also breaks out when it gets the last yielded message from here and this method returns.

neuro_san/session/direct_agent_session.py
```diff
@@ -122,7 +128,12 @@ def chat(self, request_dict: Dict[str, Any]) -> Dict[str, Any]:
             Note that responses to the chat input are asynchronous and come by polling the
             logs() method below.
         """
+        extractor = DictionaryExtractor(request_dict)
+
+        # Get the user input. Prefer the newer-style from the user_message
         user_input: str = request_dict.get("user_input")
+        user_input = extractor.get("user_message.text", user_input)
```
d1donlydfink, line 135: Prefer the user_input from the new-style request, but let the old-style request be ok as well.

neuro_san/session/direct_agent_session.py
```diff
@@ -147,9 +158,11 @@ def chat(self, request_dict: Dict[str, Any]) -> Dict[str, Any]:
             # Create an asynchronous background task to process the user input.
             # This might take a few minutes, which can be longer than some
             # sockets stay open.
-            self.asyncio_executor.submit(session_id, chat_session.chat, user_input, sly_data)
+            future: Future = self.asyncio_executor.submit(session_id, chat_session.chat,
+                                                          user_input, sly_data)
+            _ = future
```
d1donlydfink, line 163: An exercise in proper typing so I could orient myself as to how to handle the submit() for streaming chat later on. Even though we don't actually care about the Future here, at least we are consistent with out synchronous use of it.

neuro_san/session/direct_agent_session.py
```diff
@@ -281,7 +296,67 @@ def streaming_chat(self, request_dict: Dict[str, Any]) -> Generator[Dict[str, An
             Note that responses to the chat input might be numerous and will come as they
             are produced until the system decides there are no more messages to be sent.
         """
-        raise NotImplementedError
+        extractor = DictionaryExtractor(request_dict)
+
+        # Get the user input. Prefer the newer-style from the user_message
+        user_input: str = request_dict.get("user_input")
+        user_input = extractor.get("user_message.text", user_input)
```
d1donlydfink, line 303: Also prefer the new style request in the business end of the DirectAgentSession.streaming_chat() method.

neuro_san/session/direct_agent_session.py
```diff
@@ -281,7 +296,67 @@ def streaming_chat(self, request_dict: Dict[str, Any]) -> Generator[Dict[str, An
...
+                status = self.CREATED
+                chat_session = DataDrivenChatSession(registry=self.tool_registry)
+                session_id = self.chat_session_map.register_chat_session(chat_session)
+            else:
+                # We got an session_id, but this service instance has no knowledge
+                # of it.
+                status = self.NOT_FOUND
+        else:
+            # We have seen this session_id before and can register new user input.
+            status = self.FOUND
```
d1donlydfink, line 322: Same business for streaming chat about finding sessions.
When we get to returning chat history as part of the streaming response, we won't need this session business any more.

neuro_san/session/direct_agent_session.py
```diff
@@ -281,7 +296,67 @@ def streaming_chat(self, request_dict: Dict[str, Any]) -> Generator[Dict[str, An
...
+        template_response_dict = {
+            "session_id": session_id,
+            "status": status
+        }
+
+        if chat_session is None or user_input is None:
+            # Can't go on to chat, so report back early with a single value.
+            # There is no ChatMessage response in the dictionary in this case
+            yield template_response_dict
+            return
```
d1donlydfink, line 334: Handle a case where there is no session or input.

neuro_san/session/direct_agent_session.py
```diff
@@ -281,7 +296,67 @@ def streaming_chat(self, request_dict: Dict[str, Any]) -> Generator[Dict[str, An
...
+
+        # Create an asynchronous background task to process the user input.
+        # This might take a few minutes, which can be longer than some
+        # sockets stay open.
+        future: Future = self.asyncio_executor.submit(session_id, chat_session.streaming_chat,
+                                                      user_input, sly_data)
+        # Ignore the future. Live in the now.
+        _ = future
```
d1donlydfink, line 342: Actually call streaming_chat() on the ChatSession.
This gets the front man rolling.

As discussed waaaay above, we don't really care about a return value here from when streaming_chat ends.
It is being set into motion in a separate thread that handles the asynchronous event loop...

neuro_san/session/direct_agent_session.py
```diff
@@ -281,7 +296,67 @@ def streaming_chat(self, request_dict: Dict[str, Any]) -> Generator[Dict[str, An
...
+
+        # The chat_session.queue_consumer() method will hang out waiting for chat.ChatMessage
+        # dictionaries to come back asynchronously from the submit() above until there
+        # are no more from the input.
+        empty: Dict[str, Any] = {}
+        generator = AsyncToSyncGenerator(self.asyncio_executor, session_id,
+                                         generated_type=Dict,
+                                         keep_alive_result=empty,
+                                         keep_alive_timeout_seconds=10.0)
+        for message in generator.synchronously_generate(chat_session.queue_consumer):
```
d1donlydfink, line 352: ... what we do care about here in synchronous land is pulling the results off the queue via its asynchronous iterator.

Now. This part looks easy here, but this was hard-won sweaty coding here.

The for-loop above is set up to *synchronously* iterate (we are in synchronous land here) over an
*asynchronous* iterator.   You'd think there'd be plenty of StackOverflow or pypi libs out there to do something like this, but you'd be sorely mistaken.   All the guts and glory of how this translation is sequestered in that AsyncToSyncGenerator class so that this part looks simple: "Just" call the synchronous method that takes an asynchronous method returning a generator.

neuro_san/session/direct_agent_session.py
```diff
@@ -281,7 +296,67 @@ def streaming_chat(self, request_dict: Dict[str, Any]) -> Generator[Dict[str, An
...
+
+            response_dict: Dict[str, Any] = copy(template_response_dict)
+            if any(message):
+                response_dict["response"] = message
+
+            # We expect the message to be a dictionary form of chat.ChatMessage
+            yield response_dict
```
d1donlydfink, line 359: Before we yield, we have to add all the extra response business that we figured out before to each chat response dictionary.  Do that and yield ... synchronously ... so as to feed the generator/iterator that feeds the grpc streaming.

neuro_san/session/service_agent_session.py
```diff
@@ -78,7 +79,8 @@ def __init__(self, host: str = None,
                                         service_stub,
                                         use_host, use_port,
                                         timeout_in_seconds, metadata,
-                                        security_cfg, umbrella_timeout)
+                                        security_cfg, umbrella_timeout,
+                                        streaming_timeout_in_seconds)
```
d1donlydfink, line 83: Pass through a streaming timeout to AbstractServiceSession.
This timeout is only used for streaming calls, which have different lifetimes in grpc land than simple unary responses do.  As such, we can't have our own retry logic which times out after 30 seconds coming around and killing our streaming connection for results of 37 agents which takes 5 minutes to complete.  So, we set that timeout to None, which tells our leaf-common grpc infrastructure to not kill things with premature deadlines.

neuro_san/session/service_agent_session.py
```diff
@@ -236,7 +238,7 @@ def streaming_chat(self, request_dict: Dict[str, Any]) -> Generator[Dict[str, An
             "streaming_chat",
             self._streaming_chat_from_stub,
             request_dict,
-            service_messages.ChatRequest(),
+            request_instance=service_messages.ChatRequest(),
```
d1donlydfink, line 241: Be a little clearer on the args

neuro_san/utils/agent_framework_message.py
```diff
@@ -0,0 +1,25 @@
...
+from typing import Literal
+from typing import Union
+
+from langchain_core.messages.base import BaseMessage
+
+
+class AgentFrameworkMessage(BaseMessage):
+    """
+    BaseMessage implementation of a message from the agent framework
+    """
```
d1donlydfink, line 13: We create two new kinds of BaseMessage classes which interoperate with the langchain messaging hierarchy.
The idea is to have different kinds of messages to indicate where they originate from within multi-agent land.
This one is for messages that we report to the client that originate inside the agent framework.

neuro_san/utils/agent_message.py
```diff
@@ -0,0 +1,25 @@
...
+from typing import Literal
+from typing import Union
+
+from langchain_core.messages.base import BaseMessage
+
+
+class AgentMessage(BaseMessage):
+    """
+    BaseMessage implementation of a message from an agent
+    """
```
d1donlydfink, line 13: This message (which we don't use quite yet) is for messages that originate inside the agents themselves.
Eventually we can use these guys for messaging the "Should I use a tool?" kind of agent-mind messages
which do not actually come from the LLM.

neuro_san/utils/message_utils.py
```diff
@@ -87,6 +92,16 @@ def generate_response(the_messages: List[Any]) -> str:
...
 
 
+MESSAGE_TYPE_TO_ROLE: Dict[Type[BaseMessage], str] = {
+    AIMessage: "assistant",
+    HumanMessage: "user",
+    ToolMessage: "tool",
+    SystemMessage: "system",
+    AgentMessage: "agent",
+    AgentFrameworkMessage: "agent-framework",
+}
```
d1donlydfink, line 102: Translation of message class to a string role.

neuro_san/utils/message_utils.py
```diff
@@ -96,14 +111,10 @@ def get_role(message: Any) -> str:
...
-    if isinstance(message, HumanMessage):
-        return "user"
-    if isinstance(message, ToolMessage):
-        return "tool"
-    if isinstance(message, SystemMessage):
-        return "system"
+    # Check the look-up table above
+    role: str = MESSAGE_TYPE_TO_ROLE.get(type(message))
+    if role is not None:
+        return role
```
d1donlydfink, line 117: Use the lookup table above instead of the unscalable if-s.

neuro_san/utils/message_utils.py
```diff
@@ -123,3 +134,34 @@ def get_content(message: Any) -> str:
...
+MESSAGE_TYPE_TO_CHAT_MESSAGE_TYPE: Dict[Type[BaseMessage], int] = {
+    # Needs to match chat.proto
+    HumanMessage: 1,
+    SystemMessage: 2,
+    AIMessage: 3,
+    ToolMessage: 4,
+
+    AgentMessage: 100,
+    AgentFrameworkMessage: 101,
+}
```
d1donlydfink, line 148: Translation of message class to a grpc integer type. These all come from chat.proto.

neuro_san/utils/message_utils.py
```diff
@@ -123,3 +134,34 @@ def get_content(message: Any) -> str:
...
+
+    unknown: int = 0
+    chat_message: Dict[str, Any] = {
+        "type": MESSAGE_TYPE_TO_CHAT_MESSAGE_TYPE.get(type(message), unknown),
+        "text": message.content,
+        # No mime_data for now
+        # No origin for now
+    }
+
+    return chat_message
```
d1donlydfink, line 167: New conversion method to the new style ChatMessage.

requirements-private.txt
```diff
@@ -12,4 +12,4 @@
 git+https://${LEAF_SOURCE_CREDENTIALS}@github.com/leaf-ai/leaf-common.git@1.2.14
 
 # Private repos
-git+https://${LEAF_PRIVATE_SOURCE_CREDENTIALS}@github.com/leaf-ai/leaf-server-common.git@0.1.10
+git+https://${LEAF_PRIVATE_SOURCE_CREDENTIALS}@github.com/leaf-ai/leaf-server-common.git@0.1.13
```
d1donlydfink, line 15: Get asynchronous goodness from latest leaf-server-common.

neuro_san/chat/chat_session.py
ofrancon: It would be nice to have a :return in the doc string that explain what the Iterator contains. It's not obvious to me what the key and values are in the returned dictionaries. At this high level I was expecting a list of messages. I guess messages are the values in the dict, but then what are the keys? The id of the agent the message is directed to?

ofrancon: Reading further I guess the iterator returns a list of "chat messages". A chat message is a dictionary that contains a message type and a message content. Would be nice to explain a bit. Here is probably a good place, in this docstring, because it's the main / highest level API.

ofrancon, line 33: ** Not something to change for this PR, just thoughts **
Now that I understand a bit more what this `chat` function return, I feel there's an asymmetry between the input and output of this method:
1. Should the input be a "chat message" too, i.e. a dictionary? I guess it could be useful for an agent to know if the message is coming from a human or another agent.
2. Should the input be a List of messages, similarly to the output? I guess in theory I could send several messages to an agent before the agent processes them (Could be handled by a queue). Or I could bundle the messages myself, especially if they are of different mime type. In which case the input of the chat method would have to be a List of chat messages, instead of just `str`.

neuro_san/chat/data_driven_chat_session.py
```diff
@@ -109,17 +116,111 @@ async def chat(self, user_input: str, sly_data: Dict[str, Any]):
...
+        """
+        Filter chat messages from the full logs that are in/appropriate for streaming.
+
+        :param chat_message: A Chat message dictionary of the form in chat.ChatMessage proto
+        :param index: The place in the chat history the message has
+        :return: True if the given message should be streamed back. False if not.
+        """
+
+        # This comes from definitions in chat.proto
+        ai_message_type: int = 3
```
ofrancon, line 181: 3 is a magic number here. Can we use an enum instead?

neuro_san/client/agent_cli.py
ofrancon: Makes sense

neuro_san/client/agent_cli.py
```diff
@@ -179,7 +229,10 @@ def send_user_input(self, user_input: str, sly_data: Dict[str, Any]):
         if user_input != self.default_input:
             chat_request = {
                 "session_id": self.session_id,
-                "user_input": user_input
+                "user_message": {
+                    "type": 1,      # HUMAN from chat.proto
```
ofrancon, line 233: Magic number - can we use the chat.proto enum directly? If not, can we replicate it in a Python enum?

neuro_san/session/direct_agent_session.py
ofrancon: ...

neuro_san/session/service_agent_session.py
ofrancon: Does it mean we should start thinking of messages from the user like:
- "Where's the report I asked you 5 mins ago?"
and the main agent answers something like:
"We're still working on it. We've gathered the raw data and we're waiting for Bob to be done plotting nice graphs"
?

neuro_san/utils/message_utils.py
ofrancon: Could be a python `IntEnum`
I think you can make it a class and call it `MessageType`:
```
from enum import IntEnum
class MessageType(IntEnum):
    HUMAN = 1
    SYSTEM = 2
    AI = 3
    TOOL = 4
    AGENT = 100
    AGENT_FRAMEWORK = 101
```
And use it everywhere as a replacement for `int`. It still has to match chat.proto.
Then MessageType.HUMAN == 1 will return true

neuro_san/chat/chat_session.py
d1donlydfink: Fair point on the Iterator vs List confusion.  I can add such a doc in a later PR.

neuro_san/session/service_agent_session.py
d1donlydfink: Very good question.

In the previous API where chat() sent the message and logs() retrieved the responses via polling, the responsibility for saying "I'm working on that" resided within the service itself - and that code still exists for now.

In the streaming world, that responsibility gets pushed to the client.  It's will be the case with streaming that the client that knows that the results stream is still open and has to deal with the user clicking away when nothing is happening.  

That all said, there will be a much snappier and more robust stream of chat message feedback coming back as it happens for the user to get the sense "It's doing something for me".  But since the responsibility for user impatience is now on the client side, it's there that the question of "Gee, I haven't seen any kind of message for the past 30 seconds or so" needs to be addressed.

neuro_san/utils/message_utils.py
d1donlydfink: Yes. This is in the works in the Afterthought PR. I will put you on that.

neuro_san/chat/chat_session.py
d1donlydfink: 0.  You are commenting on the chat() interface, which is still set up to handle older-school polling as its basis.  But really you want to talk about the interface on streaming_chat() on the same class.

1. Yes. I think you are correct in that eventually the newer method *streaming_chat()*  should take a ChatMessage dictionary at this level.   I can repair that soon.  Note that the DirectAgentSession actually does this already and for now I do not see this ChatSession as an external-facing API (The client code uses DirectAgentSession, for instance).   That all said, ultimately the single-string input will not cut it with agents in the media-oriented world (think Dallee in the mix)

2. I'm not sure where you are going here is necessarily the best direction on a couple of fronts:
* Any List of text messages from the user can simply be compressed into a single message
* A user barraging the system with multiple messages actually takes away from the conversationally of the front-man agent.
* In my Protocol Afterthoughts PR I have the mime_data in there as a repeated field, which will allow for multiple media entities in messages.
* Lists of Messages are really more appropriate for a chat history to give context to a conversation.  I have more plans there which are perhaps better talked about with others present. In short: I think the whole stateful notion of SessionId can go away with a more complex notion of chat history going over the wire. I think there are other useful behaviors this ends up enabling besides better scalability.

neuro_san/chat/data_driven_chat_session.py
d1donlydfink: Yes. See the Afterthoughts PR.

neuro_san/client/agent_cli.py
d1donlydfink: See Afterthoughts PR for use of the enum.
(But thanks, you found a place I missed! Exactly why this should be an Enum to begin with)

neuro_san/chat/chat_session.py
ofrancon: I was indeed thinking about the `streaming_chat()` API
I was just thinking of a 'conversation like that:
- text message: USER, "in the following image, how many people stand up?"
- image message: USER, [bytes]
And the image can be sent a few seconds later, the agent patiently waiting for the image before trying to answer. Waiting either because the user is slow to send it, or because it takes several seconds to upload it in a stream.

I'm not saying we should support this use case, I was just wondering if that's what you had in mind with the `streaming_chat()` API.

PR #34, opened 2024-12-27
UN-2494 Get direct streaming client to work

d1donlydfink:
A number of changes to make the direct version of the agent_cli work with streaming clients.

* Create a new AsyncCollatingQueue to encapsulate the policy of blocking async iteration over the message queue contents for the session.
* Use that new class's iterator aspect as the basis for how we do our async-to-sync iteration translation with fixes in leaf-server-common

### DIFF
--- a/requirements-private.txt
+++ b/requirements-private.txt
@@ -12,4 +12,4 @@
 git+https://${LEAF_SOURCE_CREDENTIALS}@github.com/leaf-ai/leaf-common.git@1.2.14
 
 # Private repos
-git+https://${LEAF_PRIVATE_SOURCE_CREDENTIALS}@github.com/leaf-ai/leaf-server-common.git@0.1.13
+git+https://${LEAF_PRIVATE_SOURCE_CREDENTIALS}@github.com/leaf-ai/leaf-server-common.git@0.1.14


neuro_san/chat/chat_session.py
```diff
@@ -52,13 +52,14 @@ async def streaming_chat(self, user_input: str, sly_data: Dict[str, Any]):
...
         """
         raise NotImplementedError
 
-    async def queue_consumer(self) -> AsyncIterator[Dict[str, Any]]:
+    def get_queue_iterator(self) -> AsyncQueueIterator:
         """
-        :return: An AsyncIterator over the messages from the agent(s).
+        :return: The AsyncQueueIterator associated with this ChatSession instance.
         """
         raise NotImplementedError
```
d1donlydfink, line 64: Change the way we do the consumption of the queue to use the new AsyncCollatingQueue

neuro_san/chat/data_driven_chat_session.py
```diff
@@ -193,35 +201,6 @@ def is_streamable_message(self, chat_message: Dict[str, Any], index: int) -> boo
...
 
-    async def queue_consumer(self) -> AsyncIterator[Dict[str, Any]]:
-        """
-        Queue Consumer from this:
-            https://stackoverflow.com/questions/74130544/asyncio-yielding-results-from-multiple-futures-as-they-arrive
-
-        Loops until either the timeout is met or the end marker is seen.
-
-        :return: An AsyncIterator over the messages from the agent(s).
-        """
```
d1donlydfink, line 204: Not doing it this way any more.
We are using the AsyncQueueIterator now.

neuro_san/client/agent_cli.py
```diff
@@ -323,6 +322,7 @@ def stream_once(self, state: Dict[str, Any]) -> Dict[str, Any]:
         if sly_data is not None and len(sly_data.keys()) > 0:
             chat_request["sly_data"] = sly_data
 
+        return_state: Dict[str, Any] = {}
```
d1donlydfink, line 325: Fix a complaint when running the app by defining return_state outside the iteration loop.

neuro_san/client/agent_cli.py
```diff
@@ -336,16 +336,14 @@ def stream_once(self, state: Dict[str, Any]) -> Dict[str, Any]:
...
                 print(f"Response: {text}")
-                prompt = self.default_prompt
                 last_chat_response = text
 
-            return_state: Dict[str, Any] = {
+            return_state = {
                 "last_logs": last_logs,
                 "last_chat_response": last_chat_response,
-                "prompt": prompt,
+                "prompt": self.default_prompt,
```
d1donlydfink, line 346: Always return the default prompt as every time this guy finishes we have effectively finished the conversation.

neuro_san/registries/esp_decision_assistant.hocon
```diff
@@ -82,6 +82,7 @@ As part of your query, you will declare the needed expertise, and the
 type and values of the context as well as defining the action and outcome
 attributes.
 The prescriptor function will return the list of action values for the given context.
+When communicating to tools, never use curly braces to describe choices.
```
d1donlydfink, line 85: A minor tweak to the prompting here for esp_decision_assistant to reduce the probability of KeyErrors that arise from the bots trying to explain the choices to the tools in plain english that happens to contain curly-braces.  The curly-braces are a clue to the rest of the system/langchain to look for coded variables which are not actually there.

Needless to say, when testing, I ran into this enough times that it became annoying.

neuro_san/chat/async_collating_queue.py
```diff
@@ -0,0 +1,76 @@
...
+from typing import Dict
+
+from asyncio.queues import Queue
+
+
+class AsyncCollatingQueue(AsyncIterator):
+    """
+    AsyncIterator instance to asynchronously iterate over/consume the contents of
+    a Queue as they come in.
+    """
```
d1donlydfink, line 13: A new class to deal with asynchronously iterating over the results of the message queue for a ChatSession.

neuro_san/chat/async_collating_queue.py
```diff
@@ -0,0 +1,76 @@
...
+        :return: The Queue associated with this instance
+        """
+        return self.queue
+
+    def __aiter__(self) -> AsyncIterator[Dict[str, Any]]:
+        """
+        Self-identify as an AsyncIterator when called upon by
+        the Python async framework.
+        """
+        return self
```
d1donlydfink, line 42: Per lots of googling, this seems to be standard practice to include in these kinds of iterators.

neuro_san/chat/async_collating_queue.py
```diff
@@ -0,0 +1,76 @@
...
+        """
+        await self.put(self.END_MESSAGE)
+
+    def is_final_item(self, item: Any) -> bool:
+        """
+        :param item: An item that has just been pulled off the queue
+        :return: True if this item is considered the marker for the
+                 end of data. False otherwise.
+        """
+        return isinstance(item, Dict) and item.get(self.END_KEY) is not None
```
d1donlydfink, line 76: Use the END_MESSAGE contents to detect when we need to StopAsyncIteration.

These put_final_item() and is_final_item() methods are put together in such a way that if we wanted to 
abstract this further, they would be the purpose-specific methods on a more abstract class.

neuro_san/chat/data_driven_chat_session.py
```diff
@@ -60,12 +59,18 @@ def __init__(self, registry: AgentToolRegistry,
...
         self.last_streamed_index: int = 0
 
         if setup:
             self.set_up()
 
+    def get_queue(self) -> AsyncCollatingQueue:
+        """
+        :return: The AsyncCollatingQueue associated with this ChatSession instance.
+        """
+        return self.queue
```
d1donlydfink, line 72: Return our AsyncCollatingQueue when requested so the DirectSesion can iterate over its contents.

neuro_san/session/direct_agent_session.py
```diff
@@ -341,15 +341,15 @@ def streaming_chat(self, request_dict: Dict[str, Any]) -> Generator[Dict[str, An
...
+        # The queue_iterator() method below will hang out waiting for chat.ChatMessage
         # dictionaries to come back asynchronously from the submit() above until there
         # are no more from the input.
         empty: Dict[str, Any] = {}
         generator = AsyncToSyncGenerator(self.asyncio_executor, session_id,
                                          generated_type=Dict,
                                          keep_alive_result=empty,
                                          keep_alive_timeout_seconds=10.0)
-        for message in generator.synchronously_generate(chat_session.queue_consumer):
+        for message in generator.synchronously_iterate(chat_session.get_queue()):
```
d1donlydfink, line 352: Use the new AsyncQueueIterator to stream the results back to the client.

neuro_san/chat/async_collating_queue.py
```diff
@@ -0,0 +1,76 @@
...
+        """
+        :return: Blocks waiting to return the next item on the queue.
+                Will throw StopAsyncIteration when the final item is detected
+                via the is_final_item() method..
+        """
+        message = await self.queue.get()
+        if self.is_final_item(message):
+            raise StopAsyncIteration
+
+        return message
```
d1donlydfink, line 54: This is the magic method that will power any iterator used for pulling items off the queue.

neuro_san/chat/data_driven_chat_session.py
```diff
@@ -60,12 +59,18 @@ def __init__(self, registry: AgentToolRegistry,
         self.latest_response = None
         self.last_input_timestamp = datetime.now()
         self.sly_data: Dict[str, Any] = {}
-        self.queue: Queue[Dict[str, Any]] = Queue()
+        self.queue: AsyncCollatingQueue = AsyncCollatingQueue()
```
d1donlydfink, line 62: Use the new AsyncCollatingQueue instead of the simpler Queue we had before.
This guy encapsulates the way we want to do the Iterator over the messages on the Consumer side.
From the Producer side, we just put() stuff on it when/wherever we want.

neuro_san/chat/data_driven_chat_session.py
```diff
@@ -165,8 +171,7 @@ async def streaming_chat(self, user_input: str, sly_data: Dict[str, Any]):
 
         # Put an end-marker on the queue to tell the consumer we truly are done
         # and it doesn't need to wait for any more messages.
-        end_dict = {"end": True}
-        await self.queue.put(end_dict)
+        await self.queue.put_final_item()
```
d1donlydfink, line 174: Have the notion of an end marker be an implementation detail of the AsyncCollatingQueue.

requirements-private.txt
```diff
@@ -12,4 +12,4 @@
 git+https://${LEAF_SOURCE_CREDENTIALS}@github.com/leaf-ai/leaf-common.git@1.2.14
 
 # Private repos
-git+https://${LEAF_PRIVATE_SOURCE_CREDENTIALS}@github.com/leaf-ai/leaf-server-common.git@0.1.13
+git+https://${LEAF_PRIVATE_SOURCE_CREDENTIALS}@github.com/leaf-ai/leaf-server-common.git@0.1.14
```
d1donlydfink, line 15: Use the latest leaf-server-common with the synchronously_iterate() that is used above.

neuro_san/chat/async_collating_queue.py
ofrancon: Why not
```
return item == END_MESSAGE
```

d1donlydfink: See abstraction message above.  App policy might dictate that it's not so simple.

PR #35, opened 2024-12-27
UN-2494 streaming thinking file

d1donlydfink:
Get the client thinking file to work when streaming via a new LEGACY_LOGS message type in the protocol.
There are 30 files changed, which looks like a lot, but most of it is renames of logger -> journal and import location changes.

Also:
* Add a num_inputs and one_shot arg to the client allowing it to exit after receiving exactly N inputs. This just makes command line testing a little bit easier without having to type "quit" or ctrl-c for test cases all the time.
* Create a Journal interface.  I called it "Journal" because "logger" has low-level Python implications and "ChatHistory" has langchain implications. 🤷 
* Rename StreamForLogger to be TextJournal and have it implement the Journal interface
* Create a MessageJournal which knows how to queue LegacyLogsMessages when the Journal.write() call is made.  This gets the correct backwards-compatible logs information over the single streaming channel.
* Create a CompatibilityJournal which is really a single Journal that will do both TextJournal and MessageJournal operations for any one call to its Journal methods
* Move the messages-oriented stuff to its own package as well.

@dsargent : After this change, this stack of new streaming functionality will be good enough for unileaf consumption.
It is much much snappier than polling via Logs()

### DIFF
--- a/neuro_san/journals/text_journal.py
+++ b/neuro_san/journals/text_journal.py
@@ -28,7 +28,7 @@ class TextJournal(Journal):
 
     async def write(self, entry: Union[str, bytes]):
         """
-        :param entr: Add a string-ish entry to the logs.
+        :param entry: Add a string-ish entry to the logs.
                     Can be either a string or bytes.
         """
         # Decoding bytes to string if necessary


neuro_san/api/grpc/chat.proto
```diff
@@ -33,6 +33,7 @@ message ChatMessage {
         // These are not out-of-the-box langchain messages
         AGENT = 100;
         AGENT_FRAMEWORK = 101;
+        LEGACY_LOGS = 102;      // Deprecated. Used for backwards compatibility
```
d1donlydfink, line 36: Add a LEGACY_LOGS message type to ChatMessage proto so that we can have functional parity between streaming and polling versions of the client.

Eventually when the UI moves over to streaming messages and we are sending better informed ChatMessages over the wire (with proper filtering in place), then we can get rid of this.
Hopefully that can be soon.

neuro_san/chat/async_collating_queue.py
```diff
@@ -5,8 +5,10 @@
 
 from asyncio.queues import Queue
 
+from neuro_san.interfaces.async_hopper import AsyncHopper
 
-class AsyncCollatingQueue(AsyncIterator):
+
+class AsyncCollatingQueue(AsyncIterator, AsyncHopper):
```
d1donlydfink, line 11: This guy fulfills the new AsyncHopper interface so that lower-level packages do not have circular dependencies on this class here in chat.

neuro_san/chat/chat_session.py
```diff
@@ -63,9 +63,9 @@ def get_queue(self) -> AsyncCollatingQueue:
         """
         raise NotImplementedError
 
-    def get_logger(self) -> StreamToLogger:
+    def get_journal(self) -> Journal:
```
d1donlydfink, line 66: Typical example of a file with renaming of StreamToLogger -> Journal usage.

neuro_san/chat/data_driven_chat_session.py
```diff
@@ -38,22 +39,17 @@ class DataDrivenChatSession(ChatSession):
...
-        :param logger: The StreamToLogger that captures messages for user output
+        :param journal: The Journal that captures messages for user output
         :param setup: Whether or not set_up() should be called
                     by the constructor. Default is False.
         """
 
-        use_logger: StreamToLogger = logger
-        if logger is None:
-            use_logger = StreamToLogger()
-        self.logger: StreamToLogger = use_logger
```
d1donlydfink, line 55: This initialization of the journal (nee logger) moves down below so that the queue can be initialized first to pass in to the new queue-ing journaling capability.

neuro_san/chat/data_driven_chat_session.py
```diff
@@ -24,10 +24,11 @@
...
 from neuro_san.graph.tools.front_man import FrontMan
-from neuro_san.utils.message_utils import convert_to_chat_message
-from neuro_san.utils.message_utils import pretty_the_messages
-from neuro_san.utils.agent_framework_message import AgentFrameworkMessage
-from neuro_san.utils.stream_to_logger import StreamToLogger
+from neuro_san.journals.compatibility_journal import CompatibilityJournal
+from neuro_san.journals.journal import Journal
+from neuro_san.messages.agent_framework_message import AgentFrameworkMessage
+from neuro_san.messages.message_utils import convert_to_chat_message
+from neuro_san.messages.message_utils import pretty_the_messages
```
d1donlydfink, line 31: Also we move the messages stuff to its own package.

neuro_san/chat/data_driven_chat_session.py
```diff
@@ -62,6 +58,10 @@ def __init__(self, registry: AgentToolRegistry,
         self.queue: AsyncCollatingQueue = AsyncCollatingQueue()
         self.last_streamed_index: int = 0
 
+        self.journal: Journal = journal
+        if journal is None:
+            self.journal = CompatibilityJournal(self.queue)
```
d1donlydfink, line 63: This moves down here from above.

neuro_san/chat/data_driven_chat_session.py
```diff
@@ -75,9 +75,9 @@ async def set_up(self):
         """
         Resets or sets the instance up for the first time.
         """
-        # Reset the logger
-        self.logger = StreamToLogger()
-        self.logger.write("setting up chat agent(s)...")
+        # Reset the journal
+        self.journal = CompatibilityJournal(self.queue)
+        await self.journal.write("setting up chat agent(s)...")
```
d1donlydfink, line 80: We need to await on the write() call now. More on that later.

neuro_san/client/agent_cli.py
```diff
@@ -84,11 +85,12 @@ def main(self):
             "last_chat_response": None,
             "prompt": self.default_prompt,
             "timeout": self.input_timeout_seconds,
+            "num_input": 0,
```
d1donlydfink, line 88: initialize how many times we have sent input.

neuro_san/client/agent_cli.py
```diff
@@ -84,11 +85,12 @@ def main(self):
...
             "user_input": user_input,
             "sly_data": sly_data,
         }
 
-        while state.get("user_input") != "quit":
+        while not self.is_done(state):
```
d1donlydfink, line 93: Move the decision to exit the loop to its own method so more policy can be added as we need.

neuro_san/client/agent_cli.py
```diff
@@ -168,6 +170,10 @@ def add_args(self, arg_parser: argparse.ArgumentParser):
                                 help="Use a service connection")
         arg_parser.add_argument("--direct", dest="connection", action="store_const", const="direct",
                                 help="Use a direct/library call for the chat")
+        arg_parser.add_argument("--max_input", type=int, default=1000000,
+                                help="Maximum rounds of input to go before exiting")
+        arg_parser.add_argument("--one_shot", dest="max_input", action="store_const", const=1,
+                                help="Send one round of input, then exit")
```
d1donlydfink, line 176: New CLI arguments allow for doing limit rounds of input.
This ends up being a small but immensely convenient thing for me w/rt how I test client/server changes like this.

neuro_san/client/agent_cli.py
```diff
@@ -202,6 +208,21 @@ def get_agent_session_factory(self) -> AgentSessionFactory:
         """
         return AgentSessionFactory()
 
+    def is_done(self, state: Dict[str, Any]) -> bool:
+        """
+        :param state: The state dictionary
+        :return: True if the values in the state are considered to be sufficient
+                for terminating further conversation. False otherwise.
+        """
```
d1donlydfink, line 216: New policy method for exiting the main loop

neuro_san/client/agent_cli.py
```diff
@@ -202,6 +208,21 @@ def get_agent_session_factory(self) -> AgentSessionFactory:
...
+
+        if state.get("user_input") == "quit":
+            return True
+
+        if state.get("num_input") >= self.args.max_input:
+            return True
```
d1donlydfink, line 222: We still look for "quit" from the user, but we are also looking for the input count as well.

neuro_san/client/agent_cli.py
```diff
@@ -277,7 +299,8 @@ def get_responses(self, state: Dict[str, Any]) -> Dict[str, Any]:
                 # Might be more than 1
                 for i in range(len(last_logs), len(logs)-1):
                     thinking.write(logs[i])
-            last_logs = logs
+                    thinking.write("\n")
```
d1donlydfink, line 302: Add a newline between thinking file log entries - easier to read.

neuro_san/client/agent_cli.py
```diff
@@ -277,7 +299,8 @@ def get_responses(self, state: Dict[str, Any]) -> Dict[str, Any]:
...
+            last_logs = copy(logs)
```
d1donlydfink, line 303: This one line change actually fixes a bug I introduced in the client polling logic with my other stack of changes.

neuro_san/client/agent_cli.py
```diff
@@ -287,15 +310,18 @@ def get_responses(self, state: Dict[str, Any]) -> Dict[str, Any]:
             prompt = self.default_prompt
             timeout = self.input_timeout_seconds
             last_chat_response = chat_response
+            num_input += 1
```
d1donlydfink, line 313: Increment the number of inputs when polling.
How/when we do this is different between polling and streaming.

neuro_san/client/agent_cli.py
```diff
@@ -333,18 +360,25 @@ def stream_once(self, state: Dict[str, Any]) -> Dict[str, Any]:
...
             if text is not None:
-                print(f"Response: {text}")
-                last_chat_response = text
+                if message_type == 102:     # LEGACY_LOGS from chat.proto
+                    with open(self.args.thinking_file, "a", encoding="utf-8") as thinking:
+                        thinking.write(text)
+                        thinking.write("\n")
+                else:
+                    print(f"Response: {text}")
+                    last_chat_response = text
```
d1donlydfink, line 374: Filter where the streamed response goes based on the message type.
In a future PR I'll be introducing a whole message filtering infrastructure, but for now this will get streaming up to functional parity with polling.

neuro_san/client/agent_cli.py
```diff
@@ -333,18 +360,25 @@ def stream_once(self, state: Dict[str, Any]) -> Dict[str, Any]:
...
 
             return_state = {
                 "last_logs": last_logs,
                 "last_chat_response": last_chat_response,
                 "prompt": self.default_prompt,
-                "timeout": self.input_timeout_seconds
+                "timeout": self.input_timeout_seconds,
+                "num_input": num_input + 1
```
d1donlydfink, line 381: Increment number of inputs for streaming side.

neuro_san/graph/registry/agent_tool_registry.py
```diff
@@ -150,7 +150,7 @@ def get_agent_tool_spec(self, name: str) -> Dict[str, Any]:
 
     # pylint: disable=too-many-arguments,too-many-positional-arguments
     def create_agent_tool(self, parent_run_context: RunContext,
-                          logger: StreamToLogger,
+                          journal: Journal,
```
d1donlydfink, line 153: Boring rename from StreamToLogger -> Journal

neuro_san/graph/tools/branch_tool.py
```diff
@@ -212,23 +212,23 @@ async def build(self) -> List[Any]:
         decision_name = self.get_decision_name()
         component_name = self.get_component_name()
         assistant_name = f"{decision_name}_{component_name}"
-        self.logger.write(f"setting up {component_name} assistant...")
+        await self.journal.write(f"setting up {component_name} assistant...")
```
d1donlydfink, line 215: Here is where we start to make our first calls journal.write()
Every time I looked at code like this I had to stop and think: OK, is this Python logging? or is this sending something to the client?  Hence the name Journal for the sake of something distinct from Logger.

neuro_san/interfaces/async_hopper.py
```diff
@@ -0,0 +1,14 @@
+from typing import Any
+
+
+# pylint: disable=too-few-public-methods
+class AsyncHopper:
+    """
+    An interface whose clients store things for later use.
+    """
```
d1donlydfink, line 8: This is an interface that is implemented by the AsyncCollatingQueue in the chat package,
but in lower-level code only the put() method is ever called.  To avoid circular package dependencies,
we create this interface.

neuro_san/journals/compatibility_journal.py
```diff
@@ -0,0 +1,24 @@
...
+    """
+
+    def __init__(self, hopper: AsyncHopper):
+        """
+        Constructor
+        """
+        journals: List[Journal] = [
+            TextJournal(),
+            MessageJournal(hopper)
+        ]
```
d1donlydfink, line 23: As you will see below we have more than one implementation of the Journal interface.
One of them - TextJournal - Does exactly what StreamToLogger did.
The new one - MessageJournal - puts messages on the queue to stream.

This particular concrete class is put together so only a single instance of the Journal has to be passed around,
but both kinds of Journal-ing are happening with every call to this guy.  This is thanks to the CompoundJournal class below.

When we no longer have to support the polling API (i.e. when our unileaf UI is updated to always stream), we can get rid of this guy.

neuro_san/journals/compound_journal.py
```diff
@@ -0,0 +1,47 @@
...
+from typing import List
+from typing import Union
+
+from neuro_san.journals.journal import Journal
+
+
+class CompoundJournal(Journal):
+    """
+    A Journal implementation that can service multiple other Journal instances
+    """
```
d1donlydfink, line 11: This guy takes any number of Journal instances and wraps them together to perform like functionality on each one.

neuro_san/journals/journal.py
```diff
@@ -0,0 +1,21 @@
+from typing import Any
+from typing import List
+from typing import Union
+
+
+class Journal:
+    """
+    An interface for journaling chat messages.
+    """
```
d1donlydfink, line 9: The main Journal interface.  This is what is used in client code which doesn't really care about underlying implementation.

The methods on this guy are those that were exposed in the former StreamToLogger.
Both of them are for backwards compatibility.  Later on this interface will grow to support
chat messages as well as just strings.

neuro_san/journals/message_journal.py
```diff
@@ -0,0 +1,55 @@
...
+from neuro_san.journals.journal import Journal
+from neuro_san.messages.legacy_logs_message import LegacyLogsMessage
+from neuro_san.messages.message_utils import convert_to_chat_message
+
+
+class MessageJournal(Journal):
+    """
+    Journal implementation for putting entries into a Hopper
+    for storage for later processing.
+    """
```
d1donlydfink, line 27: So really, the underlying implementation of the Hopper for this guy is the AsyncCollatingQueue.

neuro_san/journals/message_journal.py
```diff
@@ -0,0 +1,55 @@
...
+        :param entry: Add a string-ish entry to the logs.
+                    Can be either a string or bytes.
+        """
+        # Decoding bytes to string if necessary
+        if isinstance(entry, bytes):
+            entry = entry.decode('utf-8')
+
+        legacy = LegacyLogsMessage(content=entry)
+        message_dict: Dict[str, Any] = convert_to_chat_message(legacy)
+        await self.hopper.put(message_dict)
```
d1donlydfink, line 49: So now when the Journal.write() message is called, this guy creates a LegacyLogsMessage and tosses that into the hopper (the AsyncCollatingQueue). The clients-side can sort these out later as being meant for the thinking file.

Worth noting that the hopper.put() method is async, which means anyone who calls write() for any Journal interface has to await it.

neuro_san/journals/message_journal.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def get_logs(self) -> List[str]:
+        """
+        :return: A list of strings corresponding to log entries written with write()
+        """
+        return None
```
d1donlydfink, line 55: Worth noting that this class is to satisfy the streaming case.
Therefore, when get_logs() is called on this guy, there's nothing, because he's specifically not storing anything for old-style Logs() API .. and that can die in its own time.

neuro_san/journals/text_journal.py
```diff
@@ -12,10 +12,12 @@
...
 from typing import Union
 
+from neuro_san.journals.journal import Journal
 
-class StreamToLogger:
+
+class TextJournal(Journal):
     """
-    Class for capturing output stream to a list of strings
+    Journal implementation for capturing entries as a list of strings
```
d1donlydfink, line 20: This guy was renamed from StreamToLogger.
Names are changed to protect the innocent.

neuro_san/journals/journal.py
```diff
@@ -0,0 +1,21 @@
...
+
+    async def write(self, entry: Union[str, bytes]):
+        """
+        Writes a single string entry into the journal.
+        """
+        raise NotImplementedError
```
d1donlydfink, line 15: Worth noting that write() is now async. This is because some implementation have to deal with asynchronous queues, so async-ness infects at this level.

neuro_san/journals/text_journal.py
```diff
@@ -24,15 +26,15 @@ def __init__(self):
         """
         self.log_content = []
 
-    def write(self, message: Union[str, bytes]):
+    async def write(self, entry: Union[str, bytes]):
```
d1donlydfink, line 29: "message" has a specific meaning at the langchain level.  Hence the desire for a different argument name at the Journal level.

neuro_san/messages/agent_framework_message.py
d1donlydfink, line 1: Move existing messages to their own package.

neuro_san/messages/legacy_logs_message.py
```diff
@@ -0,0 +1,26 @@
...
+from typing import Union
+
+from langchain_core.messages.base import BaseMessage
+
+
+class LegacyLogsMessage(BaseMessage):
+    """
+    BaseMessage implementation of a message that is part of neuro-san's
+    legacy logs mechanism.
+    """
```
d1donlydfink, line 14: A new message type specifically so that the agent_cli client can sort out what goes in the thinking file for the streaming case for functional compatibility.   In the future we will actually be streaming *more* thinking information to the client but via different kinds of messages with better information as to where they originate from.  But these baby steps have to come first.

neuro_san/messages/message_utils.py
```diff
@@ -99,6 +100,7 @@ def generate_response(the_messages: List[Any]) -> str:
     SystemMessage: "system",
     AgentMessage: "agent",
     AgentFrameworkMessage: "agent-framework",
+    LegacyLogsMessage: "legacy-logs",
```
d1donlydfink, line 103: Add LegacyLogsMessage to a few lookup tables.

neuro_san/journals/compound_journal.py
```diff
@@ -0,0 +1,47 @@
...
+        if self.journals is None:
+            self.journals = []
+
+    async def write(self, entry: Union[str, bytes]):
+        """
+        Writes a single string entry into each journal.
+        """
+        for journal in self.journals:
+            await journal.write(entry)
+
```
andreidenissov-cog, line 29: Naive question: in this cycle do we "await" sequentially, journal by journal?
We can await on a set of async operations, right?

neuro_san/client/agent_cli.py
```diff
@@ -333,18 +360,25 @@ def stream_once(self, state: Dict[str, Any]) -> Dict[str, Any]:
...
 
+            message_type: int = response.get("type", 0)
             text: str = response.get("text")
 
             # Update chat response and maybe prompt.
             if text is not None:
-                print(f"Response: {text}")
-                last_chat_response = text
+                if message_type == 102:     # LEGACY_LOGS from chat.proto
+                    with open(self.args.thinking_file, "a", encoding="utf-8") as thinking:
```
andreidenissov-cog, line 369: Literal constant? Horror :-)

neuro_san/journals/compound_journal.py
d1donlydfink: In general async is simply asynchronous, running in the same thread, not necessarily in parallel.

That said, it's possible for some kind of pseudo-parallelism to take place here in this loop
if the iteration over the journals is non-blocking (which in this case it should be).
In this case each async call can be put on "the list" to be executed and theoretically these could
happen in some rando order.  But I think this depends on the low-level Python implementation.
What likely happens is that normal execution happens until there's something deeper down that
actually yields the processor, and then the next thing can go while that one waits, but we would need to add extra code to engage in that level of pseudo-parallelism.

But not every async method actually has some blocking operation at the end of its tunnel of execution.
I think the right way to look at it is: await means "this _could_ give up the processor on the event loop thread" but probably doesn't until is absolutely has to.

Regardless of all that, there are async objects/methods for locking and for waiting for things to be done if we need them, but I don't think we do here.  Given what I know about what is implemented, this loop largely proceeds synchronously until it tries to put something on the AsyncQueue, at which point it *might* decide to block.

To answer your ?s more concisely:
* Yes, we do end up awaiting journal-by-journal
* I think it is possible to wait on a set of sync operations, but we haven't actually needed to do that yet.

neuro_san/client/agent_cli.py
d1donlydfink: Yes I'm not fond of that.  I have thought about this and would like to make an Enum that echoes the gRPC constants so we can operate without gRPC if need be.  Have done this before in the authorization stuff.   Will likely go there in a subsequent PR.

ofrancon: I put an example of such an `IntEnum` in a previous PR, when `MESSAGE_TYPE_TO_CHAT_MESSAGE_TYPE` was introduced.
An `IntEnum` would simplify a few things.

d1donlydfink: This is repaired in the Afterthoughts PR that lies on top of this one.

PR #36, opened 2025-01-02
UN-2494 Protocol Afterthoughts

d1donlydfink:
In taking some time to think about what the protocol is communicating, I offer some tweaks.

Also:
* create an enum as requested by @andreidenissov-cog in a previous PR and further refine it in the direction @ofrancon was heading (though not entirely)
* Use latest leaf-common and leaf-server-common and change imports to reflect that
* Add some comments to ChatSession.chat() that @ofrancon thought were missing the first time he looked at it.

### DIFF
--- a/neuro_san/chat/chat_session.py
+++ b/neuro_san/chat/chat_session.py
@@ -37,6 +37,17 @@ class ChatSession:
         :param user_input: A string with the user's input
         :param sly_data: A mapping whose keys might be referenceable by agents, but whose
                  values should not appear in agent chat text. Can be None.
+        :return: An Iterator over dictionary representation of chat messages.
+                The keys/values/structure of these chat message dictionaries will reflect
+                instances of ChatMessage from chat.proto.
+
+                Note that Iterators themselves are *not* simply lists. They are a Python
+                construct intended for use in a for-loop that is allowed to come up with
+                its content dynamically.  For our purposes, when an initiator of chat()
+                gets a handle to this Iterator, they can begin looping/waiting on its contents
+                without the content itself having been created yet.  This is a building
+                block of streaming results even though direct callers may not actually
+                be streaming.
 
         Note that nothing is returned immediately as processing
         the chat happens asynchronously and might take longer
--- a/neuro_san/chat/data_driven_chat_session.py
+++ b/neuro_san/chat/data_driven_chat_session.py
@@ -98,7 +98,17 @@ class DataDrivenChatSession(ChatSession):
         :param user_input: A string with the user's input
         :param sly_data: A mapping whose keys might be referenceable by agents, but whose
                  values should not appear in agent chat text. Can be None.
-        :return: An Iterator of chat.ChatMessage dictionaries
+        :return: An Iterator over dictionary representation of chat messages.
+                The keys/values/structure of these chat message dictionaries will reflect
+                instances of ChatMessage from chat.proto.
+
+                Note that Iterators themselves are *not* simply lists. They are a Python
+                construct intended for use in a for-loop that is allowed to come up with
+                its content dynamically.  For our purposes, when an initiator of chat()
+                gets a handle to this Iterator, they can begin looping/waiting on its contents
+                without the content itself having been created yet.  This is a building
+                block of streaming results even though direct callers may not actually
+                be streaming.
 
         Results are polled from get_latest_response() below, as this non-streaming
         version can take longer than the lifetime of a socket.


neuro_san/api/grpc/chat.proto
```diff
@@ -25,10 +25,10 @@ message ChatMessage {
...
     enum ChatMessageType {
         UNKNOWN = 0;    // Note: UNKNOWN is not a langchain thing.
-        HUMAN = 1;
-        SYSTEM = 2;
-        AI = 3;
-        TOOL = 4;
+        SYSTEM = 1;
+        HUMAN = 2;
+        TOOL = 3;
+        AI = 4;
```
d1donlydfink, line 32: Re-order these so that the least important responses come before the more important responses in the ordering.  Importance here relates more to what an end-user of a chat client would be interested in seeing.

neuro_san/api/grpc/chat.proto
```diff
@@ -47,7 +47,7 @@ message ChatMessage {
     // reference for how the data was created.  If this happens, then
     // it should be safe to assume that the text is enough to represent the
     // message in any history carried forward.
-    mime_data.MimeData mime_data = 3 [json_name="mime_data"];
+    repeated mime_data.MimeData mime_data = 3 [json_name="mime_data"];
```
d1donlydfink, line 51: Allow multiple bits of mime_data to come with a single chat message.


Rationale: You can compress a bunch of text into a single message, but you cannot do the same with media, so create the affordance for when we get there.

neuro_san/api/grpc/chat_pb2.py
```diff
@@ -15,7 +15,7 @@
 from neuro_san.api.grpc import mime_data_pb2 as neuro__san_dot_api_dot_grpc_dot_mime__data__pb2
```
d1donlydfink, line 17: Update the grpc file

neuro_san/client/agent_cli.py
```diff
@@ -365,7 +366,7 @@ def stream_once(self, state: Dict[str, Any]) -> Dict[str, Any]:
 
             # Update chat response and maybe prompt.
             if text is not None:
-                if message_type == 102:     # LEGACY_LOGS from chat.proto
+                if message_type == ChatMessageType.LEGACY_LOGS:
```
d1donlydfink, line 378: As per @andreidenissov-cog 's earlier recommendation, use an actual enum newly defined below.

neuro_san/messages/chat_message_type.py
```diff
@@ -0,0 +1,54 @@
...
+from neuro_san.messages.agent_framework_message import AgentFrameworkMessage
+from neuro_san.messages.agent_message import AgentMessage
+from neuro_san.messages.legacy_logs_message import LegacyLogsMessage
+
+
+class ChatMessageType(Enum):
+    """
+    Python enum to mimic gRPC for chat.ChatMessageType without dragging in all of gRPC.
+    These all need to match what is defined in chat.proto
+    """
```
d1donlydfink, line 22: Define an enum for integer constants used in the api.
The idea is to create convenience translations between the world and the langchain message types
without actually depending on anything gRPC itself.
(Keep in mind that grpc might eventually not be the only wire transfer mechanism to support.)

neuro_san/messages/message_utils.py
```diff
@@ -93,17 +88,6 @@ def generate_response(the_messages: List[Any]) -> str:
...
 
-MESSAGE_TYPE_TO_ROLE: Dict[Type[BaseMessage], str] = {
-    AIMessage: "assistant",
-    HumanMessage: "user",
-    ToolMessage: "tool",
-    SystemMessage: "system",
-    AgentMessage: "agent",
-    AgentFrameworkMessage: "agent-framework",
-    LegacyLogsMessage: "legacy-logs",
-}
```
d1donlydfink, line 104: This moves to ChatMessageTypes.py

neuro_san/messages/message_utils.py
```diff
@@ -138,19 +122,6 @@ def get_content(message: Any) -> str:
...
-    # Needs to match chat.proto
-    HumanMessage: 1,
-    SystemMessage: 2,
-    AIMessage: 3,
-    ToolMessage: 4,
-
-    AgentMessage: 100,
-    AgentFrameworkMessage: 101,
-    LegacyLogsMessage: 102,
-}
```
d1donlydfink, line 151: This moves to ChatMessageTypes.py

neuro_san/api/grpc/chat_pb2.py
dsargent, line 1: Can this be generated as needed rather than checked in?

neuro_san/messages/chat_message_type.py
dsargent: Bear in mind that gRPC ≠ protobuf.

TL;DR: leaking gRPC-isms into app code bad, using protobuf as a universal entity definition format okay.

The distinction is important because one is a wire communications standard, and the other is a data serialization protocol. They are different layers in the stack, like http vs JSON.

In theory (and I bet there are cases in practice) you could use protobuf outside of gRPC and you could use gRPC with a different serialization format than protobuf (eg JSON). They are orthogonal concerns.

Hot take, but I'd be okay with seeing protobuf-isms in other places in the code. However I would share your abhorrence for seeing *gRPC-isms* in the application code. 

For example, a module dealing with processing agent hocon files shouldn't be dealing with gRPC timeouts and retries -- that would be poor separation of concerns.

However, I would be okay with seeing protobuf-isms in such a module, since protobuf is a wire-agnostic serialization format that shouldn't know or care about gRPC.

Edit: when I say "protobuf-isms" I don't even mean "protobuf-isms" since there's no such thing in Python. I mean directly consuming the entities, themselves just Python types, generated by `protoc`.

neuro_san/api/grpc/chat_pb2.py
d1donlydfink: No. This is shipping as a library, and for the library to work OOB, we need to also be shipping the generated grpc.

neuro_san/chat/data_driven_chat_session.py
```diff
@@ -182,15 +183,12 @@ def is_streamable_message(self, chat_message: Dict[str, Any], index: int) -> boo
...
-        # This comes from definitions in chat.proto
-        ai_message_type: int = 3
-
         message_type: int = chat_message.get("type")
 
         # Only report messages that are important enough to send back as part of chat
         # (for now).  This includes any response for an AI (read: LLM), a tool, or
         # agent or framework messages.
-        if message_type is None or message_type < ai_message_type:
+        if message_type is None or message_type < ChatMessageType.AI.value:
```
d1donlydfink: Use the Enum. Just like @andreidenissov-cog and @ofrancon have pointed out in previous PRs.

neuro_san/client/agent_cli.py
```diff
@@ -251,7 +252,7 @@ def send_user_input(self, user_input: str, sly_data: Dict[str, Any]):
             chat_request = {
                 "session_id": self.session_id,
                 "user_message": {
-                    "type": 1,      # HUMAN from chat.proto
+                    "type": ChatMessageType.HUMAN.value,
```
d1donlydfink: Use the Enum. Just like @andreidenissov-cog and @ofrancon have pointed out in previous PRs.

neuro_san/interfaces/coded_tool.py
```diff
@@ -28,7 +28,16 @@ class CodedTool:
...
+        This method is provided as a convenience for an "easy" start to using
+        coded-tools.  This synchronous interface called by async_invoke() below.
+        when the coded tool is invoked by the agent hierarchy.
+
+        Know that any CodedTool is run within the confines of a Python asynchronous
+        EventLoop. Any synchronous blocking that happens - like making a call to a
+        web service over a socket, or something that inherently sleep()s - *will* also
+        block all other agent operations.  This is not so bad in a low-traffic or
+        test environment, but when scaling up, you really really want to embrace
+        and override the async_invoke() method below instead of this one.
```
d1donlydfink, line 40: Add a bunch of comments about the async environment these CodedTools are run in.
@ofrancon : We should make sure the 1C guys know about these implications.

neuro_san/interfaces/coded_tool.py
```diff
@@ -44,11 +53,17 @@ def invoke(self, args: Dict[str, Any], sly_data: Dict[str, Any]) -> Any:
                 adding the data is not invoke()-ed more than once.
         :return: A return value that goes into the chat stream.
         """
-        raise NotImplementedError
+        # Do not raise an exception here, but pass instead.
+        # This allows for fully asynchronous CodedTools to not have to worry about
+        # the synchronous bits.
```
d1donlydfink, line 58: Do not raise. Explained in comment.

neuro_san/interfaces/coded_tool.py
```diff
@@ -64,6 +79,7 @@ async def async_invoke(self, args: Dict[str, Any], sly_data: Dict[str, Any]) ->
                 adding the data is not invoke()-ed more than once.
         :return: A return value that goes into the chat stream.
         """
+        # DEF - Use the AsyncioExecutor for a submit() for better task tracking and handling
```
d1donlydfink, line 82: Future improvement noted.

neuro_san/api/grpc/chat_pb2.py
dsargent: Why not make that a step when building the library then?

ofrancon: Because we can also pip install it directly from the code (GitHub) (pip install -e ...)

neuro_san/chat/data_driven_chat_session.py
ofrancon: If you make it an `IntEnum` then it behaves like an `int`. 
And you can use
```or message_type < ChatMessageType.AI```
See https://docs.python.org/3/howto/enum.html#intenum

File: neuro_san/client/agent_cli.py (hunk 9 above)
Comment (ofrancon) [reply neuro_san/client/agent_cli.py]: Here too, and `IntEnum` would allow a simpler use:
```
"type": ChatMessageType.HUMAN
```

File: neuro_san/messages/chat_message_type.py [hunk 13]
```diff
@@ -0,0 +1,54 @@
...
+from langchain_core.messages.human import HumanMessage
+from langchain_core.messages.system import SystemMessage
+from langchain_core.messages.tool import ToolMessage
+
+from neuro_san.messages.agent_framework_message import AgentFrameworkMessage
+from neuro_san.messages.agent_message import AgentMessage
+from neuro_san.messages.legacy_logs_message import LegacyLogsMessage
+
+
+class ChatMessageType(Enum):
```
ofrancon: `IntEnum` is a subclass of `Enum` that behaves like an `int`

neuro_san/messages/chat_message_type.py
```diff
@@ -0,0 +1,54 @@
...
+    TOOL = 3
+    AI = 4
+
+    AGENT = 100
+    AGENT_FRAMEWORK = 101
+    LEGACY_LOGS = 102
+
+
+# Convenience mappings going between constants and class types
+MESSAGE_TYPE_TO_CHAT_MESSAGE_TYPE: Dict[Type[BaseMessage], int] = {
```
ofrancon: This could become a class method, used like that:
```
ChatMessageType.from_langchain_message(a_langchain_message)
```
See comment below for an example of the code.

neuro_san/messages/chat_message_type.py
```diff
@@ -0,0 +1,54 @@
...
+    AIMessage: ChatMessageType.AI,
+
+    AgentMessage: ChatMessageType.AGENT,
+    AgentFrameworkMessage: ChatMessageType.AGENT_FRAMEWORK,
+    LegacyLogsMessage: ChatMessageType.LEGACY_LOGS,
+}
+
+
+MESSAGE_TYPE_TO_ROLE: Dict[Type[BaseMessage], str] = {
+    AIMessage: "assistant",
```
ofrancon, line 74: The role can be a constant in the IntEnum directly if you provide a constructor.
The class is still an `int`, and can be used as an `int`, but it also has extra fields.
Here is a full example:
```
from enum import IntEnum
from enum import unique

from langchain_core.messages.ai import AIMessage
from langchain_core.messages.base import BaseMessage
from langchain_core.messages.system import SystemMessage


@unique
class ChatMessageType(IntEnum):

    UNKNOWN_MESSAGE_TYPE = (0, "Unknown")
    SYSTEM = (1, "system")
    HUMAN = (2, "user")
    TOOL = (3, "tool")
    AI = (4, "assistant")

    def __new__(cls, value, role):
        message = int.__new__(cls, value)
        message._value_ = value
        message.role = role
        return message

    @classmethod
    def from_langchain_message(cls, base_message: BaseMessage):
        if isinstance(base_message, AIMessage):
            return ChatMessageType.AI
        elif isinstance(base_message, SystemMessage):
            return ChatMessageType.SYSTEM


# Example usage:
if __name__ == "__main__":
    print(f"SYSTEM == 1: {ChatMessageType.SYSTEM == 1}")
    print(f"SYSTEM.role: {ChatMessageType.SYSTEM.role}")
    print(f"SYSTEM < HUMAN: {ChatMessageType.SYSTEM < ChatMessageType.HUMAN}")
    a_langchain_message = AIMessage(content='hello')
    print("a_langchain_message = AIMessage(content='hello')")
    print(f"ChatMessageType.from_langchain_message(a_langchain_message).role:"
          f" {ChatMessageType.from_langchain_message(a_langchain_message).role}")
```

Now the logic is well encapsulated and there can be only 1 version of the truth.

See https://stackoverflow.com/a/68400507/4821952 or ask ChatGPT, it's good at this kind of stuff.

neuro_san/messages/message_utils.py
```diff
@@ -159,9 +131,10 @@ def convert_to_chat_message(message: BaseMessage) -> Dict[str, Any]:
     :return: The ChatMessage in dictionary form
     """
 
-    unknown: int = 0
+    message_type: ChatMessageType = MESSAGE_TYPE_TO_CHAT_MESSAGE_TYPE.get(type(message),
```
ofrancon: With the IntEnum I suggested above this becomes:
```
message_type: ChatMessageType = ChatMessageType.from_langchain_message(message)
```

neuro_san/messages/message_utils.py
```diff
@@ -159,9 +131,10 @@ def convert_to_chat_message(message: BaseMessage) -> Dict[str, Any]:
...
+                                                                          ChatMessageType.UNKNOWN_MESSAGE_TYPE)
     chat_message: Dict[str, Any] = {
-        "type": MESSAGE_TYPE_TO_CHAT_MESSAGE_TYPE.get(type(message), unknown),
+        "type": message_type.value,
```
ofrancon: Works, but can become:
```
"type": message_type
```
because `message_type` is an `int`

neuro_san/chat/data_driven_chat_session.py
d1donlydfink: Done

neuro_san/client/agent_cli.py
d1donlydfink: Done

neuro_san/messages/chat_message_type.py
d1donlydfink: Done

neuro_san/api/grpc/chat_pb2.py
dsargent: Seems like that is conflating two use cases:

* If you're an "internal" dev -- `pip install -e` and since you're an internal dev, it's on you to generate the protobuf stubs
* If you just want a wheel, install the wheel, which we have previously generated including the protobuf stubs, which we generated as part of the wheel building process.

In neither case are the stubs checked into git.

Note: this is my opinion. Not a blocker for the PR (and I won't be affected by it so 🤷🏻 )

neuro_san/messages/chat_message_type.py
d1donlydfink: I used this example to riff on.   Among other things I tend to forget about IntEnum 🧀 = 🧠  So Thanks.
I am not hewing exactly to this for a couple of reasons:
1. I am not going with the compound nature of the values. I don't like the suggestion of using Python Duck-Typing to add a new member to something that isn't really supposed to be there.
2. I like the idea of having class methods to do the conversions
3. I am sticking with the MAP constants because the if/elif doesn't scale to the point where pylint won't complain about having too many branches
4. FWIW: The role stuff is on its way to obsolescence.  When Logs() API goes away from the protocol, roles can fall as well.

neuro_san/messages/message_utils.py
d1donlydfink: Done

neuro_san/messages/message_utils.py
d1donlydfink: Done

neuro_san/api/grpc/chat_pb2.py
d1donlydfink: Such a generate-only-for-the-wheel would be a new special case in the python build infrastructure that among other things also generates wheel files according to data-driven config.  This would introduce process (verbs) into something that was just converted to the standard that uses only nouns.  So ... not gunna go there - it's fine.

neuro_san/chat/chat_session.py
```diff
@@ -37,6 +37,17 @@ async def chat(self, user_input: str, sly_data: Dict[str, Any]) -> Iterator[Dict
...
+                The keys/values/structure of these chat message dictionaries will reflect
+                instances of ChatMessage from chat.proto.
+
+                Note that Iterators themselves are *not* simply lists. They are a Python
+                construct intended for use in a for-loop that is allowed to come up with
+                its content dynamically.  For our purposes, when an initiator of chat()
+                gets a handle to this Iterator, they can begin looping/waiting on its contents
+                without the content itself having been created yet.  This is a building
+                block of streaming results even though direct callers may not actually
+                be streaming.
```
d1donlydfink, line 50: Add the kind of comment @ofrancon had been looking for in a previous PR.

neuro_san/chat/data_driven_chat_session.py
```diff
@@ -97,7 +98,17 @@ async def chat(self, user_input: str, sly_data: Dict[str, Any]) -> Iterator[Dict
...
+                The keys/values/structure of these chat message dictionaries will reflect
+                instances of ChatMessage from chat.proto.
+
+                Note that Iterators themselves are *not* simply lists. They are a Python
+                construct intended for use in a for-loop that is allowed to come up with
+                its content dynamically.  For our purposes, when an initiator of chat()
+                gets a handle to this Iterator, they can begin looping/waiting on its contents
+                without the content itself having been created yet.  This is a building
+                block of streaming results even though direct callers may not actually
+                be streaming.
```
d1donlydfink, line 111: Copy the comment from its parent class

neuro_san/chat/data_driven_chat_session.py
```diff
@@ -182,15 +193,12 @@ def is_streamable_message(self, chat_message: Dict[str, Any], index: int) -> boo
...
-        # This comes from definitions in chat.proto
-        ai_message_type: int = 3
-
         message_type: int = chat_message.get("type")
 
         # Only report messages that are important enough to send back as part of chat
         # (for now).  This includes any response for an AI (read: LLM), a tool, or
         # agent or framework messages.
-        if message_type is None or message_type < ai_message_type:
+        if message_type is None or message_type < ChatMessageType.AI:
```
d1donlydfink, line 201: New IntEnum allows wonky magic number to go away.

neuro_san/client/agent_cli.py
```diff
@@ -340,7 +341,7 @@ def stream_once(self, state: Dict[str, Any]) -> Dict[str, Any]:
         chat_request = {
             "session_id": self.session_id,
             "user_message": {
-                "type": 1,      # HUMAN from chat.proto
+                "type": ChatMessageType.HUMAN,
```
d1donlydfink, line 344: New IntEnum allows wonky magic number to go away.

neuro_san/client/agent_cli.py
```diff
@@ -360,12 +361,21 @@ def stream_once(self, state: Dict[str, Any]) -> Dict[str, Any]:
...
+            # Convert the message type in the response to the enum we want to work with
+            response_message_type: str = response.get("type", "UNKNOWN_MESSAGE_TYPE")
+            message_type: int = ChatMessageType.UNKNOWN_MESSAGE_TYPE
+            try:
+                # Normal case: We have a 1:1 mapping of ChatMessageType to what is in grpc def
+                message_type = ChatMessageType[response_message_type]
+            except KeyError as exception:
+                raise ValueError(f"Got message type {response_message_type}."
+                                 " Are ChatMessageType and chat.proto out of sync?") from exception
+
```
d1donlydfink, line 373: Testing out the use of the new enum found a case where we need to convert what gRPC gives us to enum land in the client.

neuro_san/client/agent_session_factory.py
```diff
@@ -21,8 +21,7 @@ class AgentSessionFactory:
     Factory class for agent sessions.
     """
 
-    @staticmethod
-    def create_session(session_type: str,
+    def create_session(self, session_type: str,
```
d1donlydfink, line 24: This factory becomes an instance w/rt this method so that dependent repos can make subclasses that use this parent class's behavior.

neuro_san/client/direct_agent_session_factory.py
```diff
@@ -11,7 +11,7 @@
 # END COPYRIGHT
 from typing import Dict
 
-from leaf_server_common.asyncio.asyncio_executor import AsyncioExecutor
+from leaf_common.asyncio.asyncio_executor import AsyncioExecutor
```
d1donlydfink, line 14: Use the asyncio from leaf_common now.  My dev testing environment had moved on and I wasn't going back.

neuro_san/interfaces/coded_tool.py
```diff
@@ -44,11 +53,17 @@ def invoke(self, args: Dict[str, Any], sly_data: Dict[str, Any]) -> Any:
...
+        # the synchronous bits.
 
     async def async_invoke(self, args: Dict[str, Any], sly_data: Dict[str, Any]) -> Any:
         """
-        Called when the coded tool is invoked asynchronously  by the agent hierarchy.
+        Called when the coded tool is invoked asynchronously by the agent hierarchy.
+        Strongly consider overriding this method instead of the "easier" synchronous
+        version above when the possibility of making any kind of call that could block
+        (like sleep() or a socket read/write out to a web service) is within the
+        scope of your CodedTool.
```
d1donlydfink, line 66: Add a bunch of comments about the async environment these CodedTools are run in.
@ofrancon : We should make sure the 1C guys know about these implications.

neuro_san/messages/chat_message_type.py
```diff
@@ -0,0 +1,81 @@
...
+    @classmethod
+    def from_message(cls, base_message: BaseMessage) -> ChatMessageType:
+        """
+        :param base_message: A base message instance
+        :return: The ChatMessageType corresponding to the base_message
+        """
+        base_message_type: Type[BaseMessage] = type(base_message)
+        chat_message_type: ChatMessageType = \
+            _MESSAGE_TYPE_TO_CHAT_MESSAGE_TYPE.get(base_message_type, cls.UNKNOWN_MESSAGE_TYPE)
+        return chat_message_type
```
d1donlydfink, line 44: Use class methods for conversion as per @ofrancon 's suggestion. Here and below.

neuro_san/messages/chat_message_type.py
```diff
@@ -0,0 +1,81 @@
...
+    # Needs to match chat.proto
+    SystemMessage: ChatMessageType.SYSTEM,
+    HumanMessage: ChatMessageType.HUMAN,
+    ToolMessage: ChatMessageType.TOOL,
+    AIMessage: ChatMessageType.AI,
+
+    AgentMessage: ChatMessageType.AGENT,
+    AgentFrameworkMessage: ChatMessageType.AGENT_FRAMEWORK,
+    LegacyLogsMessage: ChatMessageType.LEGACY_LOGS,
+}
```
d1donlydfink, line 71: Sticking with maps to help with the conversion methods.
These scale better/easier than if/else blocks that pylint will eventually complain about.

neuro_san/messages/chat_message_type.py
```diff
@@ -0,0 +1,81 @@
...
+
+    @classmethod
+    def message_to_role(cls, base_message: BaseMessage) -> str:
+        """
+        This role stuff will be removed when the Logs() API is removed,
+        as the ChatMessageType and grpc definitions make it redundant.
+
+        :param base_message: A base message instance
+        :return: The role string corresponding to the base_message
+        """
```
d1donlydfink, line 54: Same deal here. Added comment about how the role stuff should disappear when Logs() API does.

requirements-private.txt
```diff
@@ -9,7 +9,7 @@
 # DEF 07/26/2022: Due to bug in codefresh private repo plugin, this leaf-common
 # comes in from top-level requirements.txt file for now.
 # Public repos
-git+https://${LEAF_SOURCE_CREDENTIALS}@github.com/leaf-ai/leaf-common.git@1.2.14
+git+https://${LEAF_SOURCE_CREDENTIALS}@github.com/leaf-ai/leaf-common.git@1.2.15
 
 # Private repos
-git+https://${LEAF_PRIVATE_SOURCE_CREDENTIALS}@github.com/leaf-ai/leaf-server-common.git@0.1.14
+git+https://${LEAF_PRIVATE_SOURCE_CREDENTIALS}@github.com/leaf-ai/leaf-server-common.git@0.1.15
```
d1donlydfink, line 15: Use latest leaf-server-common and leaf-common, as that is where my testing environment is now.

neuro_san/chat/data_driven_chat_session.py
ofrancon: wonderful

neuro_san/client/agent_cli.py
ofrancon: Why not using the convenient `from_message` method here?
It would become:
```
message_type = ChatMessageType.from_message(response)
```
The method can even handle invalid ids in a single place.

neuro_san/messages/chat_message_type.py
ofrancon: Sure, that's an implementation detail now.

neuro_san/messages/chat_message_type.py
ofrancon: lgtm!

neuro_san/client/agent_cli.py
d1donlydfink: What's coming in here is not a langchain BaseMessage, but a dictionary version of the grpc message. It's the grpc conversion that ends up giving us a string at this point and that is what needs to be translated.
I see what you mean though, it could be another "from_()" thing.

There is another future exercise to create reusable bits that clients can use. That can be part of that effort.
Enough work has gone into this stack, I'm anxious to move on.

PR #37, opened 2025-01-03
UN-2746 async stream external tools

d1donlydfink:
Use the new leaf-common async grpc session infrastructure to have the ExternalTool stuff call other agent services asynchronously.  This is really good, as they will no longer block everything else in the server when they are waiting for results.

When this stack of PRs go in, there is some async-ifying of other coded agents we can do more nicely now.

Tested: Used latest leaf-server-common and leaf-common on OF coded agents in unileaf repo.


neuro_san/client/direct_agent_session_factory.py
```diff
@@ -11,7 +11,7 @@
 # END COPYRIGHT
 from typing import Dict
 
-from leaf_server_common.asyncio.asyncio_executor import AsyncioExecutor
+from leaf_common.asyncio.asyncio_executor import AsyncioExecutor
```
d1donlydfink: Get asyncio stuff from leaf-common now.

neuro_san/graph/tools/external_tool.py
```diff
@@ -66,66 +65,56 @@ async def build(self) -> List[Any]:
...
 
         # Create an AgentSession if necessary
         if self.session is None:
             agent_location: Dict[str, str] = ExternalToolAdapter.parse_external_agent(self.agent_url)
             self.session = ExternalToolAdapter.create_session(agent_location)
 
         # Send off the input
-        await self.send_input(f"```json\n{json.dumps(self.arguments)}```", self.sly_data)
+        chat_request: Dict[str, Any] = self.gather_input(f"```json\n{json.dumps(self.arguments)}```",
+                                                         self.sly_data)
```
d1donlydfink, line 77: Get the chat_request from dumbed-down gather_input() method.

neuro_san/graph/tools/external_tool.py
```diff
@@ -66,66 +65,56 @@ async def build(self) -> List[Any]:
...
 
-        # Set up listening for the response
-        logs_request = {
-            "session_id": self.session_id
-        }
+        # Note that we are not await-ing the response here because what is returned is a generator.
+        # Proper await-ing for generator results is done in the "async for"-loop below.
+        chat_responses: Generator[Dict[str, Any], None, None] = self.session.streaming_chat(chat_request)
```
d1donlydfink, line 81: Get a streaming response generator from streaming_chat()

neuro_san/graph/tools/external_tool.py
```diff
@@ -66,66 +65,56 @@ async def build(self) -> List[Any]:
...
-
-        # Prepare the output
-        message: Dict[str, Any] = {
-            "role": "assistant",
-            "content": content
-        }
+        # The asynchronous generator will wait until the next response is available
+        # from the stream.  When the other side is done, the iterator will exit the loop.
+        empty = {}
+        async for chat_response in chat_responses:
```
d1donlydfink, line 86: Not the greatest diff, but we are replacing the polling loop here with the streaming loop like also happens already in the client. The difference here is that we are doing it within an asynchronous context (client is synchronous), so we have to use the ever-mysterious "async for" construct that Python offers us for just this situation.  Basically the iterator here does the await-ing internally.

neuro_san/graph/tools/external_tool.py
```diff
@@ -66,66 +65,56 @@ async def build(self) -> List[Any]:
...
+            # In terms of sending tool results back up the graph,
+            # we really only care about the AI responses.
+            if message_type == ChatMessageType.AI.value and text is not None:
+
+                # Prepare the output
+                message: Dict[str, Any] = {
+                    "role": "assistant",
+                    "content": text
+                }
+                message_list.append(message)
```
d1donlydfink, line 106: Only send AI messages back upstream to the calling tool.
At this point we do not care about anything else.

neuro_san/graph/tools/external_tool.py
```diff
@@ -66,66 +65,56 @@ async def build(self) -> List[Any]:
...
+
+        messages_str: str = json.dumps(message_list)
         return messages_str
 
-    async def send_input(self, agent_input: str, sly_data: Dict[str, Any]):
+    def gather_input(self, agent_input: str, sly_data: Dict[str, Any]) -> Dict[str, Any]:
```
d1donlydfink, line 111: Repurpose this method to put together the chat request.

neuro_san/graph/tools/external_tool.py
```diff
@@ -141,11 +130,7 @@ async def send_input(self, agent_input: str, sly_data: Dict[str, Any]):
         if sly_data is not None and len(sly_data.keys()) > 0:
             chat_request["sly_data"] = sly_data
 
-        # Send the request and get the session_id against which to query results.
-        # It'd be nice if there were an async version of this
-        chat_response = self.session.chat(chat_request)
-        if self.session_id is None:
-            self.session_id = chat_response.get("session_id")
+        return chat_request
```
d1donlydfink, line 133: No longer call the session.chat() method, as that is for old-style polling.
Instead, simply return the chat_request and let the caller do the new-style streaming.

neuro_san/run_context/utils/external_tool_adapter.py
```diff
@@ -167,10 +167,10 @@ def create_session(agent_location: Dict[str, str]) -> AgentSession:
 
         # Optimization:
         #   It's possible we might want to create a different kind of session
-        #   to minimize socket usage, but for now use the ServiceAgentSession
+        #   to minimize socket usage, but for now use the AsyncServiceAgentSession
         #   so as to ensure proper logging even on the same server (localhost).
-        session = ServiceAgentSession(host, port, agent_name=agent_name,
-                                      service_prefix=service_prefix)
+        session = AsyncServiceAgentSession(host, port, agent_name=agent_name,
+                                           service_prefix=service_prefix)
```
d1donlydfink, line 173: Always use the Async version of the ServiceAgentSession when opening up a session to talk to another server.

neuro_san/session/async_service_agent_session.py
```diff
@@ -0,0 +1,302 @@
...
+
+from neuro_san.session.agent_service_stub import AgentServiceStub
+from neuro_san.session.agent_session import AgentSession
+from neuro_san.api.grpc import agent_pb2 as service_messages
+
+
+class AsyncServiceAgentSession(AsyncAbstractServiceSession):
+    """
+    Implementation of AgentSession that talks to a gRPC service asynchronously.
+    """
```
d1donlydfink, line 28: This guy is a cut-paste version of the existing synchronous ServiceAgentSession that has been retooled for asynchronous calling.  Alas, this cut/paste is completely necessary given how async/await ends up spreading through the code like a disease.  Luckily, it's largely contained to this file.

Now that we have this guy, the ExternalToolAdapter is able to call other agent services without synchronously blocking the entirety of everything else in the event loop - much more prerformant overall.

neuro_san/session/async_service_agent_session.py
```diff
@@ -0,0 +1,302 @@
...
+
+        use_port: str = str(self.DEFAULT_PORT)
+        if port is not None:
+            use_port = port
+
+        # Normally we pass around the service_stub like a class,
+        # but AgentServiceStub has a __call__() method to intercept
+        # constructor-like behavior.
+        service_stub = AgentServiceStub(agent_name, service_prefix)
+        AsyncAbstractServiceSession.__init__(self, "Agent Server",
```
d1donlydfink, line 78: Note we use AsyncAbstractServiceSession as a base class to get its low-level async-ness.

neuro_san/session/async_service_agent_session.py
```diff
@@ -0,0 +1,302 @@
...
+                                             service_stub,
+                                             use_host, use_port,
+                                             timeout_in_seconds, metadata,
+                                             security_cfg, umbrella_timeout,
+                                             streaming_timeout_in_seconds)
+
+    async def function(self, request_dict: Dict[str, Any]) -> Dict[str, Any]:
```
d1donlydfink, line 85: All the grpc API methods we call are now labeled as async

neuro_san/session/async_service_agent_session.py
```diff
@@ -0,0 +1,302 @@
...
+                    protobufs structure. Has the following keys:
+                "function" - the dictionary description of the function
+                "status" - status for finding the function.
+        """
+        # pylint: disable=no-member
+        return await self.call_grpc_method(
+            "function",
+            self._function_from_stub,
+            request_dict,
+            service_messages.FunctionRequest())
```
d1donlydfink, line 100: And inside, everything is awaited

neuro_san/session/async_service_agent_session.py
```diff
@@ -0,0 +1,302 @@
...
+    async def _function_from_stub(stub, timeout_in_seconds,
+                                  metadata, credentials, *args):
+        """
+        Global method associated with the session that calls Function
+        given a grpc Stub already set up with a channel (socket) to call with.
+        """
+        response = await stub.Function(*args, timeout=timeout_in_seconds,
+                                       metadata=metadata,
+                                       credentials=credentials)
+        return response
```
d1donlydfink, line 254: All the static methods that use the async stub are also awaited.

requirements-private.txt
```diff
@@ -9,7 +9,7 @@
 # DEF 07/26/2022: Due to bug in codefresh private repo plugin, this leaf-common
 # comes in from top-level requirements.txt file for now.
 # Public repos
-git+https://${LEAF_SOURCE_CREDENTIALS}@github.com/leaf-ai/leaf-common.git@1.2.14
+git+https://${LEAF_SOURCE_CREDENTIALS}@github.com/leaf-ai/leaf-common.git@1.2.15
 
 # Private repos
-git+https://${LEAF_PRIVATE_SOURCE_CREDENTIALS}@github.com/leaf-ai/leaf-server-common.git@0.1.14
+git+https://${LEAF_PRIVATE_SOURCE_CREDENTIALS}@github.com/leaf-ai/leaf-server-common.git@0.1.15
```
d1donlydfink: This PR requires some updates from leaf-common and leaf-server-common.

neuro_san/graph/tools/external_tool.py
```diff
@@ -66,66 +65,56 @@ async def build(self) -> List[Any]:
...
+
+            if session_id is not None:
+                self.session_id = session_id
 
+            message_type: int = response.get("type", 0)
+            text: str = response.get("text")
+
+            # In terms of sending tool results back up the graph,
+            # we really only care about the AI responses.
+            if message_type == ChatMessageType.AI.value and text is not None:
```
ofrancon, line 99: If you make ChatMessageType an `IntEnum` instead of `Enum` then you can use `message type == ChatMessageType.AI` directly. 
See https://docs.python.org/3/howto/enum.html#intenum

neuro_san/session/async_service_agent_session.py
ofrancon: Can you add that to the docstring? If I understand correctly, with a line like that:
```
Copy/paste from ServiceAgentSession, but with async/await keywords to make it asynchronous.
WARNING: if you update this class, please also update ServiceAgentSession, and vice versa.
```

neuro_san/graph/tools/external_tool.py
```diff
@@ -29,8 +30,6 @@ class ExternalTool(CallableTool):
     another agent hierarchy as a tool.
     """
 
-    DEFAULT_SLEEP_SECONDS: int = 5
```
d1donlydfink, line 32: We no longer need to sleep with streaming.

PR #38, opened 2025-01-03
Copyright 2025

d1donlydfink:
Copyright 2025.  Easy peasy despite 97 files changed.


neuro_san/api/grpc/chat_pb2.py
```diff
@@ -1,3 +1,14 @@
...
+# All Rights Reserved.
+# Issued under the Academic Public License.
+#
+# You can be released from the terms, and requirements of the Academic Public
+# License by purchasing a commercial license.
+# Purchase of a commercial license is mandatory for any use of the
+# neuro-san SDK Software in commercial settings.
+#
+# END COPYRIGHT
 # -*- coding: utf-8 -*-
```
andreidenissov-cog, line 12: Will we keep this copyright after re-generating with protoc?

neuro_san/api/grpc/chat_pb2_grpc.py
```diff
@@ -1,3 +1,14 @@
...
+# All Rights Reserved.
+# Issued under the Academic Public License.
+#
+# You can be released from the terms, and requirements of the Academic Public
+# License by purchasing a commercial license.
+# Purchase of a commercial license is mandatory for any use of the
+# neuro-san SDK Software in commercial settings.
+#
+# END COPYRIGHT
 # Generated by the gRPC Python protocol compiler plugin. DO NOT EDIT!
```
andreidenissov-cog, line 12: Same question here.

neuro_san/api/grpc/mime_data.proto
```diff
@@ -1,4 +1,15 @@
...
+// Issued under the Academic Public License.
+//
+// You can be released from the terms, and requirements of the Academic Public
+// License by purchasing a commercial license.
+// Purchase of a commercial license is mandatory for any use of the
+// neuro-san SDK Software in commercial settings.
+//
+// END COPYRIGHT
+
 // Copyright © 2023 Cognizant Digital Business, Evolutionary AI. All rights reserved.
```
ofrancon: And here too

neuro_san/api/grpc/agent.proto
```diff
@@ -1,4 +1,15 @@
...
+// Issued under the Academic Public License.
+//
+// You can be released from the terms, and requirements of the Academic Public
+// License by purchasing a commercial license.
+// Purchase of a commercial license is mandatory for any use of the
+// neuro-san SDK Software in commercial settings.
+//
+// END COPYRIGHT
+
 // Copyright © 2024 Cognizant Digital Business, Evolutionary AI. All rights reserved.
```
ofrancon: Why do we have this line? Shouldn't you delete it, because it's replaced by the new block above?

neuro_san/api/grpc/chat.proto
```diff
@@ -1,4 +1,15 @@
...
+// Issued under the Academic Public License.
+//
+// You can be released from the terms, and requirements of the Academic Public
+// License by purchasing a commercial license.
+// Purchase of a commercial license is mandatory for any use of the
+// neuro-san SDK Software in commercial settings.
+//
+// END COPYRIGHT
+
 // Copyright © 2023 Cognizant Digital Business, Evolutionary AI. All rights reserved.
```
ofrancon: Here too, this line should be deleted

neuro_san/api/grpc/chat_pb2.py
d1donlydfink: Not likely. But the update_copyright script is easy enough to run.

neuro_san/api/grpc/chat.proto
d1donlydfink: Great catches! Thanks!

d1donlydfink: They all happened as part of a combination of cut/paste from elsewhere + not having an END COPYRIGHT line that the tools know how to replace.   Only happened in the 3 proto files.

PR #39, opened 2025-01-03
Move the guts to an internals package (imports only)

d1donlydfink:
This pretty much all file moving and subsequent import repairs.
The motivation here is to make it clearer from a file-organization standpoint what things a user of this library
should concern themselves with when they look at the top-level of code.

What remains at the top:
* api
* client
* coded_tools
* deploy
* interfaces
* internals (where everything else goes)
* registries
* service
* session

Inside the internals are now:
* chat
* errors
* graph
* interfaces
* journals
* messages
* run_context
* utils

It's possible more stuff might get shuffled around, but this is the first stab.


README.md
```diff
@@ -137,32 +137,25 @@ That gRPC service is implemented (client and server) using this interface:
...
-    Think of this as seeing all of the internal "thinking" that the agents
-    are doing talking to each other
-
-* reset()
-
-    This basically says restart this chat from nothing as if there were no
-    chat history or context
+    ChatMessages of various types will come back over the stream.
+    Anything of type AI is the front-man answering you on behalf of the rest of
+    its agent posse, so this is the kind you want to pay the most attention to.
```
d1donlydfink, line 156: Change the README to reflect the modern interface.
What's nice is that with this internals shuffling, no mentions of file paths change - by design.

neuro_san/client/agent_cli.py
```diff
@@ -23,9 +23,9 @@
 from timedinput import timedinput
 
 from neuro_san.client.agent_session_factory import AgentSessionFactory
-from neuro_san.messages.chat_message_type import ChatMessageType
+from neuro_san.internals.messages.chat_message_type import ChatMessageType
+from neuro_san.internals.utils.file_of_class import FileOfClass
 from neuro_san.session.agent_session import AgentSession
-from neuro_san.utils.file_of_class import FileOfClass
```
d1donlydfink, line 28: Import changes.

neuro_san/internals/run_context/utils/external_tool_adapter.py
```diff
@@ -18,6 +18,7 @@
 from urllib.parse import ParseResult
 from urllib.parse import urlparse
 
+# The only reach-around from internals outward.
```
ofrancon, line 21: And... is that fine or do you plan on reworking it?

d1donlydfink: No it's fine.

PR #41, opened 2025-01-07
UN-2760 Consolidate looking for response message type into a single method

d1donlydfink:
Create a ChatMessageType.from_response_type() method which handles all kinds of cases on the client side
when getting a response from the stream.   I believe this was something @ofrancon was looking for in a previous (mondo) PR. Should make it a little easier for other coded clients as well.

Also: Use the latest leaf-common with a raw grpc response streaming fix.

These and all the previous fixes were tested against OF in the matrix:
stream vs poll
direct vs service vs mdserver

### DIFF
--- a/requirements-private.txt
+++ b/requirements-private.txt
@@ -9,7 +9,7 @@
 # DEF 07/26/2022: Due to bug in codefresh private repo plugin, this leaf-common
 # comes in from top-level requirements.txt file for now.
 # Public repos
-git+https://${LEAF_SOURCE_CREDENTIALS}@github.com/leaf-ai/leaf-common.git@1.2.15
+git+https://${LEAF_SOURCE_CREDENTIALS}@github.com/leaf-ai/leaf-common.git@1.2.16
 
 # Private repos
 git+https://${LEAF_PRIVATE_SOURCE_CREDENTIALS}@github.com/leaf-ai/leaf-server-common.git@0.1.15


neuro_san/client/agent_cli.py
```diff
@@ -344,14 +344,8 @@ def stream_once(self, state: Dict[str, Any]) -> Dict[str, Any]:
...
-            response_message_type: str = response.get("type", "UNKNOWN_MESSAGE_TYPE")
-            message_type: int = ChatMessageType.UNKNOWN_MESSAGE_TYPE
-            try:
-                # Normal case: We have a 1:1 mapping of ChatMessageType to what is in grpc def
-                message_type = ChatMessageType[response_message_type]
-            except KeyError as exception:
-                raise ValueError(f"Got message type {response_message_type}."
-                                 " Are ChatMessageType and chat.proto out of sync?") from exception
+            response_type: str = response.get("type")
+            message_type: ChatMessageType = ChatMessageType.from_response_type(response_type)
```
d1donlydfink, line 348: Move the above logic into the new method on ChatMessageType

neuro_san/internals/messages/chat_message_type.py
```diff
@@ -53,6 +54,29 @@ def from_message(cls, base_message: BaseMessage) -> ChatMessageType:
...
+        if isinstance(response_type, ChatMessageType):
+            return response_type
+
+        try:
+            # Normal case: We have a 1:1 mapping of ChatMessageType to what is in grpc def
+            message_type = ChatMessageType[response_type]
+        except KeyError as exception:
+            raise ValueError(f"Got message type {response_type} (type {response_type.__class__.__name__})."
+                             " Are ChatMessageType and chat.proto out of sync?") from exception
+        return message_type
```
d1donlydfink, line 78: Move the logic from agent_cli into here so more coded clients can use it.

neuro_san/internals/messages/chat_message_type.py
```diff
@@ -53,6 +54,29 @@ def from_message(cls, base_message: BaseMessage) -> ChatMessageType:
...
+        :return: The ChatMessageType corresponding to the base_message
+        """
+        message_type: ChatMessageType = ChatMessageType.UNKNOWN_MESSAGE_TYPE
+
+        if response_type is None:
+            # Return early
+            return message_type
+
+        if isinstance(response_type, ChatMessageType):
+            return response_type
```
d1donlydfink, line 70: Add a couple of extra cases that can happen when responses come through.
The bottom one ends up fixing a bug I found in direct streaming.

requirements-private.txt
```diff
@@ -9,7 +9,7 @@
 # DEF 07/26/2022: Due to bug in codefresh private repo plugin, this leaf-common
 # comes in from top-level requirements.txt file for now.
 # Public repos
-git+https://${LEAF_SOURCE_CREDENTIALS}@github.com/leaf-ai/leaf-common.git@1.2.15
+git+https://${LEAF_SOURCE_CREDENTIALS}@github.com/leaf-ai/leaf-common.git@1.2.16
```
d1donlydfink, line 12: Use latest leaf-common with raw grpc streaming fix

neuro_san/client/agent_cli.py
ofrancon: Much nicer, thanks

PR #44, opened 2025-01-13
UN-2772 Redact sly data, as per the communicated security model (and then some)

d1donlydfink:
We have been communicating that the sly_data does not travel outside of the local network of the agent,
however when I first implemented external agent calls, I needed to pass sly_data, so the first implementation
ignored it.  I could do that because I was the only user.  No more.

* Implement a SlyDataRedactor policy class which allows specs that call external agent networks to allow certain bits of their sly_data through to the external agent.  Some examples:
    * To allow all sly data through (dumb idea, but brute force is possible):
        ```
        "allow": {
            "sly_data": true
        }
        ```
    *  To allow specific sly_data keys through:
        ```
        "allow": {
            "sly_data": {
                "allow_key": true,
                "deny_key": false
            }
        }
        ```
       Note that you don't really have to explicitly set anything to False, as by default keys not mentioned do not go through, but you at least have the option to be as explicit as you want.
    *  To allow translation of key names to have the same value:
        ```
        "allow": {
            "sly_data": {
                "allow_key": "translated_key",
                "login": "user_id"
            }
        }
        ```
    * And, the best part - by default if you don't have any of this specified in your agent_spec, nothing gets through.  This means you don't have to do anything to stay secure, but you definitely have to do *something explicit* to maybe be insecure.
* Only use the SlyDataRedactor when creating ExternalTools.  It's possible we might elect to do this for other kinds of tools as well, but it's overkill to do it for everything.
* Create a notion of an AgentSpecProvider which both RunContext and ToolCaller implement.  This is needed to avoid circular dependencies of those two knowing about each other.

So far the only agents I know of that are using sly_data are the opportunity_finder_dev_loop (which I tested against) and the analytics stuff that @deepsaia and @andreidenissov-cog put together.  It's possible @babakatwork and @hshahrzad 's recent experiment might, but I kinda doubt it.  Better to do this sooner to nip the fan-out in the bud.  That is, make 1C people be explicit from the get-go when they start to use it (which is why I am doing this now).

Also worth noting that more keys other than "sly_data" will be coming to the "allow" block when we get to specific message types being sent back to the client. 🔮

### DIFF
--- a/tests/neuro_san/test_sly_data_redactor.py
+++ b/tests/neuro_san/test_sly_data_redactor.py
@@ -12,8 +12,6 @@
 from typing import Any
 from typing import Dict
 
-import json
-
 from unittest import TestCase
 
 from neuro_san.internals.graph.registry.sly_data_redactor import SlyDataRedactor
@@ -52,7 +50,7 @@ class TestSlyDataRedactor(TestCase):
             "not_mentioned": -1,
         }
 
-        redacted = redactor.filter_config(sly_data)
+        redacted: Dict[str, Any] = redactor.filter_config(sly_data)
 
         self.assertIsNotNone(redacted.get("yes"))
         self.assertIsNone(redacted.get("no"))
@@ -75,7 +73,7 @@ class TestSlyDataRedactor(TestCase):
             "not_mentioned": -1,
         }
 
-        redacted = redactor.filter_config(sly_data)
+        redacted: Dict[str, Any] = redactor.filter_config(sly_data)
 
         self.assertIsNotNone(redacted.get("yes"))
         self.assertIsNotNone(redacted.get("no"))
@@ -98,7 +96,7 @@ class TestSlyDataRedactor(TestCase):
             "not_mentioned": -1,
         }
 
-        redacted = redactor.filter_config(sly_data)
+        redacted: Dict[str, Any] = redactor.filter_config(sly_data)
 
         self.assertIsNone(redacted.get("yes"))
         self.assertIsNone(redacted.get("no"))
@@ -118,7 +116,7 @@ class TestSlyDataRedactor(TestCase):
             "not_mentioned": -1,
         }
 
-        redacted = redactor.filter_config(sly_data)
+        redacted: Dict[str, Any] = redactor.filter_config(sly_data)
 
         self.assertIsNone(redacted.get("yes"))
         self.assertIsNone(redacted.get("no"))
@@ -144,7 +142,7 @@ class TestSlyDataRedactor(TestCase):
             "not_mentioned": -1,
         }
 
-        redacted = redactor.filter_config(sly_data)
+        redacted: Dict[str, Any] = redactor.filter_config(sly_data)
 
         self.assertIsNone(redacted.get("yes"))
         self.assertIsNone(redacted.get("no"))


neuro_san/internals/graph/registry/agent_tool_registry.py
```diff
@@ -167,7 +168,11 @@ def create_agent_tool(self, parent_run_context: RunContext,
             if not ExternalToolAdapter.is_external_agent(name):
                 raise ValueError(f"No agent_tool_spec for {name}")
 
-            agent_tool = ExternalTool(parent_run_context, journal, name, arguments, sly_data)
+            # For external tools, we want to redact the sly data based on
+            # the calling/parent's agent specs.
+            redacted_sly_data: Dict[str, Any] = self.redact_sly_data(parent_run_context, sly_data)
+
+            agent_tool = ExternalTool(parent_run_context, journal, name, arguments, redacted_sly_data)
```
d1donlydfink, line 175: Redact the sly data as per the spec of the parent run context when we go to call ExternalTools.
This prevents private data leakage at the source.

It's conceivable we might want to do this for other tools as well, but the security model at least depends on this bit working correctly: Go to another server? Don't spill the beans.

neuro_san/internals/graph/registry/agent_tool_registry.py
```diff
@@ -249,3 +254,19 @@ def merge_args(self, llm_args: Dict[str, Any], agent_tool_spec: Dict[str, Any])
...
+        :param sly_data: The internal representation of the sly_data to be redacted
+        :return: A new sly_data dictionary, redacted as per the parent spec
+        """
+        parent_spec: Dict[str, Any] = None
+        if parent_run_context is not None:
+            parent_spec = parent_run_context.get_agent_tool_spec()
+
+        redactor = SlyDataRedactor(parent_spec)
+        redacted: Dict[str, Any] = redactor.filter_config(sly_data)
+        return redacted
```
d1donlydfink, line 272: Some local policy for using the SlyDataRedactor by getting spec info via the RunContext

neuro_san/internals/graph/registry/sly_data_redactor.py
```diff
@@ -0,0 +1,79 @@
...
+from leaf_common.config.config_filter import ConfigFilter
+from leaf_common.parsers.dictionary_extractor import DictionaryExtractor
+
+
+# pylint: disable=too-few-public-methods
+class SlyDataRedactor(ConfigFilter):
+    """
+    An implementation of the ConfigFilter interface which redacts sly data
+    based on calling-agent specs.
+    """
```
d1donlydfink, line 24: Policy for redacting sly data based on the caller's agent specs.

neuro_san/internals/graph/registry/sly_data_redactor.py
```diff
@@ -0,0 +1,78 @@
...
+        """
+        self.agent_tool_spec: Dict[str, Any] = calling_agent_tool_spec
+
+    def filter_config(self, basis_config: Dict[str, Any]) -> Dict[str, Any]:
+        """
+        :param basis_config: Source of the sly_data to redact.
+        :return: A new sly_data dictionary with proper redactions per the "allow.sly_data"
+                dictionary on the agent spec.  If no such key exists, then no sly_data
+                gets through to the agent to be called.
+        """
```
d1donlydfink, line 41: Policy method for redacting sly data.

neuro_san/internals/graph/registry/sly_data_redactor.py
```diff
@@ -0,0 +1,78 @@
...
+        empty: Dict[str, Any] = {}
+
+        extractor = DictionaryExtractor(self.agent_tool_spec)
+        allow_dict: Dict[str, Any] = extractor.get("allow.sly_data", empty)
```
d1donlydfink, line 45: Look for a dictionary in the agent spec called "allow".
Within that dictionary, look for another dictionary called "sly_data" which contains the controls.

neuro_san/internals/graph/registry/sly_data_redactor.py
```diff
@@ -0,0 +1,78 @@
...
+
+        # Recall empty dictionaries evaluate to False (as well as boolean values)
+        if not bool(allow_dict):
+            # By default we don't let anything through
+            return empty
```
d1donlydfink, line 50: If there was no dictionary in the allow section, we do not let anything through by default.
The idea here is if you do nothing, you data is secure.

neuro_san/internals/graph/registry/sly_data_redactor.py
```diff
@@ -0,0 +1,78 @@
...
+
+        if isinstance(allow_dict, bool) and bool(allow_dict):
+            # The value is a simple True, so let everything through.
+            return basis_config
```
d1donlydfink, line 54: An easy (but very dumb!) way to let everything in the sly data through would be to have this in your spec:

"allow": {
    "sly_data": true
}

neuro_san/internals/graph/registry/sly_data_redactor.py
```diff
@@ -0,0 +1,78 @@
...
+
+        if not bool(basis_config) or not isinstance(basis_config, Dict):
+            # There is no dictionary content, so nothing to redact
+            return empty
```
d1donlydfink, line 58: If there is nothing in the sly_data passed in as basis_config, then our job is easy - there is no sly_data to pass on.

neuro_san/internals/graph/registry/sly_data_redactor.py
```diff
@@ -0,0 +1,78 @@
...
+
+        # Got rid of all the easy cases.
+        # Now leaf through the keys of the dictionaries.
+        # For now, just do top-level keys. Can get more complicated later if need be.
+        redacted: Dict[str, Any] = {}
+        for source_key, dest_key in allow_dict.items():
```
d1donlydfink, line 64: Now we loop through all the keys in that allow > sly_data dictionary.

neuro_san/internals/graph/registry/sly_data_redactor.py
```diff
@@ -0,0 +1,78 @@
...
+
+            source_value: Any = basis_config.get(source_key)
+            if source_value is None:
+                # We have an allowance, but no data, so nothing to do for this key.
+                continue
```
d1donlydfink, line 69: If we do not have a value in the actual sly data, just pass it up. Not passing along a None value in the sly_data dictionary should be OK.

neuro_san/internals/graph/registry/sly_data_redactor.py
```diff
@@ -0,0 +1,78 @@
...
+
+            if isinstance(dest_key, str):
+                # Translate the key
+                redacted[dest_key] = source_value
```
d1donlydfink, line 73: If the value is a string, treat it as a translation to the interface for the calling dictionary and rename the key.
That is:
```
"allow": {
    "sly_data": {
         "login": "user_id"
    }
}
```

This would translate the sly-data key called "login" to an external key called "user_id" with the same value.
This begins to allow for translational interfaces when sly_data specs aren't quite the same between agent networks.

neuro_san/internals/graph/registry/sly_data_redactor.py
```diff
@@ -0,0 +1,78 @@
...
+            elif isinstance(dest_key, bool) and bool(dest_key):
+                # Use the same key and the same value in the explicit allow
+                redacted[source_key] = source_value
```
d1donlydfink, line 76: If the value is a boolean, treat it as a license to copy the key and value to the redacted version of the sly_data.
That is:
```
"allow": {
    "sly_data": {
         "login": true
    }
}
```
Says: "The value for 'login' needs to travel outside my agent network, and I explicitly approve".

neuro_san/internals/run_context/interfaces/agent_spec_provider.py
```diff
@@ -0,0 +1,26 @@
...
+# END COPYRIGHT
+from typing import Any
+from typing import Dict
+
+
+# pylint: disable=too-few-public-methods
+class AgentSpecProvider:
+    """
+    Interface for an entity to return a copy of the agent spec that pertains to it.
+    """
```
d1donlydfink, line 20: This interface allows us to break cyclic imports.

neuro_san/internals/run_context/interfaces/run_context.py
```diff
@@ -77,3 +79,9 @@ async def delete_resources(self, parent_run_context: RunContext = None):
                         but perhaps not.  Default is None
         """
         raise NotImplementedError
+
+    def get_agent_tool_spec(self) -> Dict[str, Any]:
+        """
+        :return: the dictionary describing the data-driven agent
+        """
+        raise NotImplementedError
```
d1donlydfink, line 87: RunContexts can now give up the agent_spec associated with their agents.

neuro_san/internals/run_context/interfaces/tool_caller.py
```diff
@@ -12,11 +12,12 @@
 from typing import Any
 from typing import Dict
 
+from neuro_san.internals.run_context.interfaces.agent_spec_provider import AgentSpecProvider
 from neuro_san.internals.run_context.interfaces.agent_tool_factory import AgentToolFactory
 from neuro_san.internals.run_context.interfaces.run import Run
 
 
-class ToolCaller:
+class ToolCaller(AgentSpecProvider):
```
d1donlydfink, line 20: ToolCallers are also AgentSpecProviders.  It already implements the get_agent_tool_spec() method.

neuro_san/internals/run_context/langchain/langchain_run_context.py
```diff
@@ -385,3 +385,12 @@ async def delete_resources(self, parent_run_context: RunContext = None):
...
         self.recent_human_message = None
+
+    def get_agent_tool_spec(self) -> Dict[str, Any]:
+        """
+        :return: the dictionary describing the data-driven agent
+        """
+        if self.tool_caller is None:
+            return None
+
+        return self.tool_caller.get_agent_tool_spec()
```
d1donlydfink, line 396: Have LangChainRunContext implement get_agent_tool_spec()

neuro_san/internals/run_context/openai/openai_run_context.py
```diff
@@ -227,3 +227,12 @@ async def delete_resources(self, parent_run_context: RunContext = None):
...
                 self.thread_id = None
+
+    def get_agent_tool_spec(self) -> Dict[str, Any]:
+        """
+        :return: the dictionary describing the data-driven agent
+        """
+        if self.tool_caller is None:
+            return None
+
+        return self.tool_caller.get_agent_tool_spec()
```
d1donlydfink, line 238: Have OpenAInRunContext implement get_agent_tool_spec(), even though we really don't use this stuff any more.

tests/neuro_san/test_sly_data_redactor.py
```diff
@@ -0,0 +1,153 @@
...
+
+from unittest import TestCase
+
+from neuro_san.internals.graph.registry.sly_data_redactor import SlyDataRedactor
+
+
+class TestSlyDataRedactor(TestCase):
+    """
+    Unit tests for SlyDataRedactor class.
+    """
```
d1donlydfink, line 23: Add some unit tests

PR #46, opened 2025-01-15
UN-2770 Better behavior when there is no command given

d1donlydfink:
Babak had a very cryptic problem when there was no command specified in his agent hocon.
This smooths over that behavior to:
* Provide a reasonable default command when none is present
* Provide a slightly better exception message with more context for the original symptom of the Pydantic ValidationError

### DIFF
--- a/neuro_san/internals/run_context/langchain/langchain_run_context.py
+++ b/neuro_san/internals/run_context/langchain/langchain_run_context.py
@@ -19,7 +19,7 @@ import traceback
 
 from openai import APIError
 
-from pydantic_core._pydantic_core import ValidationError
+from pydantic_core import ValidationError
 
 from langchain.agents import Agent
 from langchain.agents import AgentExecutor


neuro_san/internals/graph/tools/branch_tool.py
```diff
@@ -174,7 +174,7 @@ def get_command(self) -> str:
         :return: A string describing the objective of the component.
         """
         agent_spec = self.get_agent_tool_spec()
-        return agent_spec.get("command")
+        return agent_spec.get("command", "Perform your instructions to the best of your ability.")
```
d1donlydfink, line 177: Provide a reasonable default command if none is specified in the agent spec.

Usually, people tend to put all their hopes and dreams into the instructions, and then the command to set things in motion is an afterthought. (I have done this before myself, though without the Pydantic error that @babakatwork found).

I think it's still a good idea to be able to customize the command, but providing a reasonable default is also, well, reasonable.

neuro_san/internals/run_context/langchain/langchain_run_context.py
```diff
@@ -200,7 +202,11 @@ async def submit_message(self, user_message: str) -> Run:
         """
         # Contruct a human message out of the text of the user message
         # Don't add this to the chat history yet.
-        self.recent_human_message = HumanMessage(user_message)
+        try:
+            self.recent_human_message = HumanMessage(user_message)
+        except ValidationError as exception:
+            message = f"ValidationError in {self.assistant_name} with message: {user_message}"
+            raise ValueError(message) from exception
```
d1donlydfink, line 209: Provide better context as to just which agent failed and with what message contents when the Pydantic ValidationError occurs - easier debugging for next time.

PR #47, opened 2025-01-16
UN-2786 Connectivity streaming to enable "constellation" view

d1donlydfink:
Create a ConnectivityReporter which sends connectivity information over the stream as AgentFramework messages.
This will enable client-side reconstruction of the agent network graph (as much as the server wants it to know, at least)
without the client-side needing access to the hocon files that live on the server.

To do this:
* Add a write_message() method to the Journal interface which will take a langchain BaseMessage. This will get used a lot in the future.
* Add a call to the new ConnectivityReporter in DataDrivenChatSession.chat() to send these messages before anything else in the stream
* Specifically skip over the connectivity messages in the streaming text client.
* Add a test
* Fix a bug in the specification of where coded_tools lives that was introduced in an earlier PR
* Also fix a case in AgentToolRegistry when PYTHONPATH was not set (like in the build)

@dsargent and @swensel : Please pay particular attention to the class comment in ConnectivityReporter
as it describes how messages are going to come over the stream to enable a "constellation" view that scales beyond the current flask app.  

Theoretically with this change, the server is now sending everything a client would need to know over the stream to recreate the information that the flask app has. That's not to say how we deliver the information will not be improved (it will), but all the same information _should_ be available now.

---

Afterthought:  After thinking about the implications of this change over the weekend there are a couple of things that I will move to change in imminent, subsequent PR(s):
* In order to start to get the the connectivity information you actually need to send some user input over the streaming_chat() method.   Seems cart-before-the-horse.
* Every single client will get connectivity information sent to them whether they want it or not.
* Connectivity information is sent in the form of text messages, forcing the client that cares to parse JSON in order to get the information they need.

I think the way to address these is to create a new Connectivity() method on the service.  Clients that want the information (if available) can use it. It's not a requirement to do so, however for the simplest clients (like agent_cli.py), thus not increasing bandwidth consumption unnecessarily. Such a method can have its own formalized messages, meaning the clients will not have to parse JSON business from text.

### DIFF
--- a/neuro_san/internals/graph/registry/agent_tool_registry.py
+++ b/neuro_san/internals/graph/registry/agent_tool_registry.py
@@ -232,7 +232,7 @@ class AgentToolRegistry(AgentToolFactory):
         # The primary way to identify a front-man is that its function has no parameters.
         for name, agent_spec in self.agent_spec_map.items():
             function: Dict[str, Any] = agent_spec.get("function")
-            if function.get("parameters") is None:
+            if function is not None and function.get("parameters") is None:
                 front_men.append(name)
 
         if len(front_men) == 0:
--- a/neuro_san/internals/run_context/langchain/langchain_openai_function_tool.py
+++ b/neuro_san/internals/run_context/langchain/langchain_openai_function_tool.py
@@ -18,6 +18,7 @@ from typing import Type
 
 import traceback
 
+from pydantic_core import ValidationError
 from langchain_core.tools import BaseTool
 
 from neuro_san.internals.run_context.interfaces.tool_caller import ToolCaller
@@ -71,6 +72,44 @@ class LangChainOpenAIFunctionTool(BaseTool):
     tool_caller: Optional[ToolCaller] = None
 
     @staticmethod
+    def verify_function_json(function_json: Dict[str, Any]):
+        """
+        :param function_json: The function json description of the tool.
+            Must have at least:
+                * A name
+                * A description
+                * A parameters dictionary with at least one entry in its properties dict
+        """
+        message: str = ""
+        name: str = function_json.get("name")
+        if name is None:
+            message = "function dictionary has no 'name' defined."
+            raise ValueError(message)
+
+        if function_json.get("description") is None:
+            message = f"Function for {name} has no description.\n"
+
+        parameters: Dict[str, Any] = function_json.get("parameters")
+        if parameters is None:
+            message = f"Function for {name} has no parameters defined.\n"
+        else:
+            if parameters.get("type") is None:
+                message = f"Function for {name} needs to have a parameters.type set to 'object'.\n"
+            properties: Dict[str, Any] = parameters.get("properties")
+            if properties is None:
+                message = f"Function for {name} needs to have a properties dictionary as part of its parameters.\n"
+            elif not isinstance(properties, Dict) or len(properties.keys()) == 0:
+                message = f"Function for {name} needs to have at least one argument dictionary in its properties.\n"
+
+        if len(message) > 0:
+            message += """
+This most often happens when calling an /external agent for the first time
+and the hocon file for the agent network does not have a full function definition
+specified for its front man for it to be called by another agent.
+"""
+            raise ValueError(message)
+
+    @staticmethod
     def from_function_json(function_json: Dict[str, Any],
                            tool_caller: ToolCaller) \
             -> LangChainOpenAIFunctionTool:
@@ -82,7 +121,16 @@ class LangChainOpenAIFunctionTool(BaseTool):
         # the actual description of the function_json that we expect to
         # be passed in which we expect to conform to an OpenAI function.
         # definition.
-        tool = LangChainOpenAIFunctionTool(**function_json)
+        LangChainOpenAIFunctionTool.verify_function_json(function_json)
+        try:
+            tool = LangChainOpenAIFunctionTool(**function_json)
+        except ValidationError as exception:
+            message: str = f"""
+Could not create tool to call extenal agent {function_json.get("name")}.
+It's function_json is described thusly:
+{function_json}
+"""
+            raise ValueError(message) from exception
 
         # Check for external tools.
         name: str = function_json.get("name")


neuro_san/internals/chat/data_driven_chat_session.py
```diff
@@ -130,6 +131,10 @@ async def chat(self, user_input: str, sly_data: Dict[str, Any]) -> Iterator[Dict
         if sly_data is not None:
             self.sly_data.update(sly_data)
 
+        # Report on network connectivity for clients that care.
+        connectivity_reporter = ConnectivityReporter(self.registry, self.front_man, self.journal)
+        await connectivity_reporter.report_network_connectivity()
```
d1donlydfink, line 136: Before any agents are called, report on the graph connectivity of the network,
so more sophisticated clients like the "constellation" view have a chance of setting up
their picture of the network before any activity comes back.

neuro_san/internals/chat/connectivity_reporter.py
```diff
@@ -0,0 +1,121 @@
...
+from neuro_san.internals.graph.tools.front_man import FrontMan
+from neuro_san.internals.journals.journal import Journal
+from neuro_san.internals.messages.agent_framework_message import AgentFrameworkMessage
+
+
+class ConnectivityReporter:
+    """
+    A class that knows how to report the connectivity of an entire
+    AgentToolRegistry to a particular Journal
+    """
```
d1donlydfink, line 53: Add a policy class that knows how to report connectivity information about an entire agent graph
back over the stream of messages.  Hopefully the comments above are detailed enough as to what is happening here.

It is likely we will eventually need to change the default sense of connectivity visibility to False for security reasons, at least on the front man. Discuss.

neuro_san/internals/graph/registry/agent_tool_registry.py
```diff
@@ -130,15 +130,23 @@ def register(self, agent_spec: Dict[str, Any]):
...
         extractor = FieldExtractor()
         name = extractor.get_field(agent_spec, "function.name")
         if name is None:
             name = agent_spec.get("name")
 
-        if self.first_agent is None:
-            self.first_agent = name
-
-        self.agent_spec_map[name] = agent_spec
+        return name
```
d1donlydfink, line 154: Not the clearest diff.  The logic which gets the name from the agent spec has some policy associated with it,
and that is now broken out into its own method.

neuro_san/internals/journals/compound_journal.py
```diff
@@ -56,3 +58,12 @@ def add_journal(self, journal: Journal):
...
         self.journals.append(journal)
+
+    async def write_message(self, message: BaseMessage, origin: str = None):
+        """
+        Writes a BaseMessage entry into the journal
+        :param message: The BaseMessage instance to write to the journal
+        :param origin: A string describing the originating agent of the information
+        """
+        for journal in self.journals:
+            await journal.write_message(message, origin)
```
d1donlydfink, line 69: Implement a new write_message() method on all of the concrete implementations of the Journal interface.
This will allow us to stream fine-grained information about langchian messages sent to/from/by specific agents to any client that cares to listen.

neuro_san/internals/journals/journal.py
```diff
@@ -30,3 +32,11 @@ def get_logs(self) -> List[Any]:
...
         """
         raise NotImplementedError
+
+    async def write_message(self, message: BaseMessage, origin: str = None):
+        """
+        Writes a BaseMessage entry into the journal
+        :param message: The BaseMessage instance to write to the journal
+        :param origin: A string describing the originating agent of the information
+        """
+        raise NotImplementedError
```
d1donlydfink, line 42: Add write_message() to the Journal interface.

neuro_san/internals/journals/message_journal.py
```diff
@@ -45,11 +47,22 @@ async def write(self, entry: Union[str, bytes]):
             entry = entry.decode('utf-8')
 
         legacy = LegacyLogsMessage(content=entry)
-        message_dict: Dict[str, Any] = convert_to_chat_message(legacy)
-        await self.hopper.put(message_dict)
+        await self.write_message(legacy)
```
d1donlydfink, line 50: Use the new write_message() method to do the dirty work.

neuro_san/internals/journals/message_journal.py
```diff
@@ -45,11 +47,28 @@ async def write(self, entry: Union[str, bytes]):
...
+
+        # Handle the origin information
+        if origin is not None:
+            origin_list: List[str] = origin
+            if isinstance(origin, str):
+                # Handle the case of a single string
+                origin_list = [origin]
+            message_dict["origin"] = origin_list
+
+        await self.hopper.put(message_dict)
```
d1donlydfink, line 74: General means by which the asynchronous innards of neuro-san chat can report a BaseMessage with an origin to the client's chat stream.

This method will eventually be key to reporting lower-level agent chat activity when we get there.

neuro_san/internals/journals/text_journal.py
```diff
@@ -41,3 +43,12 @@ def get_logs(self) -> List[str]:
...
         return self.log_content
+
+    async def write_message(self, message: BaseMessage, origin: Union[str, List[str]] = None):
+        """
+        Writes a BaseMessage entry into the journal
+        :param message: The BaseMessage instance to write to the journal
+        :param origin: A string or list of strings describing the originating agent of the information
+        """
+        # Do nothing
+        _ = message
```
d1donlydfink, line 54: When writing to the text-based logs for old-style polling chat, we simply ignore any messages that are intended for streaming clients only.

neuro_san/internals/chat/connectivity_reporter.py
```diff
@@ -0,0 +1,127 @@
...
+        """
+
+        self.registry: AgentToolRegistry = registry
+        self.front_man: FrontMan = front_man
+        self.journal: Journal = journal
+
+    async def report_network_connectivity(self):
+        """
+        Share the connectivity information of the agent network in question
+        """
```
d1donlydfink, line 70: Report the connectivity of the network as a whole.

neuro_san/internals/chat/connectivity_reporter.py
```diff
@@ -0,0 +1,127 @@
...
+        # Report connectivity for each node in the graph
+        for agent_name in agent_names:
+            agent_spec: Dict[str, Any] = self.registry.get_agent_tool_spec(agent_name)
+            await self.report_node_connectivity(agent_spec)
+
+    async def report_node_connectivity(self, agent_spec: Dict[str, Any]):
+        """
+        Share the connectivity information of a single node in the network.
+        :param agent_spec: The agent spec dictionary to report on
+        """
```
d1donlydfink, line 84: Report the connectivity of a single node in the agent network.

neuro_san/client/agent_cli.py
```diff
@@ -376,6 +376,10 @@ def stream_once(self, state: Dict[str, Any]) -> Dict[str, Any]:
                     with open(self.args.thinking_file, "a", encoding="utf-8") as thinking:
                         thinking.write(text)
                         thinking.write("\n")
+                elif message_type == ChatMessageType.AGENT_FRAMEWORK:
+                    # Skip over the connectivity messages for this text-only client
+                    if response.get("origin") is None:
+                        print(f"{text}")
```
d1donlydfink, line 382: Skip over new connectivity messages in the text client.

neuro_san/internals/chat/connectivity_reporter.py
```diff
@@ -0,0 +1,173 @@
...
+        Share the connectivity information of a single node in the network.
+        :param agent_name: The name of the agent spec dictionary to report on
+        :param reported_agents: A list of agents that have been reported already.
+                Prevents cycles.
+        """
+
+        agent_spec: Dict[str, Any] = None
+
+        if not ExternalToolAdapter.is_external_agent(agent_name):
```
d1donlydfink, line 88: We are not moving to know the internals of external agents at this point.
Leave their agent_spec None.

neuro_san/internals/chat/connectivity_reporter.py
```diff
@@ -0,0 +1,173 @@
...
+
+            # This is not an external agent, so get its spec to report on
+            agent_spec: Dict[str, Any] = self.registry.get_agent_tool_spec(agent_name)
+            if agent_spec is None:
+                # The agent referred to by the caller is not actually in the registry.
+                # As a hint, don't report anything, not even an empty tool list.
+                return
```
d1donlydfink, line 95: Do not report tool names that are not listed as proper nodes in the spec.

neuro_san/internals/chat/connectivity_reporter.py
```diff
@@ -0,0 +1,173 @@
...
+
+            # Check to see if this node in the graph actually wants its connectivity
+            # known to the outside world.
+            extractor = DictionaryExtractor(agent_spec)
+            allow_connectivity = bool(extractor.get("allow.connectivity", True))
+            if not bool(allow_connectivity):
+
+                # We are not allowing connectivity reporting from here on down
+                agent_spec = None
```
d1donlydfink, line 104: Any one node is allowed to say "Don't report connectivity of anything below me in a traversal"

neuro_san/internals/chat/connectivity_reporter.py
```diff
@@ -0,0 +1,173 @@
...
+
+        # Compile a tool list of what is referred to by the agent_spec.
+        # For many reasons, this list could be empty.
+        tool_list: List[str] = self.assemble_tool_list(agent_spec)
```
d1donlydfink, line 108: Find the list of tools for any given agent spec. For non-existent agent specs, this returns an empty list.

neuro_san/internals/chat/connectivity_reporter.py
```diff
@@ -0,0 +1,173 @@
...
+        # Report the content of the tools list as a dictionary in JSON.
+        tools_dict: Dict[str, Any] = {
+            "tools": tool_list
+        }
+        content: str = f"```json\n{json.dumps(tools_dict)}```"
+        message = AgentFrameworkMessage(content=content)
+
+        # Report the origin as the agent itself, so any client that receives
+        # the message has the correct context about the tools listed in the content.
+        await self.journal.write_message(message, origin=agent_name)
```
d1donlydfink, line 119: Report a json message in the message text about the tools it uses with the origin of the message being the agent's name for the node being reported.  These might be empty lists.

So far this needs to be in text because our chat message format does not yet support raw unstructured dictionaries.

 I think there might be a protobuf way to send unstructured dictionaries that are not text.
That will require a little research and a protocol change. When I figure that out this can change
and @dsargent can jump for joy, but that's not for this PR just yet.

neuro_san/internals/chat/connectivity_reporter.py
```diff
@@ -0,0 +1,173 @@
...
+        reported_agents.add(agent_name)
+
+        # Recurse for a bread-first search.
+        for tool in tool_list:
+            # Don't report more than once for the same node to avoid cycles.
+            if tool not in reported_agents:
+                await self.report_node_connectivity(tool, reported_agents)
```
d1donlydfink, line 126: Recurse in a breadth-first manner as per the tool_list.
This allows users some control over order of reporting.
Don't report agents more than once - avoid infinite loops caused by cycles.

neuro_san/internals/chat/connectivity_reporter.py
```diff
@@ -0,0 +1,173 @@
...
+
+    @staticmethod
+    def assemble_tool_list(agent_spec: Dict[str, Any]) -> List[str]:
+        """
+        :param agent_spec: The agent spec to assemble a tool list from.
+        :return: A list of tool names referred to by the agent spec.
+        """
```
d1donlydfink, line 133: A method to return a tool list with unique agent names in it.

tests/neuro_san/list_hopper.py
```diff
@@ -0,0 +1,40 @@
...
+from typing import List
+
+from neuro_san.internals.interfaces.async_hopper import AsyncHopper
+
+
+# pylint: disable=too-few-public-methods
+class ListHopper(AsyncHopper):
+    """
+    An AsyncHopper implementation for tests that captures items in a list
+    """
```
d1donlydfink, line 22: A test class for examining messages that are sent by a MessageJournal.

tests/neuro_san/test_connectivity_reporter.py
```diff
@@ -0,0 +1,103 @@
...
+from neuro_san.internals.journals.message_journal import MessageJournal
+from neuro_san.internals.utils.file_of_class import FileOfClass
+
+from tests.neuro_san.list_hopper import ListHopper
+
+
+class TestConnectivityReporter(TestCase):
+    """
+    Unit tests for ConnectivityReporter class.
+    """
```
d1donlydfink, line 33: Some initial tests.  Probably can expand on this over time, but I would rather have something over nothing.

neuro_san/internals/graph/registry/agent_tool_registry.py
```diff
@@ -74,7 +74,7 @@ def determine_agent_tool_path(self, agent_tool_path: str) -> str:
 
         # Try reach-around directory if still nothing to start with
         if agent_tool_path is None:
-            file_of_class = FileOfClass(__file__, "../../coded_tools")
+            file_of_class = FileOfClass(__file__, "../../../coded_tools")
```
d1donlydfink, line 77: Fix a bug that was introduced in the move to neuro_san.internals package.
This was found by a new unit test.

neuro_san/internals/graph/registry/agent_tool_registry.py
```diff
@@ -83,12 +83,17 @@ def determine_agent_tool_path(self, agent_tool_path: str) -> str:
...
-                resolved_path: str = str(Path(one_path).resolve())
-                if resolved_tool_path.startswith(resolved_path) and \
-                        len(resolved_path) > len(best_path):
-                    best_path = resolved_path
+            pythonpath: str = os.environ.get("PYTHONPATH")
+            if pythonpath is None:
+                # Trust what we have already
+                best_path = agent_tool_path
+            else:
+                pythonpath_split = pythonpath.split(":")
```
d1donlydfink, line 91: Another aspect to the fix for unit tests passing on the builds where I guess it does not have PYTHONPATH set (which is actually fine)

PR #48, opened 2025-01-17
UN-2770 Better error messages when external functions do not have proper functions defined

d1donlydfink:
Provide some details as to why a particular function_json is not up to snuff when creating a FunctionTool to call the agent from another.

### DIFF
--- a/neuro_san/internals/run_context/langchain/langchain_openai_function_tool.py
+++ b/neuro_san/internals/run_context/langchain/langchain_openai_function_tool.py
@@ -18,6 +18,7 @@ from typing import Type
 
 import traceback
 
+from pydantic_core import ValidationError
 from langchain_core.tools import BaseTool
 
 from neuro_san.internals.run_context.interfaces.tool_caller import ToolCaller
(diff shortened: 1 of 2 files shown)


neuro_san/internals/run_context/langchain/langchain_openai_function_tool.py
```diff
@@ -70,6 +71,44 @@ class LangChainOpenAIFunctionTool(BaseTool):
...
+            elif not isinstance(properties, Dict) or len(properties.keys()) == 0:
+                message = f"Function for {name} needs to have at least one argument dictionary in its properties.\n"
+
+        if len(message) > 0:
+            message += """
+This most often happens when calling an /external agent for the first time
+and the hocon file for the agent network does not have a full function definition
+specified for its front man for it to be called by another agent.
+"""
+            raise ValueError(message)
```
d1donlydfink, line 110: Provide some details as to why a particular function_json is not up to snuff when creating a FunctionTool to call the agent from another.  This most often happens when someone is calling one agent network from another for the first time.  This would have helped @babakatwork and @hshahrzad as well as the 1C people.

neuro_san/internals/run_context/langchain/langchain_openai_function_tool.py
```diff
@@ -82,7 +121,16 @@ def from_function_json(function_json: Dict[str, Any],
...
+        LangChainOpenAIFunctionTool.verify_function_json(function_json)
+        try:
+            tool = LangChainOpenAIFunctionTool(**function_json)
+        except ValidationError as exception:
+            message: str = f"""
+Could not create tool to call extenal agent {function_json.get("name")}.
+It's function_json is described thusly:
+{function_json}
+"""
+            raise ValueError(message) from exception
```
d1donlydfink, line 133: This is the initial catch-all for Pydantic validation problems.
The original exception is especially cryptic, so wrap the exceptions together with the outer one being a little more verbose.

Ideally, the new verification above shouldn't have things get this far. But never say never.

PR #54, opened 2025-01-25
Only stream the last chat message from the front man before closing out the queue

d1donlydfink:
In debugging with @dsargent , we noticed that the front-man chat history was being streamed back at the end of the stream.  We really only want the last message I think.
This actually clean up the agent_cli output at the end quite a bit.


neuro_san/internals/chat/data_driven_chat_session.py
```diff
@@ -171,14 +171,13 @@ async def streaming_chat(self, user_input: str, sly_data: Dict[str, Any]):
...
-                # The consumer await-s for self.queue.get()
-                await self.queue.put(chat_message)
-                self.last_streamed_index = index
+        message_list: List[Dict[str, Any]] = list(chat_messages)
+        index: int = len(message_list) - 1
+        chat_message: Dict[str, Any] = message_list[index]
+
+        # The consumer await-s for self.queue.get()
+        await self.queue.put(chat_message)
+        self.last_streamed_index = index
```
d1donlydfink, line 191: Instead of potentially streaming the entire chat history back, only send the very last message.
I was perhaps overzealous

PR #57, opened 2025-01-28
UN-2808 Send origin information over stream for LegacyLogs messages

d1donlydfink:
Origin information is a list of strings that tell us where in the agent hierarchy the message came from.
@dsargent had mentioned that having this information even in the LegacyLogs messages would be helpful to his parsing apparatus, so I go there first.

The agent_cli output is modified so you can see which agent is doing the talking from the origin information streamed back to the client.

There is a question at the end for neuro-san client-keepers like @ofrancon , @dsargent and @swensel 

Tested: OF pipeline and esp_decision_assistant

### DIFF
--- a/neuro_san/internals/chat/data_driven_chat_session.py
+++ b/neuro_san/internals/chat/data_driven_chat_session.py
@@ -89,9 +89,9 @@ class DataDrivenChatSession(ChatSession):
 
         run_context: RunContext = RunContextFactory.create_run_context(None, None,
                                                                        invocation_context=self.invocation_context)
-        await self.journal.write("setting up chat agent(s)...", run_context.get_origin())
-
         self.front_man = self.registry.create_front_man(self.journal, self.sly_data, run_context)
+
+        await self.journal.write("setting up chat agent(s)...", self.front_man.get_origin())
         await self.front_man.create_resources()
 
     async def chat(self, user_input: str,
--- a/neuro_san/internals/graph/tools/calling_tool.py
+++ b/neuro_san/internals/graph/tools/calling_tool.py
@@ -115,6 +115,12 @@ context with which it will proces input, essentially telling it what to do.
         """
         return self.factory
 
+    def get_origin(self) -> List[str]:
+        """
+        :return: The list of strings describing the origin associated with this node
+        """
+        return self.run_context.get_origin()
+
     async def create_resources(self, component_name: str = None,
                                specific_instructions: str = None):
         """


neuro_san/internals/run_context/factory/run_context_factory.py
```diff
@@ -74,11 +74,11 @@ def create_run_context(parent_run_context: RunContext,
             run_context = OpenAIRunContext(default_llm_config, parent_run_context,
                                            tool_caller, use_invocation_context)
         elif context_type.startswith("langchain"):
-            run_context = LangChainRunContext(default_llm_config, tool_caller,
-                                              use_invocation_context)
+            run_context = LangChainRunContext(default_llm_config, parent_run_context,
+                                              tool_caller, use_invocation_context)
```
d1donlydfink, line 78: Always send in the parent_run_context to the LangChainRunContext.
This is actually kind of nice because now OpenAI and LangChain constructors are symmetrical.

neuro_san/internals/run_context/interfaces/agent_tool_factory.py
```diff
@@ -56,3 +56,10 @@ def get_agent_tool_path(self) -> str:
         :return: The path under which tools for this registry should be looked for.
         """
         raise NotImplementedError
+
+    def get_name_from_spec(self, agent_spec: Dict[str, Any]) -> str:
+        """
+        :param agent_spec: A single agent to register
+        :return: The agent name as per the spec
+        """
+        raise NotImplementedError
```
d1donlydfink, line 65: Add to the interface. Method itself is already implemented on the concrete implementation.

neuro_san/internals/run_context/interfaces/run_context.py
```diff
@@ -93,3 +93,10 @@ def get_invocation_context(self) -> InvocationContext:
                     of the agent.
         """
         raise NotImplementedError
+
+    def get_origin(self) -> List[str]:
+        """
+        :return: A List of strings indicating the origin of the run.
+                The origin can be considered a path to the original call to the front-man.
+        """
+        raise NotImplementedError
```
d1donlydfink, line 102: Allow getting the origin list of strings from a RunContext abstractly - that is, don't care if it comes from OpenAI or LangChain

neuro_san/internals/run_context/langchain/langchain_run_context.py
```diff
@@ -62,13 +63,15 @@ class LangChainRunContext(RunContext):
     """
 
     def __init__(self, llm_config: Dict[str, Any],
+                 parent_run_context: RunContext,
```
d1donlydfink, line 67: Accept the parent RunContext in the LangChainRunContext constructor.

neuro_san/internals/run_context/langchain/langchain_run_context.py
```diff
@@ -103,15 +111,18 @@ async def create_resources(self, assistant_name: str,
         :param tool_names: The list of registered tool names to use.
                     Default is None implying no tool is to be called.
         """
+        # DEF - Remove the arg if possible
+        _ = assistant_name
```
d1donlydfink, line 116: OpenAI RunContexts need a unique name passed into create_resources(), so not sure if I can get rid of this just yet.

neuro_san/internals/run_context/langchain/langchain_run_context.py
```diff
@@ -214,7 +225,8 @@ async def submit_message(self, user_message: str) -> Run:
         try:
             self.recent_human_message = HumanMessage(user_message)
         except ValidationError as exception:
-            message = f"ValidationError in {self.assistant_name} with message: {user_message}"
+            full_name: str = self.get_full_name_from_origin()
+            message = f"ValidationError in {full_name} with message: {user_message}"
```
d1donlydfink, line 230: Another place where the full agent path is more informative than the assistant name.

neuro_san/internals/run_context/langchain/langchain_run_context.py
```diff
@@ -416,3 +428,32 @@ def get_invocation_context(self) -> InvocationContext:
                     of the agent.
         """
         return self.invocation_context
+
+    def get_origin(self) -> List[str]:
+        """
+        :return: A List of strings indicating the origin of the run.
+                The origin can be considered a path to the original call to the front-man.
+        """
+        return self.origin
```
d1donlydfink, line 438: Allow Tool/Nodes to get at the origin information even though it's the RunContexts with their parental ties that own it.

neuro_san/internals/run_context/langchain/langchain_run_context.py
```diff
@@ -416,3 +428,32 @@ def get_invocation_context(self) -> InvocationContext:
...
+    def add_name_to_origin(self):
+        """
+        Adds the agent name to the origin.
+        """
+        # Add the name from the spec to the origin, if we have it.
+        if self.tool_caller is not None:
+            agent_spec: Dict[str, Any] = self.tool_caller.get_agent_tool_spec()
+            factory: AgentToolFactory = self.tool_caller.get_factory()
+            agent_name: str = factory.get_name_from_spec(agent_spec)
+            self.origin.append(agent_name)
```
d1donlydfink, line 449: Do some necessary gymnastics to get a proper name for the agent from the spec to add to the tail-end of the origin list.

neuro_san/internals/graph/tools/branch_tool.py
```diff
@@ -139,15 +139,17 @@ async def build(self) -> List[Any]:
         instructions = self.get_instructions()
         instructions = instructions + specific_instructions
 
+        origin: List[str] = self.run_context.get_origin()
+
         decision_name = self.get_decision_name()
         component_name = self.get_component_name()
         assistant_name = f"{decision_name}_{component_name}"
-        await self.journal.write(f"setting up {component_name} assistant...")
+        await self.journal.write(f"setting up {component_name} assistant...", origin)
```
d1donlydfink, line 147: Use the origin information to pass to Journal.write() for legacy logs messages.

neuro_san/internals/journals/compound_journal.py
```diff
@@ -33,12 +33,14 @@ def __init__(self, journals: List[Journal] = None):
...
-    async def write(self, entry: Union[str, bytes]):
+    async def write(self, entry: Union[str, bytes], origin: List[str]):
         """
         Writes a single string entry into each journal.
+        :param entry: The logs entry to write
+        :param origin: A list of strings describing the originating agent of the information
         """
         for journal in self.journals:
-            await journal.write(entry)
+            await journal.write(entry, origin)
```
d1donlydfink, line 43: Have CompoundJournals pass the origin information to all of the journals they need to update.

neuro_san/internals/journals/compound_journal.py
```diff
@@ -59,11 +61,11 @@ def add_journal(self, journal: Journal):
         """
         self.journals.append(journal)
 
-    async def write_message(self, message: BaseMessage, origin: Union[str, List[str]] = None):
+    async def write_message(self, message: BaseMessage, origin: List[str]):
```
d1donlydfink, line 64: Change the typing to always expect a list. Also change the comment below.

neuro_san/internals/journals/message_journal.py
```diff
@@ -37,38 +37,35 @@ def __init__(self, hopper: AsyncHopper):
...
                     Can be either a string or bytes.
+        :param origin: A list of strings describing the originating agent of the information
         """
         # Decoding bytes to string if necessary
         if isinstance(entry, bytes):
             entry = entry.decode('utf-8')
 
         legacy = LegacyLogsMessage(content=entry)
-        await self.write_message(legacy)
+        await self.write_message(legacy, origin)
```
d1donlydfink, line 51: This is where the legacy logs message is created. Pass the origin information along to the write_message() method so it gets sent to the client.

neuro_san/internals/journals/text_journal.py
```diff
@@ -28,10 +28,11 @@ def __init__(self):
         """
         self.log_content = []
 
-    async def write(self, entry: Union[str, bytes]):
+    async def write(self, entry: Union[str, bytes], origin: List[str]):
```
d1donlydfink, line 31: Fulfill the interface, but TextJournals don't care about origin information anyway.
(This is what writes the polling logs() stream of text)

neuro_san/internals/run_context/langchain/langchain_run_context.py
```diff
@@ -80,13 +84,18 @@ def __init__(self, llm_config: Dict[str, Any],
...
         self.error_detector: ErrorDetector = None
         self.recent_human_message: HumanMessage = None
         self.tool_caller: ToolCaller = tool_caller
         self.invocation_context: InvocationContext = invocation_context
 
+        # Set up the origin by copying the list from its parent run context
+        self.origin: List[str] = []
+        if parent_run_context is not None:
+            self.origin = copy(parent_run_context.get_origin())
+        self.origin = OriginUtils.add_spec_name_to_origin(self.origin, tool_caller)
```
d1donlydfink, line 97: Set up the origin list by copying the parent's version first, then calling the new OriginUtils method.
I added this as a separate method because I have a feeling it might have to move eventually. We'll see how that goes in later PRs.

neuro_san/internals/run_context/langchain/langchain_run_context.py
```diff
@@ -103,15 +112,18 @@ async def create_resources(self, assistant_name: str,
...
                     Default is None implying no tool is to be called.
         """
+        # DEF - Remove the arg if possible
+        _ = assistant_name
+
         # Create the model we will use.
         llm: BaseLanguageModel = LlmFactory.create_llm(self.llm_config)
 
-        self.assistant_name = assistant_name
+        full_name: str = OriginUtils.get_full_name_from_origin(self.origin)
```
d1donlydfink, line 121: Take the name we pass into the error detector as the "full name" instead of the assistant_name passed in.
Use the new OriginUtils for the standard way to do that.
In the world of deeper networks this "full name" starts to look like "my_front_man.my_branch.my_leaf"

The old assistant_name has some component of LLM-injected naming that is vestigial from esp_decision_assistant, but I'm hesitant to get rid of it just yet if we add numeric suffixes for multiple calls to the same origin path.  See question in OriginUtils below.

neuro_san/internals/run_context/utils/origin_utils.py
```diff
@@ -0,0 +1,55 @@
...
+        """
+        :param origin: A list of strings determining an origin of a chat message
+        :return: A single string name given an origin path/list
+        """
+        # Connect all the elements of the origin by the delimiter "."
+        name: str = ".".join(origin)
+        # Simple replacement of local external agents.
+        # DEF - need better recognition of external agent
+        name = name.replace("./", "/")
+        return name
```
d1donlydfink, line 55: A first attempt at a full name. Will have to get better at external agent references eventually.

neuro_san/internals/run_context/utils/origin_utils.py
```diff
@@ -0,0 +1,55 @@
...
+        """
+        if origin is None:
+            origin = []
+
+        # Add the name from the spec to the origin, if we have it.
+        if tool_caller is not None:
+            agent_spec: Dict[str, Any] = tool_caller.get_agent_tool_spec()
+            factory: AgentToolFactory = tool_caller.get_factory()
+            agent_name: str = factory.get_name_from_spec(agent_spec)
+            origin.append(agent_name)
```
d1donlydfink, line 40: For now the agent name added to the origin list will come straight from the spec.

@dsargent , @swensel , @ofrancon : I have noticed that something like esp_decision_assistant will call
its prescriptor multiple times, but these are not yet distinguished from each other in the origin information.
Neither do they come in the Connectivity() information - because examining a network multiple calls is not possible there.

It's conceivable I could add a numeric at the end like "prescriptor", "prescriptor_02", "prescriptor_03" as repeated calls happen, but I feel it's worth checking with you guys who are updating various clients before going there.

neuro_san/internals/chat/data_driven_chat_session.py
```diff
@@ -92,6 +90,8 @@ async def set_up(self):
         run_context: RunContext = RunContextFactory.create_run_context(None, None,
                                                                        invocation_context=self.invocation_context)
         self.front_man = self.registry.create_front_man(self.journal, self.sly_data, run_context)
+
+        await self.journal.write("setting up chat agent(s)...", self.front_man.get_origin())
```
d1donlydfink, line 94: Pass origin information to every Journal.write() call. This is the call that produces the LegacyLog messages.

Note that the first "setting up chat agent(s)..." write() call comes down here because it's not until the FrontMan is created that we have something to get an origin from.

neuro_san/client/agent_cli.py
```diff
@@ -378,19 +380,19 @@ def stream_once(self, state: Dict[str, Any]) -> Dict[str, Any]:
...
 
             text: str = response.get("text")
+            origin: List[str] = response.get("origin")
+            origin_str: str = OriginUtils.get_full_name_from_origin(origin)
 
             # Update chat response and maybe prompt.
             if text is not None:
                 if message_type == ChatMessageType.LEGACY_LOGS:
                     with open(self.args.thinking_file, "a", encoding="utf-8") as thinking:
+                        thinking.write(f"[{origin_str}]:\n")
```
d1donlydfink, line 390: Add origin output to the streaming client (polling client does not get this information at all).
Additions are here for the thinking-file and below for the regular output.

neuro_san/client/agent_cli.py
```diff
@@ -378,19 +380,19 @@ def stream_once(self, state: Dict[str, Any]) -> Dict[str, Any]:
...
                         thinking.write(text)
                         thinking.write("\n")
-                elif message_type == ChatMessageType.AGENT_FRAMEWORK:
-                    # Skip over the connectivity messages for this text-only client
-                    if response.get("origin") is None:
-                        print(f"{text}")
```
d1donlydfink, line 391: This special elif block is not necessary any more since connectivity information is no longer coming over the stream.

neuro_san/internals/chat/data_driven_chat_session.py
```diff
@@ -194,28 +194,6 @@ async def streaming_chat(self, user_input: str,
...
         await self.queue.put_final_item()
 
-    def is_streamable_message(self, chat_message: Dict[str, Any], index: int) -> bool:
-        """
-        Filter chat messages from the full logs that are in/appropriate for streaming.
-
-        :param chat_message: A Chat message dictionary of the form in chat.ChatMessage proto
-        :param index: The place in the chat history the message has
-        :return: True if the given message should be streamed back. False if not.
-        """
```
d1donlydfink, line 204: This method is no longer used.

neuro_san/internals/graph/tools/calling_tool.py
```diff
@@ -115,6 +115,12 @@ def get_factory(self) -> AgentToolFactory:
         """
         return self.factory
 
+    def get_origin(self) -> List[str]:
+        """
+        :return: The list of strings describing the origin associated with this node
+        """
+        return self.run_context.get_origin()
```
d1donlydfink, line 122: Poke through to allow someone using a node to get its origin information.
This is not available in any interface yet, but could easily become so in future PRs.

neuro_san/internals/journals/message_journal.py
```diff
@@ -37,38 +37,31 @@ def __init__(self, hopper: AsyncHopper):
...
-        message_dict: Dict[str, Any] = convert_to_chat_message(message)
-
-        # Handle the origin information
-        if origin is not None:
-            origin_list: List[str] = origin
-            if isinstance(origin, str):
-                # Handle the case of a single string
-                origin_list = [origin]
-            message_dict["origin"] = origin_list
+        message_dict: Dict[str, Any] = convert_to_chat_message(message, origin)
```
d1donlydfink, line 65: Don't worry about the non-list case any more.

neuro_san/internals/messages/message_utils.py
```diff
@@ -134,7 +135,10 @@ def convert_to_chat_message(message: BaseMessage) -> Dict[str, Any]:
         "type": message_type,
         "text": message.content,
         # No mime_data for now
-        # No origin for now
     }
 
+    # Handle the origin information if we have it
+    if origin is not None:
+        chat_message["origin"] = origin
```
d1donlydfink, line 142: Allow passing an origin when converting a single chat message.

neuro_san/internals/run_context/openai/openai_run_context.py
```diff
@@ -182,7 +182,8 @@ async def wait_on_run(self, run: Run, journal: Journal = None) -> Run:
                 number_of_new_messages = len(latest_messages) - len(messages)
                 new_messages = latest_messages[-number_of_new_messages:]
                 if journal is not None:
-                    await journal.write(pretty_the_messages(new_messages))
+                    await journal.write(pretty_the_messages(new_messages),
+                                        self.get_origin())
```
d1donlydfink, line 186: Fulfill the Journal.write() argument requirements in OpenAI land even though it's kind of dying.

neuro_san/internals/run_context/openai/openai_run_context.py
```diff
@@ -247,3 +248,10 @@ def get_invocation_context(self) -> InvocationContext:
                     of the agent.
         """
         return self.invocation_context
+
+    def get_origin(self) -> List[str]:
+        """
+        :return: A List of strings indicating the origin of the run.
+                The origin can be considered a path to the original call to the front-man.
+        """
+        return None
```
d1donlydfink, line 257: Fulfill the RunContext.get_origin() interface requirements in OpenAI land even though it's kind of dying.

neuro_san/internals/run_context/utils/origin_utils.py
dsargent: Shooting from the hip here as this is deeper than we've been parsing anything so far, but:

We like structure and especially JSON. So ideal would be something like,

```json
{
  "tool": "prescriptor",
  "index": 2
}
```

I've no idea how feasible that is, but ideally that's what we would want to consume. Rather than some kind of string slicing.

d1donlydfink: That would require a protocol change and also whatever we do would have to be convertible into a single string anyway (doable from what you suggest).

dsargent: This is the right time to change the protocol IMO before we socialize it to the outside world. For now the only consumers are Babak's demo (?), the CLI, and the UI, so we have a closed universe. Once we start publishing it to Cog at large, the "concrete will set" and it will be harder to make changes.

I'm still not sure why we "smuggle JSON as strings" in our protobuf definitions. It requires all kinds of messy escaping to express JSON-in-JSON. I vaguely remember @andreidenissov-cog mentioned there are some tricks required due to quirks of Python to consume "JSON as first class protobuf entities" so JSON gets flattened to strings. But it's a shame -- there are first class types in protobuf for JSON so we wouldn't have the "who's on First?" escaping issues.

swensel: I agree with Darren. The other thing that would be good to consider is when Salesforce and other external API consumers look at our API, we want things to be pretty standard for them, so that on-boarding is straight forward and they don't have any blockers using our API.

d1donlydfink: OK This is all good and gives me a direction to go for the next PR.

ofrancon: In the short term you could use origin = "ESP Assistant Call #1", "ESP assitant Call #2" etc. if it's "just" a log.
But I agree with Darren and Sean, we should not build parsers and logic on top of that. A json structure would be much better.

d1donlydfink: This will be my next PR.  Thanks for the comments all!

PR #58, opened 2025-01-29
Updating dependencies to support Python 3.12

kxk302:


requirements.txt
```diff
@@ -12,14 +12,14 @@ grpcio-health-checking>=1.62.0
...
+protobuf>=4.25.5
 
 # Downgraded secondary dependencies for error-free pip installs
 pyhocon>=0.3.60
 pyOpenSSL>=24.0.0
 
-boto3>=1.34.51
-botocore>=1.34.51
+botocore==1.35.81
+boto3==1.35.81
```
d1donlydfink, line 22: Do not pin libraries in repos that export libraries.

This file requirements.txt gets picked up by the packaging (wheel) process and broadcasts out to all who are installing it: "These are the dependencies I need".  When we pin them we pin them for everyone including the people outside our group who are using it.

For any client repo which is using neuro-san, the >= designation already equivalent to what you want to change things to, but it more on the client library's terms than ours.  

That is, this change is unnecessary as it stands.

If there were security problems with the lower-bound versions, then that's the cause for upgrading.

PR #59, opened 2025-01-31
UN-2846 Re-type Origin components as structures / Separate chat by origin into --thinking_dir

d1donlydfink:
Not as bad as the 21 files changed might look.

The big things:
* Re-type the components of the origin array in terms of ChatMessage proto to be structures
* Make an Origination object carry per-AgentSession state which keeps track of how many times each tool has been called 
* Now that we have distinct tool origins, add a --thinking_dir argument to the client that separates out chat from each tool into its own file for easier debugging.

The little things:
* Re-type the python code for origin arrays to be List[Dict[str, Any]] instead of just List[str]
* Add logic to the do_generate.sh script to always add the copyright notice at the top. Was getting annoying doing that by hand.
* Move OriginUtils -> messages.Origination 
* Allow this Origination object to hang off the InvocationContext so guts can all access the same state
* Expose a get_name() call on the ToolCaller interface so that RunContext can correctly get the name.

Tested: esp_decision_assistant on the agent_cli client.  This is nice because that guy often invokes 2 predictors and I can see the tool count in action with it.

### DIFF
--- a/neuro_san/client/agent_cli.py
+++ b/neuro_san/client/agent_cli.py
@@ -57,7 +57,7 @@ class AgentCli:
         self.session_id: str = None
         self.thinking_dir: str = None
 
-    # pylint: disable=too-many=branches
+    # pylint: disable=too-many-branches
     def main(self):
         """
         Main entry point for command line user interaction


neuro_san/api/grpc/agent_pb2.py
```diff
@@ -1,3 +1,4 @@
+
```
d1donlydfink, line 1: Whitespace is added by new extra logic in the do_generate.sh script to preserve the copyrights,
cuz restoring them manually was getting annoying.

neuro_san/api/grpc/chat.proto
```diff
@@ -22,8 +22,19 @@ package dev.cognizant_ai.neuro_san.api.grpc.chat;
...
 
+message Origin {
+    // String name of the originating tool, as per the agent spec.
+    string tool = 1 [json_name="tool"];
+
+    // Some tools can be called more than once my one or more different paths.
+    // Allow for an instantiation index to distinguish these in the chat stream.
+    // Index counting starts at 0.
+    int32 instantiation_index = 2 [json_name="instantiation_index"];
+}
```
d1donlydfink, line 36: New origin structure, per the squad's recommendations.

neuro_san/api/grpc/chat.proto
```diff
@@ -63,16 +74,19 @@ message ChatMessage {
...
+    repeated Origin origin = 4 [json_name="origin"];
+
+    // Optional structure for a message whose contents are parsed JSON.
+    // The idea is to have the server side do the parsing when requested
+    // by the agent spec.
+    //
+    // As of 1/30/25 this is a forward-looking, experimental field not likely
+    // to be used in regular operation until we can get proper plumbing of such
+    // data in place.
+    google.protobuf.Struct structure = 5;
```
d1donlydfink, line 91: I did a little research and I think this will be the way to communicate arbitrary structure information over the wire ala JSON.  Adding this now with caveat comment to minimize the number of times the protocol itself has to change.

It is beyond the scope of this PR to actually populate this field right now.
Baby steps, @dsargent

neuro_san/internals/graph/tools/branch_tool.py
```diff
@@ -139,7 +139,7 @@ async def build(self) -> List[Any]:
         instructions = self.get_instructions()
         instructions = instructions + specific_instructions
 
-        origin: List[str] = self.run_context.get_origin()
+        origin: List[Dict[str, Any]] = self.run_context.get_origin()
```
d1donlydfink, line 142: Re-type the origin information.

neuro_san/internals/run_context/langchain/langchain_run_context.py
```diff
@@ -430,20 +430,13 @@ def get_invocation_context(self) -> InvocationContext:
...
-    def add_name_to_origin(self):
-        """
-        Adds the agent name to the origin.
-        """
-        # Add the name from the spec to the origin, if we have it.
-        if self.tool_caller is not None:
-            agent_spec: Dict[str, Any] = self.tool_caller.get_agent_tool_spec()
-            factory: AgentToolFactory = self.tool_caller.get_factory()
-            agent_name: str = factory.get_name_from_spec(agent_spec)
-            self.origin.append(agent_name)
```
d1donlydfink, line 449: This method has been moved to OriginUtils and is no longer used

neuro_san/internals/graph/tools/calling_tool.py
```diff
@@ -115,9 +118,14 @@ def get_factory(self) -> AgentToolFactory:
...
-    def get_origin(self) -> List[str]:
+    def get_origin(self) -> List[Dict[str, Any]]:
         """
-        :return: The list of strings describing the origin associated with this node
+        :return: A List of origin dictionaries indicating the origin of the run.
+                The origin can be considered a path to the original call to the front-man.
+                Origin dictionaries themselves each have the following keys:
+                    "tool"                  The string name of the tool in the spec
+                    "instantiation_index"   An integer indicating which incarnation
+                                            of the tool is being dealt with.
```
d1donlydfink, line 128: Re-type the origin information, and add a standard comment describing the format of the origin dictionaries

neuro_san/client/agent_cli.py
```diff
@@ -381,7 +381,7 @@ def stream_once(self, state: Dict[str, Any]) -> Dict[str, Any]:
 
             text: str = response.get("text")
             origin: List[str] = response.get("origin")
-            origin_str: str = OriginUtils.get_full_name_from_origin(origin)
+            origin_str: str = Origination.get_full_name_from_origin(origin)
```
d1donlydfink, line 426: Rename OriginUtils -> Origination since it is no longer a static utility class.  More on that below.

neuro_san/internals/graph/tools/branch_tool.py
```diff
@@ -139,10 +139,10 @@ async def build(self) -> List[Any]:
...
 
         decision_name = self.get_decision_name()
-        component_name = self.get_component_name()
+        component_name = self.get_name()
```
d1donlydfink, line 145: Rename get_component_name() -> get_name()

neuro_san/internals/graph/tools/calling_tool.py
```diff
@@ -87,19 +87,22 @@ def get_agent_tool_spec(self) -> Dict[str, Any]:
...
+    def get_name(self) -> str:
         """
-        :return: The string name of the component
+        :return: the name of the data-driven agent as it comes from the spec
         """
-        return self.agent_tool_spec.get("name")
+        agent_spec: Dict[str, Any] = self.get_agent_tool_spec()
+        factory: AgentToolFactory = self.get_factory()
+        agent_name: str = factory.get_name_from_spec(agent_spec)
+        return agent_name
```
d1donlydfink, line 97: Rename get_component_name() -> get_name()
Go to the factory to get the name from the spec. There can be a little more policy logic around it than first meets the eye, and this is a single, consolidated, more-proper way.

neuro_san/internals/interfaces/invocation_context.py
```diff
@@ -33,3 +34,10 @@ def get_asyncio_executor(self) -> AsyncioExecutor:
         :return: The AsyncioExecutor associated with the invocation
         """
         raise NotImplementedError
+
+    def get_origination(self) -> Origination:
+        """
+        :return: The Origination instance carrying state about tool instantation
+                during the course of the AgentSession.
+        """
+        raise NotImplementedError
```
d1donlydfink, line 43: Add a stateful Origination object to the InvocationContext interface

Gee I'm glad I did all the plumbing for that InvocationContext thing.
It's coming in really handy and I can already see more uses for it.

neuro_san/internals/journals/compound_journal.py
```diff
@@ -33,11 +34,16 @@ def __init__(self, journals: List[Journal] = None):
...
         """
         Writes a single string entry into each journal.
         :param entry: The logs entry to write
-        :param origin: A list of strings describing the originating agent of the information
+        :param origin: A List of origin dictionaries indicating the origin of the run.
+                The origin can be considered a path to the original call to the front-man.
+                Origin dictionaries themselves each have the following keys:
+                    "tool"                  The string name of the tool in the spec
+                    "instantiation_index"   An integer indicating which incarnation
+                                            of the tool is being dealt with.
```
d1donlydfink, line 46: More re-typing of origin

neuro_san/internals/run_context/interfaces/tool_caller.py
```diff
@@ -43,3 +43,9 @@ def get_agent_tool_spec(self) -> Dict[str, Any]:
         :return: the dictionary describing the data-driven agent
         """
         raise NotImplementedError
+
+    def get_name(self) -> str:
+        """
+        :return: the name of the data-driven agent as it comes from the spec
+        """
+        raise NotImplementedError
```
d1donlydfink, line 51: Expose the get_name() call off of ToolCaller so RunContexts can set up their origination information properly.

neuro_san/internals/run_context/langchain/langchain_run_context.py
```diff
@@ -90,11 +89,13 @@ def __init__(self, llm_config: Dict[str, Any],
...
-        self.origin: List[str] = []
+        self.origin: List[Dict[str, Any]] = []
         if parent_run_context is not None:
-            self.origin = copy(parent_run_context.get_origin())
-        self.origin = OriginUtils.add_spec_name_to_origin(self.origin, tool_caller)
+
+            # Settle the origin.
+            agent_name: str = tool_caller.get_name()
+            origination: Origination = self.invocation_context.get_origination()
+            self.origin = origination.add_spec_name_to_origin(parent_run_context.get_origin(), agent_name)
```
d1donlydfink, line 98: 1. Re-type the origin.
2. Use the new ToolCaller.get_name() to pass the agent name along instead of the ToolCaller itself (which ends up creating an import cycle anyway)
3. Use the Origination object now hanging off the InvocationContext which has state to determine what should be added to the origin list
4. Note that the copy() of the origin list above is now handled inside add_spec_name_to_origin() as an implementation detail of Origin stuff.

neuro_san/internals/run_context/langchain/langchain_run_context.py
```diff
@@ -118,7 +119,7 @@ async def create_resources(self, assistant_name: str,
         # Create the model we will use.
         llm: BaseLanguageModel = LlmFactory.create_llm(self.llm_config)
 
-        full_name: str = OriginUtils.get_full_name_from_origin(self.origin)
+        full_name: str = Origination.get_full_name_from_origin(self.origin)
```
d1donlydfink, line 122: This is still a non-stateful, static method on the Origination object. Makes sense to keep all the policy in that one place.

neuro_san/internals/run_context/openai/openai_run_context.py
```diff
@@ -249,9 +249,13 @@ def get_invocation_context(self) -> InvocationContext:
...
-        :return: A List of strings indicating the origin of the run.
+        :return: A List of origin dictionaries indicating the origin of the run.
                 The origin can be considered a path to the original call to the front-man.
+                Origin dictionaries themselves each have the following keys:
+                    "tool"                  The string name of the tool in the spec
+                    "instantiation_index"   An integer indicating which incarnation
+                                            of the tool is being dealt with.
         """
-        return None
+        return []
```
d1donlydfink, line 261: Let's not have to check for None everywhere.

neuro_san/session/session_invocation_context.py
```diff
@@ -36,6 +37,7 @@ def __init__(self, async_session_factory: AsyncAgentSessionFactory,
 
         self.async_session_factory: AsyncAgentSessionFactory = async_session_factory
         self.asyncio_executor: AsyncioExecutor = asyncio_executor
+        self.origination: Origination = Origination()
```
d1donlydfink, line 40: Keep an Origination object as part of the InvocationContext implementation.

neuro_san/api/grpc/chat.proto
```diff
@@ -63,16 +74,19 @@ message ChatMessage {
...
-    //          [host:]/agent1/agent2/...
-    //
-    // ... where the host is optional and each agent name represents an agent
-    // in a single agent hierarchy.
     //
     // For each top-level agent/front-man (perhaps on another server) that is called,
-    // an extra string is added to the list.
-    repeated string origin = 4 [json_name="origin"];
+    // an extra structure is added to the list.
+    repeated Origin origin = 4 [json_name="origin"];
```
d1donlydfink, line 82: Directly re-type the same field in the protobuf.
Normally this would not be OK, but I am assuming no other client has been using this field just yet because there has been no information in it yet.

neuro_san/api/grpc/chat_pb2.py
```diff
@@ -23,19 +23,22 @@
...
-  _globals['_CHATMESSAGE']._serialized_start=112
-  _globals['_CHATMESSAGE']._serialized_end=457
-  _globals['_CHATMESSAGE_CHATMESSAGETYPE']._serialized_start=337
-  _globals['_CHATMESSAGE_CHATMESSAGETYPE']._serialized_end=457
+  _globals['_ORIGIN']._serialized_start=141
+  _globals['_ORIGIN']._serialized_end=219
+  _globals['_CHATMESSAGE']._serialized_start=222
+  _globals['_CHATMESSAGE']._serialized_end=661
+  _globals['_CHATMESSAGE_CHATMESSAGETYPE']._serialized_start=541
+  _globals['_CHATMESSAGE_CHATMESSAGETYPE']._serialized_end=661
```
d1donlydfink, line 43: Regenerated grpc biz.

neuro_san/internals/messages/origination.py
```diff
@@ -0,0 +1,118 @@
...
+    An instance for common code and state that manipulates and keeps track of origin information
+    on a per-AgentSession basis.
+
+    A full origin description is a List of Dictionaries indicating the origin of a chat message.
+    An origin can be considered a path to the original call to the front-man.
+    Origin dictionaries themselves each have the following keys:
+        "tool"                  The string name of the tool in the spec
+        "instantiation_index"   An integer indicating which incarnation
+                                of the tool is being dealt with.
+    """
```
d1donlydfink, line 29: Guts of OriginUtils moves here to become a stateful object.

neuro_san/internals/messages/origination.py
```diff
@@ -0,0 +1,118 @@
...
+                    "instantiation_index"   An integer indicating which incarnation
+                                            of the tool is being dealt with.
+        :param agent_name: The agent name to be added to the list.
+        :return: The new origin with the agent name at the end of the list
+        """
+        new_origin: List[Dict[str, Any]] = []
+
+        # Add the name from the spec to the origin, if we have it.
+        if origin is None or agent_name is None:
+            return new_origin
```
d1donlydfink, line 57: Slightly new logic for add_spec_name_to_origin() as it:
a) no longer takes a ToolCaller (cuz that now introduces import cycles) and takes a plain-old string agent_name instead
b) Depends on instance state.

neuro_san/internals/messages/origination.py
```diff
@@ -0,0 +1,118 @@
...
+
+        # Make a shallow copy of the list, as we are going to modify it
+        # and don't want to muck with the original
+        new_origin = copy(origin)
```
d1donlydfink, line 61: The shallow copy of the origin now happens in here instead of in LangChainRunContext.
That it's a shallow copy is more an implementation detail of origin biz.

neuro_san/internals/messages/origination.py
```diff
@@ -0,0 +1,118 @@
...
+
+        # Find the current instantiation index for the tool
+        # and increment it in the map for later use
+        instantiation_index: int = self.tool_to_index_map.get(agent_name, Origination.INSTANTIATION_START)
+        self.tool_to_index_map[agent_name] = instantiation_index + 1
```
d1donlydfink, line 66: Here is where the stateful object comes in.
Every time a single named tool is invoked for origin information, keep track of how many times we have done so,
and here we also increment for the next time around.

We start the count at 1 because we want this to be human readable (see INSTANTIATION_START constant above).

The idea is that some nodes in the graph can be called more than once and we want clients to be able to distinguish which version of the tool instantiation is doing the talking. This ends up being useful even in
the esp_decision_assistant agent, which I used for testing this functionality.

neuro_san/internals/messages/origination.py
```diff
@@ -0,0 +1,118 @@
...
+
+        # Prepare the origin dictionary to append
+        origin_dict: Dict[str, Any] = {
+            "tool": agent_name,
+            "instantiation_index": instantiation_index
+        }
+        new_origin.append(origin_dict)
```
d1donlydfink, line 73: Add one of those new-fangled dictionaries with all the right info in it to get over the grpc wire.

neuro_san/internals/messages/origination.py
```diff
@@ -0,0 +1,118 @@
...
+        # Connect all the elements of the origin by the delimiter "."
+        origin_list: List[str] = []
+        for origin_dict in origin:
+
+            # Get basic fields from the dict
+            instantiation_index: int = origin_dict.get("instantiation_index", Origination.INSTANTIATION_START)
+            tool: str = origin_dict.get("tool")
+            if tool is None:
+                # No information of value will be conveyed with no tool set in the dict.
+                raise ValueError("tool name in origin_dict is None")
```
d1donlydfink, line 100: New logic for this static function to get a single string out of the origin structure list. (It is also used by the client).
Here we take apart the dictionary for a component and crap out if we don't have any tool information (new behavior we should never hit).

neuro_san/internals/messages/origination.py
```diff
@@ -0,0 +1,118 @@
...
+
+            # Figure out how we will deal with the index
+            index_str: str = ""
+            if instantiation_index > Origination.INSTANTIATION_START:
+                # zfill() adds leading 0's up to the number of characters provided
+                index_str = f"-{str(instantiation_index).zfill(Origination.NUM_INSTANTIATION_INDEX_DIGITS)}"
```
d1donlydfink, line 106: Some policy around when we tack on a number for tools in the single origin string.
We don't want to label things "-01" all the time because most of the time agents don't repeat.
But "-02" and onwards starts to be interesting.

Note that the CLI client uses this code and uses this as its display policy, but any other client
can do whatever it wants with the raw information in the structures.

neuro_san/api/scripts/do_generate.sh
```diff
@@ -102,4 +104,14 @@ do
...
         "${PROTO_FILE}"
+
+    # Prepend copyrights to output files
+    MESSAGE_FILE="${PROTO_FILE/.proto/_pb2.py}"
+    SERVICE_FILE="${PROTO_FILE/.proto/_pb2_grpc.py}"
+
+    TEMPFILE=$(mktemp)
+    cat "${COPYRIGHT_FILE}" "${MESSAGE_FILE}" > "${TEMPFILE}" && mv "${TEMPFILE}" "${MESSAGE_FILE}"
+    TEMPFILE=$(mktemp)
+    cat "${COPYRIGHT_FILE}" "${SERVICE_FILE}" > "${TEMPFILE}" && mv "${TEMPFILE}" "${SERVICE_FILE}"
```
d1donlydfink, line 115: Add the ability to preserve the copyrights when regenerating the python files for export.
Was getting annoying replacing them by hand.

neuro_san/client/agent_cli.py
```diff
@@ -162,26 +169,42 @@ def add_args(self, arg_parser: argparse.ArgumentParser):
         Adds arguments.  Allows subclasses a chance to add their own.
         :param arg_parser: The argparse.ArgumentParser to add.
         """
+        # What agent are we talking to?
+        arg_parser.add_argument("--agent", type=str, default="esp_decision_assistant",
+                                help="Name of the agent to talk to")
```
d1donlydfink, line 176: Separate the args into sections. Agent name moves up here from below.

neuro_san/client/agent_cli.py
```diff
@@ -101,8 +105,11 @@ def main(self):
         initial_prompt: str = function.get("description")
         print(f"\n{initial_prompt}\n")
 
-        print("To see the thinking involved with the agent:\n"
-              f"    tail -f {self.args.thinking_file}\n")
+        print("To see the thinking involved with the agent:\n")
+        if not self.args.thinking_dir:
+            print(f"    tail -f {self.args.thinking_file}\n")
+        else:
+            print(f"    See any one of the files in {self.thinking_dir} for agent network chat details.\n")
```
d1donlydfink, line 114: Different inspection hints depending on how we are capturing chat output.

neuro_san/client/agent_cli.py
```diff
@@ -162,26 +169,42 @@ def add_args(self, arg_parser: argparse.ArgumentParser):
...
+
+        # How are we capturing output?
         arg_parser.add_argument("--thinking_file", type=str, default="/tmp/agent_thinking.txt",
                                 help="File that captures agent thinking. "
                                      "This is a separate text stream from the user/assistant chat")
+        arg_parser.add_argument("--thinking_dir", default=False, action="store_true",
+                                help="Use the basis of the thinking_file as a directory to capture "
+                                     "internal agent chatter in separate files. "
+                                     "This is a separate text stream from the user/assistant chat. "
+                                     "Only available when streaming (the default).")
```
d1donlydfink, line 194: Add an argument to split up chat capture by origin.

neuro_san/client/agent_cli.py
```diff
@@ -221,16 +248,29 @@ def open_session(self):
...
+            # basis for the thinking directory
+            self.thinking_dir, extension = os.path.splitext(self.args.thinking_file)
+            _ = extension
+
+            # Remove any contents that might be there already.
+            # Writing over will existing dir will just confuse output.
+            if os.path.exists(self.thinking_dir):
+                shutil.rmtree(self.thinking_dir)
+            # Create the directory anew
+            os.makedirs(self.thinking_dir)
```
d1donlydfink, line 275: Initialize the thinking_dir, if that's what we are doing.

neuro_san/client/agent_cli.py
```diff
@@ -381,15 +421,12 @@ def stream_once(self, state: Dict[str, Any]) -> Dict[str, Any]:
...
+            origin_str: str = Origination.get_full_name_from_origin(origin)
 
             # Update chat response and maybe prompt.
             if text is not None:
                 if message_type == ChatMessageType.LEGACY_LOGS:
-                    with open(self.args.thinking_file, "a", encoding="utf-8") as thinking:
-                        thinking.write(f"[{origin_str}]:\n")
-                        thinking.write(text)
-                        thinking.write("\n")
+                    self.write_text(text, origin_str)
```
d1donlydfink, line 431: Consolidate this into new write_text() method below

neuro_san/client/agent_cli.py
```diff
@@ -405,6 +442,27 @@ def stream_once(self, state: Dict[str, Any]) -> Dict[str, Any]:
...
+    def write_text(self, text: str, origin_str: str):
+        """
+        Writes a line of text attributable to the origin, however we are doing that.
+        :param text: The text of the message
+        :param origin_str: The string representing the origin of the message
+        """
+
+        filename: str = self.args.thinking_file
+        if self.args.thinking_dir:
+            filename = os.path.join(self.thinking_dir, origin_str)
```
d1donlydfink, line 456: Choose the file name to write to based on how the args say the user wanted to capture output.
If we are doing a thinking_dire, then the file we want to write to is:

```<dir>/<origin-string>```

File: neuro_san/client/agent_cli.py [hunk 33]
```diff
@@ -405,6 +442,27 @@ def stream_once(self, state: Dict[str, Any]) -> Dict[str, Any]:
...
+
+        how_to_open_file: str = "a"
+        if not os.path.exists(filename):
+            how_to_open_file = "w"
+
+        with open(filename, how_to_open_file, encoding="utf-8") as thinking:
+            if not self.args.thinking_dir:
+                thinking.write(f"[{origin_str}]:\n")
+            thinking.write(text)
+            thinking.write("\n")
```
d1donlydfink, line 466: Send output to the appropriate file, as it exists or not at the beginning of this operation.
If we are doing thinking dirs we do not need to repeat the origin information in the file itself.

neuro_san/api/grpc/chat.proto
```diff
@@ -22,8 +22,19 @@ package dev.cognizant_ai.neuro_san.api.grpc.chat;
...
 option go_package = "github.com/leaf-ai/neuro_san/internal/gen/dev.cognizant_ai/neuro_san/api/grpc/chat/v1;chat";
 
+import "google/protobuf/struct.proto";
 import "neuro_san/api/grpc/mime_data.proto";
 
+message Origin {
+    // String name of the originating tool, as per the agent spec.
+    string tool = 1 [json_name="tool"];
+
+    // Some tools can be called more than once my one or more different paths.
```
ofrancon, line 32: typo: my -> by

neuro_san/api/grpc/chat.proto
ofrancon: We haven't published a new release. So no one has used it.

neuro_san/client/agent_cli.py
```diff
@@ -162,26 +171,42 @@ def add_args(self, arg_parser: argparse.ArgumentParser):
...
+                                     "Only available when streaming (the default).")
+
+        # How can we get input to the chat client without typing it in?
         arg_parser.add_argument("--first_prompt_file", type=str,
                                 help="File that captures the first response to the input prompt")
         arg_parser.add_argument("--sly_data", type=str,
                                 help="JSON string containing data that is out-of-band to the chat stream, "
                                      "but is still essential to agent function")
+
+        # How to we receive messages?
```
ofrancon, line 203: typo: how do we, or how to receive messages but not to we

neuro_san/client/agent_cli.py
```diff
@@ -162,26 +171,42 @@ def add_args(self, arg_parser: argparse.ArgumentParser):
...
         arg_parser.add_argument("--stream", default=True, action="store_true",
                                 help="Use streaming chat instead of polling")
         arg_parser.add_argument("--poll", dest="stream", action="store_false",
                                 help="Use polling chat instead of streaming")
+
+        # How to we set up our session connection?
```
ofrancon, line 209: typo: to we

neuro_san/api/grpc/chat.proto
d1donlydfink: Fixed in subsequent PR

neuro_san/client/agent_cli.py
d1donlydfink: Fixed in subsequent PR

neuro_san/client/agent_cli.py
d1donlydfink: Fixed in subsequent PR

PR #60, opened 2025-02-01
Handy website search agent along with example for how to use it.

babakatwork:
Used Duck-duck-go free API to write a tool for searching the web. Called the tool from a reusable agent added to the manifest. Also added a simple example agent network for how the website search agent should be called.

@d1donlydfink is the main reviewer.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -48,4 +48,7 @@ httpx==0.27.2
 # For input validation
 validators>=0.22.0
 
+# For chat-based cli tools not used in deployment
+timedinput>=0.1.1
+
 duckduckgo_search>=7.3.0


neuro_san/registries/website_search.hocon
babakatwork, line 1: This agent is pretty generic and meant for reuse within different agent networks.

neuro_san/registries/website_search_usage_example.hocon
babakatwork, line 1: Here is an example of how an agent network can call the reusable website_search agent.

requirements.txt
```diff
@@ -49,4 +49,10 @@ httpx==0.27.2
...
 
 # For chat-based cli tools not used in deployment
-timedinput
+timedinput~=0.1.1
+
+setuptools~=65.5.0
+pyparsing~=3.2.0
+langchain-core~=0.3.19
+pydantic_core~=2.23.4
+duckduckgo_search~=7.3.0
```
d1donlydfink: Theoretically all the other requirements changes except duckduckgo should not be necessary.

The bigger issue is: Should we really have neuro-san require duckduckgo_search?  
Probably not in the long run. Consuming clients will not likely use it (like unileaf and 1C), but it is a good example overall.

What do you know about what is dragged in along with duckduckgo_search?  Is it going to mess with unileaf dependencies for 3.12 at all?

neuro_san/registries/website_search_usage_example.hocon
```diff
@@ -0,0 +1,54 @@
...
+                    },
+                },
+                "required": ["user_inquiry"]
+            }
+         },
+
+            "instructions": """
+Use your tool to respond to the inquiry.
+""",
+            "tools": ["/website_search"]
```
d1donlydfink, line 51: This is actually kinda nice to have an example of an external tool.
You should add comments to that effect (like those that are seen in opportunity_finder_dev_loop in unileaf)

neuro_san/coded_tools/website_search/website_search.py
```diff
@@ -0,0 +1,93 @@
...
+    # Example usage
+    URL_TO_SCRAPE = "https://www.amazon.com"
+    SEARCH_TERMS = "10.5 white men sneakers"
+    TOP_N = 5
+
+    links = search_web(URL_TO_SCRAPE + " " + SEARCH_TERMS, TOP_N)
+
+    print(f"Top {TOP_N} search result links for '{SEARCH_TERMS}' from {URL_TO_SCRAPE}:")
+    for i, link in enumerate(links, start=1):
+        print(f"{i}. {link}")
```
d1donlydfink: Do not leave main entry points in here.  Better to make a test program under tests/ for it.
Reason: Then we don't have to have an exception when the security guys flag it as a possible attack point.

neuro_san/coded_tools/website_search/website_search.py
```diff
@@ -0,0 +1,93 @@
+from typing import Any
+from typing import Dict
+from typing import Union
+import duckduckgo_search
```
d1donlydfink: Looks like you only use DDGS from this package.  Try not to import the world and import only what you use.

neuro_san/coded_tools/website_search/website_search.py
```diff
@@ -0,0 +1,93 @@
...
+    # Extract and return only the URLs from the returned list of dictionaries
+    returned_links = [res["href"] for res in results if "href" in res]
+
+    return returned_links
+
+
+if __name__ == "__main__":
+    # Example usage
+    URL_TO_SCRAPE = "https://www.amazon.com"
+    SEARCH_TERMS = "10.5 white men sneakers"
```
d1donlydfink: Ah. This is why I cannot jump. There are sneakers that are specific to me. ;)

neuro_san/coded_tools/website_search/website_search.py
```diff
@@ -0,0 +1,93 @@
...
+    :param query: The search query (e.g., "10.5 white men sneakers").
+    :param num_results: Number of links to retrieve (default=5).
+    :return: List of hyperlink strings.
+    """
+    # Use duckduckgo_search to retrieve results
+    search = duckduckgo_search.DDGS()
+    results = search.text(query, max_results=num_results)
+
+    # Extract and return only the URLs from the returned list of dictionaries
+    returned_links = [res["href"] for res in results if "href" in res]
```
d1donlydfink, line 78: Comprehensions are generally incomprehensible. Consider breaking this up into > 1 line for easier reading
so it takes < 5 seconds to understand what is happening.

requirements.txt
d1donlydfink: Also FWIW: We are treating this as a library repo, so all dependency versions should float upwards whenever possible.  Without knowing pip dependency details, this should probably should look like:

```
duckduckgo_search>=7.3.0
```

... as an addition and that's it.

neuro_san/coded_tools/website_search/website_search.py
babakatwork: LOL!

neuro_san/coded_tools/website_search/website_search.py
```diff
@@ -0,0 +1,93 @@
...
+#     # Example usage
+#     URL_TO_SCRAPE = "https://www.amazon.com"
+#     SEARCH_TERMS = "10.5 white men sneakers"
+#     TOP_N = 5
+#
+#     links = search_web(URL_TO_SCRAPE + " " + SEARCH_TERMS, TOP_N)
+#
+#     print(f"Top {TOP_N} search result links for '{SEARCH_TERMS}' from {URL_TO_SCRAPE}:")
+#     for i, link in enumerate(links, start=1):
+#         print(f"{i}. {link}")
```
d1donlydfink, line 93: Seriously, make a test someday.

PR #62, opened 2025-02-04
UN-2542 Add LLMs to llm_factory

deepsaia:
Models added:
- gpt-4o-mini
- mistral-nemo
- llama3.1
- qwen2.5:14b
- deepseek-r1:14b
- nvidia-llama-3.1-405b-instruct
- nvidia-llama-3.3-70b-instruct

### DIFF
--- a/neuro_san/internals/run_context/langchain/llm_factory.py
+++ b/neuro_san/internals/run_context/langchain/llm_factory.py
@@ -52,14 +52,14 @@ LLM_ENTRIES = {
     "gpt-4o": {
         "class": ChatOpenAI,
         "api_key": "openai_api_key",
-        "max_tokens": 4096  # https://platform.openai.com/docs/models says 128,000
+        "max_tokens": 4096  # https://platform.openai.com/docs/models says 16,384
                             # but that is for input, and empirical evidence shows this is
                             # the number required for output.
     },
     "gpt-4o-mini": {
         "class": ChatOpenAI,
         "api_key": "openai_api_key",
-        "max_tokens": 4096  # https://platform.openai.com/docs/models says 128,000
+        "max_tokens": 4096  # https://platform.openai.com/docs/models says 16,384
                             # but that is for input, and empirical evidence shows this is
                             # the number required for output.
     },


neuro_san/internals/run_context/langchain/llm_factory.py
```diff
@@ -16,7 +16,8 @@
 import os
 
 from langchain_anthropic.chat_models import ChatAnthropic
-from langchain_community.chat_models.ollama import ChatOllama
+from langchain_ollama import ChatOllama
```
deepsaia, line 19: New source for ChatOllama

neuro_san/internals/run_context/langchain/llm_factory.py
```diff
@@ -50,7 +52,14 @@
...
         "api_key": "openai_api_key",
-        "max_tokens": 4096  # https://platform.openai.com/docs/models/gpt-4-turbo says 12800
+        "max_tokens": 4096  # https://platform.openai.com/docs/models says 128,000
+                            # but that is for input, and empirical evidence shows this is
+                            # the number required for output.
+    },
+    "gpt-4o-mini": {
+        "class": ChatOpenAI,
+        "api_key": "openai_api_key",
+        "max_tokens": 4096  # https://platform.openai.com/docs/models says 128,000
```
deepsaia: 4o-mini

neuro_san/internals/run_context/langchain/llm_factory.py
```diff
@@ -144,7 +153,7 @@
         "class": ChatOllama,
         "max_tokens": 4096  # Scant from https://github.com/ollama/ollama
     },
-    "llama3": {
+    "llama3.1": {
         "class": ChatOllama,
```
deepsaia: llama3.1

neuro_san/internals/run_context/langchain/llm_factory.py
```diff
@@ -160,10 +169,34 @@
...
         "max_tokens": 4096  # Scant from https://github.com/ollama/ollama
     },
+    "mistral-nemo": {
+        "class": ChatOllama,
+        "max_tokens": 4096  # Scant from https://github.com/ollama/ollama
+    },
```
deepsaia, line 179: mistral-nemo

neuro_san/internals/run_context/langchain/llm_factory.py
```diff
@@ -160,10 +169,34 @@
...
         "max_tokens": 4096  # Scant from https://github.com/ollama/ollama
     },
+    "qwen2.5:14b": {
+        "class": ChatOllama,
+        "max_tokens": 4096  # Scant from https://github.com/ollama/ollama
+    },
+    "deepseek-r1:14b": {  # Does not support tools
+        "class": ChatOllama,
+        "max_tokens": 4096
+    },
```
deepsaia, line 191: qwen2.5:14b, deepseek-r1:14b

neuro_san/internals/run_context/langchain/llm_factory.py
```diff
@@ -160,10 +169,34 @@
...
+    "nvidia-llama-3.1-405b-instruct": {
+        "use_model_name": "meta/llama-3.1-405b-instruct",
```
deepsaia, line 193: NIM hosted model

neuro_san/internals/run_context/langchain/llm_factory.py
```diff
@@ -160,10 +169,34 @@
...
+        "class": ChatNVIDIA,
+        "api_key": "nvidia_api_key",
+        "max_tokens": 4096  # From https://python.langchain.com/docs/integrations/chat/nvidia_ai_endpoints/
+    },
+    "nvidia-llama-3.3-70b-instruct": {
+        "use_model_name": "meta/llama-3.3-70b-instruct",
```
deepsaia, line 199: NIM hosted model

neuro_san/internals/run_context/langchain/llm_factory.py
d1donlydfink: Should we keep "llama3" ?

neuro_san/internals/run_context/langchain/llm_factory.py
d1donlydfink: Is DeepSeek really Ollama based?

neuro_san/internals/run_context/langchain/llm_factory.py
deepsaia: llama3 hasn't been updated in a while, we now have newer versions of llama3.1 in the same 8b parameter size.

neuro_san/internals/run_context/langchain/llm_factory.py
deepsaia: Yes the Deeseek-R1 is Ollama based. So we can use these models freely locally. 
We can use other providers as well such as [huggingface](https://huggingface.co/deepseek-ai/DeepSeek-R1) or [deepseek.com](https://api-docs.deepseek.com/) (but it'd cost us tokens)

neuro_san/internals/run_context/langchain/llm_factory.py
```diff
@@ -50,7 +52,14 @@
     "gpt-4o": {
         "class": ChatOpenAI,
         "api_key": "openai_api_key",
-        "max_tokens": 4096  # https://platform.openai.com/docs/models/gpt-4-turbo says 12800
+        "max_tokens": 4096  # https://platform.openai.com/docs/models says 128,000
```
ofrancon: https://platform.openai.com/docs/models#gpt-4o now says 16,384 tokens

neuro_san/internals/run_context/langchain/llm_factory.py
ofrancon: https://platform.openai.com/docs/models#gpt-4o-mini says 16,384 max output tokens

d1donlydfink: comments say this is for output so we should probably change.
Makes me wonder if we should distinguish between input/output token max so that it's clearer what to enter here.

neuro_san/internals/run_context/langchain/llm_factory.py
d1donlydfink: Right, but what about compatibility with older hocon files?

neuro_san/internals/run_context/langchain/llm_factory.py
deepsaia: That's right, I read the context_window there by mistake. 
The `max_output` tokens needs to be updated for multiple models, but I think thats an exercise for a later PR.

requirements.txt
```diff
@@ -37,6 +37,8 @@ langchain==0.3.7
 langchain-anthropic==0.2.4
 langchain-community==0.3.5
 langchain-openai==0.2.5
+langchain-nvidia-ai-endpoints==0.3.8
+langchain-ollama==0.2.3
```
d1donlydfink, line 41: FWIW: I just checked the licenses on these and they are both MIT, so nothing to worry about from security scans in that department.

neuro_san/internals/run_context/langchain/llm_factory.py
d1donlydfink: OK fair. If it's already whacked, then here we go.

neuro_san/internals/run_context/langchain/llm_factory.py
```diff
@@ -52,14 +52,14 @@
     "gpt-4o": {
         "class": ChatOpenAI,
         "api_key": "openai_api_key",
-        "max_tokens": 4096  # https://platform.openai.com/docs/models says 128,000
+        "max_tokens": 4096  # https://platform.openai.com/docs/models says 16,384
```
ofrancon, line 55: Context window is `128,000` tokens but output is `16,384`. So I think we should update `max_tokens` here to `16,384`. One of the earlier versions of gpt-4o (the first version?), gpt-4o-2024-05-13, had a limit at `4096`, so that's probably where this number and comment are coming from.
But I wonder if that could cause the `max_iteration` problem we're dealing with. Maybe the output gets truncated, so the agent tries again, and ends up in an infinite loop.

neuro_san/internals/run_context/langchain/llm_factory.py
```diff
@@ -52,14 +52,14 @@
...
                             # but that is for input, and empirical evidence shows this is
                             # the number required for output.
     },
     "gpt-4o-mini": {
         "class": ChatOpenAI,
         "api_key": "openai_api_key",
-        "max_tokens": 4096  # https://platform.openai.com/docs/models says 128,000
+        "max_tokens": 4096  # https://platform.openai.com/docs/models says 16,384
```
ofrancon, line 62: We should fix the message to says `128,000` context window and `16,384` max output tokens. 
And point to https://platform.openai.com/docs/models#gpt-4o-mini 
Can be another PR, no worries. Ideally that should be configurable directly from the LLM config params.

PR #68, opened 2025-02-05
Un-2812 http API for neuro-san service

andreidenissov-cog:
This PR adds a simple tornado http-server running alongside "main" service process as a sub-process.
tornado http server runs as a client to gRPC service, translating incoming http requests to gRPC API calls
and sending API results back.
It's possible we'll have to tune up tornado configuration as needed, but right now it seems to work out of the box.
Tested by running agent_cli against http service endpoint with "hello world" and "six hats" configurations.

### DIFF
--- a/neuro_san/http_sidecar/handlers/base_request_handler.py
+++ b/neuro_san/http_sidecar/handlers/base_request_handler.py
@@ -12,6 +12,7 @@
 """
 See class comment for details
 """
+import logging
 from tornado.web import RequestHandler
 
 from neuro_san.interfaces.agent_session import AgentSession
@@ -38,6 +39,7 @@ class BaseRequestHandler(RequestHandler):
         factory: AgentSessionFactory = self.application.get_session_factory()
         self.grpc_session: AgentSession = \
             factory.create_session("grpc", self.agent_name, hostname="localhost", port=self.port)
+        self.logger = logging.getLogger(self.__class__.__name__)
 
     def data_received(self, chunk):
         """
--- a/neuro_san/http_sidecar/handlers/connectivity_handler.py
+++ b/neuro_san/http_sidecar/handlers/connectivity_handler.py
@@ -14,6 +14,7 @@ See class comment for details
 """
 from typing import Any, Dict
 import json
+import traceback
 
 from neuro_san.http_sidecar.handlers.base_request_handler import BaseRequestHandler
 
@@ -36,16 +37,14 @@ class ConnectivityHandler(BaseRequestHandler):
             self.set_header("Content-Type", "application/json")
             self.write(result_dict)
 
-            # Log the response
-            print(f"Sent response: {result_dict}")
-
         except json.JSONDecodeError:
             # Handle invalid JSON input
             self.set_status(400)
             self.write({"error": "Invalid JSON format"})
-        except Exception as exc:  # pylint: disable=broad-exception-caught
+        except Exception:  # pylint: disable=broad-exception-caught
             # Handle unexpected errors
             self.set_status(500)
-            self.write({"error": f"Internal server error: {exc}"})
+            self.write({"error": "Internal server error"})
+            self.logger.error("Internal server error: %s", traceback.format_exc())
         finally:
             self.flush()
--- a/neuro_san/http_sidecar/handlers/streaming_chat_handler.py
+++ b/neuro_san/http_sidecar/handlers/streaming_chat_handler.py
@@ -14,6 +14,7 @@ See class comment for details
 """
 from typing import Any, Dict, Generator
 import json
+import traceback
 
 from google.protobuf.json_format import MessageToDict
 from google.protobuf.json_format import Parse
@@ -36,7 +37,6 @@ class StreamingChatHandler(BaseRequestHandler):
         try:
             # Parse JSON body
             data = json.loads(self.request.body)
-            print(f"Received POST request with data: {data}")
 
             grpc_request = Parse(json.dumps(data), ChatRequest())
 
@@ -52,7 +52,6 @@ class StreamingChatHandler(BaseRequestHandler):
             for result_message in result_generator:
                 result_dict: Dict[str, Any] = MessageToDict(result_message)
                 result_str: str = json.dumps(result_dict) + "\n"
-                print(f"CHAT HANDLER: |{result_dict}|\n\n")
                 self.write(result_str)
                 self.flush()
 
@@ -60,10 +59,11 @@ class StreamingChatHandler(BaseRequestHandler):
             # Handle invalid JSON input
             self.set_status(400)
             self.write({"error": "Invalid JSON format"})
-        except Exception as exc:  # pylint: disable=broad-exception-caught
+        except Exception:  # pylint: disable=broad-exception-caught
             # Handle unexpected errors
             self.set_status(500)
-            self.write({"error": f"Internal server error: {exc}"})
+            self.write({"error": "Internal server error"})
+            self.logger.error("Internal server error: %s", traceback.format_exc())
         finally:
             self.flush()
         # We are done with response stream:
--- a/neuro_san/http_sidecar/handlers/function_handler.py
+++ b/neuro_san/http_sidecar/handlers/function_handler.py
@@ -14,6 +14,7 @@ See class comment for details
 """
 from typing import Any, Dict
 import json
+import traceback
 
 from neuro_san.http_sidecar.handlers.base_request_handler import BaseRequestHandler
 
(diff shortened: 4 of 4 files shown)


neuro_san/http_sidecar/handlers/connectivity_handler.py
```diff
@@ -0,0 +1,68 @@
...
+        :param agent_name: name of agent this API method is implemented for;
+        :param method_name: name of API method;
+        :return: dynamically constructed Python type to be used by Tornado application
+            for API method handling.
+        """
+
+        def get(self):
+            """
+            Implementation of GET request handler for "connectivity" API call.
+            """
```
d1donlydfink: Why is get() only local to the build() method?
Highly unusual, please add comments here and other 2 handlers doing similar things
or make the structure flat everywhere (seems better and more than possible)

d1donlydfink: FWIW: looks like there is room for consolidation between these handlers

neuro_san/http_sidecar/handlers/streaming_chat_handler.py
```diff
@@ -0,0 +1,87 @@
...
+                self.set_header("Transfer-Encoding", "chunked")
+                # Flush headers immediately
+                self.flush()
+
+                for result_message in result_generator:
+                    result_dict: Dict[str, Any] = MessageToDict(result_message)
+                    result_str: str = json.dumps(result_dict) + "\n"
+                    print(f"CHAT HANDLER: |{result_dict}|\n\n")
+                    self.write(result_str)
+                    self.flush()
```
d1donlydfink: I guess this is the equivalent of the end of the road for yield

neuro_san/session/http_service_agent_session.py
```diff
@@ -0,0 +1,172 @@
...
+from leaf_common.time.timeout import Timeout
+
+from neuro_san.interfaces.agent_session import AgentSession
+
+
+class HttpServiceAgentSession(AgentSession):
+    """
+    Implementation of AgentSession that talks to an HTTP service.
+    This is largely only used by command-line tests.
+    """
```
d1donlydfink, line 29: I like that you call this HttpServiceAgentSession. 
Makes me want to call what we already have a GrpcServiceAgentSession.

Will this guy handle https as well?

requirements.txt
```diff
@@ -47,6 +47,9 @@ pydantic==2.9.2
 # httpx pinned only for this reason: https://github.com/openai/openai-python/issues/1903
 httpx==0.27.2
 
+# HTTP server part
+tornado>=6.4.2
+
```
andreidenissov-cog, line 52: Add tornado server to requirements: 6.4.2 is current version we use.

neuro_san/client/agent_cli.py
```diff
@@ -188,11 +188,12 @@ def add_args(self, arg_parser: argparse.ArgumentParser):
...
         group.add_argument("--connection", default="direct", type=str,
-                           choices=["grpc", "service", "direct"],
+                           choices=["grpc", "service", "direct", "http"],
                            help="""
 The type of connection to initiate. Choices are to connect to:
     "grpc"      - an agent service via gRPC. (The default).  Needs host and port.
     "service"   - compatibility synonym for 'grpc' above.
+    "http"      - an agent service via HTTP. Needs host and port.
     "direct"    - a session via library.
 All choices require an agent name.
```
andreidenissov-cog, line 198: Add "http" connectivity option to selection.

neuro_san/http_sidecar/handlers/base_request_handler.py
```diff
@@ -0,0 +1,49 @@
...
+        :param port: gRPC service port.
+        """
+        #
+        self.agent_name: str = agent_name
+        self.port: int = port
+        factory: AgentSessionFactory = self.application.get_session_factory()
+        self.grpc_session: AgentSession = \
+            factory.create_session("grpc", self.agent_name, hostname="localhost", port=self.port)
+        self.logger = logging.getLogger(self.__class__.__name__)
+
```
andreidenissov-cog, line 43: Yes, all this is done outside of __init__() constructor,
because we cannot define our own constructor for RequestHandler.

neuro_san/http_sidecar/handlers/connectivity_handler.py
```diff
@@ -0,0 +1,50 @@
...
+        except json.JSONDecodeError:
+            # Handle invalid JSON input
+            self.set_status(400)
+            self.write({"error": "Invalid JSON format"})
+        except Exception:  # pylint: disable=broad-exception-caught
+            # Handle unexpected errors
+            self.set_status(500)
+            self.write({"error": "Internal server error"})
+            self.logger.error("Internal server error: %s", traceback.format_exc())
+        finally:
```
andreidenissov-cog, line 49: Here and in all handler classes: as suggested by Github security bot, we send back over a wire a generic error message only, and dump all the details into local log.

neuro_san/http_sidecar/handlers/connectivity_handler.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def get(self):
+        """
+        Implementation of GET request handler for "connectivity" API call.
+        """
+
+        try:
+            data: Dict[str, Any] = {}
+            result_dict: Dict[str, Any] = self.grpc_session.connectivity(data)
+
```
andreidenissov-cog, line 35: This self.grpc_session instance member has been setup by initialize() call
executed by Tornado machinery (see class BaseRequestHandler, which is a super-class for all our specific handlers)

neuro_san/http_sidecar/http_server_application.py
```diff
@@ -0,0 +1,37 @@
...
+    additional context for requests handling:
+    our internal session factory to connect to agents gRPC service.
+    """
+    def __init__(self, *args, **kwargs):
+        # Initialize the base Application
+        super().__init__(*args, **kwargs)
+
+        # Initialize and store dependencies:
+        self.session_factory: AgentSessionFactory = AgentSessionFactory()
+
```
andreidenissov-cog, line 32: Put sessions factory into our own Application sub-class, so it could be accessed by request handlers we implement. Application object is available to all its RequestHandler instances.

neuro_san/http_sidecar/http_sidecar.py
```diff
@@ -0,0 +1,78 @@
...
+from neuro_san.http_sidecar.handlers.function_handler import FunctionHandler
+from neuro_san.http_sidecar.handlers.streaming_chat_handler import StreamingChatHandler
+
+
+class HttpSidecar:
+    """
+    Class provides simple http endpoint for neuro-san API,
+    working as a client to neuro-san gRPC service.
+    """
+    def __init__(self, port: int, agents: Dict[str, Any]):
```
andreidenissov-cog, line 33: This class is a Tornado "application", meaning http server with request paths and corresponding handlers defined.
It can be run in different modes, but we will run it in a separate process alongside "main" gRPC service process.

neuro_san/session/http_service_agent_session.py
andreidenissov-cog: https - a good question. For this, we'll need to properly setup Tornado server we are running,
certificates and all. Not sure what should be done on client side, but we can find out.

neuro_san/session/http_service_agent_session.py
```diff
@@ -0,0 +1,172 @@
...
+            are produced until the system decides there are no more messages to be sent.
+        """
+        path: str = self._get_request_path("streaming_chat")
+        with requests.post(path, json=request_dict,
+                           stream=True,
+                           timeout=self.timeout_in_seconds) as response:
+            response.raise_for_status()
+
+            for line in response.iter_lines(decode_unicode=True):
+                if line.strip():  # Skip empty lines
```
andreidenissov-cog, line 169: Kinda weird that we need to iterate response line by line, but this is what seems to get it really streaming,
instead of buffering all of the output and then dumping out the whole thing.

neuro_san/service/agent_main_loop.py
```diff
@@ -133,7 +135,17 @@ def main_loop(self):
                                   server_name_for_logs=self.server_name_for_logs,
                                   request_limit=self.request_limit,
                                   service_prefix=self.service_prefix)
-        self.server.serve()
+
+        # Start HTTP server side-car:
+        http_sidecar = HttpSidecar(self.port, self.tool_registries)
+        http_server_process = multiprocessing.Process(target=http_sidecar)
+        http_server_process.start()
+
```
andreidenissov-cog, line 143: Start a http-serving sub-process together with starting gRPC service.

neuro_san/http_sidecar/handlers/base_request_handler.py
d1donlydfink: Worth adding to the method comment

neuro_san/session/http_service_agent_session.py
```diff
@@ -0,0 +1,172 @@
...
+        :param umbrella_timeout: A Timeout object under which the length of all
+                        looping and retries should be considered
+        :param streaming_timeout_in_seconds: timeout to use when streaming to/from
+                        the service. Default is None, indicating connection should
+                        stay open until the (last) result is yielded.
+        :param agent_name: The name of the agent to talk to
+        :param service_prefix: The service prefix to use. Default is None,
+                        implying the policy in AgentServiceStub takes over.
+        """
+        _ = metadata
```
d1donlydfink, line 69: At some point you will want to forward the metadata.
For instance, the current agent_cli client populates the user_id metadata field with $USER and it will show up in the server logs.

neuro_san/session/http_service_agent_session.py
d1donlydfink: I think darren discovered it send 1 streamed response per line

PR #75, opened 2025-02-18
Add a README for the internals

d1donlydfink:


neuro_san/internals/README.md
```diff
@@ -0,0 +1,131 @@
+Here is how streaming_chat() works for the langchain implementation in perhaps too much detail:
```
d1donlydfink, line 1: A description of the algorithm I needed to write up for patent purposes

neuro_san/internals/README.md
```diff
@@ -0,0 +1,131 @@
...
+            * A placeholder message which holds the (impending) chat history
+            * A "human" message which holds the input to the calle (user or agent generated)
+            * A placeholder message for the "agent scratchpad" to allow a place for agent 
+                decisions to be kept.
+        * Finally, a langchain Agent instance itself is created passing it the LLM instance,
+            the tools list and the prompt template.
+        * All of the above are kept as state in the RunContext instance
+    * A timestamp is taken as to when the user-input came in.
+    * Any last-response information is cleared for the polling chat() case.
+    * the sly_data (private data) dictionary is prepared for dissemination to other agents
```
andreidenissov-cog, line 59: Are we still using polling chats? I thought they are deprecated by now?

neuro_san/internals/README.md
```diff
@@ -0,0 +1,131 @@
...
+    * the method is passed a Run result which holds a list of tool names that
+        the agent logic wishes to invoke and with what parameters.
+        (These are all determined by lower-level langchain Agent logic in conjuction
+         with the LLM selected. That is - the LLM itself decides what tools to call.)
+    * For each tool that the agent wants to be called:
+        * We get the name of the tool to be called  
+        * We get the arguments to be passed to the tool
+        * We instantiate a new node in our graph via the AgentToolFactory that corresponds 
+            to the agent_spec of the tool to be called.  The Factory will create one of the
+            following nodes based on the spec:
```
andreidenissov-cog, line 87: Is each node in the graph instantiated only once? What happens if some tool has to be used several times?
CodedTool for example?

neuro_san/internals/README.md
d1donlydfink: They are deprecated but not yet deleted.
When @ofrancon gives the signal that he has entirely moved to streaming chat, we can start to delete.

neuro_san/internals/README.md
d1donlydfink: Great question. Nodes can be instantiated several times for the same tool references.  This is precisely why the components of our Origin descriptions have a "tool" and an "instantiation_index" to keep these straight.
Easiest place to see multiple instantiations of the same tool is esp_decision_assistant which often invokes multiple predictors.

andreidenissov-cog: So what about resource management? Will "done" tool instantiation release all the corresponding resources?
Also: several tool X instantiations are not strictly sequential in general case, right?

d1donlydfink: > So what about resource management? Will "done" tool instantiation release all the corresponding resources?

Put it this way: RunContext.delete_resources() gives an opportunity to free up resources. If that has to be deferred after someone in the middle can create, then it's possible we might have to build a mechanism to delete ad-hoc shared resources.   This is what UN-2871 gets at. https://leaf-ai.atlassian.net/browse/UN-2871

> Also: several tool X instantiations are not strictly sequential in general case, right?

Correct. more or less.  Our instantiation_index for a given tool is only ever incremented by one, but when a tool
gets instantiated, it might be done by several different paths, and each time that index gets a bump:
A -> B -> C (1)
A -> B -> C (2)
A -> D -> C (3)
A-> EFGHIJK -> C (4)

PR #76, opened 2025-02-19
UN-2813 create event loop executor per chat request.

andreidenissov-cog:
This PR addresses an issue of having a single global event loop running all async tasks in neuro-san server.
Now a separate event loop (in the form of AsyncioExecutor instance) will be created for each incoming chat request
and will be owned by an instance of SessionInvocationContext.

### DIFF
--- a/neuro_san/client/agent_cli.py
+++ b/neuro_san/client/agent_cli.py
@@ -188,12 +188,13 @@ Some suggestions:
         group = arg_parser.add_argument_group(title="Session Type",
                                               description="How will we connect to neuro-san?")
         group.add_argument("--connection", default="direct", type=str,
-                           choices=["grpc", "service", "direct", "http"],
+                           choices=["grpc", "service", "direct", "http", "https"],
                            help="""
 The type of connection to initiate. Choices are to connect to:
     "grpc"      - an agent service via gRPC. (The default).  Needs host and port.
     "service"   - compatibility synonym for 'grpc' above.
     "http"      - an agent service via HTTP. Needs host and port.
+    "https"     - an agent service via secure HTTP. Needs host and port.
     "direct"    - a session via library.
 All choices require an agent name.
 """)
@@ -205,6 +206,10 @@ All choices require an agent name.
                            help="Use a direct/library call for the chat")
         group.add_argument("--http", dest="connection", action="store_const", const="http",
                            help="Use a HTTP service connection")
+        group.add_argument("--https", dest="connection", action="store_const", const="https",
+                           help="Use a secure HTTP service connection. "
+                                "Requires your agent server to be set up with certificates that are well known. "
+                                "This is not something that our basic server setup supports out-of-the-box.")
         self.arg_groups[group.title] = group
 
         # How will we connect to a server?
@@ -263,7 +268,7 @@ Have external tools that can be found in the local agent manifest use a service 
         group.add_argument("--thinking_file", type=str, default="/tmp/agent_thinking.txt",
                            help="File that captures agent thinking. "
                                 "This is a separate text stream from the user/assistant chat")
-        group.add_argument("--thinking_dir", default=False, action="store_true",
+        group.add_argument("--thinking_dir", default=True, action="store_true",
                            help="Use the basis of the thinking_file as a directory to capture "
                                 "internal agent chatter in separate files. "
                                 "This is a separate text stream from the user/assistant chat. "
--- a/neuro_san/client/agent_session_factory.py
+++ b/neuro_san/client/agent_session_factory.py
@@ -9,6 +9,7 @@
 # neuro-san SDK Software in commercial settings.
 #
 # END COPYRIGHT
+from typing import Any
 from typing import Dict
 
 from neuro_san.client.direct_agent_session_factory import DirectAgentSessionFactory
@@ -52,15 +53,19 @@ class AgentSessionFactory:
         elif session_type in ("service", "grpc"):
             session = GrpcServiceAgentSession(host=hostname, port=port, agent_name=agent_name,
                                               metadata=metadata)
-        elif session_type == "http":
+        elif session_type in ("http", "https"):
 
             # If there is no port really specified, use the other default port
             use_port = port
             if port is None or port == AgentSession.DEFAULT_PORT:
                 use_port = AgentSession.DEFAULT_HTTP_PORT
 
+            security_cfg: Dict[str, Any] = None
+            if session_type == "https":
+                # For now, to get the https scheme
+                security_cfg = {}
             session = HttpServiceAgentSession(host=hostname, port=use_port, agent_name=agent_name,
-                                              metadata=metadata)
+                                              security_cfg=security_cfg, metadata=metadata)
         else:
             # Incorrectly flagged as destination of Trust Boundary Violation 2
             #   Reason: This is the place where the session_type enforced-string argument is
--- a/neuro_san/deploy/Dockerfile
+++ b/neuro_san/deploy/Dockerfile
@@ -14,7 +14,7 @@
 # and check it into the neuro-san repo or modify it in place with the
 # neuro-san source code, chances are you shouldn't. Instead, strongly
 # consider creating your own repo with your own agents and using this
-# Dockerfile and accompanying scipts as a basis to riff on for your
+# Dockerfile and accompanying scripts as a basis to riff on for your
 # own projects.
 #
 # Use ./neuro_san/deploy/build.sh to build this container only *after*
--- a/neuro_san/internals/run_context/langchain/langchain_token_counter.py
+++ b/neuro_san/internals/run_context/langchain/langchain_token_counter.py
@@ -66,6 +66,11 @@ class LangChainTokenCounter:
                 "completion_tokens": callback.completion_tokens,
                 "successful_requests": callback.successful_requests,
                 "total_cost": callback.total_cost,
+                "caveats": [
+                    "Langchain only allows accounting for OpenAI and Anthropic models.",
+                    "Each LLM Branch Node also includes accounting for each of its callees.",
+                    "Each LLM Branch Node does not yet properly account for its own tokens.",
+                ]
             }
 
         return token_dict
--- a/neuro_san/session/http_service_agent_session.py
+++ b/neuro_san/session/http_service_agent_session.py
@@ -66,10 +66,10 @@ class HttpServiceAgentSession(AgentSession):
         :param service_prefix: The service prefix to use. Default is None,
                         implying the policy in AgentServiceStub takes over.
         """
-        _ = security_cfg
         _ = umbrella_timeout
         _ = streaming_timeout_in_seconds
 
+        self.security_cfg: Dict[str, Any] = security_cfg
         self.service_prefix: str = service_prefix
         self.use_host: str = "localhost"
         if host is not None:
@@ -84,9 +84,13 @@ class HttpServiceAgentSession(AgentSession):
         self.metadata: Dict[str, str] = metadata
 
     def _get_request_path(self, function: str):
+        scheme: str = "http"
+        if self.security_cfg is not None:
+            scheme = "https"
+
         if self.service_prefix is None or len(self.service_prefix) == 0:
-            return f"http://{self.use_host}:{self.use_port}/api/v1/{self.agent_name}/{function}"
-        return f"http://{self.use_host}:{self.use_port}/api/v1/{self.service_prefix}.{self.agent_name}/{function}"
+            return f"{scheme}://{self.use_host}:{self.use_port}/api/v1/{self.agent_name}/{function}"
+        return f"{scheme}://{self.use_host}:{self.use_port}/api/v1/{self.service_prefix}/{self.agent_name}/{function}"
 
     def help_message(self, path: str) -> str:
         """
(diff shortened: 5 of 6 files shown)


neuro_san/service/agent_service.py
```diff
@@ -37,7 +36,7 @@
 from neuro_san.session.session_invocation_context import SessionInvocationContext
 
 # A list of methods to not log requests for
-# Some of these can be way to chatty
+# Some of these can be way too chatty
 DO_NOT_LOG_REQUESTS = [
```
andreidenissov-cog, line 41: typo.

neuro_san/session/external_agent_session_factory.py
```diff
@@ -32,7 +32,7 @@ class ExternalAgentSessionFactory(AsyncAgentSessionFactory):
 
     def __init__(self, use_direct: bool = False):
         """
-        Constructuor
+        Constructor
```
andreidenissov-cog, line 36: Typo.

neuro_san/session/session_invocation_context.py
```diff
@@ -31,22 +33,21 @@ class SessionInvocationContext(InvocationContext):
...
                          the header. Default is None. Preferred format is a
                          dictionary of string keys to string values.
         """
 
         self.async_session_factory: AsyncAgentSessionFactory = async_session_factory
-        self.asyncio_executor: AsyncioExecutor = asyncio_executor
+        # Get an async executor to run all tasks for this session instance:
+        self.asyncio_executor: AsyncioExecutor = AsyncioExecutor()
+        self.asyncio_executor.start()
         self.origination: Origination = Origination()
```
andreidenissov-cog, line 51: This is basically the main part of the change:
instead of taking externally provided AsyncioExecutor as a constructor parameter,
create separate new AsyncioExecutor to use in each SessionInvocationContext instance.
That will ensure that different client requests will not affect each other by mutual blocking or going into infinite loops etc.

neuro_san/session/session_invocation_context.py
```diff
@@ -108,3 +109,11 @@ def reset_origination(self):
         Resets the origination
         """
         self.origination = Origination()
+
+    def close(self):
+        """
+        Release resources owned by this context
```
andreidenissov-cog, line 124: AsyncioExecutor instance is now a resource owned by SessionInvocationContext.
We need a method to release and clean up this resource.

neuro_san/session/chat_session_map.py
```diff
@@ -53,7 +53,10 @@ def __init__(self, init_argument: Dict[str, Any]):
 
         # Store access to the globals
         self.chat_sessions: Dict[str, ChatSession] = init_argument.get("chat_sessions")
-        self.asyncio_executor: AsyncioExecutor = init_argument.get("executor")
+
+        # Run our own async executor for cleanup tasks
+        self.asyncio_executor: AsyncioExecutor = AsyncioExecutor()
+        self.asyncio_executor.start()
```
andreidenissov-cog, line 60: Don't take AsyncioExecutor intended for cleanup tasks from somewhere.
Run your own. And we are OK for it to run as long as our service runs.

neuro_san/session/chat_session_map.py
```diff
@@ -142,7 +145,7 @@ def prune(self):
 
             # We want to delete the resources of all the existing chat_sessions
             # that have not had any activity but need to keep their references
-            # around after we clear them out from the list so we can do the
+            # around after we clear them out from the list, so we can do the
             # delete_resources() outside the lock.
```
andreidenissov-cog, line 149: Typos.

neuro_san/service/agent_service.py
```diff
@@ -54,7 +53,6 @@ def __init__(self,
                  request_logger: RequestLogger,
                  security_cfg: Dict[str, Any],
                  chat_session_map: ChatSessionMap,
-                 asyncio_executor: AsyncioExecutor,
                  agent_name: str,
                  tool_registry: AgentToolRegistry):
         """
```
andreidenissov-cog: No externally provided AsyncioExecutors.

neuro_san/service/agent_server.py
```diff
@@ -16,7 +16,6 @@
 import logging
 import os
 
-from leaf_common.asyncio.asyncio_executor import AsyncioExecutor
 from leaf_server_common.logging.logging_setup import setup_logging
```
andreidenissov-cog, line 19: No externally provided AsyncioExecutors.

neuro_san/client/direct_agent_session_factory.py
```diff
@@ -62,7 +62,7 @@ def create_session(self, agent_name: str, use_direct: bool = False,
         factory = ExternalAgentSessionFactory(use_direct=use_direct)
         tool_registry: AgentToolRegistry = factory.get_tool_registry(agent_name, self.manifest_tool_registries)
 
-        invocation_context = SessionInvocationContext(factory, self.asyncio_executor, metadata)
+        invocation_context = SessionInvocationContext(factory, metadata)
         session: DirectAgentSession = DirectAgentSession(chat_session_map=self.chat_session_map,
```
andreidenissov-cog, line 60: SessionInvocationContext will make its very own AsyncioExecutor.

neuro_san/service/agent_main_loop.py
```diff
@@ -63,12 +59,10 @@ def __init__(self, service_prefix: str = DEFAULT_SERVICE_PREFIX):
                         one configurable item that is likely to be shared with other code.
         """
         self.port: int = 0
-        self.asyncio_executor = ASYNCIO_EXECUTOR
```
andreidenissov-cog, line 64: This is where we created a single global AsyncioExecutor and then were passing it all over the place.
Not anymore.

requirements-private.txt
```diff
@@ -8,7 +8,7 @@
 # Do the source-available stuff first to fast-fail if credentials are not right
 
 # Public repos
-git+https://${LEAF_SOURCE_CREDENTIALS}@github.com/leaf-ai/leaf-common.git@1.2.18
+git+https://${LEAF_SOURCE_CREDENTIALS}@github.com/leaf-ai/leaf-common.git@1.2.19
```
andreidenissov-cog, line 12: Use latest leaf-common version with AsyncioExecutor  fixes.

neuro_san/service/agent_main_loop.py
```diff
@@ -43,9 +42,6 @@
 # string keys -> ChatSession implementations
 CHAT_SESSIONS: Dict[str, ChatSession] = {}
 
-# A single global for executing asyncio stuff in the background
-ASYNCIO_EXECUTOR: AsyncioExecutor = AsyncioExecutor()
```
d1donlydfink, line 48: Ahhh!
If you can make this go away and everything still works, you are most certainly on the right path.

neuro_san/service/agent_main_loop.py
```diff
@@ -130,14 +124,10 @@ def main_loop(self):
         """
         self.parse_args()
 
-        # Start up the background thread which will process asyncio stuff
-        self.asyncio_executor.start()
```
d1donlydfink, line 149: Where do start() here and shutdown() below equivalent functionality migrate to for each new AsyncioExecutor?

neuro_san/client/direct_agent_session_factory.py
d1donlydfink: If this guy created an AsynioExecutor or had one passed in, I would suspect that that guy
could be deleted up the chain as well.  Is that not the case?

neuro_san/session/session_invocation_context.py
d1donlydfink: Style nit:
It might be better to create a separate start() method on the SessionInvocationContext.

It's usually a good idea to have constructors be there to arrange the set up of member variables,
but if there is policy that is heavier-weight, that stuff is often better relegated to a separate method
(like start( ) or init()).

For me: Constructors say "I am making a new instance".
Methods like start() say "I am doing something with my new instance".

This way, if they are copy.copy()-ed in any way, it's very clear that it's the data inside that gets duplicated
and not the extra doing-somethings.

Provides for better symmetry with your close() method below as well.

neuro_san/session/chat_session_map.py
d1donlydfink: This is really nice because eventually this ChatSesionMap will go away and with it the need for this extra executor.

neuro_san/service/agent_service.py
```diff
@@ -223,7 +218,7 @@ def Chat(self, request: service_messages.ChatRequest,
 
         # Delegate to Direct*Session
         factory = ExternalAgentSessionFactory(use_direct=False)
-        invocation_context = SessionInvocationContext(factory, self.asyncio_executor)
+        invocation_context = SessionInvocationContext(factory)
```
d1donlydfink, line 224: It is fitting that the AsyncioExecutor gets created in the Service for each call.

neuro_san/session/session_invocation_context.py
d1donlydfink: ... and we need to actually call it!

TBH: With a global executor, I never paid too much attention to whether or not things were cleaned up properly (threads joined, event loops destroyed, etc).  With this we will need to look more closely at that so we do not leak resources.

neuro_san/internals/interfaces/invocation_context.py
```diff
@@ -65,3 +72,9 @@ def get_metadata(self) -> Dict[str, str]:
         :return: The metadata to pass along with any request
         """
         raise NotImplementedError
+
+    def close(self):
+        """
+        Release resources owned by this context
+        """
+        raise NotImplementedError
```
andreidenissov-cog, line 81: Added these life-cycle management methods to abstract interface.

neuro_san/session/async_direct_agent_session.py
```diff
@@ -63,17 +63,8 @@ def __init__(self,
...
         self.tool_registry: AgentToolRegistry = tool_registry
 
-        # For convenience
-        if self.invocation_context is not None and \
-                self.invocation_context.get_asyncio_executor() is None:
-            self.we_created_executor = True
-            asyncio_executor = AsyncioExecutor()
-            self.invocation_context.set_asyncio_executor(asyncio_executor)
-            asyncio_executor.start()
-
```
andreidenissov-cog, line 76: With our new way of doing things, lifetime of AsyncioExecutor is exactly the same as lifetime of  SessionInvocationContext (ownership relation and all that). Simpler to manage things.
So if we have self.invocation_context, we have AsyncioExecutor alive and running inside.

neuro_san/session/async_direct_agent_session.py
```diff
@@ -237,8 +228,5 @@ def close(self):
         """
         if self.invocation_context is None:
             return
-
-        asyncio_executor: AsyncioExecutor = self.invocation_context.get_asyncio_executor()
-        if self.we_created_executor and asyncio_executor is not None:
-            asyncio_executor.shutdown()
-            self.invocation_context.set_asyncio_executor(None)
+        self.invocation_context.close()
+        self.invocation_context = None
```
andreidenissov-cog, line 232: More interesting question is: what is lifetime of SessionInvocationContext relative to Session object?
It seems to me from the code that DirectSession instance actually owns its SessionInvocationContext (?)
and so when we end-of-life Session object, we should close its SessionInvocationContext.
But how do we close Sessions? 
I see that you are retiring (closing) them if they are not used for a while?
Custom GC of sorts - am I missing something?

neuro_san/session/direct_agent_session.py
```diff
@@ -64,17 +64,8 @@ def __init__(self,
...
 
-        # For convenience
-        if self.invocation_context is not None and \
-                self.invocation_context.get_asyncio_executor() is None:
-            self.we_created_executor = True
-            asyncio_executor = AsyncioExecutor()
-            self.invocation_context.set_asyncio_executor(asyncio_executor)
-            asyncio_executor.start()
-
     def function(self, request_dict: Dict[str, Any]) -> Dict[str, Any]:
```
andreidenissov-cog, line 69: Same logic here - if we have self.invocation_context, we have AsyncioExecutor running inside.

neuro_san/session/session_invocation_context.py
```diff
@@ -31,28 +31,34 @@ class SessionInvocationContext(InvocationContext):
...
         self.request_reporting: Dict[str, Any] = {}
 
+    def start(self):
+        """
+        Starts the active components of this invocation context.
+        Do this separately from constructor for more control.
+        Currently, we only start internal AsyncioExecutor.
+        """
+        self.asyncio_executor.start()
+
```
d1donlydfink, line 64: Add something like this here:

```
        # Set up logging fields within the thread so we have consistent logging from async calls.
        # Ignore the future that is returned - we trust it will get done.
        _: Future = self.asyncio_executor.submit("logging_setup", setup_extra_logging_fields,
                                                 metadata_dict=self.metadata)
```
... with these imports above ...

```
from asyncio import Future
from leaf_server_common.logging.logging_setup import setup_extra_logging_fields
```

... and we should start to see consistent structured logging from the async calls I think.
This will be important to the 1C guys.

neuro_san/session/async_direct_agent_session.py
d1donlydfink: SessionInvocationContext gets created just before/outside a DirectAgentSession gets invoked for asynchronous calls (like streaming_chat() - Not necessarily for function() or connectivity()).  This tends to happen in AgentService but also happens in the client with --direct flag.

Tricky bit: It's possible that DAS can create/call other DAS when being used as a library and "nearby" external agents (networks in same manifest file) are invoked.   So close()-ing SIC when any DAS is done is not quite the right answer.  Best to close() SIC just after its corresponding initial DAS is done being used.  That is - SIC has slightly wider scope than DAS.

As for sessions themselves, yes there is a close() on them today, but those should probably go away.
I don't think the close() gets called anywhere in top-of-trunk.

neuro_san/internals/interfaces/invocation_context.py
d1donlydfink: It's likely that we don't need close() on the interface definition.
Whoever constructs them will also need to close() them and at those levels it's a concrete class, not an interface being passed around trying to avoid dependency tangles because it's used at low levels of the code.
That is to say that we don't really want that low-level code ever doing the close()-ing.

PR #78, opened 2025-02-22
UN-2851 Token counting per request

d1donlydfink:
Start reporting token accounting stats for each request. This was a request from the 1C guys.

Reporting actually can happen on a per-agent level, and when available, that now comes over as a _structure_ on an AgentMessage that has no string text/content.

There are some caveats here though:
1. Only certain kinds of LLMs - OpenAI and Anthropic actually have the token reporting in place
2. Branch LLM nodes that use tool-calling agents seem to report the cumulative token accounting for the entire depth of their callees, yet they are not able to provide information on themselves - So we still have some missing information that will require quite a bit of digging to get good numbers (I have tried and am giving up for now).

Still the idea is that something is better than nothing, and for now this means that the final front-man token stats is the best we can do for a single-request's worth of accounting.

In the future, when everything is actually accounted for, we might get more sophisticated in terms of our breakdowns in this reporting.  I can imagine per-agent, per-model, per-request and periodic summaries of per-service, per-server stats.
But let's not get ahead of ourselves. Baby steps.

@ofrancon , @dsargent , @swensel : With this change for the 1st time some messages will start to have the structure field filled in instead of the text/content field

### DIFF
--- a/neuro_san/internals/run_context/langchain/langchain_run_context.py
+++ b/neuro_san/internals/run_context/langchain/langchain_run_context.py
@@ -389,13 +389,13 @@ class LangChainRunContext(RunContext):
         if token_dict is not None and bool(token_dict):
 
             # Accumulate what we learned about tokens to request reporting.
-            # For now we just overwrite the token_counting key because we know
+            # For now we just overwrite the one key because we know
             # the last one out will be the front man, and as of 2/21/25 his stats
             # are cumulative.  At some point we might want a finer-grained breakdown
             # that perhaps contributes to a service/er-wide periodic token stats breakdown
             # of some kind.  For now, get something going.
             request_reporting: Dict[str, Any] = self.invocation_context.get_request_reporting()
-            request_reporting["token_counting"] = token_dict
+            request_reporting["token_accounting"] = token_dict
 
             # We actually have a token dictionary to report, so go there.
             agent_message = AgentMessage(structure=token_dict)


neuro_san/client/thinking_file_message_processor.py
```diff
@@ -71,6 +73,11 @@ def write_message(self, response: Dict[str, Any], origin_str: str):
         message_type_str: str = ChatMessageType.to_string(message_type)
 
         text: str = response.get("text")
+        structure: Dict[str, Any] = response.get("structure")
+
+        if structure is not None:
+            # There is no real text, but there is a structure. JSON-ify it.
+            text = json.dumps(structure, indent=4, sort_keys=True)
```
d1donlydfink, line 80: Allow the client to process messages that have structures associated with them.

neuro_san/internals/messages/agent_message.py
```diff
@@ -25,12 +25,15 @@ class AgentMessage(BaseMessage):
 
     type: Literal["agent"] = "agent"
 
-    def __init__(self, content: Union[str, List[Union[str, Dict]]], **kwargs: Any) -> None:
+    def __init__(self, content: Union[str, List[Union[str, Dict]]] = "",
+                 structure: Dict[str, Any] = None, **kwargs: Any) -> None:
```
d1donlydfink, line 29: Allow AgentMessages to have structures (dictionaries) instead of string content if they want.

neuro_san/internals/messages/message_utils.py
```diff
@@ -159,7 +159,8 @@ def convert_to_chat_message(message: BaseMessage, origin: List[Dict[str, Any]] =
     optionals: Dict[str, str] = {
         "content": "text",
         "chat_context": "chat_context",
-        "tool_result_origin": "tool_result_origin"
+        "tool_result_origin": "tool_result_origin",
+        "structure": "structure",
```
d1donlydfink, line 163: Pass along the structure to the right field when converting the AgentMessage to chat.ChatMessage

neuro_san/internals/run_context/langchain/journaling_callback_handler.py
```diff
@@ -69,8 +69,6 @@ async def on_llm_end(self, response: LLMResult,
         generations = response.generations[0]
         first_generation = generations[0]
         if isinstance(first_generation, ChatGeneration):
-            # DEF   There can be token counts in the usage_metadata field as well
-            #       we might be able to use later
```
d1donlydfink, line 73: This comment is true ... sometimes, but it turns out it's not quite how langchain is currently directing us to count tokens

neuro_san/internals/run_context/langchain/langchain_run_context.py
```diff
@@ -167,7 +171,7 @@ async def create_resources(self, assistant_name: str,
             callbacks.append(LoggingCallbackHandler(logging.getLogger(full_name)))
 
         # Create the model we will use.
-        llm: BaseLanguageModel = LlmFactory.create_llm(self.llm_config, callbacks=callbacks)
+        self.llm = LlmFactory.create_llm(self.llm_config, callbacks=callbacks)
```
d1donlydfink, line 173: Keep a reference to the LLM. We will need it later now.

neuro_san/internals/run_context/langchain/langchain_run_context.py
```diff
@@ -358,6 +362,33 @@ async def wait_on_run(self, run: Run, journal: Journal = None) -> Run:
...
+        return run
+
+    async def ainvoke(self, agent_executor: AgentExecutor, inputs: Dict[str, Any], invoke_config: Dict[str, Any]):
+        """
+        Set the agent in motion
+
+        :param agent_executor: The AgentExecutor to invoke
+        :param inputs: The inputs to the agent_executor
+        :param invoke_config: The invoke_config to send to the agent_executor
+        """
```
d1donlydfink, line 413: Break out what we were doing before for the tail end of wait_on_run()  into a new invoke() method.

neuro_san/internals/run_context/langchain/langchain_run_context.py
```diff
@@ -412,8 +443,6 @@ async def wait_on_run(self, run: Run, journal: Journal = None) -> Run:
         # Chat history is updated in write_message
         await self.journal.write_message(return_message)
 
-        return run
```
d1donlydfink, line 415: Because of the new function, this return moves up to the end of wait_on_run()

neuro_san/internals/run_context/langchain/langchain_token_counter.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+
+    @staticmethod
+    def get_callback_for_llm(llm: BaseLanguageModel) -> Any:
+        """
+        :param llm: A BaseLanguageModel returned from LlmFactory.
+        :return: A handle to a no-args function, that when called will
+                open up a context manager for token counting.
+                Can be None if no such entity exists for the llm type
+        """
```
d1donlydfink, line 39: Given the LLM, return the correct context-manager method that returns the callback.

neuro_san/internals/run_context/langchain/langchain_token_counter.py
```diff
@@ -0,0 +1,71 @@
...
+
+        return None
+
+    @staticmethod
+    def normalize_token_count(callback: BaseCallbackHandler) -> Dict[str, Any]:
+        """
+        Normalizes the values in the token counting callback into a standard dictionary
+
+        :param callback: A BaseCallbackHandler instance that contains token counting information
+        """
```
d1donlydfink, line 55: Given the actual callback from the above context-manager, convert that into a standardized dictionary.
So far OpenAI and Anthropic are the only ones we can handle. It's possible new versions of langchain will
give us more options for this kind of reporting, but there's no guarantee they will have the same format.

neuro_san/internals/run_context/langchain/langchain_run_context.py
```diff
@@ -358,6 +361,47 @@ async def wait_on_run(self, run: Run, journal: Journal = None) -> Run:
             }
         }
 
+        # Attempt to count tokens/costs while invoking the agent.
+        # The means by which this happens is on a per-LLM basis, so get the right hook
+        # given the LLM we've got.
+        token_dict: Dict[str, Any] = {}
+        token_counter_context_manager = LangChainTokenCounter.get_callback_for_llm(self.llm)
```
d1donlydfink, line 367: Point of code flow: We are inserting new logic at the end of wait_on_run() and to make it more manageable,
what had been left here moves into a new invoke() method below.

Get the correct context manager for token counting given the LLM type.

neuro_san/internals/run_context/langchain/langchain_run_context.py
```diff
@@ -358,6 +361,47 @@ async def wait_on_run(self, run: Run, journal: Journal = None) -> Run:
...
+            #   https://python.langchain.com/docs/how_to/llm_token_usage_tracking/#using-callbacks
+            #
+            # Caveats:
+            # * In using this context manager approach, any tool that is called
+            #   also has its token counts contributing to its callers for better or worse.
+            # * As of 2/21/25, it seems that tool-calling agents (branch nodes) are not
+            #   registering their tokens correctly. Not sure if this is a bug in langchain
+            #   or there is something we are not doing in that scenario that we should be.
+            with token_counter_context_manager() as callback:
+                await self.ainvoke(agent_executor, inputs, invoke_config)
```
d1donlydfink, line 381: If we actually can count tokens, do so by opening a context/callback from what we got before.
The comments above carry some caveats - I  believe we have incomplete accounting
and I am not sure there is much we can do about it at the moment. 

I have already spent a few days looking into this to no avail so far. The remaining choices are:
* upgrade langchain version to see if stuff starts to work - easy, but risky and maybe no fix is yet available.
* Go digging in langchain-land to see if we can hack/fix it ourselves - will take a good while.
🤷

neuro_san/internals/run_context/langchain/langchain_run_context.py
```diff
@@ -358,6 +361,46 @@ async def wait_on_run(self, run: Run, journal: Journal = None) -> Run:
...
+            # No token counting was available for the LLM, but we still need to invoke.
+            await self.ainvoke(agent_executor, inputs, invoke_config)
+
+        # Token counting results are collected in the callback, if there are any.
+        # Different LLMs can count things in different ways, so normalize.
+        token_dict: Dict[str, Any] = LangChainTokenCounter.normalize_token_count(callback)
+        if token_dict is not None and bool(token_dict):
+            # We actually have a token dictionary to report, so go there.
+            agent_message = AgentMessage(structure=token_dict)
+            await self.journal.write_message(agent_message)
```
d1donlydfink, line 402: Normalize the token count into something that we can accumulate over multiple LLM providers if there is anything at all.  This dictionary gets reported as an AgentMessage back to the client and added to the request_reporting dictionary.

Now, because of the python context-manager-y way this is done (per langchain's docs),
the reporting for the root node will encompass all the tokens counted for all its child nodes.
This is a reasonable start to get something going.   

When we get to the point that where we can properly count tokens for branch nodes, we might want
this to report differently so that the files in the thinking-dir on the client report only what that agent stream is doing.  We'll see when we get there.

neuro_san/internals/run_context/langchain/langchain_run_context.py
```diff
@@ -510,6 +551,8 @@ async def delete_resources(self, parent_run_context: RunContext = None):
         self.chat_history = []
         self.agent = None
         self.recent_human_message = None
+        self.llm = None
+        self.journal = None
```
d1donlydfink, line 565: Clear out some references on delete_resources()

neuro_san/internals/interfaces/invocation_context.py
```diff
@@ -65,3 +66,9 @@ def get_metadata(self) -> Dict[str, str]:
         :return: The metadata to pass along with any request
         """
         raise NotImplementedError
+
+    def get_request_reporting(self) -> Dict[str, Any]:
+        """
+        :return: The request reporting dictionary
+        """
+        raise NotImplementedError
```
d1donlydfink, line 74: Add the notion of a per-request reporting dictionary to accumulate stats.

neuro_san/service/agent_service.py
```diff
@@ -393,6 +393,11 @@ def StreamingChat(self, request: service_messages.ChatRequest,
         # Iterator has finally signaled that there are no more responses to be had.
         # Log that we are done.
         if request_log is not None:
+            request_reporting: Dict[str, Any] = invocation_context.get_request_reporting()
+            reporting: str = None
+            if request_reporting is not None:
+                reporting = json.dumps(request_reporting, indent=4, sort_keys=True)
+            request_log.metrics("Request reporting: %s", reporting)
```
d1donlydfink, line 400: Add a log message for per-request reporting of token accounting.

neuro_san/session/session_invocation_context.py
```diff
@@ -108,3 +109,9 @@ def reset_origination(self):
         Resets the origination
         """
         self.origination = Origination()
+
+    def get_request_reporting(self) -> Dict[str, Any]:
+        """
+        :return: The request reporting dictionary
+        """
+        return self.request_reporting
```
d1donlydfink, line 117: Keep request_reporting dictionary state on the InvocationContext.

neuro_san/client/thinking_file_message_processor.py
dsargent, line 80: I wouldn't bother neatly indenting it here for transmission over the wire. That will just add bloat to the payload. Let the consumer pretty-format it if they so choose before displaying to a human.

d1donlydfink: This also shows up in the logs now

PR #81, opened 2025-02-25
UN-2917 Properly forward metadata to external tools

d1donlydfink:
Fix a crash found by 1C guys.

Verified fixed with OF dev_loop. I can also see the metadata passed along in the external agent's logging.

### DIFF
--- a/neuro_san/session/external_agent_session_factory.py
+++ b/neuro_san/session/external_agent_session_factory.py
@@ -73,7 +73,9 @@ class ExternalAgentSessionFactory(AsyncAgentSessionFactory):
 
         # Note: It's possible we might want some filtering/translation of
         #       metadata keys not unlike what we are doing for sly_data.
-        metadata: Dict[str, str] = invocation_context.get_metadata()
+        metadata: Dict[str, str] = None
+        if invocation_context is not None:
+            metadata = invocation_context.get_metadata()
 
         session: AsyncAgentSession = None
         if self.use_direct and (host is None or len(host) == 0 or host == "localhost"):


neuro_san/internals/run_context/langchain/langchain_run_context.py
```diff
@@ -221,7 +221,7 @@ async def _create_base_tool(self, name: str) -> BaseTool:
             #   network calls.
             session_factory: AsyncAgentSessionFactory = self.invocation_context.get_async_session_factory()
             adapter = ExternalToolAdapter(session_factory, name)
-            function_json = await adapter.get_function_json()
+            function_json = await adapter.get_function_json(self.invocation_context)
```
d1donlydfink, line 224: Pass the invocation context to where it is needed so we can pass metadata along when it is needed.

neuro_san/internals/run_context/utils/external_tool_adapter.py
```diff
@@ -37,15 +38,16 @@ def __init__(self, session_factory: AsyncAgentSessionFactory, agent_url: str):
...
         """
+        :param invocation_context: The context policy container that pertains to the invocation
         :return: The function json for the agent, as specified by the external agent.
         """
         if self.function_json is None:
 
             # Lazily get the information about the service
             session: AgentSession = self.session_factory.create_session(self.agent_url,
-                                                                        invocation_context=None)
+                                                                        invocation_context=invocation_context)
```
d1donlydfink, line 50: More plumbing

neuro_san/session/external_agent_session_factory.py
```diff
@@ -71,6 +71,12 @@ def create_session_from_location_dict(self, agent_location: Dict[str, str],
         agent_name = agent_location.get("agent_name")
         service_prefix = agent_location.get("service_prefix")
 
+        # Note: It's possible we might want some filtering/translation of
+        #       metadata keys not unlike what we are doing for sly_data.
+        metadata: Dict[str, str] = None
+        if invocation_context is not None:
+            metadata = invocation_context.get_metadata()
```
d1donlydfink, line 78: Get metadata from invocation context if there is something to get it from.
(So we don't crash again).  

Add a comment as to future feature that would be inserted at this point.

neuro_san/session/external_agent_session_factory.py
```diff
@@ -83,12 +89,12 @@ def create_session_from_location_dict(self, agent_location: Dict[str, str],
...
             tool_registry: AgentToolRegistry = self.get_tool_registry(agent_name, manifest_tool_registries)
             session = AsyncDirectAgentSession(chat_session_map, tool_registry, invocation_context,
-                                              metadata=invocation_context.get_metadata())
+                                              metadata=metadata)
 
         if session is None:
             session = AsyncGrpcServiceAgentSession(host, port, agent_name=agent_name,
                                                    service_prefix=service_prefix,
-                                                   metadata=invocation_context.get_metadata())
+                                                   metadata=metadata)
```
d1donlydfink, line 97: Pass along any metadata we got from invocation context.

The original crash itself was here, as invocation_context was None at the time.
Now it's not.

PR #86, opened 2025-02-27
UN-2926 Standardized unavailable agent messaging

d1donlydfink:
Requested by 1C guys.

We really have 2 scenarios to cover:

1. Most likely - we go to make first contact for the agent and we get the hand for whatever reason. This happens when we want to call function() to get the agent description.  Log a standardized message on the server and send that via an AGENT message to the client, but continue on without including the tool in the mix.
2. Less likely from race condition - we passed (1) above, but agent went down when we were trying to use streaming_chat().  Return a standardized message as a value from the agent.

Tested: direct, http and grpc in chicken scenario where unrealized /predictor agent was part of the network.

### DIFF
--- a/neuro_san/internals/run_context/langchain/langchain_run_context.py
+++ b/neuro_san/internals/run_context/langchain/langchain_run_context.py
@@ -176,7 +176,7 @@ class LangChainRunContext(RunContext):
         if isinstance(verbose, str) and verbose.lower() in ("extra", "logging"):
             # This particular class adds a *lot* of very detailed messages
             # to the logs.  Add this because some people are interested in it.
-            callbacks.append(LoggingCallbackHandler(getLogger(full_name)))
+            callbacks.append(LoggingCallbackHandler(self.logger))
 
         # Create the model we will use.
         self.llm = LlmFactory.create_llm(self.llm_config, callbacks=callbacks)


neuro_san/internals/graph/tools/external_tool.py
```diff
@@ -89,9 +93,19 @@ async def build(self) -> List[Any]:
...
+            # Proper await-ing for generator results is done in the "async for"-loop below.
+            chat_responses: AsyncGenerator[Dict[str, Any], None] = self.session.streaming_chat(chat_request)
+        except ValueError:
+            # Could not reach the server for the external agent, so tell about it
+            messages_str: str = f"Agent/tool {self.agent_url} was unreachable. " + \
+                                "Cannot rely on results from it as a tool."
+            full_name: str = Origination.get_full_name_from_origin(self.run_context.get_origin())
+            logger: Logger = getLogger(full_name)
+            logger.info(messages_str)
+            return messages_str
```
d1donlydfink, line 108: Less likely, but still possible because of a race condition of agent going down after we got its function definition.
Return a standardized string as external agent results when we cannot reach the agent.

neuro_san/internals/run_context/langchain/langchain_run_context.py
```diff
@@ -130,6 +134,11 @@ def __init__(self, llm_config: Dict[str, Any],
 
         self.update_from_chat_context(self.chat_context)
 
+        # Set up so local logging gives origin info.
+        if self.origin is not None and len(self.origin) > 0:
+            full_name: str = Origination.get_full_name_from_origin(self.origin)
+            self.logger = getLogger(full_name)
```
d1donlydfink, line 140: Set up the logger within the run context to use the name from the origin.

neuro_san/internals/run_context/langchain/langchain_run_context.py
```diff
@@ -167,7 +176,7 @@ async def create_resources(self, assistant_name: str,
         if isinstance(verbose, str) and verbose.lower() in ("extra", "logging"):
             # This particular class adds a *lot* of very detailed messages
             # to the logs.  Add this because some people are interested in it.
-            callbacks.append(LoggingCallbackHandler(logging.getLogger(full_name)))
+            callbacks.append(LoggingCallbackHandler(self.logger))
```
d1donlydfink, line 179: Use the logger we already made in the constructor.

neuro_san/internals/run_context/langchain/langchain_run_context.py
```diff
@@ -221,7 +230,14 @@ async def _create_base_tool(self, name: str) -> BaseTool:
...
             adapter = ExternalToolAdapter(session_factory, name)
-            function_json = await adapter.get_function_json(self.invocation_context)
+            try:
+                function_json = await adapter.get_function_json(self.invocation_context)
+            except ValueError:
+                # Could not reach the server for the external agent, so tell about it
+                message: str = f"Agent/tool {name} was unreachable. Not including it as a tool."
+                agent_message = AgentMessage(content=message)
+                await self.journal.write_message(agent_message)
+                self.logger.info(message)
```
d1donlydfink, line 240: Most likely place for agent access to fail is when we first try to contact the server and we get the hand.
Log on the server side, but also send a message back to the client that it is not included as a tool.

PR #91, opened 2025-03-07
UN-2810 LlmFactory facelift - Phase 1

d1donlydfink:
* Break out llm information to be in its own default_llm_info.hocon file - separate from the code
* Have the data in this file come in 3 sections:
    * llm entries.  These are what had been in code before and the data for these is extended to serve us into the future
    * chat classes.  Less likely to be modified by users. These are common definitions on a per-provider basis. 
    * default_config. What we set as defaults can now go in the hocon for users to override as they see fit.
    * _I'd like to be sure we get this file format as right as possible the first time out. Please scrutinize as to future extensibility_
* Allow the env var "AGENT_LLM_INFO_FILE" to specify a user-defined llm info file that extends what is in the default.
* Extend the Dockerfiles to announce the existence of the new env var.

There is still more to do in the extensibility of LlmFactory.  In particular:
* I'd like users to be able to specify their own LlmFactory class so they can add fringe ChatModel classes to their mix
* There is some refactoring for better performance w/rt just when all the files are read. 
... but there's enough going on in here already.

### DIFF
--- a/neuro_san/internals/run_context/langchain/default_llm_info.hocon
+++ b/neuro_san/internals/run_context/langchain/default_llm_info.hocon
@@ -16,7 +16,6 @@
     # Model context and output tokens: https://platform.openai.com/docs/models
     # Model compatibility: https://platform.openai.com/docs/models#model-endpoint-compatibility
     "o3-mini": {
-        "class": "openai",
         "use_model_name": "o3-mini-2025-01-31",
     },
     "o3-mini-2025-01-31-17": {
@@ -33,7 +32,6 @@
     },
 
     "o1": {
-        "class": "openai",
         "use_model_name": "o1-2024-12-17",
     },
     "o1-2024-12-17": {
@@ -49,7 +47,6 @@
         "knowledge_cutoff": "09/30/2023",
     },
     "o1-preview": {
-        "class": "openai",
         "use_model_name": "o1-preview-2024-09-12",
     },
     "o1-preview-2024-09-12": {
@@ -66,7 +63,6 @@
     },
 
     "gpt-4o": {
-        "class": "openai",
         "use_model_name": "gpt-4o-2024-08-06",
     },
     "gpt-4o-2024-05-13": {
@@ -107,7 +103,6 @@
     },
 
     "gpt-4o-mini": {
-        "class": "openai",
         "use_model_name": "gpt-4o-2024-07-18",
     },
     "gpt-4o-mini-2024-07-18": {
@@ -124,7 +119,6 @@
     },
 
     "gpt-4-turbo": {
-        "class": "openai",
         "use_model_name": "gpt-4-turbo-2024-04-09",
     },
     "gpt-4-turbo-2024-04-09": {
@@ -142,7 +136,6 @@
 
     "gpt-4-turbo-preview": {
         # FYI: LLMs that do not support tools cannot be used as front-men or branch nodes
-        "class": "openai",
         "use_model_name": "gpt-4-0125-preview",
     },
     "gpt-4-0125-preview": {
@@ -161,7 +154,6 @@
 
     "gpt-4": {
         # FYI: LLMs that do not support tools cannot be used as front-men or branch nodes
-        "class": "openai",
         "use_model_name": "gpt-4-0613",
     },
     "gpt-4-0613": {
@@ -193,12 +185,10 @@
 
     "gpt-3.5-turbo": {
         # FYI: LLMs that do not support tools cannot be used as front-men or branch nodes
-        "class": "openai",
         "use_model_name": "gpt-3.5-turbo-0125",
     },
     "gpt-3.5-turbo-16k": {
         # FYI: LLMs that do not support tools cannot be used as front-men or branch nodes
-        "class": "openai",
         "use_model_name": "gpt-3.5-turbo-0125",
     },
     "gpt-3.5-turbo-0125": {
@@ -281,7 +271,6 @@
     # Anthropic models
 
     "claude-3-haiku": {
-        "class": "anthropic",
         "use_model_name": "claude-3-haiku-20240307",
     },
     "claude-3-haiku-20240307": {
@@ -299,7 +288,6 @@
     },
 
     "claude-3-opus": {
-        "use_model_name": "claude-3-opus-20240229",
         "class": "anthropic",
     },
     "claude-3-opus-20240229": {
@@ -317,7 +305,6 @@
     },
 
     "claude-3-5-haiku": {
-        "class": "anthropic",
         "use_model_name": "claude-3-5-haiku-20241022",
     },
     "claude-3-5-haiku-20241022": {
@@ -335,7 +322,6 @@
     },
 
     "claude-3-5-sonnet": {
-        "class": "anthropic",
         "use_model_name": "claude-3-5-sonnet-20241022",
     },
     "claude-3-5-sonnet-20241022": {
@@ -353,7 +339,6 @@
     },
 
     "claude-3-7-sonnet": {
-        "class": "anthropic",
         "use_model_name": "claude-3-7-sonnet-20250219",
     },
     "claude-3-7-sonnet-20250219": {


neuro_san/internals/run_context/langchain/default_llm_info.hocon
```diff
@@ -0,0 +1,708 @@
...
+
+{
+    # OpenAI models
+
+    # Model context and output tokens: https://platform.openai.com/docs/models
+    # Model compatibility: https://platform.openai.com/docs/models#model-endpoint-compatibility
+    "o3-mini": {
+        "class": "openai",
+        "use_model_name": "o3-mini-2025-01-31",
+    },
```
d1donlydfink, line 20: FWIW: o3-mini looks like it has a larger context window and is overall cheaper (~50% of 4o cost).
Has anyone tried it? Might be a good candidate for a new default over gpt-4o.

Also, when speed is of the essence, there is a new gpt-4o-realtime preview in the works.
For 2x cost of gpt-4o, the results are supposed to come back a lot faster.

neuro_san/internals/run_context/langchain/default_llm_info.hocon
```diff
@@ -0,0 +1,708 @@
...
+        "model_info_url": "https://platform.openai.com/docs/models/o3-mini",
+        "modalities": {
+            "input": [ "text" ],
+            "output": [ "text" ],
+        },
+        "capabilities": [ "tools" ],
+        "context_window_size": 200000,
+        "max_output_tokens": 100000,
+        "knowledge_cutoff": "09/30/2023",
+    },
```
d1donlydfink, line 32: This is an example of a new forward-looking llm description.
At the moment we are only really using the following fields:
* class
* max_output_tokens
... but I want to start off the extensibility with people really starting to think about these things
when they input new Llms.  It should begin to set us up for being able to dynamically choosing them
at some point in the future.

The extra information can also allow us to throw better exceptions/errors when, for instance:
* someone is trying to use a non-tools-capable llm for a branch node
* input tokens exceed model capabilities
* someone tries to shove an image down the throat of an llm that doesn't do it.
... in all these cases, we could someday look for these scenarios and say
"Go look at this model url to see for yourself that it doesn't do what you are trying to do."

In the future we can use this information to be more dynamic w/rt how we choose LLMs.
For instance, we could someday just have an agent specify "capabilities": [ "tools" ]  and "optimize": "price" without specifying a name and the system could go and pick the best llm for the job given api keys available and cost info that already exists within langchain libs.

Another example of a future capability might be specifying endpoints to fine-tunes llms by subject matter. Extra capabilities might be "medical", "legal", etc.  

Some llms are already describing themselves in terms of certain features like "reasoning", etc. These could also be in the capabilities section for future dynamic selection.  But for now, I add what we need.

neuro_san/internals/run_context/langchain/default_llm_info.hocon
```diff
@@ -0,0 +1,708 @@
...
+
+    "o1": {
+        "class": "openai",
+        "use_model_name": "o1-2024-12-17",
+    },
```
d1donlydfink, line 36: This is an example of a model alias (effectively)

neuro_san/internals/run_context/langchain/default_llm_info.hocon
```diff
@@ -0,0 +1,708 @@
...
+    "azure-o1": {
+        "class": "azure-openai",
+        "use_model_name": "o1-2024-12-17",
+        "model_info_url": "https://learn.microsoft.com/en-us/azure/ai-services/openai/concepts/models?tabs=global-standard%2Cstandard-chat-completions#o-series-models"
+    },
+    "azure-o3-mini": {
+        "class": "azure-openai",
+        "use_model_name": "o3-mini-2025-01-31",
+        "model_info_url": "https://learn.microsoft.com/en-us/azure/ai-services/openai/concepts/models?tabs=global-standard%2Cstandard-chat-completions#o-series-models"
+    },
```
d1donlydfink, line 268: Added a couple of azure models for the 1C guys.

neuro_san/internals/run_context/langchain/default_llm_info.hocon
```diff
@@ -0,0 +1,708 @@
...
+    "gpt-3.5-turbo-0125": {
+        # FYI: LLMs that do not support tools cannot be used as front-men or branch nodes
+        "class": "openai",
+        "model_info_url": "https://platform.openai.com/docs/models/gpt-3.5-turbo",
+        "modalities": {
+            "input": [ "text" ],
+            "output": [ "text" ],
+        },
+        "capabilities": [],
```
d1donlydfink, line 202: FWIW: OpenAI no longer says gpt-3.5-turbo supports tools.
It did when it first came out last year, but by our own experience the support was poor.
Maybe they don't want to set that expectation any more 🤷

neuro_san/internals/run_context/langchain/default_llm_info.hocon
```diff
@@ -0,0 +1,708 @@
...
+#
+# You can be released from the terms, and requirements of the Academic Public
+# License by purchasing a commercial license.
+# Purchase of a commercial license is mandatory for any use of the
+# neuro-san SDK Software in commercial settings.
+#
+# END COPYRIGHT
+
+{
+    # OpenAI models
```
d1donlydfink, line 14: This first section we describe all the models.  There are different sections for different model providers.
I did a bunch of initial research to fill out all the fields. Kinda tedious.

neuro_san/internals/run_context/langchain/default_llm_info.hocon
```diff
@@ -0,0 +1,708 @@
...
+        "class": "ollama",
+        "model_info_url": "https://ollama.com/library/llama2",
+        "modalities": {
+            "input": [ "text" ],
+            "output": [ "text" ],
+        },
+        "capabilities": [ ],
+        "context_window_size": 4096,    # https://github.com/meta-llama/llama/issues/267
+        "max_output_tokens": 2048,      # ???
+        "knowledge_cutoff": "unknown",
```
d1donlydfink, line 372: FWIW: Ollama models are fairly terrible at documenting their context window size and output token max.
Some of these numbers are from wonky websites who claim some authority.  Better than nothing.
Where the values are still unknown I put ??? in the comments.

neuro_san/internals/run_context/langchain/default_llm_info.hocon
```diff
@@ -0,0 +1,708 @@
...
+        "capabilities": [],
+        "context_window_size": 64000,  # https://www.reddit.com/r/LocalLLaMA/comments/1ialsnx/deepseek_r1_api_real_world_experience/
+        "max_output_tokens": 8192,
+        "knowledge_cutoff": "unknown",
+    },
+
+    # These are data for each LLM class, so they don't have to be repeated for each
+    # LLM info entry above.
+    "classes": {
+        "openai": {
```
d1donlydfink, line 545: This next section is to define aspects in a data-driven fashion for each of the model classes.
Here we have OpenAI models.

neuro_san/internals/run_context/langchain/default_llm_info.hocon
```diff
@@ -0,0 +1,708 @@
...
+            "token_counting": "callback",   # Uses OpenAICallbackHandler
```
d1donlydfink, line 546: I'm looking for some DD indication to say "This is how you do token counting", because that is intimately
related to chat class.

neuro_san/internals/run_context/langchain/default_llm_info.hocon
```diff
@@ -0,0 +1,708 @@
...
+                "seed": null,
+                "logprobs": null,
+                "top_logprobs": null,
+                "logit_bias": null,
+                # streaming is always on. Without it token counting will not work.
+                # n is always 1.  neuro-san will only ever consider a single chat completion.
+                "top_p": null,
+                "max_tokens": null,         # This is always for output
+                "tiktoken_model_name": null,
+                "stop": null,
```
d1donlydfink, line 568: Allow people to go stone nuts with parameters they can set as default.
We set this dictionary up as a list of defaults.  Conceivably people can override all this stuff now.
1C guys went fishing in the source code to get at what they wanted, so I figure may as well
open everything up so that doesn't happen again.

neuro_san/internals/run_context/langchain/default_llm_info.hocon
```diff
@@ -0,0 +1,708 @@
...
+
+                # If you really need more parameters, you will likely have to create your own LlmFactory.
+            }
+        },
+        "azure-openai": {
+            "token_counting": "callback",   # Uses OpenAICallbackHandler
+            "extends": "openai",
```
d1donlydfink, line 575: All other chat classes are similar except azure-openai ends up being an extension of openai.
Let that happen more naturally by calling that out.

neuro_san/internals/run_context/langchain/default_llm_info.hocon
```diff
@@ -0,0 +1,708 @@
...
+                "max_retires": 2,
+                "stop_sequences": null,
+                "anthropic_api_url": "https://api.anthropic.com",
+                "anthropic_api_key": null,
+                # streaming is always set to True so that ...
+                # stream_usage is always set to True so that token counting can work correctly
+            }
+        },
+        "ollama": {
+            "token_counting": null,
```
d1donlydfink, line 612: Not much love for token counting on Ollama models. 🤷

neuro_san/internals/run_context/langchain/default_llm_info.hocon
```diff
@@ -0,0 +1,708 @@
...
+                "stop": null,
+                "tfs_z": null,
+                "top_k": null,
+                "top_p": null,
+                "keep_alive": null,
+                "base_url": null,
+            }
+        },
+        "nvidia": {
+            "token_counting": null,     # For now. Seems there is hope.
```
d1donlydfink, line 636: I think there might be some love for token counting on nvidia models, but not going there just yet.
More important to get this extendable stuff out there.

neuro_san/internals/run_context/langchain/default_llm_info.hocon
```diff
@@ -0,0 +1,704 @@
...
+    },
+    "deepseek-r1:14b": {
+        # FYI: LLMs that do not support tools cannot be used as front-men or branch nodes
+        "class": "ollama",
+        "model_info_url": "https://ollama.com/library/deepseek-r1",
+        "modalities": {
+            "input": [ "text" ],
+            "output": [ "text" ],
+        },
+        "capabilities": [],
```
d1donlydfink, line 481: Interesting to note that with all the hype around deepseek, we can't use it much cuz no tools support.
Maybe someday.

neuro_san/deploy/Dockerfile
```diff
@@ -122,6 +122,10 @@ ENV APP_ENTRYPOINT ${PACKAGE_DEPLOY}/entrypoint.sh
 # files to serve up from this server instance.
 ENV AGENT_MANIFEST_FILE=${APP_SOURCE}/registries/manifest.hocon
 
+# An llm_info hocon file with user-provided llm descriptions that are to be used
+# in addition to the neuro-san defaults.
+ENV AGENT_LLM_INFO_FILE=""
```
d1donlydfink, line 127: Allow for deployments to supply a hocon file that will augment what is in the default_llm_info.hocon file.

neuro_san/internals/run_context/langchain/langchain_run_context.py
```diff
@@ -179,7 +179,9 @@ async def create_resources(self, assistant_name: str,
             callbacks.append(LoggingCallbackHandler(self.logger))
 
         # Create the model we will use.
-        self.llm = LlmFactory.create_llm(self.llm_config, callbacks=callbacks)
+        llm_factory = LlmFactory()
+        llm_factory.load()
+        self.llm = llm_factory.create_llm(self.llm_config, callbacks=callbacks)
```
d1donlydfink, line 184: Specifically load the LlmFactory now. There might be more than 1 file to load.
In the near future we will likely more this around so it only happens once per-server.
For now let's get it working and out there so people can extend.

neuro_san/internals/run_context/langchain/llm_factory.py
```diff
@@ -24,251 +24,9 @@
...
-DEFAULT_PROMPT_TOKEN_FRACTION = 0.5
-
-LLM_ENTRIES = {
-    # Model context and output tokens: https://platform.openai.com/docs/models
-    # Model compatibility: https://platform.openai.com/docs/models#model-endpoint-compatibility
-    "gpt-3.5-turbo": {
-        "class": ChatOpenAI,
-        "api_key": "openai_api_key",
-        "max_tokens": 4096  # From https://platform.openai.com/docs/models/gpt-3-5
-    },
```
d1donlydfink, line 48: All this static data moves to the hocon file.

neuro_san/internals/run_context/langchain/llm_factory.py
```diff
@@ -24,251 +24,9 @@
 from langchain_openai.chat_models.base import ChatOpenAI
 from langchain_openai.chat_models.azure import AzureChatOpenAI
 
-from tiktoken import Encoding
-from tiktoken import encoding_for_model
-from tiktoken import get_encoding
-
 from leaf_common.config.dictionary_overlay import DictionaryOverlay
 
-DEFAULT_MODEL = "gpt-3.5-turbo"
```
d1donlydfink, line 33: Worth noting that in the default hocon file our default model is now gpt-4o, as OpenAI
now says gpt-3.5-turbo does not support tools.

neuro_san/internals/run_context/langchain/llm_factory.py
```diff
@@ -24,251 +24,9 @@
...
-    "nvidia-deepseek-r1": {
-        "use_model_name": "deepseek-ai/deepseek-r1",
-        "class": ChatNVIDIA,
-        "api_key": "nvidia_api_key",
-        "max_tokens": 4096  # From https://python.langchain.com/docs/integrations/chat/nvidia_ai_endpoints/
-    },
-}
-
-
-DEFAULT_CONFIG = {
```
d1donlydfink, line 215: Notion of defaults also moves to the hocon file.

neuro_san/internals/run_context/langchain/llm_factory.py
```diff
@@ -300,19 +54,29 @@ class LlmFactory:
...
+        restorer = LlmInfoRestorer()
+        self.llm_infos = restorer.restore()
 
-    @staticmethod
-    def create_llm(config: Dict[str, Any], callbacks: List[BaseCallbackHandler] = None) -> BaseLanguageModel:
+        # Mix in user-specified llm info, if available.
+        llm_info_file: str = os.getenv("AGENT_LLM_INFO_FILE")
+        if llm_info_file is not None and len(llm_info_file) > 0:
+            extra_llm_infos: Dict[str, Any] = restorer.restore(file_reference=llm_info_file)
+            self.llm_infos = self.overlayer.overlay(self.llm_infos, extra_llm_infos)
```
d1donlydfink, line 77: Not the greatest diff. The new load() method reads in the existing default hocon and allows for user-defined data to be overlayed on top of it.

neuro_san/internals/run_context/langchain/llm_factory.py
```diff
@@ -324,160 +88,260 @@ def create_llm(config: Dict[str, Any], callbacks: List[BaseCallbackHandler] = No
                 Can raise a ValueError if the config's model_name value is
                 unknown to this method.
         """
-        overlayer = DictionaryOverlay()
-        use_config = overlayer.overlay(DEFAULT_CONFIG, config)
+        full_config: Dict[str, Any] = self.create_full_llm_config(config)
+        llm: BaseLanguageModel = self.create_base_chat_model(full_config, callbacks)
+        return llm
```
d1donlydfink, line 93: Break up create_llm() into 2 simple steps:
1. create the "full" config from the default config, default args from classes, and user config.
2. Pass that into something that is more strictly a factory taking that amalgamated config as read-only.

Note: For the next PR we will also allow people a chance to override this factory aspect so they can
instantiate fringe Chat classes as they like.  One step at a time though.

neuro_san/internals/run_context/langchain/llm_factory.py
```diff
@@ -324,160 +88,260 @@ def create_llm(config: Dict[str, Any], callbacks: List[BaseCallbackHandler] = No
...
 
-        model_name = use_config.get("model_name", DEFAULT_MODEL)
+    def create_full_llm_config(self, config: Dict[str, Any]) -> Dict[str, Any]:
+        """
+        :param config: The llm_config from the user
+        :return: The fully specified config with defaults filled in.
+        """
```
d1donlydfink, line 99: Policy method for creating the "full" config.

neuro_san/internals/run_context/langchain/llm_factory.py
```diff
@@ -324,160 +88,260 @@ def create_llm(config: Dict[str, Any], callbacks: List[BaseCallbackHandler] = No
...
         if llm_entry is None:
             raise ValueError(f"No llm entry for model_name {model_name}")
 
         # Get some bits from the llm_entry
-        base_class = llm_entry.get("class")
-        api_key = llm_entry.get("api_key")
         use_model_name = llm_entry.get("use_model_name", model_name)
+        if len(llm_entry.keys()) <= 2 and use_model_name is not None:
+            # We effectively have an alias. Switch out the llm entry.
+            llm_entry = self.llm_infos.get(use_model_name)
```
d1donlydfink, line 113: This little bit allows for 1 level of indirection for the "use_model_name" aliasing.

neuro_san/internals/run_context/langchain/llm_factory.py
```diff
@@ -324,160 +88,260 @@ def create_llm(config: Dict[str, Any], callbacks: List[BaseCallbackHandler] = No
...
+
+        return full_config
+
+    def get_chat_class_args(self, chat_class_name: str, use_model_name: str) -> Dict[str, Any]:
+        """
+        :param chat_class_name: string name of the chat class to look up.
+        :param use_model_name: the original model name that prompted the chat class lookups
+        :return: A dictionary of default arguments for the chat class.
+                Can throw an exception if the chat class does not exist.
+        """
```
d1donlydfink, line 142: Method to get the default args for the chat class. Will recurse if "extends" is present (like for azure)

neuro_san/internals/run_context/langchain/llm_factory.py
```diff
@@ -324,160 +88,260 @@ def create_llm(config: Dict[str, Any], callbacks: List[BaseCallbackHandler] = No
...
+                               callbacks: List[BaseCallbackHandler] = None) -> BaseLanguageModel:
+        """
+        Create a BaseLanguageModel from the fully-specified llm config.
+        :param config: The fully specified llm config which is a product of
+                    _create_full_llm_config() above.
+        :param callbacks: A list of BaseCallbackHandlers to add to the chat model.
+        :return: A BaseLanguageModel (can be Chat or LLM)
+                Can raise a ValueError if the config's class or model_name value is
+                unknown to this method.
+        """
```
d1donlydfink, line 172: Now this method becomes more of a straight-ahead factory.
When we get to extensible factories, this will be the entry point.

neuro_san/internals/run_context/langchain/llm_factory.py
```diff
@@ -324,160 +88,260 @@ def create_llm(config: Dict[str, Any], callbacks: List[BaseCallbackHandler] = No
...
+                            seed=config.get("seed"),
+                            logprobs=config.get("logprobs"),
+                            top_logprobs=config.get("top_logprobs"),
+                            logit_bias=config.get("logit_bias"),
+                            streaming=True,     # streaming is always on. Without it token counting will not work.
+                            n=1,                # n is always 1.  neuro-san will only ever consider one chat completion.
+                            top_p=config.get("top_p"),
+                            max_tokens=config.get("max_tokens"),    # This is always for output
+                            tiktoken_model_name=config.get("tiktoken_model_name"),
+                            stop=config.get("stop"),
```
d1donlydfink, line 206: Worth noting that all these Chat constructors now have extensive support for the wonkiest parameters
in config, as long as those params can be "just data".

neuro_san/internals/run_context/langchain/llm_factory.py
```diff
@@ -324,160 +88,260 @@ def create_llm(config: Dict[str, Any], callbacks: List[BaseCallbackHandler] = No
...
-        """
-        Creates a tiktoken tokenizer based on the 'model_name' value of
-        the config passed in.
-
-        :param config: A dictionary which describes which LLM to use.
-                See the class comment for details.
-        :return: A Encoding
-                Can raise a ValueError if the config's model_name value is
-                unknown to this method.
-        """
```
d1donlydfink, line 421: This was never used within neuro-san

neuro_san/internals/run_context/langchain/llm_factory.py
```diff
@@ -324,160 +88,260 @@ def create_llm(config: Dict[str, Any], callbacks: List[BaseCallbackHandler] = No
...
         if llm_entry is None:
             raise ValueError(f"No llm entry for model_name {model_name}")
 
-        entry_max_tokens = llm_entry.get("max_tokens")
-        prompt_token_fraction = use_config.get("prompt_token_fraction",
-                                               DEFAULT_PROMPT_TOKEN_FRACTION)
+        use_model_name = llm_entry.get("use_model_name", model_name)
+        if len(llm_entry.keys()) <= 2 and use_model_name is not None:
+            # We effectively have an alias. Switch out the llm entry.
+            llm_entry = self.llm_infos.get(use_model_name)
```
d1donlydfink, line 331: Allow for  1 level of aliasing.

neuro_san/internals/run_context/langchain/llm_factory.py
```diff
@@ -324,160 +88,260 @@ def create_llm(config: Dict[str, Any], callbacks: List[BaseCallbackHandler] = No
...
+
+        entry_max_tokens = llm_entry.get("max_output_tokens")
+        prompt_token_fraction = config.get("prompt_token_fraction")
```
d1donlydfink, line 334: Specifically look at the llm entry's max_output_tokens.
This is much less confusing because what we had before was not clear about the number
being for input or output tokens.

Also worth noting that the default fraction here is 1.0, where it had been 0.5 before.
0.5 was from a time where input and output were conflated in all the docs, and so splitting the
difference worked with most models.  Now that most models are clear about the difference between
input and output max, we can jack the fraction here to 1.0 fairly safely.

neuro_san/internals/run_context/langchain/llm_info_restorer.py
```diff
@@ -0,0 +1,66 @@
...
+
+from neuro_san.internals.utils.file_of_class import FileOfClass
+
+
+# pylint: disable=too-few-public-methods
+class LlmInfoRestorer(Restorer):
+    """
+    Implementation of the Restorer interface to read in an LlmInfo dictionary
+    instance given a hocon file name.
+    """
```
d1donlydfink, line 32: Restores a hocon file describing the llm info.

neuro_san/internals/run_context/langchain/default_llm_info.hocon
ofrancon: Why do we need `"class"` then? That opens the door to having 2 different classes being specified. An alias should me nothing more than a redirection.

neuro_san/internals/run_context/langchain/default_llm_info.hocon
```diff
@@ -0,0 +1,704 @@
...
+        "context_window_size": 4096,
+        "max_output_tokens": 4096,
+        "knowledge_cutoff": "08/31/2021",
+    },
+
+    # Azure models
+
+    "azure-gpt-3.5-turbo": {
```
ofrancon, line 236: Not sure we should list any "Azure models" here. In Azure models names are **deployment** model names, i.e. whatever name the person that deploys the model chooses. They are most probably custom, as well as the end point url and other params, so they don't belong to this "default" hocon file.

neuro_san/internals/run_context/langchain/default_llm_info.hocon
ofrancon: Not sure that will really help them: it just means they'll have to use one of these names when they deploy their model on Azure. Calling it "neuro-san-azure-model" would be as useful.

neuro_san/internals/run_context/langchain/default_llm_info.hocon
d1donlydfink: It's possible we might not need it. Let me see.

neuro_san/internals/run_context/langchain/default_llm_info.hocon
d1donlydfink: Not so. Something needs to tell the innards of the system to use the ChatAzureOpenAI class.
Some are privately deployed models with a deployment name (which is now specifyable under this regime)
but others are not with fixed names as are found on the azure web site quoted.. In either case people can
now set up their models (private or otherwise) and call them "george" if they want.

neuro_san/internals/run_context/langchain/default_llm_info.hocon
d1donlydfink: That is up to them.  Here they have the opportunity to specify in their llm config:
"model_name": "azure-o3-mini"
"deployment_name": "george" ,
+ other azure parameters that openai on its own won't take
if they want.

neuro_san/internals/run_context/langchain/default_llm_info.hocon
d1donlydfink: Removed and all is still well. Thanks!

PR #92, opened 2025-03-07
UN-2928 async HTTP side-car for neuro-san service.

andreidenissov-cog:
This PR switches to async implementation for http service side-car for neuro-san.
We now talk to async gRPC service session.
The idea is to address overall service scalability by switching to an event loop instead of multiple request threads. 
Tested by running neuro-san server with "hello_world" and "six_thinking_hats" clients.


neuro_san/http_sidecar/handlers/base_request_handler.py
```diff
@@ -56,18 +56,17 @@ def get_metadata(self) -> Dict[str, Any]:
...
-                                   hostname="localhost",
-                                   port=self.port,
-                                   metadata=metadata)
+        grpc_session: AsyncAgentSession = \
+            AsyncGrpcServiceAgentSession(
+                host="localhost",
+                port=self.port,
+                metadata=metadata,
+                agent_name=self.agent_name)
         return grpc_session
```
andreidenissov-cog, line 70: Straightforward switch from "sync" to "async".
Also, we don't use session factory anymore - we know exactly what we need from gRPC host.

neuro_san/http_sidecar/handlers/connectivity_handler.py
```diff
@@ -17,24 +17,24 @@
...
         try:
             data: Dict[str, Any] = {}
             metadata: Dict[str, Any] = self.get_metadata()
-            grpc_session: AgentSession = self.get_grpc_session(metadata)
-            result_dict: Dict[str, Any] = grpc_session.connectivity(data)
+            grpc_session: AsyncAgentSession = self.get_grpc_session(metadata)
+            result_dict: Dict[str, Any] = await grpc_session.connectivity(data)
 
             # Return gRPC response to the HTTP client
             self.set_header("Content-Type", "application/json")
```
andreidenissov-cog, line 40: Here and below for "function" handler - switch to asynchronous API.

neuro_san/http_sidecar/handlers/health_check_handler.py
```diff
@@ -38,7 +38,7 @@ def get(self):
             self.set_status(500)
             self.write({"error": "Internal server error"})
         finally:
-            self.flush()
+            await self.flush()
```
andreidenissov-cog, line 42: Same here: switch to async API.

neuro_san/http_sidecar/handlers/streaming_chat_handler.py
```diff
@@ -41,22 +41,27 @@ def post(self):
...
             grpc_request = Parse(json.dumps(data), ChatRequest())
             metadata: Dict[str, Any] = self.get_metadata()
-            grpc_session: AgentSession = self.get_grpc_session(metadata)
+            grpc_session: AsyncAgentSession = self.get_grpc_session(metadata)
 
-            result_generator: Generator[Dict[str, Any], None, None] =\
+            # Mind the type hint:
+            # here we are getting Generator of Generators of ChatResponses!
+            result_generator: Generator[Generator[ChatResponse, None, None], None, None] =\
                 grpc_session.streaming_chat(grpc_request)
```
andreidenissov-cog, line 49: Added a type hint to clarify this non-obvious (for me at least) fact that what we are getting back
is async generator of async generators of ChatResponse objects. Go figure.

neuro_san/http_sidecar/handlers/streaming_chat_handler.py
```diff
@@ -41,22 +41,27 @@ def post(self):
...
-                self.write(result_str)
-                self.flush()
+            # async for result_message in result_generator:
+            # result_generator = self.async_generator()
+            async for sub_generator in result_generator:
+                async for result_message in sub_generator:
+                    result_dict: Dict[str, Any] = MessageToDict(result_message)
+                    result_str: str = json.dumps(result_dict) + "\n"
+                    self.write(result_str)
+                    await self.flush()
```
andreidenissov-cog, line 65: Nested loop seems to just work here.

neuro_san/http_sidecar/http_server_application.py
```diff
@@ -26,12 +26,3 @@ class HttpServerApplication(Application):
...
         super().__init__(*args, **kwargs)
-
-        # Initialize and store dependencies:
-        self.session_factory: AgentSessionFactory = AgentSessionFactory()
-
-    def get_session_factory(self):
-        """
-        Get factory for agent sessions.
-        """
-        return self.session_factory
```
andreidenissov-cog, line 37: No factory needed - we go directly for async gRPC session for "external" clients,
since we are running in separate process.

neuro_san/http_sidecar/http_sidecar.py
```diff
@@ -87,4 +87,4 @@ def make_app(self):
             handlers.append((route, StreamingChatHandler, request_data))
             self.logger.info("Registering URL path: %s", route)
```
andreidenissov-cog, line 89: Can be simplified - we don't need our own Application sub-class anymore.

neuro_san/http_sidecar/handlers/connectivity_handler.py
```diff
@@ -17,24 +17,24 @@
...
+from neuro_san.interfaces.async_agent_session import AsyncAgentSession
 
 
 class ConnectivityHandler(BaseRequestHandler):
     """
     Handler class for neuro-san "connectivity" API call.
     """
 
-    def get(self):
+    async def get(self):
```
d1donlydfink, line 28: There is something off here.

3 possibilities:

1. If get() is called by the BaseRequestHandler superclass, I would expect to see an await in that class when doing a get(). This would indicate that the interface was completely within our control and we are handling asynchronous stuff properly. I don't see any await on get() in this PR

2. get() is called as part of the RequestHandler interface we do not control.  If that were the case I'd expect to see a different name for the method like aget() or async_get() so as not to clash with the synchronous version.  It's my experience that you can't intermingle asynchronous and sync method signatures.

3. The library that RequestHandler comes from has some vastly superior async-fu that we could learn from w/rt CodedTools and now MessageProcessors.

 I would also expect to see a log message during operation as to "asynchronous function get() not awaited" from at least one of the first 2 scenarios.

neuro_san/http_sidecar/handlers/streaming_chat_handler.py
```diff
@@ -20,18 +20,18 @@
...
+from neuro_san.interfaces.async_agent_session import AsyncAgentSession
 
 
 class StreamingChatHandler(BaseRequestHandler):
     """
     Handler class for neuro-san streaming chat API call.
     """
 
-    def post(self):
+    async def post(self):
```
d1donlydfink, line 34: Same deal for post() as above w/rt async

neuro_san/http_sidecar/http_sidecar.py
d1donlydfink: Is this the magic that makes formerly sync API like get() and post() async?

neuro_san/http_sidecar/handlers/connectivity_handler.py
andreidenissov-cog: Good questions all :-)

So:
- This code works, as tested by myself with couple of agent cli clients. So something is right;
-   >>> I don't see any await on get() in this PR
    because you shouldn't. There is an "await" when calling corresponding methods from AsyncGrpcClientSession    inside our handlers. But actual call into our handlers get/post are done somewhere deep inside the bowels of Tornado machinery, which I'm not super-eager to dig into right now. 

- Python3 language being cool as it is, it does allow you to define say "async X()" function in sub-class B, if base class A has it as "regular X()". And vice versa. The trick here is to be able to call function X() both ways, sync and async, depending on run-time function kind. Which kind can be queried dynamically. I suspect that's what Tornado does internally, because both versions of http sidecar work.

d1donlydfink: Hmm. So that's case #3.  We could learn from that.

PR #93, opened 2025-03-11
Fix logging initialization when AGENT_FORWARDED_REQUEST_METADATA is set

d1donlydfink:
1C guys found a problem and this should fix it as well as take care of the case where AGENT_FORWARDED_REQUEST_METADATA is not set at all.

### DIFF
--- a/neuro_san/deploy/Dockerfile
+++ b/neuro_san/deploy/Dockerfile
@@ -150,6 +150,7 @@ ENV AGENT_SERVER_NAME_FOR_LOGS="Agent Server"
 # A space-delimited list of http metadata request keys to forward to logs/other requests
 # You can see how these are used in the AGENT_SERVICE_LOG_JSON file (see above) and
 # customize this and the AGENT_SERVER_LOG_JSON file to your needs.
+# Note that any metadata key needs to be all lowercase.
 ENV AGENT_FORWARDED_REQUEST_METADATA="request_id user_id"
 
 # Port number for the grpc service endpoint
--- a/neuro_san/deploy/internal/Dockerfile
+++ b/neuro_san/deploy/internal/Dockerfile
@@ -172,6 +172,7 @@ ENV AGENT_SERVER_NAME="neuro-san.Agent"
 ENV AGENT_SERVER_NAME_FOR_LOGS="Agent Server"
 
 # A space-delimited list of http metadata request keys to forward to logs/other requests
+# Note that any metadata key needs to be all lowercase.
 ENV AGENT_FORWARDED_REQUEST_METADATA="request_id user_id"
 
 # Port number for the grpc service endpoint


neuro_san/service/agent_server.py
```diff
@@ -85,9 +85,22 @@ def __init__(self, port: int,
...
+        # call to a logger (which is below!).
+        self.forwarded_request_metadata: List[str] = forwarded_request_metadata.split(" ")
+        extra_logging_defaults: Dict[str, str] = {
+            "source": server_name_for_logs,
+            "user_id": "None",
+            "request_id": "None",
+        }
+        if len(self.forwarded_request_metadata) > 0:
+            for key in self.forwarded_request_metadata:
+                extra_logging_defaults[key] = "None"
```
d1donlydfink, line 98: Initialize a dictionary of default values for logging when AGENT_FORWARDED_REQUEST_METADATA is set.

neuro_san/service/agent_server.py
```diff
@@ -99,7 +112,6 @@ def __init__(self, port: int,
         self.server_name_for_logs: str = server_name_for_logs
         self.max_concurrent_requests: int = max_concurrent_requests
         self.request_limit: int = request_limit
-        self.forwarded_request_metadata: List[str] = forwarded_request_metadata.split(" ")
```
d1donlydfink, line 102: This moves up higher to take care of the logging initialization sooner

neuro_san/service/agent_server.py
```diff
@@ -85,9 +85,22 @@ def __init__(self, port: int,
...
+        }
+        if len(self.forwarded_request_metadata) > 0:
+            for key in self.forwarded_request_metadata:
+                extra_logging_defaults[key] = "None"
+
         setup_logging(server_name_for_logs, current_dir,
                       'AGENT_SERVICE_LOG_JSON',
-                      'AGENT_SERVICE_LOG_LEVEL')
+                      'AGENT_SERVICE_LOG_LEVEL',
+                      extra_logging_defaults)
```
d1donlydfink, line 103: Pass these initial values in when we are setting up logging

neuro_san/deploy/Dockerfile
```diff
@@ -150,6 +150,7 @@ ENV AGENT_SERVER_NAME_FOR_LOGS="Agent Server"
 # A space-delimited list of http metadata request keys to forward to logs/other requests
 # You can see how these are used in the AGENT_SERVICE_LOG_JSON file (see above) and
 # customize this and the AGENT_SERVER_LOG_JSON file to your needs.
+# Note that any metadata key needs to be all lowercase.
```
d1donlydfink, line 153: Add a note in the Docerfile that any metadata key must be all lowercase.

PR #95, opened 2025-03-11
UN-2986 Better error reporting from HTTP side-car service.

andreidenissov-cog:
This PR provides a mapping between possible gRPC-level errors when talking to neuro-san gRPC session
and HTTP request codes we return to a client.
More detailed error message is also included instead of generic "Internal Server Error".

### DIFF
--- a/neuro_san/http_sidecar/handlers/function_handler.py
+++ b/neuro_san/http_sidecar/handlers/function_handler.py
@@ -17,6 +17,7 @@ from typing import Any, Dict
 from neuro_san.http_sidecar.handlers.base_request_handler import BaseRequestHandler
 from neuro_san.interfaces.async_agent_session import AsyncAgentSession
 
+
 class FunctionHandler(BaseRequestHandler):
     """
     Handler class for neuro-san "function" API call.


neuro_san/http_sidecar/handlers/base_request_handler.py
```diff
@@ -28,6 +32,17 @@ class BaseRequestHandler(RequestHandler):
...
+        grpc.StatusCode.INVALID_ARGUMENT: 400,
+        grpc.StatusCode.UNAUTHENTICATED: 401,
+        grpc.StatusCode.PERMISSION_DENIED: 403,
+        grpc.StatusCode.NOT_FOUND: 404,
+        grpc.StatusCode.ALREADY_EXISTS: 409,
+        grpc.StatusCode.INTERNAL: 500,
+        grpc.StatusCode.UNAVAILABLE: 503,
+        grpc.StatusCode.DEADLINE_EXCEEDED: 504
+    }
+
```
andreidenissov-cog, line 45: Put in a (partial) map of grpc error codes to HTTP status codes.

neuro_san/http_sidecar/handlers/base_request_handler.py
```diff
@@ -69,6 +84,46 @@ def get_grpc_session(self, metadata: Dict[str, Any]) -> AsyncAgentSession:
...
+        """
+        code = exc.code()
+        http_code = BaseRequestHandler.grpc_to_http.get(code, 500)
+        return http_code, code.name, exc.details()
+
+    def process_exception(self, exc: Exception):
+        """
+        Process exception raised during request handling
+        """
+        if exc is None:
```
andreidenissov-cog, line 104: General logic for request exception processing, shared by all our request handlers.
Depending on specific exception type, set HTTP response status,
write back some error details information and log them locally also.

neuro_san/http_sidecar/handlers/connectivity_handler.py
```diff
@@ -40,14 +38,7 @@ async def get(self):
             self.set_header("Content-Type", "application/json")
```
andreidenissov-cog, line 38: Here and below:
replace big copy-pasted code segment with smaller one,
which delegates request exception handling to common base class method.

neuro_san/http_sidecar/handlers/base_request_handler.py
dsargent: This must exist somewhere already? Can't believe we have to reinvent something like this.

neuro_san/http_sidecar/handlers/base_request_handler.py
```diff
@@ -69,6 +84,46 @@ def get_grpc_session(self, metadata: Dict[str, Any]) -> AsyncAgentSession:
...
+            return
+        if isinstance(exc, json.JSONDecodeError):
+            # Handle invalid JSON input
+            self.set_status(400)
```
dsargent, line 108: If we must have the hand-coded map I mentioned above, then at least let's use it rather than cryptic codes?

```python
set.set_status(grpc_to_http[grpc.StatusCode.INVALID_ARGUMENT])
```

neuro_san/http_sidecar/handlers/base_request_handler.py
d1donlydfink: Kinda agree. Maybe the value for the enum is actually the integer? That'd be most logical thing.
Optimization tho.

PR #97, opened 2025-03-12
UN-2874 "Concierge" service definition.

andreidenissov-cog:
This is a draft for what we potentially want from neuro-san Concierge service.
Open to any kind of extensions/modifications/suggestions.

### DIFF
--- a/neuro_san/api/grpc/concierge.proto
+++ b/neuro_san/api/grpc/concierge.proto
@@ -1,0 +1,49 @@
+
+// Copyright (C) 2023-2025 Cognizant Digital Business, Evolutionary AI.
+// All Rights Reserved.
+// Issued under the Academic Public License.
+//
+// You can be released from the terms, and requirements of the Academic Public
+// License by purchasing a commercial license.
+// Purchase of a commercial license is mandatory for any use of the
+// neuro-san SDK Software in commercial settings.
+//
+// END COPYRIGHT
+
+// To obtain information concerning the code generation that is used by this Interface Definition Language
+// please see the following:
+//
+// Go - https://developers.google.com/protocol-buffers/docs/reference/go-generated
+// Python - https://developers.google.com/protocol-buffers/docs/reference/python-generated
+//
+syntax = "proto3";
+
+package dev.cognizant_ai.neuro_san.api.grpc.concierge;
+
+// The service
+service ConciergeService {
+
+    // Called when a client needs the information about available agents.
+    rpc Describe (ConciergeRequest) returns (ConciergeResponse) {}
+}
+
+// Request structure for Describe gRPC method
+message ConciergeRequest {
+}
+
+// Description of an agent's function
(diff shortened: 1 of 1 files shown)


neuro_san/api/grpc/concierge.proto
```diff
@@ -0,0 +1,49 @@
...
+message ConciergeRequest {
+}
+
+// Description of an agent's function
+message AgentInfo {
+    // Agent name
+    string agent_name =  1 [json_name="agent_name"];
+
+    // Outward-facing description of what the agent does.
+    string description = 2 [json_name="description"];
```
ofrancon, line 40: If the agent is a tool, should we also have a dictionary of parameters and their description?

neuro_san/api/grpc/concierge.proto
```diff
@@ -0,0 +1,49 @@
...
+//
+syntax = "proto3";
+
+package dev.cognizant_ai.neuro_san.api.grpc.concierge;
+
+// The service
+service ConciergeService {
+
+    // Called when a client needs the information about available agents.
+    rpc Describe (ConciergeRequest) returns (ConciergeResponse) {}
```
dsargent, line 27: I like the clean, simple API.

Perhaps `List` would be a better endpoint name? I would think of `Describe` as giving me _all_ the info about a particular agent, whereas `List` would give a, well, _list_ of agents.

neuro_san/api/grpc/concierge.proto
```diff
@@ -0,0 +1,49 @@
...
+}
+
+// Request structure for Describe gRPC method
+message ConciergeRequest {
```
ofrancon, line 31: What would a ConciergeRequest contain? A string like "who can provide this service?"

d1donlydfink: No. It's more a request to query "What agents does this server dish out?"
The idea is to probe for "what agents exist?"

neuro_san/api/grpc/concierge.proto
d1donlydfink: After existence of the agent is known, then later calls to that agent's Function() method can get those kinds of  details.

dsargent: This is another case where REST would be a lot more intuitive IMO.

`.../Agents` - get a list of agents
`.../Agents/123/Tools` - get tools for agent 123
`.../Agents/123/Tools/456/` - get details on Tool #456

Rather than having separate APIs to learn about. REST is _guessable_ and _discoverable_.

REST even has the (terribly named) [HATEOAS](https://en.wikipedia.org/wiki/HATEOAS?useskin=vector) extension which guides developers in a structured, formalized way how to use the API. It generates (in a well-known format) responses like, "you got the info on an agent, if you want to get the list of tools, use this URL: <url>".

Great for reducing support calls.

PR #101, opened 2025-03-14
UN-2940 Filter chat stream - protocol change 

d1donlydfink:
Looks big but it's not too bad.
Reason this looks big:
* add too-few-public-methods to the pylint disable list and clean up the uses in this codebase
* Remove polling api stuff from agent_cli client and DataDrivenChat.

But the meat is:
* Add a chat_filter parameter to the ChatRequest that allows clients whether or not they want minimal or maximal messages from the server.  
    * Note: **MINIMAL is the default for the server.** because the default end-user experience really only cares about what the answer is from the agent and the ability to continue the conversation.    
    * The MAXIMAL chat_filter is useful for developers, which interestingly enough is all the clients we want to show with fancy bells and blinking lights.
    *  @ofrancon @deepsaia @swensel @dsargent This means that upon taking the next release, a little bit of extra code will be required to continue the light shows.  The agent_cli.py is modified in this way to get you started.
    * At some point we might want to make this more complex than these 2, but not going there for now.
* Add the notion of a MessageFilter interface used on the service side which has a few implementations to support this:
    * A MaximalMessageFilter lets everything through
    * A MinimalMessageFilter lets the bare minimum through with the following components to help it:
        * An AnswerMessageFilter to allow sending the AI's final answer to the client
        * A ChatContextMessageFilter to allow sending the info that allows the user to continue the conversation
        * A CompoundMessageFilter which is an abstract guy that can group several MessageFilters together
    * A MessageFilterFactory to allow creating the right version given the request.
* agent_cli.py gets modified to support:
    * MAXIMAL chat_filter by default
    * --minimal and --maximal arguments for specifying this on the command line

Also worth noting:    
* The user_input field is removed from the ChatRequest.  Use a ChatMessage whose text field is set to the user input instead.  This is more versatile going forward.  @deepsaia @ofrancon @dsargent @swensel your clients might be effected.

I am hoping this can be among the last PRs before a version 0.5.0 gets released. Trying to get all the protocol changes in before a larger fan-out happens and there are too many clients to simply eliminate things from the protocol.

### DIFF
--- a/neuro_san/internals/filters/message_filter_factory.py
+++ b/neuro_san/internals/filters/message_filter_factory.py
@@ -19,8 +19,10 @@ from neuro_san.internals.filters.minimal_message_filter import MinimalMessageFil
 
 TYPE_TO_MESSAGE_FILTER_CLASS: Dict[Any, Type[MessageFilter]] = {
     0:  MinimalMessageFilter,
-    1:  MaximalMessageFilter,
+    1:  MinimalMessageFilter,
+    2:  MaximalMessageFilter,
 
+    "UNKNOWN":  MinimalMessageFilter,
     "MINIMAL":  MinimalMessageFilter,
     "MAXIMAL":  MaximalMessageFilter,
 }


.pylintrc
```diff
@@ -42,3 +42,4 @@ suggestion-mode=yes
 disable=
     duplicate-code,
     missing-module-docstring,
+    too-few-public-methods,
```
d1donlydfink, line 45: We have a lot of interfaces that only have one method to anticipate scaling up the API.
Becomes a pain to add this to so many classes, so add this to the pylint disable list

neuro_san/api/grpc/agent.proto
```diff
@@ -75,20 +75,33 @@ message FunctionResponse {
...
+}
+
+// Allows for controlling the messages that get streamed via StreamingChat.
+message ChatFilter {
+
+    // For now allow for an enum to describe how we want chat messages streamed.
+    // In the future the description of this server-side filter might offer more
+    // fine-grained control (hence an encapsulating structure).
+    ChatFilterType chat_filter_type = 1 [json_name="chat_filter_type"];
+}
```
d1donlydfink, line 92: Add a ChatFilter message to the ChatRequest.
We make this its own message even though there is only one field because this could grow to be more complicated beyond the basic functionality required here.

neuro_san/api/grpc/agent.proto
```diff
@@ -75,20 +75,33 @@ message FunctionResponse {
...
+
 // Request structure for Chat gRPC method
 message ChatRequest {
 
-    // User input to initiate or continue the chat
-    string user_input = 2 [json_name="user_input"];
```
d1donlydfink, line 82: Remove the user_input field to force everyone to use the ChatMessage

@dsargent @swensel @deepsaia @ofrancon : Be sure your clients are sending a user_message whose text field is filled in for user input

neuro_san/api/grpc/agent.proto
```diff
@@ -104,6 +117,10 @@ message ChatRequest {
     //      to continue the conversation (your new user_input adding to the exchange).
     //      No client-side parsing of the ChatContext is explicitly required.
     chat.ChatContext chat_context = 5 [json_name="chat_context"];
+
+    // A structure defining a server-side filter on what kind of messages come
+    // over the stream.
+    ChatFilter chat_filter = 6 [json_name="chat_filter"];
```
d1donlydfink, line 124: New chat_filter field on ChatRequest.
@dsargent @swensel @ofrancon @deepsaia : Note that if this is not present by default, the default will be minimal messages coming over which means "answer" and "chat_context".  All of your clients really want the maximal behavior, so you'll need to be sure your clients are re-tooled appropriately.

neuro_san/api/grpc/agent.proto
```diff
@@ -75,20 +75,33 @@ message FunctionResponse {
     Function function = 1 [json_name="function"];
 }
 
+// Type of filtering for StreamingChat
+enum ChatFilterType {
+    MINIMAL = 0;    // Stream the bare minimum messages to provide an answer and chat_context
+    MAXIMAL = 1;    // Stream everything possible for maximum debugging information.
+}
```
d1donlydfink, line 83: Note that MINIMAL is the default behavior.

neuro_san/client/abstract_input_processor.py
```diff
@@ -1,67 +0,0 @@
...
-        """
-        Employ the subclass's strategy to communicate with agent.
-        :param state: The state dictionary to pass around
-        :return: An updated state dictionary
-        """
-        raise NotImplementedError
-
-    def formulate_chat_request(self, user_input: str,
-                               sly_data: Dict[str, Any] = None,
-                               chat_context: Dict[str, Any] = None) -> Dict[str, Any]:
```
d1donlydfink, line 42: This method moves to StreamingInputProcessor.

neuro_san/client/agent_cli.py
```diff
@@ -51,7 +49,6 @@ def __init__(self):
         self.arg_groups: Dict[str, Any] = {}
 
         self.session: AgentSession = None
-        self.session_id: str = None
```
d1donlydfink, line 54: session_id is no longer needed in the client.

neuro_san/client/agent_cli.py
```diff
@@ -111,29 +116,19 @@ def main(self):
             print(f"    See any one of the files in {self.thinking_dir} for agent network chat details.\n")
 
         state: Dict[str, Any] = {
-            "last_logs": [],
```
d1donlydfink, line 114: last_logs is no longer needed.

neuro_san/client/agent_cli.py
```diff
@@ -111,29 +116,19 @@ def main(self):
...
             "last_chat_response": None,
             "prompt": self.default_prompt,
             "timeout": self.input_timeout_seconds,
             "num_input": 0,
             "user_input": user_input,
             "sly_data": sly_data,
+            "chat_filter": chat_filter,
```
d1donlydfink, line 130: Add chat_filter to the client state

neuro_san/client/agent_cli.py
```diff
@@ -111,29 +116,19 @@ def main(self):
...
-            input_processor = PollingInputProcessor(self.default_prompt,
-                                                    self.default_input,
-                                                    self.input_timeout_seconds,
-                                                    self.args.thinking_file,
-                                                    self.session,
-                                                    self.poll_timeout_seconds)
+        input_processor = StreamingInputProcessor(self.default_input,
+                                                  self.args.thinking_file,
+                                                  self.session,
+                                                  self.thinking_dir)
```
d1donlydfink, line 136: No more polling.

neuro_san/client/agent_cli.py
```diff
@@ -254,12 +249,12 @@ def add_args(self, arg_parser: argparse.ArgumentParser):
...
-        group.add_argument("--stream", default=True, action="store_true",
-                           help="Use streaming chat instead of polling")
-        group.add_argument("--poll", dest="stream", action="store_false",
-                           help="Use polling chat instead of streaming")
+        group = arg_parser.add_argument_group(title="Message Filtering",
+                                              description="What kind of messages will we receive?")
+        group.add_argument("--maximal", dest="chat_filter", action="store_true",
+                           help="Allow all messages to come from the server")
+        group.add_argument("--minimal", dest="chat_filter", action="store_false",
+                           help="Allow only the bare minimum of messages to come from the server")
```
d1donlydfink, line 262: Remove the stream/poll args since now there is only one.
Replace them with minimal/maximal args to influence the chat filter.

neuro_san/client/agent_session_factory.py
```diff
@@ -18,7 +18,7 @@
 from neuro_san.session.http_service_agent_session import HttpServiceAgentSession
 
 
-# pylint: disable=too-few-public-methods,too-many-arguments,too-many-positional-arguments
+# pylint: disable=too-many-arguments,too-many-positional-arguments
```
d1donlydfink, line 21: Remove the too-few-public-methods here and many many other places.

neuro_san/client/polling_input_processor.py
d1donlydfink, line 1: This class can go away.

neuro_san/client/streaming_input_processor.py
```diff
@@ -64,19 +64,16 @@ def process_once(self, state: Dict[str, Any]) -> Dict[str, Any]:
         # Note that by design, a client does not have to interpret the
         # chat_context at all. It merely needs to pass it along to continue
         # the conversation.
-        chat_request: Dict[str, Any] = self.formulate_chat_request(user_input, sly_data, chat_context)
+        chat_request: Dict[str, Any] = self.formulate_chat_request(user_input, sly_data,
+                                                                   chat_context, chat_filter)
```
d1donlydfink, line 68: Pass along the chat filter

neuro_san/client/streaming_input_processor.py
```diff
@@ -96,3 +93,35 @@ def process_once(self, state: Dict[str, Any]) -> Dict[str, Any]:
         print(f"\nResponse from {origin_str}:")
         print(f"{last_chat_response}")
         return return_state
+
+    def formulate_chat_request(self, user_input: str,
+                               sly_data: Dict[str, Any] = None,
+                               chat_context: Dict[str, Any] = None,
+                               chat_filter: Dict[str, Any] = None) -> Dict[str, Any]:
```
d1donlydfink, line 100: This method comes from AbstractInputProcessor.

neuro_san/internals/chat/data_driven_chat_session.py
```diff
@@ -51,10 +48,7 @@ def __init__(self, registry: AgentToolRegistry):
         self.front_man: FrontMan = None
 
         self.registry: AgentToolRegistry = registry
-        self.latest_response = None
-        self.last_input_timestamp = datetime.now()
         self.sly_data: Dict[str, Any] = {}
-        self.last_streamed_index: int = 0
```
d1donlydfink, line 57: Can remove a bunch of polling related member variables.

neuro_san/internals/chat/data_driven_chat_session.py
```diff
@@ -196,19 +176,6 @@ async def streaming_chat(self, user_input: str,
...
-        :return: The most recent response to the user from the chat agent.
-                Can be None if the chat agent is still chewing on the previous user input.
-        """
-        return self.latest_response
-
-    def clear_latest_response(self):
-        """
-        Clears out the latest response so as not to return duplicates.
-        """
-        self.latest_response = None
```
d1donlydfink, line 210: Remove methods that are no longer needed for polling.

neuro_san/internals/filters/answer_message_filter.py
```diff
@@ -0,0 +1,50 @@
...
+from typing import List
+
+from neuro_san.internals.filters.message_filter import MessageFilter
+from neuro_san.internals.messages.chat_message_type import ChatMessageType
+
+
+class AnswerMessageFilter(MessageFilter):
+    """
+    MessageFilter implementation for a message with "the answer" in it.
+    """
```
d1donlydfink, line 23: AnswerMessageFilter to help with MinimalMessageFilter implementation.

neuro_san/internals/filters/chat_context_message_filter.py
```diff
@@ -0,0 +1,40 @@
...
+from typing import Dict
+
+from neuro_san.internals.filters.message_filter import MessageFilter
+from neuro_san.internals.messages.chat_message_type import ChatMessageType
+
+
+class ChatContextMessageFilter(MessageFilter):
+    """
+    MessageFilter implementation for a message with "the chat_context" in it.
+    """
```
d1donlydfink, line 22: Helps with the MinimalMessageFilter implementation.

neuro_san/internals/filters/compound_message_filter.py
```diff
@@ -0,0 +1,55 @@
...
+from typing import List
+
+from neuro_san.internals.filters.message_filter import MessageFilter
+from neuro_san.internals.messages.chat_message_type import ChatMessageType
+
+
+class CompoundMessageFilter(MessageFilter):
+    """
+    A MessageFilter implementation that can service multiple other MessageFilter instances
+    """
```
d1donlydfink, line 23: Helps with the MinimalMessageFilter implementation.

neuro_san/client/agent_cli.py
```diff
@@ -79,6 +76,19 @@ def main(self):
...
+        #       developers, we specifically want a default MAXIMAL client to
+        #       show all the bells and whistles of the output that a typical
+        #       end user will not care about and not appreciate the extra
+        #       data charges on their cell phone.
+        chat_filter: Dict[str, Any] = {
+            "chat_filter_type": "MAXIMAL"
+        }
+        if self.args.chat_filter is not None and \
+                not bool(self.args.chat_filter):
+            chat_filter["chat_filter_type"] = "MINIMAL"
```
d1donlydfink, line 90: Fill in the chat_filter.  Big comment to say that by default this particular client wants maximal, but most clients will want minimal.

neuro_san/internals/filters/maximal_message_filter.py
```diff
@@ -0,0 +1,36 @@
...
+from typing import Dict
+
+from neuro_san.internals.filters.message_filter import MessageFilter
+from neuro_san.internals.messages.chat_message_type import ChatMessageType
+
+
+class MaximalMessageFilter(MessageFilter):
+    """
+    MessageFilter implementation that lets everything through.
+    """
```
d1donlydfink, line 22: What gets chosen for MAXIMAL - let's everything through.

neuro_san/internals/filters/message_filter.py
```diff
@@ -0,0 +1,52 @@
...
+from typing import Any
+from typing import Dict
+
+from neuro_san.internals.messages.chat_message_type import ChatMessageType
+
+
+class MessageFilter:
+    """
+    An interface for filtering a single message.
+    """
```
d1donlydfink, line 21: Interface that everything derives from.

neuro_san/internals/filters/message_filter.py
```diff
@@ -0,0 +1,52 @@
...
+    @staticmethod
+    def get_message_type(chat_message_dict: Dict[str, Any]) -> ChatMessageType:
+        """
+        Convert the message type in the ChatMessage dictionary to the enum we want to work with
+        :param chat_message_dict: The ChatMessage dictionary to consider.
+        :return: The ChatMessageType of the chat_message_dict
+        """
+        response_type: str = chat_message_dict.get("type")
+        message_type: ChatMessageType = ChatMessageType.from_response_type(response_type)
+        return message_type
```
d1donlydfink, line 40: This moves here from MessageProcessor.

neuro_san/internals/filters/message_filter_factory.py
```diff
@@ -0,0 +1,60 @@
...
+    """
+    Class for creating MessageFilters
+    """
+
+    @staticmethod
+    def create_message_filter(chat_filter: Dict[str, Any]) -> MessageFilter:
+        """
+        :param chat_filter: The ChatFilter dictionary to process.
+        :return: A MessageFilter that corresponds to the contents
+        """
```
d1donlydfink, line 41: Figures out what the user wanted from the chat_filter field

neuro_san/internals/filters/minimal_message_filter.py
```diff
@@ -0,0 +1,34 @@
...
+from neuro_san.internals.filters.chat_context_message_filter import ChatContextMessageFilter
+from neuro_san.internals.filters.compound_message_filter import CompoundMessageFilter
+from neuro_san.internals.filters.message_filter import MessageFilter
+
+
+class MinimalMessageFilter(CompoundMessageFilter):
+    """
+    A CompoundMessageFilter that lets the minimal messages needed for an agent interaction
+    go through.
+    """
```
d1donlydfink, line 24: Brings together al the MessageFilter pieces above to be sure that only the minimum required messages go over the wire to the client.

neuro_san/message_processing/answer_message_processor.py
```diff
@@ -55,21 +57,12 @@ def process_message(self, chat_message_dict: Dict[str, Any], message_type: ChatM
         :param chat_message_dict: The ChatMessage dictionary to process.
         :param message_type: The ChatMessageType of the chat_message_dictionary to process.
         """
-        if message_type != ChatMessageType.AI:
-            # Final answers are only ever AI Messages
+        if not self.filter.allow_message(chat_message_dict, message_type):
+            # Does not pass the criteria for a message holding a final answer
```
d1donlydfink, line 61: Use the AnswerMessageFilter for the policy

neuro_san/message_processing/chat_context_message_processor.py
```diff
@@ -47,8 +49,8 @@ def process_message(self, chat_message_dict: Dict[str, Any], message_type: ChatM
         :param chat_message_dict: The ChatMessage dictionary to process.
         :param message_type: The ChatMessageType of the chat_message_dictionary to process.
         """
-        if message_type != ChatMessageType.AGENT_FRAMEWORK:
-            # chat_context only ever comes from AGENT_FRAMEWORK Messages
+        if not self.filter.allow_message(chat_message_dict, message_type):
+            # Filter says no
```
d1donlydfink, line 53: Use the ChatContextMessageFilter for the common policy.

neuro_san/message_processing/composite_message_processor.py
```diff
@@ -47,16 +48,6 @@ def add_processor(self, message_processor: MessageProcessor):
...
 
-    def get_message_type(self, chat_message_dict: Dict[str, Any]) -> ChatMessageType:
-        """
-        Convert the message type in the ChatMessage dictionary to the enum we want to work with
-        :param chat_message_dict: The ChatMessage dictionary to consider.
-        :return: The ChatMessageType of the chat_message_dict
-        """
-        response_type: str = chat_message_dict.get("type")
-        message_type: ChatMessageType = ChatMessageType.from_response_type(response_type)
-        return message_type
```
d1donlydfink, line 58: This moves to MessageFilter

neuro_san/session/async_direct_agent_session.py
```diff
@@ -127,13 +129,10 @@ async def streaming_chat(self, request_dict: Dict[str, Any]) -> Generator[Dict[s
...
         extractor = DictionaryExtractor(request_dict)
 
-        # Get the user input. Prefer the newer-style from the user_message
-        user_input: str = request_dict.get("user_input")
-        user_input = extractor.get("user_message.text", user_input)
-
-        chat_context: Dict[str, Any] = request_dict.get("chat_context")
-        sly_data: Dict[str, Any] = request_dict.get("sly_data")
+        # Get the user input.
+        user_input = extractor.get("user_message.text")
```
d1donlydfink, line 133: A bunch of these move down closer to where they are used.

neuro_san/session/async_direct_agent_session.py
```diff
@@ -146,6 +145,13 @@ async def streaming_chat(self, request_dict: Dict[str, Any]) -> Generator[Dict[s
             yield template_response_dict
             return
 
+        # Create a message filter so as to minimize network traffic per what the user wants
+        chat_filter: Dict[str, Any] = request_dict.get("chat_filter")
+        message_filter: MessageFilter = MessageFilterFactory.create_message_filter(chat_filter)
```
d1donlydfink, line 150: Create a MessageFilter using the factory based on what the user indicated in the request.

neuro_san/session/async_direct_agent_session.py
```diff
@@ -163,7 +169,7 @@ async def streaming_chat(self, request_dict: Dict[str, Any]) -> Generator[Dict[s
         async for message in generator:
 
             response_dict: Dict[str, Any] = copy(template_response_dict)
-            if any(message):
+            if message_filter.allow(message):
```
d1donlydfink, line 172: Use the MessageFilter to determine what messages get yielded to the client.

neuro_san/session/direct_agent_session.py
```diff
@@ -130,13 +132,10 @@ def streaming_chat(self, request_dict: Dict[str, Any]) -> Generator[Dict[str, An
...
         extractor = DictionaryExtractor(request_dict)
 
-        # Get the user input. Prefer the newer-style from the user_message
-        user_input: str = request_dict.get("user_input")
-        user_input = extractor.get("user_message.text", user_input)
-
-        chat_context: Dict[str, Any] = request_dict.get("chat_context")
-        sly_data: Dict[str, Any] = request_dict.get("sly_data")
+        # Get the user input.
+        user_input = extractor.get("user_message.text")
```
d1donlydfink, line 136: Same thing here as in the asynchronous client above.

neuro_san/api/grpc/agent.proto
dsargent: The first item in the enum should be `UNSPECIFIED` or `UNKNOWN`. 

Note: *not* the same as the default value. 

See [doc](https://protobuf.dev/programming-guides/proto3/#enum):

```
In proto3, the first value defined in an enum definition must have the value zero and should have the name ENUM_TYPE_NAME_UNSPECIFIED or ENUM_TYPE_NAME_UNKNOWN. This is because:

There must be a zero value, so that we can use 0 as a numeric [default value](https://protobuf.dev/programming-guides/proto3/#default).
The zero value needs to be the first element, for compatibility with the [proto2](https://protobuf.dev/programming-guides/proto2) semantics where the first enum value is the default unless a different value is explicitly specified.
It is also recommended that this first, default value have no semantic meaning other than “this value was unspecified”.
```

neuro_san/api/grpc/agent.proto
dsargent: We will adapt to specify `MAXIMUM`.

neuro_san/api/grpc/agent.proto
dsargent: You also might want to avoid `MAXIMAL` as that doesn't leave any "room to grow". If there are new message types in future you'll end up with an arms race of `SUPER_MAXIMAL`, `SUPER_EXTRA_MAXIMAL` etc.

Consider "verbosity levels" like logging systems have instead.

d1donlydfink: I am having the 0 value be MINIMUM on purpose.
If it's not there, then someone is not thinking about adding it.
If they are not thinking about it, then give them bare minimum messages by default which is MINIMAL.
I thought about UNKNOWN, MINIMAL and MAXIMAL, but in all those scenarios, the right thing
to do is default to MINIMAL behavior when UNKNOWN is encountered anyway. So why have 2 things
mean the same thing?

The only answer I came up with to that question was "well, I could then change the default behavior
later for UNKNOWN".  But I really want to inflict that pain on fewer clients earlier, so I deliberately
decided against it.

As for logging levels, I am on purpose not going any more complex than MAXIMAL and MINIMAL at the moment.
I think there are many ways to specify in-betweens which we will have to give much more thought to
than what we are doing now.  This is precisely the reason ChatFilter is a structure with room to expand
if/when we get there.

dsargent: > I am having the 0 value be MINIMUM on purpose. If it's not there, then someone is not thinking about adding it.

What you are describing is the _default_ -- but that's not how gRPC is supposed to work. See the doc link I posted.
Enum value 0 is a de facto "reserved value" for "not specified", _not_ the default value.

dsargent: In particular it goes against this tenet of gRPC:

`It is also recommended that this first, default value have no semantic meaning other than “this value was unspecified”.`

dsargent: image

dsargent: > So why have 2 things mean the same thing?

They don't. See image I posted above. There is a difference between "not specified" and "specified, but specified the default value" and it's important to preserve that distinction.

If you really want you can check for the "forbidden" 0-value and make it an error -- "sorry, you have to specify the filter type you want". But usurping slot 0 for a semantically meaningful value is contrary to gRPC recommendations (and all our other gRPC enums too).

Edit: it's in Google's gRPC [best practices](https://protobuf.dev/best-practices/dos-donts/) too:

>Enums should include a default FOO_UNSPECIFIED value as the first value in the declaration . When new values are added to a proto2 enum, old clients will see the field as unset and the getter will return the default value or the first-declared value if no default exists . For consistent behavior with [proto enums](https://protobuf.dev/programming-guides/proto2#enum), the first declared enum value should be a default FOO_UNSPECIFIED value and should use tag 0. It may be tempting to declare this default as a semantically meaningful value but as a general rule, do not, to aid in the evolution of your protocol as new enum values are added over time. All enum values declared under a container message are in the same C++ namespace, so prefix the unspecified value with the enum’s name to avoid compilation errors. If you’ll never need cross-language constants, an int32 will preserve unknown values and generates less code. Note that [proto enums](https://protobuf.dev/programming-guides/proto2#enum) require the first value to be zero and can round-trip (deserialize, serialize) an unknown enum value.

d1donlydfink: OK I will relent.

d1donlydfink: Done

PR #107, opened 2025-03-20
remove ConversationalAgent

Noravee:


neuro_san/internals/run_context/langchain/langchain_run_context.py
```diff
@@ -211,8 +211,8 @@ async def create_resources(self, assistant_name: str,
             # Per empirical experience, this is "last".
             self.agent.last = JournalingToolsAgentOutputParser(self.journal)
         else:
-            self.agent = ConversationalAgent.from_llm_and_tools(self.llm, self.tools,
-                                                                prefix=instructions)
+            # There's no need for "intermediate step" since llm has no tools.
+            self.agent = prompt_template | self.llm | JournalingToolsAgentOutputParser(self.journal)
```
Noravee, line 223: Remove ConverstionalAgent since it is deprecated.
this chain
prompt_template | self.llm | JournalingToolsAgentOutputParser(self.journal)
is almost the same with tool_calling agent except we do not bind tools to llm (since there is no tools) and there is no intermediate_steps.

d1donlydfink: Please add a comment as the pipe notation is a non-obvious langchain-ism.

Also: Why don't we need an agent here?
The LLM might not have any tools, but that doesn't mean it doesn't need an agent, does it?
Without an agent, the JournalingToolsAgentOutputParser would likely need to be something different
(If I remembered what it does!)

Noravee: It is still an agent. The only difference is that it has no intermediete_steps since it does not have to invoke a tool.

neuro_san/internals/run_context/langchain/langchain_run_context.py
```diff
@@ -211,8 +210,17 @@ async def create_resources(self, assistant_name: str,
             # Per empirical experience, this is "last".
             self.agent.last = JournalingToolsAgentOutputParser(self.journal)
         else:
-            self.agent = ConversationalAgent.from_llm_and_tools(self.llm, self.tools,
-                                                                prefix=instructions)
+            # This uses LangChain Expression Language (LCEL), which enables a functional, pipeline-style composition
+            # using "|". Here, we pass `agent_scratchpad` in the input message, but since we don't explicitly assign it
+            # to `intermediate_steps` (as done in `create_tool_calling_agent`), it remains unused by the prompt.
```
Noravee, line 215: Add more comments regarding langchain syntax and agent with tools.

neuro_san/internals/run_context/langchain/langchain_run_context.py
```diff
@@ -211,8 +210,17 @@ async def create_resources(self, assistant_name: str,
...
+            #
+            # In contrast, `create_tool_calling_agent` can be written in LCEL as
+            # RunnablePassthrough | prompt | llm_with_tools | ToolsAgentOutputParser
+            # where RunnablePassthrough `agent scratchpad` convert (AgentAction, tool output) tuples into ToolMessages.
```
Noravee, line 219: Add more comments regarding langchain syntax and agent with tools.

neuro_san/internals/run_context/langchain/langchain_run_context.py
```diff
@@ -211,8 +210,17 @@ async def create_resources(self, assistant_name: str,
...
+            #
+            # By skipping this step, our agent functions as a pure LLM-driven system with a defined role,
+            # without tool invocation logic influencing its decision-making.
+            self.agent = prompt_template | self.llm | JournalingToolsAgentOutputParser(self.journal)
```
Noravee, line 223: Add more comments regarding langchain syntax and agent with tools.

PR #108, opened 2025-03-20
UN-3013 Add simple logging to http server.

andreidenissov-cog:
This PR adds basic logging to HTTP side-car server to enable tracking incoming service requests.

### DIFF
--- a/neuro_san/http_sidecar/handlers/function_handler.py
+++ b/neuro_san/http_sidecar/handlers/function_handler.py
@@ -17,6 +17,7 @@ from typing import Any, Dict
 from neuro_san.http_sidecar.handlers.base_request_handler import BaseRequestHandler
 from neuro_san.interfaces.async_agent_session import AsyncAgentSession
 
+
 class FunctionHandler(BaseRequestHandler):
     """
     Handler class for neuro-san "function" API call.
--- a/neuro_san/http_sidecar/logging/http_logger.py
+++ b/neuro_san/http_sidecar/logging/http_logger.py
@@ -20,6 +20,7 @@ from leaf_server_common.logging.logging_setup import setup_logging
 
 from neuro_san.http_sidecar.logging.log_context_filter import LogContextFilter
 
+
 class HttpLogger:
     """
     Custom logger class for use by Http server.
--- a/neuro_san/http_sidecar/logging/log_context_filter.py
+++ b/neuro_san/http_sidecar/logging/log_context_filter.py
@@ -16,6 +16,7 @@ See class comment for details
 import contextvars
 import logging
 
+
 class LogContextFilter(logging.Filter):
     """
     Custom logging filter for Http server.


neuro_san/http_sidecar/handlers/concierge_handler.py
```diff
@@ -27,7 +27,8 @@ def get(self):
         """
         Implementation of GET request handler for "concierge" API call.
         """
-
+        request_id: int = self.get_request_id()
+        self.logger.info("Start GET %s/list request [%d]", self.agent_name, request_id)
```
d1donlydfink: FWIW: request_id is allowed to come in from the client as request metadata as well. 1C guys use this to track their client-side requests.

neuro_san/http_sidecar/handlers/base_request_handler.py
```diff
@@ -45,6 +45,8 @@ class BaseRequestHandler(RequestHandler):
         grpc.StatusCode.DEADLINE_EXCEEDED: 504
     }
 
+    request_id: int = 0
```
d1donlydfink, line 51: FWIW: There is an AtomicCounter class in leaf-server-common

neuro_san/http_sidecar/http_sidecar.py
```diff
@@ -48,14 +50,30 @@ def __init__(self, port: int, http_port: int,
...
-        self.logger = logging.getLogger(self.__class__.__name__)
+        self.logger = None
         self.forwarded_request_metadata: List[str] = forwarded_request_metadata.split(" ")
 
+    def setup_logging(self):
+        """
+        Setup logging from configuration file.
+        """
+        server_dir = Path(__file__).parent
+        log_config = server_dir / "logging.json"
```
d1donlydfink: Use FileOfClass instead. Examples abound.

neuro_san/http_sidecar/http_sidecar.py
```diff
@@ -48,14 +50,30 @@ def __init__(self, port: int, http_port: int,
...
+        try:
+            with open(log_config, "r", encoding='utf-8') as f_log:
+                config = json.load(f_log)
+                logging.config.dictConfig(config)
+        except Exception as exc:  # pylint: disable=broad-exception-caught
+            print(f"FAILED to setup http server logging: {exc}")
```
d1donlydfink: This is precisely what the leaf-server-common logging_setup is supposed to do.

neuro_san/http_sidecar/logging.json
```diff
@@ -0,0 +1,36 @@
+{
```
d1donlydfink: We do not need yet another logging file.
Use the one set up in containers as the AGENT_SERVICE_LOG_JSON env var.
 That way any server (like the 1C guys') needs to change 1 thing.

Which brings me to a corollary:
Always try to anticipate how users will complain. Because they will. That's what people do.

neuro_san/http_sidecar/http_sidecar.py
```diff
@@ -48,14 +50,30 @@ def __init__(self, port: int, http_port: int,
...
+
     def __call__(self):
         """
         Method to be called by a process running tornado HTTP server
         to actually start serving requests.
         """
+        self.setup_logging()
```
d1donlydfink: Also: 1C guys will want their metadata forwarded to the logger because they have their own logging format they worry about and this will need to happen in any asynchronous threads too.  See agent_server_logging for details.

neuro_san/http_sidecar/handlers/base_request_handler.py
andreidenissov-cog: Yeah, I'm aware of atomic counters. But I don't think they are necessary in a single-thread event loop.

neuro_san/http_sidecar/http_sidecar.py
andreidenissov-cog: Done.

neuro_san/http_sidecar/http_sidecar.py
andreidenissov-cog: Yeah OK. Didn't want to drag in this whole pile of leaf-server-common stuff,
but we already have it anyway.

neuro_san/http_sidecar/handlers/base_request_handler.py
```diff
@@ -81,6 +85,20 @@ def get_metadata(self) -> Dict[str, Any]:
...
+        """
+        # Try to extract "user_id" from meta-data:
+        user_id: str = metadata.get("user_id", "None")
+        # If request metadata has "request_id" already specified, use it:
+        req_id: str = metadata.get("request_id", None)
+        if not req_id:
+            req_id = f"request-{BaseRequestHandler.request_id}"
+            BaseRequestHandler.request_id += 1
+        return f"user_id: {user_id} request_id: {req_id}"
+
```
andreidenissov-cog: For each incoming request, use metadata if available to construct a unique id string
to track request flow through http server.

neuro_san/http_sidecar/handlers/base_request_handler.py
```diff
@@ -61,7 +65,7 @@ def initialize(self, agent_name, port, forwarded_request_metadata):
         self.agent_name: str = agent_name
         self.port: int = port
         self.forwarded_request_metadata: List[str] = forwarded_request_metadata
-        self.logger = logging.getLogger(self.__class__.__name__)
+        self.logger = logging.getLogger(BaseRequestHandler.HTTP_LOGGER_NAME)
```
andreidenissov-cog, line 67: Get logger by a fixed name - our http sidecar logging is specified under this name in logging.json.

neuro_san/http_sidecar/handlers/base_request_handler.py
```diff
@@ -48,6 +48,10 @@ class BaseRequestHandler(RequestHandler):
         grpc.StatusCode.DEADLINE_EXCEEDED: 504
     }
 
+    HTTP_LOGGER_NAME: str = "HttpSidecar"
+
```
andreidenissov-cog: We define an explicit name for http server logger.
Must match a name in logging.json file.

neuro_san/http_sidecar/http_sidecar.py
```diff
@@ -45,17 +49,44 @@ def __init__(self, port: int, http_port: int,
...
+    def setup_logging(self):
+        """
+        Setup logging from configuration file.
+        """
+
+        # Need to initialize the forwarded metadata default values
+        extra_logging_defaults: Dict[str, str] = {
+            "source": self.server_name_for_logs
+        }
+
```
andreidenissov-cog: Logging setup for Http server differs from the other parts of the code:
here we are running a single-threaded event loop, so what is put to extra_logging_defaults must be common for all incoming client requests. 
All request-specific data (in metadata or maybe request parameters) necessary for logging
must be processed in each request and directly provided to HttpServer logger.

neuro_san/http_sidecar/http_sidecar.py
andreidenissov-cog: Note that self.setup_logging() must be called here - when we are already running in the context of separate 
http server process, not in HttpSidecar constructor, which is called in "main" neuro-san process.

neuro_san/http_sidecar/handlers/concierge_handler.py
```diff
@@ -27,10 +27,11 @@ def get(self):
         """
         Implementation of GET request handler for "concierge" API call.
         """
-
+        metadata: Dict[str, Any] = self.get_metadata()
+        request_id: str = self.get_request_id(metadata)
+        self.logger.info("Start GET %s/list %s", self.agent_name, request_id)
         try:
```
andreidenissov-cog, line 32: Here and below in all request handlers:
to log entry and exit for each handler, we construct request-specific id string using incoming metadata,
and we use it directly in the logger,
since we cannot rely of thread-local logging default data, except things common to all requests.

PR #111, opened 2025-03-24
UN-2854 - Initial swagger.json and optional CORS headers

d1donlydfink:
Get enough of a manual swagger.json going so that swagger-ui container can largely do what it needs to.
@andreidenissov-cog reports that Postman largely does better with the streaming.

Will eventually need to push a new release for this for the external-facing server.

### DIFF
--- a/neuro_san/internals/chat/data_driven_chat_session.py
+++ b/neuro_san/internals/chat/data_driven_chat_session.py
@@ -20,7 +20,6 @@ import traceback
 from openai import BadRequestError
 
 from neuro_san.internals.chat.async_collating_queue import AsyncCollatingQueue
-from neuro_san.internals.filters.answer_message_filter import AnswerMessageFilter
 from neuro_san.internals.graph.registry.agent_tool_registry import AgentToolRegistry
 from neuro_san.internals.graph.tools.front_man import FrontMan
 from neuro_san.internals.interfaces.invocation_context import InvocationContext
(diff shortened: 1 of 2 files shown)


neuro_san/api/grpc/swagger.json
```diff
@@ -0,0 +1,429 @@
+{
+  "openapi": "3.0.3",
```
d1donlydfink, line 2: Manual edit of an OpenAPI v3.0 file that @andreidenissov-cog got from gnostic (?)
Function and Connectivity worked fine, but I had to then manually tweak it and the server to get StreamingChat
working.

neuro_san/http_sidecar/handlers/streaming_chat_handler.py
```diff
@@ -37,7 +37,7 @@ async def stream_out(self,
         :param generator: async gRPC generator
         """
         # Set up headers for chunked response
-        self.set_header("Content-Type", "application/json")
+        self.set_header("Content-Type", "application/json-lines")
```
d1donlydfink, line 40: https://github.com/OAI/OpenAPI-Specification/issues/3730

Says that json-seq is not yet supported until OpenAPI 3.2, but another issue mentioned that json-lines was supported. Close enough.

neuro_san/http_sidecar/handlers/streaming_chat_handler.py
```diff
@@ -76,3 +76,9 @@ async def post(self):
             await self.flush()
         # We are done with response stream:
         await self.finish()
+
+    async def options(self):
+        """
+        Implementation of OPTIONS request handler for streaming chat API call.
+        """
+        await self.flush()
```
d1donlydfink, line 84: Handle OPTIONS requests for streaming_chat.  This allows for swagger-ui container to work w/ CORS fixes.
No idea why. 🤷

neuro_san/http_sidecar/handlers/base_request_handler.py
```diff
@@ -60,6 +63,11 @@ def initialize(self, agent_name, port, forwarded_request_metadata):
         self.forwarded_request_metadata: List[str] = forwarded_request_metadata
         self.logger = logging.getLogger(self.__class__.__name__)
 
+        if os.environ.get("AGENT_ALLOW_CORS_HEADERS") is not None:
+            self.set_header("Access-Control-Allow-Origin", "*")
+            self.set_header("Access-Control-Allow-Methods", "GET, POST, OPTIONS")
+            self.set_header("Access-Control-Allow-Headers", "Content-Type, Transfer-Encoding")
```
d1donlydfink, line 69: Put the CORS headers behind an env var that is not on by default.
@sidcog is going to configure these for the externally facing server on the load balancer.
For localhost to work with swagger-ui container though, these need to come from inside the house.

According to all the docs, this "*" is definitely not the right thing to do.
That said, I don't know what the right thing to do is just yet.

Be sure OPTIONS is part of that number as well.

neuro_san/api/grpc/swagger.json
ofrancon: Wow that's very useful. We should probably make it available in the `neuro-san-demos` docs / tutorials at some point.

d1donlydfink: The next step is to be able to generate this automatically when things change.

PR #113, opened 2025-03-25
UN-3019 Repair chat_context use in client DirectAgentSessions

d1donlydfink:
@ofrancon had found that in direct sessions, the chat_context seemed to be ignored.
Turns out it wasn't ignored at all, but the real problem was a mismatch between the origins provided in that chat_context and the new agents coming up in the subsequent conversation.  So now we reset that on the client side if we are using a direct session.  The server side effectively does this for every request already which is why the problem was localized to DirectAgentSessions only and not http or grpc sessions

### DIFF
--- a/neuro_san/client/streaming_input_processor.py
+++ b/neuro_san/client/streaming_input_processor.py
@@ -20,6 +20,7 @@ from neuro_san.interfaces.agent_session import AgentSession
 from neuro_san.internals.messages.chat_message_type import ChatMessageType
 from neuro_san.internals.messages.origination import Origination
 from neuro_san.message_processing.basic_message_processor import BasicMessageProcessor
+from neuro_san.session.direct_agent_session import DirectAgentSession
 
 
 class StreamingInputProcessor:
@@ -66,7 +67,7 @@ class StreamingInputProcessor:
         # the conversation.
         chat_request: Dict[str, Any] = self.formulate_chat_request(user_input, sly_data,
                                                                    chat_context, chat_filter)
-        self.processor.reset()
+        self.reset()
 
         return_state: Dict[str, Any] = copy(state)
         chat_responses: Generator[Dict[str, Any], None, None] = self.session.streaming_chat(chat_request)
@@ -125,3 +126,15 @@ class StreamingInputProcessor:
             chat_request["chat_filter"] = chat_filter
 
         return chat_request
+
+    def reset(self):
+        """
+        Reset for a new exchange
+        """
+        # Reset the message processor to recieve a new answer
+        self.processor.reset()
+
+        if isinstance(self.session, DirectAgentSession):
+            # Direct sessions need their Origination reset otherwise chat_context
+            # origins do not match up.
+            self.session.reset()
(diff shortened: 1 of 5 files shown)


neuro_san/client/streaming_input_processor.py
```diff
@@ -66,7 +67,7 @@ def process_once(self, state: Dict[str, Any]) -> Dict[str, Any]:
         # the conversation.
         chat_request: Dict[str, Any] = self.formulate_chat_request(user_input, sly_data,
                                                                    chat_context, chat_filter)
-        self.processor.reset()
+        self.reset()
```
d1donlydfink, line 70: Put a few elements of resetting for a new exchange into the same method.

neuro_san/client/streaming_input_processor.py
```diff
@@ -125,3 +126,15 @@ def formulate_chat_request(self, user_input: str,
...
+        """
+        Reset for a new exchange
+        """
+        # Reset the message processor to recieve a new answer
+        self.processor.reset()
+
+        if isinstance(self.session, DirectAgentSession):
+            # Direct sessions need their Origination reset otherwise chat_context
+            # origins do not match up.
+            self.session.reset()
```
d1donlydfink, line 140: Reset the message processor, and a DirectAgentSession's invocation context.

neuro_san/internals/messages/origination.py
```diff
@@ -120,3 +120,9 @@ def get_full_name_from_origin(origin: List[Dict[str, Any]]) -> str:
         full_path = full_path.replace("./", "/")
 
         return full_path
+
+    def reset(self):
+        """
+        Resets the origination tracking
+        """
+        self.tool_to_index_map = {}
```
d1donlydfink, line 128: This is the end-game of the reset() - to clear out the Originations map of agents it has already seen.
A new exchange brings a new set of agents.

neuro_san/session/async_direct_agent_session.py
```diff
@@ -174,6 +174,12 @@ async def streaming_chat(self, request_dict: Dict[str, Any]) -> Generator[Dict[s
                 response_dict["response"] = message
                 yield response_dict
 
+    def reset(self):
+        """
+        Allows for re-use of the same instance for clients
+        """
+        self.invocation_context.reset()
```
d1donlydfink, line 181: Plumbing in the asynchronous and sync versions of DirectAgentSession

neuro_san/session/session_invocation_context.py
```diff
@@ -130,3 +130,13 @@ def get_llm_factory(self) -> ContextTypeLlmFactory:
...
+
+    def reset(self):
+        """
+        Resets the instance for a subsequent use for another exchange with the agent network.
+        """
+        # Origination needs to be reset so that origin information can match up
+        # with what is in the chat_context. If we do not reset this, then library calls
+        # to DirectAgentSession do not properly carry forward any memory of the conversation
+        # in subsequent interactions with the same network.
+        self.origination.reset()
```
d1donlydfink, line 142: Plumbing to reset the Origination.

PR #115, opened 2025-03-25
UN-2854 Automate generation of OpenAPI service specification.

andreidenissov-cog:
This PR extends our do_generate.sh script with the logic to also generate OpenAPI v3 specification for AgentService.

### DIFF
--- a/neuro_san/api/grpc/agent.proto
+++ b/neuro_san/api/grpc/agent.proto
@@ -21,7 +21,7 @@ syntax = "proto3";
 package dev.cognizant_ai.neuro_san.api.grpc.agent;
 
 import "google/protobuf/struct.proto";
-import "neuro_san/api/grpc/google/api/annotations.proto";
+import "google/api/annotations.proto";
 
 import "neuro_san/api/grpc/chat.proto";
 
@@ -41,7 +41,6 @@ service AgentService {
         };
     }
 
-    // Unidirectional streaming method which would supercede Chat() and Logs() above.
     // Most important semantics of the streaming:
     //  1) The "answer" to a query of any agent network is the *last* streamed
     //     AI message whose origin list is of length 1 - this indicates it is from
--- a/neuro_san/api/grpc/agent_pb2.py
+++ b/neuro_san/api/grpc/agent_pb2.py
@@ -24,11 +24,11 @@ _sym_db = _symbol_database.Default()
 
 
 from google.protobuf import struct_pb2 as google_dot_protobuf_dot_struct__pb2
-from neuro_san.api.grpc.google.api import annotations_pb2 as neuro__san_dot_api_dot_grpc_dot_google_dot_api_dot_annotations__pb2
+from google.api import annotations_pb2 as google_dot_api_dot_annotations__pb2
 from neuro_san.api.grpc import chat_pb2 as neuro__san_dot_api_dot_grpc_dot_chat__pb2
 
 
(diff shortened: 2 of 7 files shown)


neuro_san/api/grpc/google/api/annotations.proto
```diff
@@ -0,0 +1,31 @@
+// Copyright 2015 Google LLC
```
andreidenissov-cog: Files under google/api are necessary for compiling "http options" part of our protobufs.

neuro_san/api/grpc/agent_pb2.py
```diff
@@ -24,37 +24,44 @@
 
 
 from google.protobuf import struct_pb2 as google_dot_protobuf_dot_struct__pb2
```
andreidenissov-cog, line 26: agent.proto is recompiled with new "http" options added, hence the difference with previous version.

neuro_san/api/grpc/agent.proto
```diff
@@ -21,6 +21,8 @@ syntax = "proto3";
 package dev.cognizant_ai.neuro_san.api.grpc.agent;
```
andreidenissov-cog, line 22: Adding "http" options to our service definition.
This should go (somehow) into specification of http service endpoint
in OpenAPI document.

neuro_san/api/scripts/check_openapi_plugin.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+
+set -e
+
```
andreidenissov-cog, line 4: The script (generated by GPT) which will check
if required plugin protoc-gen-openapi is available and if not,
it will provide some instructions for getting and installing it.
"go install" part is tested.

neuro_san/api/scripts/convert_yaml_to_json.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+
+# Usage: ./convert_yaml_to_json.sh input.yaml output.json
```
andreidenissov-cog, line 3: This script (again generated mostly by GPT) will convert generated .yaml file (the only option for now)
into .json file.
We assume using json format for our OpenAPI document.

neuro_san/api/grpc/agent_service.json
```diff
@@ -0,0 +1,417 @@
+{
+  "openapi": "3.0.3",
```
andreidenissov-cog, line 2: And this is the result of OpenAPI document generation.
We check it in along with compiled protobuf files of neuro-san service.

neuro_san/api/scripts/do_generate.sh
```diff
@@ -16,6 +16,9 @@
 # 2) Modify the code generated by protoc to conform to Python 3
 #    by fully specifying package paths for local imports.
 
+# Exit on any error during execution
+set -e
+
 # Find out where we are
```
andreidenissov-cog, line 22: Make the script bail out immediately on any error.

neuro_san/api/scripts/do_generate.sh
```diff
@@ -30,7 +33,7 @@ TOP_LEVEL=${ANOTHER_LEVEL_UP%/*}
 GENERATED_DIR=neuro_san/api/grpc
 
 # Ordering matters w/rt where generated file is output
-PROTO_PATH="--proto_path=${TOP_LEVEL}"
+PROTO_PATH="--proto_path=${TOP_LEVEL} --proto_path=${TOP_LEVEL}/${GENERATED_DIR}"
 echo "PROTO_PATH is ${PROTO_PATH}"
```
andreidenissov-cog, line 37: That is needed for proper handling of added google/api protobuf files.

neuro_san/api/scripts/do_generate.sh
```diff
@@ -115,3 +118,26 @@ do
     cat "${COPYRIGHT_FILE}" "${SERVICE_FILE}" > "${TEMPFILE}" && mv "${TEMPFILE}" "${SERVICE_FILE}"
 
 done
+
+# Now generate OpenAI specification for our service
```
andreidenissov-cog, line 134: The part doing OpenAPI document generation.
Tested to work if openapi plug-in is installed and available.

neuro_san/api/scripts/do_generate.sh
```diff
@@ -115,3 +118,26 @@ do
...
+
+# Check that we have necessary protoc plug-in installed:
+${TOP_LEVEL}/neuro_san/api/scripts/check_openapi_plugin.sh
+
```
andreidenissov-cog, line 137: If plug-in is not available, this script will exit with instructions on how to get it.

neuro_san/api/grpc/google/api/annotations.proto
d1donlydfink: Can't we install these from someplace?
... like this :  googleapis-common-protos
https://pypi.org/project/googleapis-common-protos/

andreidenissov-cog: Yes we actually can. Switched to getting them directly from google repo.

neuro_san/api/scripts/do_generate.sh
```diff
@@ -93,6 +96,18 @@ touch "${GENERATED_DIR}"/__init__.py
...
+GOOGLE_API_DIR="${TOP_LEVEL}"/"${GENERATED_DIR}"/google/api
+mkdir -p "${GOOGLE_API_DIR}"
+
+for PROTO_FILE in ${GOOGLE_API_FILES}
+do
+    curl --header "Accept: application/vnd.github.raw+json" \
+        --output "${GOOGLE_API_DIR}"/"${PROTO_FILE}" \
+    --location --show-error --silent --fail \
+    "https://api.github.com/repos/googleapis/googleapis/contents/google/api"/"${PROTO_FILE}"
+done
```
andreidenissov-cog, line 111: Pull 2 of google API protobuf files we need directly from google repo.

neuro_san/api/grpc/agent.proto
```diff
@@ -21,6 +21,8 @@ syntax = "proto3";
...
 
 import "google/protobuf/struct.proto";
+import "neuro_san/api/grpc/google/api/annotations.proto";
```
d1donlydfink: In other repos like unileaf, we only have to do this:
```
import "google/api/annotations.proto";
```

What prevents us from doing the same thing here?

neuro_san/api/grpc/agent_service.json
d1donlydfink: ooo.
How does it compare to the swagger.json?

neuro_san/api/grpc/agent_service.json
```diff
@@ -0,0 +1,417 @@
...
+              }
+            }
+          },
+          "required": true
+        },
+        "responses": {
+          "200": {
+            "description": "OK",
+            "content": {
+              "application/json": {
```
d1donlydfink, line 122: I believe this needs to be "application/json-lines" in order to be recognized as a stream.

There are other proposals for similar functionality out there but from what I read they are only proposals
for OpenAPI 3.2-ish.   That all said, the result here is most certainly not a single json struct as implied by current content-type.

neuro_san/api/grpc/agent.proto
```diff
@@ -33,7 +35,11 @@ option go_package = "github.com/leaf-ai/neuro_san/internal/gen/dev.cognizant_ai/
...
 
     // Called when a client needs the function description of an agent.
-    rpc Function (FunctionRequest) returns (FunctionResponse) {}
+    rpc Function (FunctionRequest) returns (FunctionResponse) {
+        option (google.api.http) = {
+            get: "/api/v1/{agent_name}/function"
+        };
+    }
 
     // Unidirectional streaming method which would supercede Chat() and Logs() above.
```
d1donlydfink: Since we are changing these files, it'd be nice to get rid of the Chat() and Logs() references here.

neuro_san/api/grpc/agent_service.json
```diff
@@ -0,0 +1,417 @@
...
+          }
+        }
+      }
+    },
+    "/api/v1/{agent_name}/streaming_chat": {
+      "post": {
+        "tags": [
+          "AgentService"
+        ],
+        "description": "Unidirectional streaming method which would supercede Chat() and Logs() above.\n Most important semantics of the streaming:\n  1) The \"answer\" to a query of any agent network is the *last* streamed\n     AI message whose origin list is of length 1 - this indicates it is from\n     the FrontMan of the agent network.\n  2) To RESTfully continue your conversation with the agent network:\n     The very last AGENT_FRAMEWORK message before the stream closes will\n     have its chat_context field filled in with a structure. You can\n     copy this whole-cloth to the chat_context of your next StreamingChat\n     request to continue the conversation.",
```
d1donlydfink: Getting rid of Chat() and Logs() reference in agent.proto above would make this comment less confusing - as they are no longer there!

neuro_san/api/scripts/check_openapi_plugin.sh
```diff
@@ -0,0 +1,51 @@
...
+  echo "🛠 macOS Installation:"
+  echo "1. Install Go (version >= $REQUIRED_GO_VERSION):"
+  echo "   brew install go"
+  echo ""
+  echo "2. Install protoc-gen-openapi plug-in:"
+  echo "   go install github.com/google/gnostic/cmd/protoc-gen-openapi@latest"
+  echo ""
+  echo "3. Add to PATH:"
+  echo "   export PATH=\"\$PATH:\$HOME/go/bin\""
+elif [[ "$OS" == "Linux" ]]; then
```
d1donlydfink, line 31: Nice! Directions even I could follow! Thanks!

neuro_san/api/scripts/convert_yaml_to_json.sh
```diff
@@ -0,0 +1,37 @@
...
+  echo "✅ Using yq..."
+  yq -o=json "$INPUT" > "$OUTPUT"
+elif command -v python3 &>/dev/null; then
+  echo "⚠️ yq not found — falling back to Python..."
+  python3 - <<EOF
+import sys, json, yaml
+with open("$INPUT") as f:
+    data = yaml.safe_load(f)
+with open("$OUTPUT", "w") as f:
+    json.dump(data, f, indent=2)
```
d1donlydfink, line 27: We know people will already have Python.
Maybe just use the python to have one less dependency?

neuro_san/api/scripts/do_generate.sh
d1donlydfink: OK. If we are doing this, then wouldn't we still need to check in the files here so that path to neuro_san/blah/blah/annotations.proto is actually able to ship with the wheel file?

Still, if there is a way to just pip install these guys (proto + .py) and find them from the venv, that would be even more ideal.

andreidenissov-cog: These files used to be checked in. But no problemo, we'll check them in again.

neuro_san/api/grpc/agent.proto
andreidenissov-cog: Good catch. With extended PROTO_PATH, we now can. Fixed.

neuro_san/api/grpc/agent_service.json
andreidenissov-cog: That's interesting. 
With my generated agent_service.json, "application/json" specified as content type, I do observe that POST request goes through and I receive 2 response json structures back.
Maybe it's Postman being smart, I dunno.
I guess we need to see how it'll go with SF people.

neuro_san/api/grpc/agent.proto
andreidenissov-cog: Done.

neuro_san/api/grpc/agent_service.json
andreidenissov-cog: No longer mentioned.

neuro_san/api/scripts/convert_yaml_to_json.sh
andreidenissov-cog: OK, out goes yq. I liked it one-liner function.

neuro_san/api/grpc/agent_service.json
andreidenissov-cog: Actually, this is what my streaming_chat handler does:

self.set_header("Content-Type", "application/json-lines")
self.set_header("Transfer-Encoding", "chunked") 

Probably it's enough for any decent http client to handle it correctly.

PR #117, opened 2025-03-26
UN-3020 Fix curly-braces in chat_context causing service-side crashes

d1donlydfink:
* Modify music_nerd_pro to return an answer in JSON format.   This ensures curly-braces from multiple sources for testing.
* Add a ChatHistoryMessageProcessor which does a few things:
    * Removes messages that will never be re-ingested into chat history - like any AGENT_* message. @ofrancon's original problem was that these messages had curly braces in them.
    * Redacts the initial system prompt which can give away the goods w/rt how the agent works.  That and it's not necessary to re-transmit it when we have the data server-side.  All we need is a SYSTEM message placeholder, and we can optimize that requirements further later on.
    * Replace any output text with curly braces with double-curly braces so that reading them in again will be fine.
* Use the new guy described above in DataDrivenChatSession
* Create some convenience methods on MessageProcessor so we don't have to redo the loop structure on lists of messages.

I think this will help the 1C guys as well.

### DIFF
--- a/neuro_san/internals/chat/chat_history_message_processor.py
+++ b/neuro_san/internals/chat/chat_history_message_processor.py
@@ -94,7 +94,7 @@ class ChatHistoryMessageProcessor(MessageProcessor):
         # JSON spec does not allow control characters in strings and newlines in particular
         # can be a problem for http clients that expect one full JSON message per line.
         # Replace any lurking newlines with the 2 raw characters \ and n.
-        text = text.replace("\n", r"\n")
+        # DEF - for the future.
 
         transformed["text"] = text
         return transformed


neuro_san/internals/chat/chat_history_message_processor.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def process_message(self, chat_message_dict: Dict[str, Any], message_type: ChatMessageType):
+        """
+        Process the message.
+        :param chat_message_dict: The ChatMessage dictionary to process.
+        :param message_type: The ChatMessageType of the chat_message_dictionary to process.
+        """
+        if message_type not in (ChatMessageType.HUMAN, ChatMessageType.SYSTEM, ChatMessageType.AI):
+            # Don't send any messages over the wire that won't be re-ingestable.
+            return
```
d1donlydfink, line 49: Do not keep anything in the chat history that will not be re-ingested.
Waste of bandwidth.

neuro_san/internals/chat/chat_history_message_processor.py
```diff
@@ -0,0 +1,95 @@
...
+
+from neuro_san.internals.messages.chat_message_type import ChatMessageType
+from neuro_san.message_processing.message_processor import MessageProcessor
+
+
+class ChatHistoryMessageProcessor(MessageProcessor):
+    """
+    MessageProcessor implementation for processing a single message
+    in a chat history.
+    """
```
d1donlydfink, line 26: MessageProcessor class to filter chat history before we send it over the wire for chat_context consumption.

neuro_san/internals/chat/chat_history_message_processor.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def redact_instructions(self, chat_message_dict: Dict[str, Any]) -> Dict[str, Any]:
+        """
+        Redacts the text instructions of the given system message.
+        These will get replaced by the agent's instructions anyway to preserve
+        server-side intent.
+        """
+        redacted: Dict[str, Any] = copy(chat_message_dict)
+        redacted["text"] = "<redacted>"
+        return redacted
```
d1donlydfink, line 72: Method used to redact the initial system prompt from chat_context

neuro_san/internals/chat/chat_history_message_processor.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def escape_message(self, chat_message_dict: Dict[str, Any]) -> Dict[str, Any]:
+        """
+        Prepare a message such that it can be re-ingested by the system nicely.
+        This means properly escaping any text that is sent.
+        """
```
d1donlydfink, line 78: Method to properly escape message text that has curly braces in it on the way out.

neuro_san/internals/chat/data_driven_chat_session.py
```diff
@@ -189,9 +191,12 @@ def prepare_chat_context(self, chat_message_history: List[Dict[str, Any]]) -> Di
                 the conversation such that it could be taken up on a different
                 server instance
         """
+        processor: MessageProcessor = ChatHistoryMessageProcessor()
+        processor.process_messages(chat_message_history)
+
         chat_history: Dict[str, Any] = {
             "origin": self.front_man.get_origin(),
-            "messages": chat_message_history
+            "messages": processor.get_message_history()
```
d1donlydfink, line 199: Process the chat history messages and include that filtered version instead of the raw list of messages.

neuro_san/message_processing/message_processor.py
```diff
@@ -53,3 +54,21 @@ async def async_process_message(self, chat_message_dict: Dict[str, Any], message
         :param message_type: The ChatMessageType of the chat_message_dictionary to process.
         """
         self.process_message(chat_message_dict, message_type)
+
+    def process_messages(self, chat_message_dicts: List[Dict[str, Any]]):
+        """
+        Convenience method for processing lists of messages.
+        :param chat_message_dicts: The messages to process.
+        """
```
d1donlydfink, line 62: Add a couple of convenience methods to loop over a list of messages.
Over time I will replace other loops with this.

neuro_san/registries/music_nerd_pro.hocon
```diff
@@ -48,8 +48,9 @@ You’re Music Nerd Pro, the go-to brain for all things rock, pop, and everythin
 You’re equal parts playlist curator, music historian, and pop culture mythbuster—with a sixth sense for sonic nostalgia and a deep respect for the analog gods.
 
 This service comes for a fee. For each question you're about to answer, use your Accountant tool to calculate the
-running fees and finish your answer with a new line that says:
-"**Running cost**: $X.XX"
+running fees. 
+
+Return your answer and the running cost in a JSON message.
```
d1donlydfink, line 53: Modify the music nerd pro to return its answers in JSON format -  yet another source of curly braces to contend with.

neuro_san/internals/chat/chat_history_message_processor.py
```diff
@@ -0,0 +1,99 @@
...
+        text = text.replace("{{", "{")
+        text = text.replace("}}", "}")
+
+        # Now replace normal braces with escaped braces.
+        # Idea is to catch everything pre-escaped or not
+        text = text.replace("{", "{{")
+        text = text.replace("}", "}}")
+
+        # Newlines can be a problem for http clients that expect one full JSON message per line.
+        text = text.replace(r"\n", "\n")
```
dsargent: I don't understand how this works -- an extra comment would be good. 

>Newlines can be a problem for http clients that expect one full JSON message per line.

I think it's more accurate to say, "the JSON spec does not allow control characters in strings".

It feels a bit "If Dr Seuss wrote Python" though I understand it's implementing something inherently confusing, escapes vs literal control characters. To the casual observer it looks like it's doing a thing, then immediately reversing the thing.

d1donlydfink: Deciding not to attempt the newline thing at all - too tweaky to combine with this change.

neuro_san/internals/chat/chat_history_message_processor.py
ofrancon: Is there any method in LangChain that we could reuse?

PR #118, opened 2025-03-26
UN-2874 Concierge service implementation.

andreidenissov-cog:
Very simple implementation of Concierge service.
All service wiring is already in place, here is the actual data we publish.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,5 @@ site/
 #       exported via GitHub zip/tar.gz and wheel file mechanisms.
 # **/grpc/generated/*
 
+# macOS folder metadata
+.DS_Store
--- a/build_scripts/Dockerfile
+++ b/build_scripts/Dockerfile
@@ -74,6 +74,4 @@ COPY --chown=${USERNAME}:${USERNAME} . ${APP_SOURCE}
 
 # Generate any grc
 USER ${USERNAME}
-RUN cd ${APP_SOURCE} ; ./neuro_san/api/scripts/do_generate.sh
-
 WORKDIR ${APP_HOME}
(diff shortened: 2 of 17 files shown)


neuro_san/session/direct_concierge_session.py
```diff
@@ -49,6 +51,10 @@ def list(self, request_dict: Dict[str, Any]) -> Dict[str, Any]:
                     protobuf structure. Has the following keys:
                 "agents" - the sequence of dictionaries describing available agents
         """
-        return {"agents": [
-            {"agent_name": "Agent-1", "description": "Doing Agent_1 thing"},
-            {"agent_name": "Agent-2", "description": "Doing Agent_2 thing"}]}
+        manifest_restorer = RegistryManifestRestorer()
+        manifest_registries: Dict[str, AgentToolRegistry] = manifest_restorer.restore()
+
```
andreidenissov-cog, line 56: Idea here is to read "current" version of registry in case it could potentially dynamically change during server lifetime.

neuro_san/session/direct_concierge_session.py
```diff
@@ -49,6 +51,10 @@ def list(self, request_dict: Dict[str, Any]) -> Dict[str, Any]:
...
+        agents_list = []
+        for agent_name, _ in manifest_registries.items():
+            agents_list.append({"agent_name": agent_name, "description": ""})
+        return {"agents": agents_list}
```
andreidenissov-cog, line 60: We provide empty "description" for each agent currently. Options are: copy "function" agent description here or introduce some separate agents descriptions somewhere. Which seems like over-complication.

PR #124, opened 2025-03-28
Minimal MINIMAL + manually add concierge service to agent_service.json

d1donlydfink:
* Now, the "answer" also comes in the last AGENT_FRAMEWORK message
* For the MINIMAL chat_filter_type:
    * Do not include the request in any ChatResponse
    * Only allow the last AGENT_FRAMEWORK message with the chat_context and answer text as the only message streamed. This is much simpler for minimal user-facing clients
 * Update the agent.proto docs for semantics on the answer and downstream files
 * Hand-tweak (for now) the agent_service.json to include the Concierge service and some servers. Tested this via swagger-ui container

### DIFF
--- a/neuro_san/api/grpc/agent_service.json
+++ b/neuro_san/api/grpc/agent_service.json
@@ -5,7 +5,46 @@
     "description": "The service comprises all the exchanges to the backend in support of a single agent's\n services.  Routing is done by way of agent name on the grpc service hosting the agent,\n so as to keep info about which agents are hosted private (grpc gives the hand when a\n particular agent is unknown.",
     "version": "0.0.1"
   },
+  "servers": [
+    {
+      "url": "https://neuro-san.decisionai.ml" 
+    },
+    {
+      "url": "http://localhost:8081" 
+    }
+  ],
   "paths": {
+    "/api/v1/list": {
+      "get": {
+        "tags": [
+          "ConciergeService"
+        ],
+        "description": "List the agents available on this server",
+        "operationId": "ConciergeService_List",
+        "responses": {
+          "200": {
+            "description": "OK",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/ConciergeResponse"
+                }
+              }
+            }
+          },
+          "default": {
+            "description": "Default error response",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/Status"
+                }
+              }
+            }
+          }
+        }
+      }
+    },
     "/api/v1/{agent_name}/connectivity": {
       "get": {
         "tags": [
@@ -406,6 +445,30 @@
           }
         },
         "description": "The `Status` type defines a logical error model that is suitable for different programming environments, including REST APIs and RPC APIs. It is used by [gRPC](https://github.com/grpc). Each `Status` message contains three pieces of data: error code, error message, and error details. You can find out more about this error model and how to work with it in the [API Design Guide](https://cloud.google.com/apis/design/errors)."
+      },
+      "ConciergeResponse": {
+        "type": "object",
+        "properties": {
+          "connectivity_info": {
+            "type": "array",
+            "items": {
+              "$ref": "#/components/schemas/AgentInfo"
+            }
+          }
+        },
+        "description": "Response structure for List gRPC method"
+      },
+      "AgentInfo": {
+        "type": "object",
+        "properties": {
+          "agent_name": {
+            "type": "string"
+          },
+          "description": {
+            "type": "string"
+          }
+        },
+        "description": "Single listing structure for List gRPC method"
       }
     }
   },


neuro_san/client/streaming_input_processor.py
```diff
@@ -91,6 +91,9 @@ def process_once(self, state: Dict[str, Any]) -> Dict[str, Any]:
         }
         return_state.update(update)
 
+        if origin_str is None or len(origin_str) == 0:
+            origin_str = "agent network"
```
d1donlydfink, line 95: If we do not get an origin from the message processor, that's OK. We were minimal and we don't get an origin from that case.

neuro_san/internals/chat/data_driven_chat_session.py
```diff
@@ -161,9 +162,19 @@ async def streaming_chat(self, user_input: str,
         chat_messages: Iterator[Dict[str, Any]] = await self.chat(user_input, invocation_context, sly_data)
         message_list: List[Dict[str, Any]] = list(chat_messages)
 
-        # Stream over chat state as the last message
+        # Determine the chat_context to enable continuing the conversation
         return_chat_context: Dict[str, Any] = self.prepare_chat_context(message_list)
-        message = AgentFrameworkMessage(content="", chat_context=return_chat_context)
```
d1donlydfink, line 166: Constructing the message moves to the end of the next block.

neuro_san/internals/chat/data_driven_chat_session.py
```diff
@@ -161,9 +162,19 @@ async def streaming_chat(self, user_input: str,
...
+
+        # Find "the answer" and have that be the content of the last message we send
+        answer_processor = AnswerMessageProcessor()
+        answer_processor.process_messages(message_list)
+        answer: str = answer_processor.get_answer()
+        if answer is None:
+            # Can't have content as None or problems arise.
+            answer = ""
```
d1donlydfink, line 174: Find the answer from the full message list. Use that for the content of the AGENT_FRAMEWORK message below.
This allows us to send a single message when using a MINIMAL chat filter.

neuro_san/internals/filters/answer_message_filter.py
```diff
@@ -30,12 +30,12 @@ def allow_message(self, chat_message_dict: Dict[str, Any], message_type: ChatMes
         :param message_type: The ChatMessageType of the chat_message_dictionary to process.
         :return: True if the message should be allowed through to the client. False otherwise.
         """
-        if message_type != ChatMessageType.AI:
-            # Final answers are only ever AI Messages
+        if message_type not in (ChatMessageType.AI, ChatMessageType.AGENT_FRAMEWORK):
+            # Final answers are only ever AI or AgentFramework Messages
```
d1donlydfink, line 34: Answers can come from an AI message or an AGENT_FRAMEWORK message.

neuro_san/internals/filters/answer_message_filter.py
```diff
@@ -30,12 +30,12 @@ def allow_message(self, chat_message_dict: Dict[str, Any], message_type: ChatMes
...
             return False
 
         origin: List[Dict[str, Any]] = chat_message_dict.get("origin")
-        if origin is None or len(origin) != 1:
+        if origin is not None and len(origin) > 1:
```
d1donlydfink, line 38: AGENT_FRAMEWORK messages won't have any origin. Allow them.

neuro_san/internals/filters/minimal_message_filter.py
```diff
@@ -28,7 +27,6 @@ def __init__(self):
         Constructor
         """
         filters: List[MessageFilter] = [
-            AnswerMessageFilter(),
```
d1donlydfink, line 31: Don't use the AnswerMessageFilter in the MinimalMessageFilter.
The ChatContextMessageFilter is enough to get the last AGENT_FRAMEWORK message.

neuro_san/message_processing/answer_message_processor.py
```diff
@@ -67,5 +67,7 @@ def process_message(self, chat_message_dict: Dict[str, Any], message_type: ChatM
         # Record what we got.
         # We might get another as we go along, but the last message in the stream
         # meeting the criteria above is our final answer.
-        self.answer = text
-        self.answer_origin = origin
+        if text is not None:
+            self.answer = text
+        if origin is not None:
+            self.answer_origin = origin
```
d1donlydfink, line 73: Accumulate state about the answer when it comes to us.
This allows the origin for the answer to be part of this processor state when MAXIMAL

neuro_san/service/agent_service.py
```diff
@@ -239,8 +239,16 @@ def StreamingChat(self, request: service_messages.ChatRequest,
...
+        chat_filter_dict: Dict[str, Any] = {}
+        chat_filter_dict = request_dict.get("chat_filter", chat_filter_dict)
+        chat_filter_type: str = chat_filter_dict.get("chat_filter_type", "MINIMAL")
+
         for response_dict in response_dict_iterator:
-            response_dict["request"] = request_dict
+
+            # Do not return the request when the filter is MINIMAL
+            if chat_filter_type != "MINIMAL":
+                response_dict["request"] = request_dict
```
d1donlydfink, line 251: Don't add the request when the chat filter is MINIMAL

.gitignore
```diff
@@ -37,3 +37,6 @@ site/
 
 # Temporary integrtion test files
 tmp_*
+
+# Transient openapi file
+neuro_san/api/grpc/openapi.yaml
```
d1donlydfink, line 42: Do not include transient file in creating agent_service.json.

neuro_san/api/grpc/agent.proto
```diff
@@ -42,9 +42,8 @@ service AgentService {
     }
 
     // Most important semantics of the streaming:
-    //  1) The "answer" to a query of any agent network is the *last* streamed
-    //     AI message whose origin list is of length 1 - this indicates it is from
-    //     the FrontMan of the agent network.
+    //  1) The "answer" to a query of any agent network is in the *last* streamed
+    //     AGENT_FRAMEWORK message.
```
d1donlydfink, line 46: Update how the answer comes to the client.

neuro_san/api/grpc/swagger.json
d1donlydfink, line 1: Don't need this if we have agent_service.json

neuro_san/api/grpc/agent_service.json
```diff
@@ -5,7 +5,46 @@
...
     "version": "0.0.1"
   },
+  "servers": [
+    {
+      "url": "https://neuro-san.decisionai.ml" 
+    },
+    {
+      "url": "http://localhost:8081" 
+    }
+  ],
```
d1donlydfink, line 15: Hand tweak this file to include a couple of servers.
Idea is to allow this file for external usage.
Later on, these aspects can be piled on automatically later.

neuro_san/api/grpc/agent_service.json
```diff
@@ -5,7 +5,46 @@
...
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/Status"
+                }
+              }
+            }
+          }
+        }
+      }
+    },
```
d1donlydfink, line 47: Add the Concierge service's list method.

neuro_san/api/grpc/agent_service.json
```diff
@@ -93,7 +132,7 @@
         "tags": [
           "AgentService"
         ],
-        "description": "Most important semantics of the streaming:\n  1) The \"answer\" to a query of any agent network is the *last* streamed\n     AI message whose origin list is of length 1 - this indicates it is from\n     the FrontMan of the agent network.\n  2) To RESTfully continue your conversation with the agent network:\n     The very last AGENT_FRAMEWORK message before the stream closes will\n     have its chat_context field filled in with a structure. You can\n     copy this whole-cloth to the chat_context of your next StreamingChat\n     request to continue the conversation.",
+        "description": "Most important semantics of the streaming:\n  1) The \"answer\" to a query of any agent network is in the *last* streamed\n     AGENT_FRAMEWORK message.\n  2) To RESTfully continue your conversation with the agent network:\n     The very last AGENT_FRAMEWORK message before the stream closes will\n     have its chat_context field filled in with a structure. You can\n     copy this whole-cloth to the chat_context of your next StreamingChat\n     request to continue the conversation.",
```
d1donlydfink, line 135: Automatically updated docs.

neuro_san/api/grpc/agent_service.json
```diff
@@ -406,6 +445,30 @@
...
+        "type": "object",
+        "properties": {
+          "agent_name": {
+            "type": "string"
+          },
+          "description": {
+            "type": "string"
+          }
+        },
+        "description": "Single listing structure for List gRPC method"
```
d1donlydfink, line 471: Concierge service schema

neuro_san/client/streaming_input_processor.py
ofrancon: I might have to account for that when I process the `origin` field in the web client then

ofrancon: Actually web client should be fine because it only processes `origin` for AGENT messages. AGENT messages always have an origin, right?

neuro_san/internals/filters/answer_message_filter.py
ofrancon: That logic is complex. If there is an origin and it contains more than 1 item we don't allow this message because we want to allow AGENT_FRAMEWORK messages that don't have an origin? The following comment doesn't make sense then because this if guarantees length is NOT 1. I'm lost. I'd need a unit test.

ofrancon: Why not a block:
```
if message_type == ChatMessageType.AGENT_FRAMEWORK:
    return True
```
?

neuro_san/client/streaming_input_processor.py
d1donlydfink: Yes. AGENT messages always have an origin.  AGENT_FRAMEWORK messages do not have an origin. They come down from on-high.

W/rt origin information here - the only change you would see is if you were using MINIMAL. Then you would 't get an origin, but the assumption there is that you care more about succinct messages and bandwidth and less about the insides of what is going on.   For your MAXIMAL client, you won't really see a change.

neuro_san/internals/filters/answer_message_filter.py
d1donlydfink: This is a decent point.  This class is overloaded in its current use - it is used on the client side to pick up answers but it is also used on the server side to pick up answers to put them in that last AGENT_FRAMEWORK message. If i divided the responsibilities in 2, this would become as clear as you would hope.  Can clean that up later.

PR #125, opened 2025-03-28
Add timing stats to token reporting

d1donlydfink:
Just like the title sez.
Easy peasy.

### DIFF
--- a/neuro_san/internals/run_context/langchain/langchain_token_counter.py
+++ b/neuro_san/internals/run_context/langchain/langchain_token_counter.py
@@ -131,6 +131,7 @@ class LangChainTokenCounter:
             # No token counting was available for the LLM, but we still need to invoke.
             retval = await awaitable
 
+        # Figure out how much time our agent took.
         end_time: float = time()
         time_taken_in_seconds: float = end_time - start_time
 


neuro_san/internals/run_context/langchain/langchain_token_counter.py
```diff
@@ -121,12 +125,16 @@ async def count_tokens(self, awaitable: Awaitable) -> Any:
                 task: Task = new_context.run(self.create_task, awaitable)
                 retval = await task
 
-            callback_var.set(old_callback)
+            if callback_var is not None:
+                callback_var.set(old_callback)
```
d1donlydfink, line 129: Prevent a crash that 1C guys saw when callback_var had been None.

neuro_san/internals/run_context/langchain/langchain_token_counter.py
```diff
@@ -121,12 +125,17 @@ async def count_tokens(self, awaitable: Awaitable) -> Any:
...
         else:
             # No token counting was available for the LLM, but we still need to invoke.
             retval = await awaitable
 
-        await self.report(callback)
+        # Figure out how much time our agent took.
+        end_time: float = time()
+        time_taken_in_seconds: float = end_time - start_time
+
+        await self.report(callback, time_taken_in_seconds)
```
d1donlydfink, line 138: Figure out how much time it took to call the awaitable whose tokens we were measuring.

neuro_san/internals/run_context/langchain/langchain_token_counter.py
```diff
@@ -214,14 +225,17 @@ def get_context_var_for_llm(llm: BaseLanguageModel) -> ContextVar:
...
         Normalizes the values in the token counting callback into a standard dictionary
 
         :param callback: A BaseCallbackHandler instance that contains token counting information
+        :param time_taken_in_seconds: The amount of time the awaitable took in count_tokens()
         """
 
-        token_dict: Dict[str, Any] = {}
+        token_dict: Dict[str, Any] = {
+            "time_taken_in_seconds": time_taken_in_seconds
+        }
```
d1donlydfink, line 238: Report the time taken. Always

neuro_san/internals/run_context/langchain/langchain_token_counter.py
```diff
@@ -233,8 +247,9 @@ def normalize_token_count(callback: BaseCallbackHandler) -> Dict[str, Any]:
                 "completion_tokens": callback.completion_tokens,
                 "successful_requests": callback.successful_requests,
                 "total_cost": callback.total_cost,
+                "time_taken_in_seconds": time_taken_in_seconds,
                 "caveats": [
-                    "Only doing token accounting for OpenAI and Anthropic models for now.",
+                    "Only doing token accounting for OpenAI, AzureOpenAI and Anthropic models for now.",
```
d1donlydfink, line 252: Re-add AzureOpenAI as supported since this has been fixed with Santhosh's gracious help.

neuro_san/internals/run_context/langchain/langchain_token_counter.py
dsargent: I went down a mini 5 minute rabbit hole of "how to time things in Python" and some seem to recommend [time.monotonic()](https://docs.python.org/3/library/time.html#time.monotonic) to avoid glitches during clock adjustments (daylight savings, leap seconds...) where clocks going _backwards_ could cause a negative elapsed time.

You'd have to be stupendously unlucky to encounter such a case in the wild though, especially at our scale.

ofrancon: Well, it will probably happen twice a year :)

dsargent: It would have to happen _during_ processing a particular request to be an issue. If it happened during "spring forward" it would appear that the request took over an hour; during "fall back", it would appear the request took _negative_ time. During a leap second...no idea. Maybe a phantom extra second on the "request duration".

d1donlydfink: An optimization to be had later if needed.  Even coarse grained timing data is valuable.

PR #126, opened 2025-03-31
Pass origin and origin_str via args to CodedTools

d1donlydfink:
As long as there are already no origin or origin_str keys in the arguments, pass what the system knows for them to the CodedTool so it can know where it is in the grand scheme of things.

This was requested by the 1C guys.

### DIFF
--- a/neuro_san/internals/graph/tools/class_tool.py
+++ b/neuro_san/internals/graph/tools/class_tool.py
@@ -16,6 +16,7 @@ from typing import List
 import json
 
 from asyncio import AbstractEventLoop
+from copy import deepcopy
 from logging import getLogger
 from logging import Logger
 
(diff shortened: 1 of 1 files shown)


neuro_san/internals/graph/tools/class_tool.py
```diff
@@ -62,12 +63,20 @@ def __init__(self, parent_run_context: RunContext,
...
             self.arguments = arguments
 
-        full_name: str = Origination.get_full_name_from_origin(self.run_context.get_origin())
-        self.logger: Logger = getLogger(full_name)
+        # Set some standard args so CodedTool can know about origin, but only if they are
+        # not already set by other infrastructure.
+        if self.arguments.get("origin") is None:
+            self.arguments["origin"] = deepcopy(self.run_context.get_origin())
+        if self.arguments.get("origin_str") is None:
+            self.arguments["origin_str"] = self.full_name
```
d1donlydfink, line 79: Do a deep copy for the origin information itself so CodedTool does not have any chance to mess things up for downstream tools.

neuro_san/internals/graph/tools/class_tool.py
```diff
@@ -62,12 +63,20 @@ def __init__(self, parent_run_context: RunContext,
...
+        self.full_name: str = Origination.get_full_name_from_origin(self.run_context.get_origin())
+        self.logger: Logger = getLogger(self.full_name)
+
+        # Put together the arguments to pass to the CodedTool
         self.arguments: Dict[str, Any] = {}
         if arguments is not None:
             self.arguments = arguments
 
-        full_name: str = Origination.get_full_name_from_origin(self.run_context.get_origin())
-        self.logger: Logger = getLogger(full_name)
```
d1donlydfink, line 70: These move up top

neuro_san/internals/graph/tools/class_tool.py
d1donlydfink: FWIW: The alternatives would be:
1) Add new parameters to the CodedTool interface.  This would provide a backwards incompatibility.
2) Go through sly_data.  Problem here is sly_data is used among all tools while these args are only directed at a single tool.

PR #127, opened 2025-03-31
UN-3013 Add logging to Http server.

andreidenissov-cog:
This PR adds simple logging to HTTP side-car server to enable tracking incoming service requests.
The effort was made to preserve current logging format used for the rest of neuro-san,
including logging of client-specific request metadata.


neuro_san/http_sidecar/logging/http_logger.py
```diff
@@ -0,0 +1,96 @@
+
+# Copyright (C) 2023-2025 Cognizant Digital Business, Evolutionary AI.
+# All Rights Reserved.
+# Issued under the Academic Public License.
+#
+# You can be released from the terms, and requirements of the Academic Public
+# License by purchasing a commercial license.
+# Purchase of a commercial license is mandatory for any use of the
+# neuro-san SDK Software in commercial settings.
+#
```
andreidenissov-cog, line 10: Custom logger class, that will:
- take request-specific metadata, already filtered by the list of metadata keys provided by neuro-san configuration;
- setup logging filter using this metadata;
- and then delegate logging proper to underlying "regular" logger defined by our logging.json.

neuro_san/http_sidecar/logging/http_logger.py
```diff
@@ -0,0 +1,96 @@
...
+        Constructor
+        """
+        LogContextFilter.set_log_context()
+        self.setup_logging()
+        # For our Http server, we have separate logging setup,
+        # because things like "user_id" and "request_id"
+        # can only be extracted and used on per-request basis.
+        # Async Http server is single-threaded.
+        self.logger = logging.getLogger(HttpLogger.HTTP_LOGGER_NAME)
+
```
andreidenissov-cog, line 42: We are getting a logger by its fixed name - this name must match what we define in logging.json

neuro_san/http_sidecar/logging/http_logger.py
```diff
@@ -0,0 +1,96 @@
...
+                      'AGENT_SERVICE_LOG_JSON',
+                      'AGENT_SERVICE_LOG_LEVEL')
+
+        # This module within openai library can be quite chatty w/rt http requests
+        logging.getLogger("httpx").setLevel(logging.WARNING)
+
+    def prepare_filter(self, metadata: Dict[str, Any]):
+        """
+        Prepare logging filter using request metadata,
+        """
```
andreidenissov-cog, line 94: Copy request metadata into context to be used by our special logger;
set up "source" to be http server.

neuro_san/http_sidecar/logging/log_context_filter.py
```diff
@@ -0,0 +1,40 @@
...
+# neuro-san SDK Software in commercial settings.
+#
+# END COPYRIGHT
+"""
+See class comment for details
+"""
+
+import contextvars
+import logging
+
```
andreidenissov-cog, line 18: Custom logging filter (as suggested by GPT).
It's function is to setup current logging record with data provided by a "log_context",
which is essentially a global object in the scope of our Http server.

neuro_san/http_sidecar/handlers/streaming_chat_handler.py
```diff
@@ -50,32 +52,36 @@ async def stream_out(self,
                 result_str: str = json.dumps(result_dict) + "\n"
                 self.write(result_str)
                 await self.flush()
+                sent_out += 1
+        return sent_out
```
andreidenissov-cog, line 57: Count chat messages we send back. We'll log it out for tracking and debugging.

neuro_san/http_sidecar/handlers/base_request_handler.py
```diff
@@ -61,7 +63,7 @@ def initialize(self, agent_name, port, forwarded_request_metadata):
         self.agent_name: str = agent_name
         self.port: int = port
         self.forwarded_request_metadata: List[str] = forwarded_request_metadata
-        self.logger = logging.getLogger(self.__class__.__name__)
+        self.logger = HttpLogger()
```
andreidenissov-cog, line 67: Use our custom logger for incoming requests.

neuro_san/http_sidecar/handlers/base_request_handler.py
```diff
@@ -79,6 +81,12 @@ def get_metadata(self) -> Dict[str, Any]:
         for item_name in self.forwarded_request_metadata:
             if item_name in headers.keys():
                 result[item_name] = headers[item_name]
+            elif item_name == "request_id":
+                # Generate unique id so we have some way to track this request:
+                result[item_name] = f"request-{BaseRequestHandler.request_id}"
+                BaseRequestHandler.request_id += 1
+            else:
+                result[item_name] = "None"
         return result
```
andreidenissov-cog, line 90: Construct metadata dictionary for our custom logger out of http request headers;
if no "request_id" is present, make up our own unique name for tracking purposes.

neuro_san/http_sidecar/handlers/base_request_handler.py
```diff
@@ -129,7 +137,7 @@ def process_exception(self, exc: Exception):
             # Handle invalid JSON input
             self.set_status(400)
             self.write({"error": "Invalid JSON format"})
-            self.logger.error("error: Invalid JSON format")
+            self.logger.error({}, "error: Invalid JSON format")
             return
```
andreidenissov-cog, line 141: If there is no request metadata at this point, just put in empty dictionary,
all custom fields will come out as "None".

neuro_san/http_sidecar/handlers/concierge_handler.py
```diff
@@ -27,10 +27,10 @@ def get(self):
         """
         Implementation of GET request handler for "concierge" API call.
         """
-
+        metadata: Dict[str, Any] = self.get_metadata()
+        self.logger.info(metadata, "Start GET %s/list", self.agent_name)
         try:
```
andreidenissov-cog, line 32: Here and all request handlers below:
prepare metadata out of request headers;
pass it to our custom logger.
Note that internally, we have to setup logger filter every time we call logging operation,
right before we use an underlying logger.
Reason: we are running in a single-thread event loop.

neuro_san/deploy/logging.json
```diff
@@ -16,9 +23,23 @@
...
+            "stream": "ext://sys.stdout"
         }
     },
     "loggers": {
+        "HttpServer": {
+            "level": "INFO",
+            "handlers": [
+                "http_console"
+            ],
+            "propagate": false
```
andreidenissov-cog, line 41: Define custom logger for Http server - that includes a special Filter for logging records.
We use the same structured formatting as the rest of neuro-san service.

neuro_san/http_sidecar/handlers/concierge_handler.py
d1donlydfink: FWIW: We have a custom logger level of "API" via leaf-server-common infrastructure that might also suit here.  Currently it's above INFO=20 but below WARN=30 ... API=27

neuro_san/http_sidecar/logging/log_context_filter.py
d1donlydfink: It's possible this could be worth eventually pushing down to leaf-server-common. Totally appropriate to have it start here.

PR #132, opened 2025-04-02
UN-3038 Generate a single OpenAPI spec document for neuro-san.

andreidenissov-cog:
Changes to our do_generate.sh script to produce a single unified OpenAPI service spec,
so all our neuro-san services are included (that is, both of them).

### DIFF
--- a/neuro_san/api/scripts/do_generate.sh
+++ b/neuro_san/api/scripts/do_generate.sh
@@ -99,13 +99,13 @@ touch "${GENERATED_DIR}"/__init__.py
 ALL_PROTO_FILES=""
 ALL_PROTO_FILES="${ALL_PROTO_FILES} ${LOCAL_PROTO_FILES}"
 
-# Install google API related protobufs:
-GOOGLE_API_FILES="annotations.proto http.proto"
-GOOGLE_API_DIR="${TOP_LEVEL}"/"${GENERATED_DIR}"/google/api
-mkdir -p "${GOOGLE_API_DIR}"
-
 # Google API proto files are not automatically updated,
 # but if this needs to be done, uncomment the lines below:
+## Install google API related protobufs:
+#GOOGLE_API_FILES="annotations.proto http.proto"
+#GOOGLE_API_DIR="${TOP_LEVEL}"/"${GENERATED_DIR}"/google/api
+#mkdir -p "${GOOGLE_API_DIR}"
+#
 #for PROTO_FILE in ${GOOGLE_API_FILES}
 #do
 #    curl --header "Accept: application/vnd.github.raw+json" \
@@ -113,6 +113,8 @@ mkdir -p "${GOOGLE_API_DIR}"
 #    --location --show-error --silent --fail \
 #    "https://api.github.com/repos/googleapis/googleapis/contents/google/api"/"${PROTO_FILE}"
 #done
+# End of lines to uncomment if you need
+# to update Google API files with latest versions.
 
 # Generate the python files and make them happy w/rt Python 3
 for PROTO_FILE in ${ALL_PROTO_FILES}


neuro_san/api/grpc/concierge.proto
```diff
@@ -20,11 +20,20 @@ syntax = "proto3";
...
 service ConciergeService {
 
     // Called when a client needs the information about available agents.
-    rpc List (ConciergeRequest) returns (ConciergeResponse) {}
+    rpc List (ConciergeRequest) returns (ConciergeResponse) {
+        option (google.api.http) = {
+            get: "/api/v1/list"
+        };
+    }
 }
```
andreidenissov-cog, line 37: Add options necessary for compiling this source with open-api plug-in.

neuro_san/api/grpc/agent_service.json
```diff
@@ -1,25 +1,16 @@
 {
   "openapi": "3.0.3",
   "info": {
```
andreidenissov-cog, line 3: Regenerated combined OpenAPI spec document.

neuro_san/api/scripts/do_generate.sh
```diff
@@ -44,6 +44,9 @@ COPYRIGHT_FILE="${TOP_LEVEL}/build_scripts/source_available_copyright.txt"
 # Where output files go.
 PYTHON_OUT=${TOP_LEVEL}
 
+# Configuration values for OpenAPI specification document
+SERVICE_TITLE=NeuroSan
+SERVICE_VERSION=0.0.1
```
andreidenissov-cog, line 50: Explicitly define service title and version number that will go into OpenAPI specification.

neuro_san/api/scripts/do_generate.sh
```diff
@@ -96,18 +99,22 @@ touch "${GENERATED_DIR}"/__init__.py
...
+#
+#for PROTO_FILE in ${GOOGLE_API_FILES}
+#do
+#    curl --header "Accept: application/vnd.github.raw+json" \
+#        --output "${GOOGLE_API_DIR}"/"${PROTO_FILE}" \
+#    --location --show-error --silent --fail \
+#    "https://api.github.com/repos/googleapis/googleapis/contents/google/api"/"${PROTO_FILE}"
+#done
+# End of lines to uncomment if you need
+# to update Google API files with latest versions.
```
andreidenissov-cog, line 118: We don't want to update Google API protobufs every time we generate.
If we do, we need to uncomment this block.

neuro_san/api/scripts/do_generate.sh
```diff
@@ -139,7 +146,11 @@ done
 echo "generating OpenAPI specification for ${TOP_LEVEL}/${GENERATED_DIR}/agent.proto"
 # shellcheck disable=SC2086    # PROTO_PATH is compilation of cmd line args
 python -m grpc_tools.protoc ${PROTO_PATH} \
-    --openapi_out="${TOP_LEVEL}"/${GENERATED_DIR} "${TOP_LEVEL}"/${GENERATED_DIR}/agent.proto
+    --openapi_out="${TOP_LEVEL}"/${GENERATED_DIR} \
+    --openapi_opt title="${SERVICE_TITLE}" \
+    --openapi_opt version="${SERVICE_VERSION}" \
+    "${TOP_LEVEL}"/${GENERATED_DIR}/agent.proto \
+    "${TOP_LEVEL}"/${GENERATED_DIR}/concierge.proto
```
andreidenissov-cog, line 154: This is how combined OpenAPI spec is generated.

neuro_san/api/scripts/do_generate.sh
d1donlydfink: Hmm. Makes me wonder how we will want to increment this service version as we go on.
Don't have to answer that now.

neuro_san/api/scripts/do_generate.sh
d1donlydfink: ah. neat

So we also include the /api/v1/doc somehow?  Or is that redundant in that - If you have to ask how did you get this json representation in the first place?

andreidenissov-cog: That's actually a good question. Right now we just define an http-level handler to serve this document.
Do we want to include this request on grpc level?
Into Concierge service API maybe?

d1donlydfink: I don't know what the standard is.  This is progress in an of itself, so no need to block it over this.

PR #135, opened 2025-04-03
UN-3049  1C: Crash in create_base_tool()

Noravee:
Check for base_tool after checking that agent_spec is not None.


neuro_san/internals/run_context/langchain/langchain_run_context.py
```diff
@@ -258,11 +257,13 @@ async def _create_base_tool(self, name: str) -> BaseTool:
                 agent_message = AgentMessage(content=message)
                 await self.journal.write_message(agent_message)
                 self.logger.info(message)
-        elif base_tool:
-            base_tool_factory = BaseToolFactory()
-            return base_tool_factory.get_agent_tool(base_tool, agent_spec.get('args'))
         else:
-            function_json = agent_spec.get("function")
+            base_tool: str = agent_spec.get('base_tool')
```
Noravee, line 261: Check for base_tool only when agent_spec is not None.

PR #136, opened 2025-04-03
UN-3023 Eliminate duplicate AGENT messages

d1donlydfink:
There exists a source of AGENT messages which sometimes, not always, produces duplicate content the same as a subsequent AI message.  Add a mechanism to filter that duplicate out.

### DIFF
--- a/neuro_san/internals/journals/originating_journal.py
+++ b/neuro_san/internals/journals/originating_journal.py
@@ -38,6 +38,7 @@ class OriginatingJournal(Journal):
         self.wrapped_journal: Journal = wrapped_journal
         self.origin: List[Dict[str, Any]] = origin
         self.chat_history: List[BaseMessage] = chat_history
+        self.pending: BaseMessage = None
 
     def get_origin(self) -> List[Dict[str, Any]]:
         """
@@ -65,6 +66,13 @@ class OriginatingJournal(Journal):
 
         if self.chat_history is not None and is_relevant_to_chat_history(message):
             self.chat_history.append(message)
+
+        if self.pending is not None:
+            # Avoid cases where two different kinds of message hold the same content.
+            if self.pending.content != message.content:
+                await self.wrapped_journal.write_message(self.pending, use_origin)
+            self.pending = None
+
         await self.wrapped_journal.write_message(message, use_origin)
 
     def get_chat_history(self) -> List[BaseMessage]:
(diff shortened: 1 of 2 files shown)


neuro_san/internals/journals/originating_journal.py
```diff
@@ -65,10 +66,29 @@ async def write_message(self, message: BaseMessage, origin: List[Dict[str, Any]]
...
+        """
+        Writes a BaseMessage entry into the wrapped_journal
+        and appends to the chat history as long as the next message does not have the same content.
+
+        :param message: The BaseMessage instance to write to the wrapped_journal
+        """
+        if self.pending is not None:
+            # Flush if anything is already waiting.
+            await self.write_message(self.pending)
+        self.pending = message
```
d1donlydfink, line 94: Allow a mechanism where a single message is not sent immediately when the source knows there is the possibility of a duplicate message going out from another source.

neuro_san/internals/run_context/langchain/journaling_callback_handler.py
```diff
@@ -46,19 +46,19 @@ class JournalingCallbackHandler(AsyncCallbackHandler):
     # Declarations of member variables here satisfy Pydantic style,
     # which is a type validator that langchain is based on which
     # is able to use JSON schema definitions to validate fields.
-    journal: Journal
+    journal: OriginatingJournal
```
d1donlydfink, line 49: Change the typing to OriginatgJournal, as that is the guy that has the new method for pending messages.

neuro_san/internals/run_context/langchain/journaling_callback_handler.py
```diff
@@ -73,7 +73,10 @@ async def on_llm_end(self, response: LLMResult,
             if content is not None and len(content) > 0:
                 # Package up the thinking content as an AgentMessage to stream
                 message = AgentMessage(content=content.strip())
-                await self.journal.write_message(message)
+                # Some AGENT messages that come from this source end up being dupes
+                # of AI messages that can come later.
+                # Use this method to put the message on hold for later comparison.
+                await self.journal.write_message_if_next_not_dupe(message)
```
d1donlydfink, line 79: Use the new method to put a message from this source in a pending state before writing it out.
Tricky part here is that not all messages from this source produce duplicate AGENT messages. Only sometimes.

PR #137, opened 2025-04-04
Fixed issue with running the integration test

kxk302:
- Added build.sh and run.sh scripts to be able to create an image and run a container locally
- Simplified the integration test by using run() instead of Popen(). No need to do process cleanup now as it is handled by run().

### DIFF
--- a/tests/client/test_client_response.py
+++ b/tests/client/test_client_response.py
@@ -16,8 +16,6 @@ class TestClientResponse(unittest.TestCase):
     and get the response in another file. This helps with integration testing
     """
 
-    music_nerd_pro_agent = "music_nerd_pro"
-
     @staticmethod
     def run_agent_cli_subprocess(agent, input_file, response_file):
         """


build_scripts/build.sh
```diff
@@ -0,0 +1,128 @@
+#!/bin/bash -e
```
kxk302, line 1: Script to build Docker image so we can run the integration test locally.

build_scripts/run.sh
```diff
@@ -0,0 +1,79 @@
+#!/bin/bash
```
kxk302, line 1: Script to run locally created Docker image

tests/client/test_client_response.py
```diff
@@ -20,36 +19,23 @@ class TestMusicNerdProClient(unittest.TestCase):
...
-    @staticmethod
-    def destruct_agent_cli_subprocess(agent_process):
-        """
-        :param agent_process: a Popen object representing the running process to be destructed
-        :return: None
-        """
-        if agent_process:
-            agent_process.terminate()
-            agent_process.wait()
+        result = subprocess.run(["python3", "-m", "neuro_san.client.agent_cli",
```
kxk302, line 29: Using run() instead of Popen(). run() waits for the process to finish and returns a result object. No need to do process clean up with Popen.

tests/client/test_client_response.py
```diff
@@ -73,14 +59,17 @@ def test_beatles(self):
...
 
             try:
-                agent_process = TestMusicNerdProClient.get_agent_cli_subprocess(TestMusicNerdProClient.agent,
-                                                                                input_file, response_file.name)
-                self.assert_response(response_file, response_keyword)
-            finally:
-                TestMusicNerdProClient.destruct_agent_cli_subprocess(agent_process)
+                result = TestMusicNerdProClient.run_agent_cli_subprocess(TestMusicNerdProClient.agent,
+                                                                         input_file, response_file.name)
+                if result.returncode == 0:
```
kxk302, line 66: If return code is non zero, CalledProcessError exception is thrown.

build_scripts/run.sh
```diff
@@ -0,0 +1,79 @@
...
+
+    # Run the docker container in interactive mode
+    #   Mount the current user's aws credentials in the container
+    #   Mount the 1st command line arg as the place where input files come from
+    #   Mount the directory where the vault certs live. In production this would be
+    #       set up to what kubernetes does natively.
+    #   Slurp in the rest as environment variables, all of which are optional.
+    # XXX 11/17/23 There might be more env vars to add here for DB stuff to function properly.
+    #   It'd be ideal if we could have a single cmd line arg that referenced the deployment
+    #   we wanted (with a default of dev) and then all the DB information could come from Vault.
```
d1donlydfink: We don't have db stuff in neuro-san

tests/client/test_client_response.py
```diff
@@ -20,36 +19,23 @@ class TestMusicNerdProClient(unittest.TestCase):
     agent = "music_nerd_pro"
 
     @staticmethod
-    def get_agent_cli_subprocess(agent, input_file, response_file):
+    def run_agent_cli_subprocess(agent, input_file, response_file):
```
d1donlydfink, line 20: How does this get run inside the container?

tests/client/test_client_response.py
```diff
@@ -73,14 +59,17 @@ def test_beatles(self):
...
             response_keyword = "Beatles"
-            agent_process = None
 
             try:
-                agent_process = TestMusicNerdProClient.get_agent_cli_subprocess(TestMusicNerdProClient.agent,
-                                                                                input_file, response_file.name)
-                self.assert_response(response_file, response_keyword)
-            finally:
-                TestMusicNerdProClient.destruct_agent_cli_subprocess(agent_process)
+                result = TestMusicNerdProClient.run_agent_cli_subprocess(TestMusicNerdProClient.agent,
```
d1donlydfink: The class name does not match the file name. Please make this match.

build_scripts/run.sh
kxk302: What exactly is the db stuff? The comment? or some env variable on the command line below?

tests/client/test_client_response.py
kxk302: Renamed the class

tests/client/test_client_response.py
kxk302: CodeFresh builds Docker image, and runs the container with pytest command.

PR #145, opened 2025-04-10
UN-3077 Fix the problem with setting default field values for Http server logging.

andreidenissov-cog:
This PR makes sure that for any http logger call, our logger filter sets up logging record at least with some default value ("None") for every key from forwarded request meta-data.
The problem shows up actually when we use Http logger outside of request handling context,
but this case is now fixed too.

### DIFF
--- a/neuro_san/http_sidecar/http_sidecar.py
+++ b/neuro_san/http_sidecar/http_sidecar.py
@@ -62,7 +62,7 @@ class HttpSidecar:
         Method to be called by a process running tornado HTTP server
         to actually start serving requests.
         """
-        self.logger = HttpLogger()
+        self.logger = HttpLogger(self.forwarded_request_metadata)
 
         app = self.make_app()
         app.listen(self.http_port)
--- a/neuro_san/http_sidecar/logging/http_logger.py
+++ b/neuro_san/http_sidecar/logging/http_logger.py
@@ -12,9 +12,10 @@
 """
 See class comment for details
 """
+import copy
 import logging
 import pathlib
-from typing import Any, Dict
+from typing import Any, Dict, Sequence
 
 from leaf_server_common.logging.logging_setup import setup_logging
 
@@ -28,7 +29,7 @@ class HttpLogger:
 
     HTTP_LOGGER_NAME: str = "HttpServer"
 
-    def __init__(self):
+    def __init__(self, forwarded_metadata: Sequence[str]):
         """
         Constructor
         """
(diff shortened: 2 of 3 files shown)


neuro_san/http_sidecar/http_sidecar.py
```diff
@@ -62,7 +62,7 @@ def __call__(self):
         Method to be called by a process running tornado HTTP server
         to actually start serving requests.
         """
-        self.logger = HttpLogger()
+        self.logger = HttpLogger(self.forwarded_request_metadata)
```
andreidenissov-cog, line 66: Pass forwarded meta-data list to our custom Http logger.

neuro_san/http_sidecar/logging/http_logger.py
```diff
@@ -39,6 +40,12 @@ def __init__(self):
         # can only be extracted and used on per-request basis.
         # Async Http server is single-threaded.
         self.logger = logging.getLogger(HttpLogger.HTTP_LOGGER_NAME)
+        # Initialize minimal metadata dictionary to contain some value
+        # for each metadata key we expect to be used.
+        self.base_metadata: Dict[str, Any] = {}
+        for key in forwarded_metadata:
+            self.base_metadata[key] = "None"
+        self.base_metadata["source"] = HttpLogger.HTTP_LOGGER_NAME
```
andreidenissov-cog, line 49: Build a "minimal" dictionary for forwarded meta-data keys, so logging record will be initialized for everything we might use.

neuro_san/http_sidecar/logging/http_logger.py
```diff
@@ -92,5 +99,6 @@ def prepare_filter(self, metadata: Dict[str, Any]):
         """
         Prepare logging filter using request metadata,
         """
-        metadata["source"] = HttpLogger.HTTP_LOGGER_NAME
-        LogContextFilter.log_context.set(metadata)
+        use_metadata: Dict[str, Any] = copy.copy(self.base_metadata)
+        use_metadata.update(metadata)
+        LogContextFilter.log_context.set(use_metadata)
```
andreidenissov-cog, line 104: Update dictionary to be used by logger filter with actual meta-data coming from request being processed.
Now this will work even if incoming meta-data is empty.

d1donlydfink (commented):
In general, seems good.
It'd be good to know how you tested.

d1donlydfink (approved):
@andreidenissov-cog says:

Quick test - extend logging.json formatter with "query_id",
export AGENT_FORWARDED_REQUEST_METADATA="request_id user_id query_id"
start neuro-san server - no crash;
manually put "query_id": "xyz" into some metadata passed to the Http logger,
observe logs out with and without this addition.

PR #148, opened 2025-04-11
UN-3075 max_message_history

d1donlydfink:
Requested by 1C guys.

Allow a "max_message_history" on the front man agent spec to potentially limit the number of messages sent back in the chat history. Preserved context favors the N most recent messages.

This becomes a somewhat simple/reasonable way to deal with input context limits for long conversations like 1C has without getting into the nitty-gritty of token counting and input context limits when agents can now have fallback llms of varying capabilities.  Some of the 1C responses get pretty long and the number of interactions can be pretty high as well.

Tested: music_nerd_pro with and without a max_message_history of 2.
Q: Who did Yellow Submarine?
A: Beatles
Q: Where were they from?
A: Liverpool
Q: What album was it on?
A: Revolver     <- No max_message_history
A: What are you talking about? <- max_message_history = 2  (Yellow Submarine is out of context now)

### DIFF
--- a/neuro_san/internals/chat/chat_history_message_processor.py
+++ b/neuro_san/internals/chat/chat_history_message_processor.py
@@ -9,6 +9,7 @@
 # neuro-san SDK Software in commercial settings.
 #
 # END COPYRIGHT
+from numbers import Number
 from typing import Any
 from typing import Dict
 from typing import List
@@ -25,10 +26,24 @@ class ChatHistoryMessageProcessor(MessageProcessor):
     in a chat history.
     """
 
-    def __init__(self):
+    def __init__(self, max_message_history: int = None):
         """
         Constructor
+
+        :param max_message_history: The maximum number of messages to preserve
+                in the message history, not including the instructions.
+                The default value of None implies there is no maximum.
+                Non-positive numbers revert to no-maximum behavior.
         """
+        self.max_message_history: int = max_message_history
+        if self.max_message_history is not None:
+            if not isinstance(self.max_message_history, Number):
+                # If we don't have a number we don't have a max.
+                self.max_message_history = None
+            else:
+                # Be sure we are dealing with an integer
+                self.max_message_history = int(self.max_message_history)
+
         self.message_history: List[Dict[str, Any]] = []
         self.saw_first_system: bool = False
 
@@ -37,6 +52,27 @@ class ChatHistoryMessageProcessor(MessageProcessor):
         :return: The filtered message history
         """
         return self.message_history
+
+    def process_messages(self, chat_message_dicts: List[Dict[str, Any]]):
+        """
+        Convenience method for processing lists of messages.
+        :param chat_message_dicts: The messages to process.
+        """
+        super().process_messages(chat_message_dicts)
+
+        # See if we need to curtail chat history at all.
+        if self.max_message_history is None or self.max_message_history <= 1:
+            # Nothing to do.
+            return
+
+        # Save the first item in the list. This is the redacted placeholder for instructions
+        instructions: Dict[str, Any] = self.message_history.pop(0)
+
+        # Preserve the most recent n elements in the message history
+        self.message_history = self.message_history[-self.max_message_history:]
+
+        # Prepend the instructions to the list
+        self.message_history.insert(0, instructions)
 
     def process_message(self, chat_message_dict: Dict[str, Any], message_type: ChatMessageType):
         """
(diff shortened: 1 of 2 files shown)


neuro_san/internals/chat/chat_history_message_processor.py
```diff
@@ -25,10 +26,24 @@ class ChatHistoryMessageProcessor(MessageProcessor):
...
+                Non-positive numbers revert to no-maximum behavior.
         """
+        self.max_message_history: int = max_message_history
+        if self.max_message_history is not None:
+            if not isinstance(self.max_message_history, Number):
+                # If we don't have a number we don't have a max.
+                self.max_message_history = None
+            else:
+                # Be sure we are dealing with an integer
+                self.max_message_history = int(self.max_message_history)
```
d1donlydfink, line 45: Be sure we get either None or an integer for the new max_message_history parameter.

neuro_san/internals/chat/chat_history_message_processor.py
```diff
@@ -38,6 +53,27 @@ def get_message_history(self) -> List[Dict[str, Any]]:
         """
         return self.message_history
 
+    def process_messages(self, chat_message_dicts: List[Dict[str, Any]]):
+        """
+        Convenience method for processing lists of messages.
+        :param chat_message_dicts: The messages to process.
+        """
```
d1donlydfink, line 60: Override the method from the superclass to allow for some post-processing of the message_history

neuro_san/internals/chat/data_driven_chat_session.py
```diff
@@ -212,7 +212,9 @@ def prepare_chat_context(self, chat_message_history: List[Dict[str, Any]]) -> Di
                 the conversation such that it could be taken up on a different
                 server instance
         """
-        processor: MessageProcessor = ChatHistoryMessageProcessor()
+        # OK if this is None
+        max_message_history: int = self.front_man.get_agent_tool_spec().get("max_message_history")
+        processor: MessageProcessor = ChatHistoryMessageProcessor(max_message_history)
```
d1donlydfink, line 217: Get the max_message_history from the front-man.

I thought a bit about where the parameter should come from:
* A server env var was too harsh because it was global.
* Different hocon files will have different avg. length answers
* At some point in the future we might have multiple chat histories coming back.

... so I settled with "max_message_history" on the front-man to contain the scope, but also be forward-compatible.

neuro_san/internals/chat/chat_history_message_processor.py
andreidenissov-cog: And said integer is positive?

neuro_san/internals/chat/chat_history_message_processor.py
andreidenissov-cog: If we have a limit on this history length, do we need to process all messages before cutting off
the tail?

neuro_san/internals/chat/data_driven_chat_session.py
ofrancon: You should document it somewhere. No way I remember if someone asks.

ofrancon (approved):
lgtm. Would be nice to document in the agent tool spec, for the frontman only.

ofrancon:
You could have a unit test for that feature too

neuro_san/internals/chat/chat_history_message_processor.py
```diff
@@ -25,10 +26,24 @@ class ChatHistoryMessageProcessor(MessageProcessor):
...
+    def __init__(self, max_message_history: int = None):
         """
         Constructor
+
+        :param max_message_history: The maximum number of messages to preserve
+                in the message history, not including the instructions.
+                The default value of None implies there is no maximum.
+                Non-positive numbers revert to no-maximum behavior.
         """
+        self.max_message_history: int = max_message_history
```
kxk302, line 38: Not a biggie, but you can initialize the field to None, and set it if the input parameter is an instance of a Number (and have an if instead of if/else)

neuro_san/internals/chat/chat_history_message_processor.py
d1donlydfink: That is checked below.

neuro_san/internals/chat/chat_history_message_processor.py
d1donlydfink: We are actually cutting off the head.  There is more relevance w/rt current conversation towards the tail.
In regards to "process all messages": yes we process all the messages before cutting off the head.
That's the call to the superclass method just below.

neuro_san/internals/chat/data_driven_chat_session.py
d1donlydfink: I think there is a test case to be had (see beatles example above) when the infrastructure comes.
At some point it will be worth putting together a pretty extensive hocon document. Would be good to see what we have going in demos before starting on that.

PR #149, opened 2025-04-11
UN-3077 Properly initialize base-level logger for Http server.

andreidenissov-cog:
This PR fixes initialization for "base level" logger in Http neuro-san server.
Http server uses 2 kinds of loggers: one is custom logger specifically tuned to work for code running in Tornado event loop,
the other(s) are loggers used by gRPC service stack from leaf-common, which rely on thread-specific custom meta-data
we use. This is the second kind which wasn't properly initialized for user-specified requests meta-data.

### DIFF
--- a/neuro_san/http_sidecar/logging/http_logger.py
+++ b/neuro_san/http_sidecar/logging/http_logger.py
@@ -33,6 +33,12 @@ class HttpLogger:
         """
         Constructor
         """
+        # Initialize minimal metadata dictionary to contain some value
+        # for each metadata key we expect to be used.
+        self.base_metadata: Dict[str, Any] = {}
+        for key in forwarded_metadata:
+            self.base_metadata[key] = "None"
+        self.base_metadata["source"] = HttpLogger.HTTP_LOGGER_NAME
         LogContextFilter.set_log_context()
         self.setup_logging()
         # For our Http server, we have separate logging setup,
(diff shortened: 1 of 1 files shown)


neuro_san/http_sidecar/logging/http_logger.py
```diff
@@ -33,19 +33,19 @@ def __init__(self, forwarded_metadata: Sequence[str]):
...
         # because things like "user_id" and "request_id"
         # can only be extracted and used on per-request basis.
         # Async Http server is single-threaded.
         self.logger = logging.getLogger(HttpLogger.HTTP_LOGGER_NAME)
-        # Initialize minimal metadata dictionary to contain some value
-        # for each metadata key we expect to be used.
-        self.base_metadata: Dict[str, Any] = {}
-        for key in forwarded_metadata:
-            self.base_metadata[key] = "None"
-        self.base_metadata["source"] = HttpLogger.HTTP_LOGGER_NAME
```
andreidenissov-cog, line 49: Move this code up so that self.base_metadata is visible in setup_logging() scope.

neuro_san/http_sidecar/logging/http_logger.py
```diff
@@ -87,10 +87,13 @@ def setup_logging(self):
...
         Setup logging from configuration file.
         """
+        # Need to initialize the forwarded metadata default values before our first
+        # call to a logger.
         current_dir: str = pathlib.Path(__file__).parent.parent.resolve()
         setup_logging(HttpLogger.HTTP_LOGGER_NAME, current_dir,
                       'AGENT_SERVICE_LOG_JSON',
-                      'AGENT_SERVICE_LOG_LEVEL')
+                      'AGENT_SERVICE_LOG_LEVEL',
+                      self.base_metadata)
```
andreidenissov-cog, line 97: Initialize thread-specific logger record fields. They are not likely to contain anything non-default,
but at least logger will not crash.

d1donlydfink (approved):

PR #152, opened 2025-04-13
UN-3085-neuro-san-anthropic-models-type-error-write-argument-must-be-str-not-list

Noravee:
### Changes
- Check if "text" is str or list in write_message in thinking_file_message_processor
- Check if "text" in escape_message in chat_history_message_processor is a str or list.
- If it is a list, "text" is a key of a dictionary which is the first element of the list.

### Tests
- Run agent hocon with various anthropic models and specifically talk to frontman to avoid having it calling tools (see UN-3086 https://leaf-ai.atlassian.net/jira/software/projects/UN/boards/51?selectedIssue=UN-3086)

### DIFF
--- a/neuro_san/client/thinking_file_message_processor.py
+++ b/neuro_san/client/thinking_file_message_processor.py
@@ -13,7 +13,6 @@
 from typing import Any
 from typing import Dict
 from typing import List
-from typing import Union
 
 import json
 import os
@@ -69,13 +68,7 @@ class ThinkingFileMessageProcessor(MessageProcessor):
         message_type: ChatMessageType = ChatMessageType.from_response_type(response_type)
         message_type_str: str = ChatMessageType.to_string(message_type)
 
-        text: Union[str, List] = response.get("text")
-        # For OpenAI and Ollama, content of AI message is a string but content from
-        # Anthropic AI message can either be a single string or a list of content blocks.
-        # If it is a list, "text" is a key of a dictionary which is the first element of
-        # the list. For more details: https://python.langchain.com/docs/integrations/chat/anthropic/#content-blocks
-        if isinstance(text, list):
-            text = text[0].get("text")
+        text: str = response.get("text")
         structure: Dict[str, Any] = response.get("structure")
 
         if structure is not None:
--- a/neuro_san/client/thinking_file_message_processor.py
+++ b/neuro_san/client/thinking_file_message_processor.py
@@ -74,7 +74,7 @@ class ThinkingFileMessageProcessor(MessageProcessor):
         # Anthropic AI message can either be a single string or a list of content blocks.
         # If it is a list, "text" is a key of a dictionary which is the first element of
         # the list. For more details: https://python.langchain.com/docs/integrations/chat/anthropic/#content-blocks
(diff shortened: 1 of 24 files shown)


neuro_san/client/thinking_file_message_processor.py
```diff
@@ -68,7 +69,13 @@ def write_message(self, response: Dict[str, Any], origin_str: str):
...
         message_type_str: str = ChatMessageType.to_string(message_type)
 
-        text: str = response.get("text")
+        text: Union[str, List] = response.get("text")
+        # For OpenAI and Ollama, content of AI message is a string but content from
+        # Anthropic AI message can either be a single string or a list of content blocks.
+        # If it is a list, "text" is a key of a dictionary which is the first element of
+        # the list. For more details: https://python.langchain.com/docs/integrations/chat/anthropic/#content-blocks
+        if isinstance(text, list):
+            text = text[0].get("text")
```
d1donlydfink: We should do this on the service side. This would make all clients easier to write if the clients never had to think about this.

Most likely places for this would be MessageUtils.get_content() and MessageUtils.convert_to_chat_message(), but there could conceivably be a few more.

What other information is in the subsequent content blocks?

d1donlydfink (changes requested):
We shouldn't have any model-specific processing in the client-side code. This will just proliferate policy in  multiple ways that could be consolidated.
See the PR comments for suggestions (and a question)

neuro_san/client/thinking_file_message_processor.py
Noravee: There is also "type". In general, the content block is like
```
[{'text': string we want,
  'type': 'text'},
 {'id': tool_id,
  'input': tool_input_dict,
  'name': tool_name,
  'type': 'tool_use'}]
```

Noravee:
### Changes
- Modify MessageUtils.get_content() and MessageUtils.convert_to_chat_message() to handle list content

### Tests
- Run website_search and prebuilt_search_tool hocon with anthropic and openai models

neuro_san/internals/messages/message_utils.py
```diff
@@ -172,6 +180,13 @@ def convert_to_chat_message(message: BaseMessage, origin: List[Dict[str, Any]] =
...
 
+        # For OpenAI and Ollama, content of AI message is a string but content from
+        # Anthropic AI message can either be a single string or a list of content blocks.
+        # If it is a list, "text" is a key of a dictionary which is the first element of
+        # the list. For more details: https://python.langchain.com/docs/integrations/chat/anthropic/#content-blocks
+        if src == "content" and isinstance(value, list):
+            value = value[0].get("text", "")
+
         if value is not None:
             chat_message[dest] = value
```
d1donlydfink, line 191: There is another method in this same file called "convert_to_base_message()".
Did you not have to do anything with that one?

Noravee: No. Once we put the valid "content" in the chat message, I don't think we have to modify anything when we convert chat message to base message.

d1donlydfink (approved):

PR #154, opened 2025-04-15
UN-3093 Properly initialize direct client session with everything a server would do.

andreidenissov-cog:
This PR puts back initialization for direct client session with everything
that is normally done on the server side.
If we are using neuro-san as a library, this needs to be done in client session.

Tested by running "web_search" and "music_nerd" agents as direct client session.


neuro_san/client/direct_agent_session_factory.py
```diff
@@ -27,9 +28,22 @@ class DirectAgentSessionFactory:
...
+        """
+        Constructor
+        """
+        manifest_restorer = RegistryManifestRestorer()
+        self.manifest_tool_registries: Dict[str, AgentToolRegistry] = manifest_restorer.restore()
+        tool_factory: ServiceToolFactoryProvider =\
+            ServiceToolFactoryProvider.get_instance()
+        for agent_name, tool_registry in self.manifest_tool_registries.items():
+            tool_factory.add_agent_tool_registry(agent_name, tool_registry)
+
```
Noravee, line 46: LGTM.

Noravee (approved):

PR #155, opened 2025-04-16
UN-3088 Beginnings of data-driven test infrastructure

d1donlydfink:
First stab at data-driven test infrastructure. Currently this is feeding a single new Beatles-oriented unit test and there is plenty more to do to continue to flesh out functionality in this regard (see hocon file for plans) and more and more tests themselves.
Eventually with proper feedback and input, this can potentially feed:
* integration tests to check for continuing valid server operation
* a higher-level tool which can begin to examine efficacy of existing agent prompting
* a scoring mechanism that could help automate fine-tuning of llms.

Special guest appearance by @donn-leaf to get the secrets right for happy unit tests.

@kxk302 and @vince-leaf are primary reviewers.


.github/workflows/tests.yml
```diff
@@ -37,3 +37,5 @@ jobs:
 
       - name: Run pytest (excluding integration tests)
         run: pytest --verbose -m "not integration" --timer-top-n 10
+        env:
+          OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
```
d1donlydfink, line 41: @donn-leaf adds the secret so that the new unit test here can pass.

tests/fixtures/music_nerd/beatles_with_history.hocon
```diff
@@ -0,0 +1,33 @@
...
+# Issued under the Academic Public License.
+#
+# You can be released from the terms, and requirements of the Academic Public
+# License by purchasing a commercial license.
+# Purchase of a commercial license is mandatory for any use of the
+# neuro-san SDK Software in commercial settings.
+#
+# END COPYRIGHT
+
+# This is file defines everthing necessary for a data-driven test.
```
d1donlydfink: First file for data-driven test of agents.
In file comments tell the story of what is going on now.
 PR comments will tell the story of where this can go.

tests/fixtures/music_nerd/beatles_with_history.hocon
```diff
@@ -0,0 +1,49 @@
...
+{
+    # Describes what agent to test against.
+    "agent": "music_nerd",
+
+    # Interactions are a series of dictionaries with request elements paired with
+    # descriptions of response checks.
+    "interactions": [
+        {
+            # This is what we send as input to streaming_chat()
+            "text": "Who did Yellow Submarine?",
```
d1donlydfink, line 22: Currently we are only specifying text, but in short order the future we will specify other aspects of the ChatMessage such as sly_data.

tests/fixtures/music_nerd/beatles_with_history.hocon
```diff
@@ -0,0 +1,49 @@
...
+
+            # The response block treats how we are going to test what comes back
+            "response": {
+
+                # Text block says how we are going to examine the text of the response.
+                "text": {
```
d1donlydfink, line 28: Here we are testing only text but soon we will also test sly_data coming back as well.

tests/fixtures/music_nerd/beatles_with_history.hocon
```diff
@@ -0,0 +1,49 @@
...
+
+                    # Keywords says we are going to look for exact matches for each
+                    # element in a list of strings.  All elements need to show up
+                    # in order to pass.
+                    "keywords": ["Beatles"]
```
d1donlydfink, line 33: The idea about the keys in this block is that they will map to AgentEvaluator implementations that
check specific aspects of the response in a particular way.

Future AgentEvaluators might include:
* not_keywords - keywords that should not be in the response
* keys/not_keys - see if particular keys do/not exist in a returned sly data dictionary
* gist - Take the output and a description of what the gist of the output should be and hand that to an agent to give a true/false value back as to whether or not the response is satisfactory - LLMs testing LLMs.
* score - Similar to "gist" above, but will produce a more expensive semantic score that could be used as a data-point to fine-tune LLMs.

tests/fixtures/music_nerd/beatles_with_history.hocon
```diff
@@ -0,0 +1,49 @@
...
+# License by purchasing a commercial license.
+# Purchase of a commercial license is mandatory for any use of the
+# neuro-san SDK Software in commercial settings.
+#
+# END COPYRIGHT
+
+# This is file defines everthing necessary for a data-driven test.
+{
+    # Describes what agent to test against.
+    "agent": "music_nerd",
```
d1donlydfink, line 15: In addition to what is here at the top level, we are considering blocks that describe different aspects of iteration,
like:
* different connections to try: direct (default now), http, grpc
* A ratio of success : attempt that is to be considered successful. This will be useful in testing troublesome agents whose prompts are not yet 100% foolproof.

tests/neuro_san/client/agent_evaluator.py
```diff
@@ -0,0 +1,29 @@
...
+
+from neuro_san.message_processing.basic_message_processor import BasicMessageProcessor
+
+
+class AgentEvaluator:
+    """
+    Interface definition for evaluating part of an agent's response
+    """
+
+    def evaluate(self, processor: BasicMessageProcessor, verify_for: Any):
```
d1donlydfink, line 22: Interface for code that describes how we are going to test an aspect of agent response.

Worth noting that a BasicMessageProcessor is what is needed for now.
This might evolve in the future.

tests/neuro_san/client/agent_evaluator_factory.py
```diff
@@ -0,0 +1,36 @@
...
+from tests.neuro_san.client.keywords_agent_evaluator import KeywordsAgentEvaluator
+
+
+class AgentEvaluatorFactory:
+    """
+    Factory that creates AgentEvaluators
+    """
+
+    @staticmethod
+    def create_evaluator(asserts: AssertForwarder, evaluation_type: str, test_key: str) -> AgentEvaluator:
```
d1donlydfink, line 23: Given a test entry for a response, this factory creates an evaluator which corresponds to keys found in the response block of an interaction.

tests/neuro_san/client/assert_forwarder.py
```diff
@@ -0,0 +1,145 @@
...
+# neuro-san SDK Software in commercial settings.
+#
+# END COPYRIGHT
+from typing import Any
+
+
+class AssertForwarder:
+    """
+    Interface definition for forwarding asserts to different test infrastructure systems.
+    """
```
d1donlydfink, line 17: A interface that allows the AgentEvaluators to assert things about what they are testing without necessarily hooking into the python UnitTest infrastructure and/or actually stopping test operation.

tests/neuro_san/client/assert_forwarder.py
```diff
@@ -0,0 +1,145 @@
...
+    # pylint: disable=invalid-name
+    def assertNotIsInstance(self, obj: Any, cls: Any, msg: str = None):
+        """
+        Assert that the obj is not an instance of the cls
+
+        :param obj: object instance comparison element
+        :param cls: Class comparison element
+        :param msg: optional string message
+        """
+        raise NotImplementedError
```
d1donlydfink, line 145: It's possible we might want to add more later on to more or less match what unittest.TestCase gives us today.

tests/neuro_san/client/data_driven_agent_test.py
```diff
@@ -0,0 +1,159 @@
...
+        :param asserts: The AssertForwarder instance to use to integrate failures
+                        back into the test system.
+        :param fixtures: Optional path to the fixtures root.
+        """
+        self.asserts: AssertForwarder = asserts
+        self.fixtures: FileOfClass = fixtures
+        if self.fixtures is None:
+            self.fixtures = FileOfClass(__file__, path_to_basis="../../fixtures")
+
+    def one_test(self, hocon_file: str):
```
d1donlydfink, line 53: Engine that drives the reading of a single hocon file for a round of testing a single agent.

tests/neuro_san/client/data_driven_agent_test.py
```diff
@@ -0,0 +1,159 @@
...
+
+            session: AgentSession = AgentSessionFactory().create_session(connection,
+                                                                         agent,
+                                                                         use_direct=use_direct,
+                                                                         metadata=metadata)
+            chat_context: Dict[str, Any] = None
+            for interaction in interactions:
+                if isinstance(session, DirectAgentSession):
+                    session.reset()
+                chat_context = self.interact(agent, session, interaction, chat_context)
```
d1donlydfink, line 95: It's quite likely these nested loops will get broken up in the future. For now it allows for @vince-leaf 's requirements of testing different connection mechanisms with the same test.

tests/neuro_san/client/data_driven_agent_test.py
```diff
@@ -0,0 +1,159 @@
...
+        :param hocon_file: The name of the hocon from the fixtures directory.
+        """
+        test_path: str = self.fixtures.get_file_in_basis(hocon_file)
+        hocon = EasyHoconPersistence()
+        test_case: Dict[str, Any] = hocon.restore(file_reference=test_path)
+        return test_case
+
+    # pylint: disable=too-many-locals
+    def interact(self, agent: str, session: AgentSession, interaction: Dict[str, Any],
+                 chat_context: Dict[str, Any]) -> Dict[str, Any]:
```
d1donlydfink, line 110: Engine that drives how to deal with a single interaction block of a test-case hocon file.

tests/neuro_san/client/keywords_agent_evaluator.py
```diff
@@ -0,0 +1,59 @@
...
+        self.asserts: AssertForwarder = asserts
+        self.test_key: str = test_key
+
+    def evaluate(self, processor: BasicMessageProcessor, verify_for: Any):
+        """
+        Evaluate the contents of the BasicMessageProcessor
+
+        :param processor: The BasicMessageProcessor to evaluate
+        :param verify_for: The data to evaluate the response against
+        """
```
d1donlydfink, line 42: Implementation of keyword search in response output.

tests/neuro_san/client/test_music_nerd.py
```diff
@@ -0,0 +1,29 @@
...
+    Tests basic functionality via the music_nerd agent.
+    """
+
+    def test_conversation_history(self):
+        """
+        Tests a basic conversation to see if chat context is being carried over correctly.
+        """
+        asserts = UnitTestAssertForwarder(self)
+        agent_test = DataDrivenAgentTest(asserts)
+        agent_test.one_test("music_nerd/beatles_with_history.hocon")
```
d1donlydfink, line 29: First unit test to use this new system.

tests/neuro_san/client/unit_test_assert_forwarder.py
```diff
@@ -0,0 +1,147 @@
...
+
+from unittest import TestCase
+
+from tests.neuro_san.client.assert_forwarder import AssertForwarder
+
+
+class UnitTestAssertForwarder(AssertForwarder):
+    """
+    AssertForwarder implemetation for python unittest.TestCase
+    """
```
d1donlydfink, line 21: Implementation of the AssertForwarder hooking into the python unit test infrastructure.

hshahrzad (dismissed):
Looks good to me.

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 # These bits are assembled from (largely) following the guidance here:
 # https://packaging.python.org/en/latest/overview/
-
+# no-op
```
kxk302: Is this needed? or can it be reverted?

tests/fixtures/music_nerd/beatles_with_history.hocon
kxk302: Nit picking, but a couple of typos here. Should be "This file defines everything necessary for a data-driven test."

tests/fixtures/music_nerd/beatles_with_history.hocon
kxk302: We might want to add "ranking" to the list of things a future AgentEvaluator evaluates. Given a prompt, and 2 responses, it should rank the responses.

tests/neuro_san/client/agent_evaluator_factory.py
```diff
@@ -0,0 +1,36 @@
...
+        Creates AgentEvaluators
+
+        :param asserts: The AssertForwarder instance to handle failures
+        :param evaluation_type: A string key describing how the evaluation will take place
+        :param test_key: The test key describing the portion of the ChatMessage contents to evaluate
+        """
+        evaluator: AgentEvaluator = None
+
+        if evaluation_type == "keywords":
+            evaluator = KeywordsAgentEvaluator(asserts, test_key)
```
kxk302, line 34: Too early, but raise a ValueError is evaluation_type not found?

kxk302 (approved):
Some comments and potentially minor changes.

tests/fixtures/music_nerd/beatles_with_history.hocon
vince-leaf: - Maybe the default would be this list of all connections. On the other side, you know what connects to the test therefore you can set it.

- A ratio of success is helpful but needs fine-tuning to get there. In some test cases, it does not need this maybe we have a boolean.

tests/fixtures/music_nerd/beatles_with_history.hocon
vince-leaf: Also, a unitest check to confirm response block is not empty or null.

tests/fixtures/music_nerd/beatles_with_history.hocon
vince-leaf: Maybe my comment above can be another option here also.

> Also, a unitest check to confirm response block is not empty or null.

tests/fixtures/music_nerd/beatles_with_history.hocon
```diff
@@ -0,0 +1,49 @@
...
+            }
+        },
+        {
+            # This next interaction block tests the ability for an agent
+            # to continue its conversation with the context from the previous
+            # interaction.
+            "text": "Where were they from?",
+            "response": {
+                "text": {
+                    "keywords": ["Liverpool"]
```
vince-leaf, line 44: Just wonder, in some cases that I have seen. The response to deal with case sensitive. Do list all here or handle it at the test case level?

tests/fixtures/music_nerd/beatles_with_history.hocon
vince-leaf: What about sometimes the response is not in a text string? Like a JSON.

tests/neuro_san/client/data_driven_agent_test.py
```diff
@@ -0,0 +1,159 @@
...
+        request: Dict[str, Any] = input_processor.formulate_chat_request(text, sly_data, chat_context, chat_filter)
+
+        # Call streaming_chat()
+        chat_responses: Generator[Dict[str, Any], None, None] = session.streaming_chat(request)
+        for chat_response in chat_responses:
+            message = chat_response.get("response", empty)
+            processor.process_message(message, chat_response.get("type"))
+
+        # Evaluate response
+        response: Dict[str, Any] = interaction.get("response", empty)
```
vince-leaf, line 142: In the case the response is null or emty, should we have an option to ignore or call the retry?

vince-leaf (approved):
LGTM on this 1st start; added some comment.

tests/fixtures/music_nerd/beatles_with_history.hocon
d1donlydfink: We don't need to test the tests.

tests/fixtures/music_nerd/beatles_with_history.hocon
d1donlydfink: We want the default to be the most usable, which is a direct connection only.
As for success ratio, it makes sense that "1/1" would be the default, implying 1 try 100% success.

pyproject.toml
d1donlydfink: reverting

tests/neuro_san/client/agent_evaluator_factory.py
d1donlydfink: Yes early.

tests/fixtures/music_nerd/beatles_with_history.hocon
d1donlydfink: Good point. There are ways to search for strings in a case-insensitive manner.
We can modify keywords implementation to do that as it's needed.

tests/neuro_san/client/data_driven_agent_test.py
d1donlydfink: We ignore the response block if it's empty.
This is effectively what happens with the existing logic anyway.

tests/fixtures/music_nerd/beatles_with_history.hocon
d1donlydfink: Also eventually I'd like to be able to eventually have finer-grained inspection of the MAXIMAL message stream at some point. That would allow us to test things like "are we getting token accounting data from the places we need?"

tests/fixtures/music_nerd/beatles_with_history.hocon
d1donlydfink: Fixed typos.

vince-leaf (approved):
LGTM

tests/fixtures/music_nerd/beatles_with_history.hocon
ofrancon: `nb_tokens` could be another one. Either a maximum number of tokens or a distribution (expected mean and standard deviation). That could be used to detect a model / prompts that become more verbose than usual or necessary.

PR #156, opened 2025-04-16
UN-3087 neuro san enable coded tools to be used across different networks

Noravee:
### Changes
- Modify ClassTool to check one level above the agent network folder
- Modify error message to include information about global CodedTool
- Add 2 global CodedTools
     - rag_global
     - website_search_global 
- Add an example agent network that uses global CodedTools
     - website_rag 
- Modify manifest.hocon accordingly

### Tests
- Run website_search that uses local CodedTools
- Run website_rag that uses global CodedTools


neuro_san/coded_tools/rag_global.py
```diff
@@ -0,0 +1,84 @@
...
+        retriever: VectorStoreRetriever = vectorstore.as_retriever()
+        page_contents: List[Document] = retriever.invoke(query)
+
+        rag_str: str = ""
+        for content in page_contents:
+            rag_str += content.page_content + "\n\n"
+
+        return rag_str
+
+    async def async_invoke(self, args: Dict[str, Any], sly_data: Dict[str, Any]) -> Union[Dict[str, Any], str]:
```
ofrancon: I guess adding the web pages to a vector store and querying it can be a bit slow, so it would be nice to have an async implementation instead.
Can be done later.
Would be nice to have a unit test too. Can be added later too.

neuro_san/coded_tools/website_search_global.py
```diff
@@ -0,0 +1,84 @@
...
+        the_links = self.search_web(search_terms, self.top_n)
+        links_str = ""
+        for index, the_link in enumerate(the_links, start=1):
+            links_str += f"{index}. {the_link} ; "
+            logger.info("%s. %s", str(index), str(the_link))
+        logger.info(">>>>>>>>>>>>>>>>>>>DONE !!!>>>>>>>>>>>>>>>>>>")
+        return links_str
+
+    async def async_invoke(self, args: Dict[str, Any], sly_data: Dict[str, Any]) -> Union[Dict[str, Any], str]:
+        raise NotImplementedError
```
ofrancon: Same here: would be nice it was async and if it had unit tests.

ofrancon (approved):
works for me!
Consider:
- making the tools async
- adding unit tests

neuro_san/coded_tools/rag_global.py
```diff
@@ -0,0 +1,84 @@
...
+                Keys expected for this implementation are:
+                    None
+
+        :return:
+            In case of successful execution:
+                The output string from RAG.
+            otherwise:
+                a text string an error message in the format:
+                "Error: <error message>"
+        """
```
d1donlydfink: Please add some more comments in here as to what different sections are doing.
Also: Please document which calls are synchronous to answer the question: why can't this be in async_invoke() instead?  Further work on moving those to be asynchronous could help move this to a happier place.

neuro_san/coded_tools/website_search_global.py
d1donlydfink: Yes. Add comments to the methods above that are explicitly synchronous so later we can work on them being asynchronous.

neuro_san/internals/graph/tools/class_tool.py
```diff
@@ -107,31 +107,43 @@ async def build(self) -> List[Any]:
...
+                agent_name: str = self.factory.get_name_from_spec(self.agent_tool_spec)
+                message = f"""
+    Could not find class "{class_name}"
+    in module "{module_name}"
+    under AGENT_TOOL_PATH "{agent_tool_path}"
+    for the agent called "{agent_name}"
+    in the agent network "{agent_network}".
+
+    Check these things:
+    1.  Is there a typo in your AGENT_TOOL_PATH? Your current AGENT_TOOL_PATH is {parent_path}.
```
d1donlydfink: Seems as if you are giving false information about what the value of AGENT_TOOL_PATH is here by
producing parent_path. That's not the raw value.  Also we try to specifically avoid giving information
about environment variable values because we get dinged for that in security scans.

neuro_san/internals/graph/tools/class_tool.py
```diff
@@ -107,31 +107,43 @@ async def build(self) -> List[Any]:
...
+    2.  Expected to find a specific CodedTool for the given agent network in:
+        <AGENT_TOOL_PATH>/<agent_network>/<coded_tool_name>.py
+        Global CodedTools (shared across networks) should be located at:
+        <AGENT_TOOL_PATH>/<coded_tool_name>.py
```
d1donlydfink, line 136: FWIW: Eventually when there are libraries of CodedTools available we will just want anything that lives in PYTHONPATH to be accessible.

d1donlydfink (changes requested):
See PR comments

neuro_san/internals/graph/tools/class_tool.py
Noravee: The parts about parent_path and environment variable have been removed.

neuro_san/coded_tools/rag_global.py
deepsaia: Agree with Olivier and I would suggest segregating the two: `add to db` and `query from the db` as two activities., where `add to db` can be handled separately to save time.
For later.

neuro_san/coded_tools/website_search_global.py
Noravee: Add comments that;
Currently `duckduckgo_search` does not support async operations.
However, `asyncio.to_thread method` maybe useful here
https://docs.python.org/3/library/asyncio-task.html

I can adding unit tests in a later PR.

neuro_san/coded_tools/rag_global.py
Noravee: rag_global is now async.

unit test will be add in later PR.

Noravee:
### Changes
- Remove comment about env var in error message of ClassTool.
- Add a comment in website_search_global that duckduckgo does not support async operations.
- Make rag_global async and add more comments.

### Tests
- Run website_rag with gpt-4o.

neuro_san/coded_tools/rag_global.py
Noravee: rag_global is now async.

neuro_san/internals/graph/tools/class_tool.py
Noravee: Ok.

neuro_san/coded_tools/rag_global.py
```diff
@@ -0,0 +1,130 @@
...
+from langchain_community.vectorstores import InMemoryVectorStore
+from langchain_core.documents import Document
+from langchain_core.vectorstores.base import VectorStoreRetriever
+from langchain_openai import OpenAIEmbeddings
+from langchain_text_splitters import RecursiveCharacterTextSplitter
+
+from neuro_san.interfaces.coded_tool import CodedTool
+
+
+class RAG(CodedTool):
```
d1donlydfink: 2 things:
* Filename should always reflect the class name.
* The convention is to only capitalize the first letter of an acronym, otherwise it looks like a constant.

So for this class one way to conform would be to rename the class RagGlobal.
Or you could rename the class to be Rag and the file to be rag.py

neuro_san/coded_tools/website_search_global.py
```diff
@@ -0,0 +1,100 @@
...
+from typing import Union
+
+import logging
+
+from duckduckgo_search import DDGS
+
+from neuro_san.interfaces.coded_tool import CodedTool
+
+
+class WebsiteSearch(CodedTool):
```
d1donlydfink: Same deal here. This should be WebsiteSearchGlobal or rename the file.

d1donlydfink (changes requested):
1. Conform to class/filename consistency
2. Only capitalize first letter of acronym.
3. Update the Rag class comment to reflect what it actually does.  If you were your own first reviewer, you would have caught this ;)

neuro_san/coded_tools/rag_global.py
```diff
@@ -0,0 +1,130 @@
...
+    """
+    CodedTool implementation which provides a way to utilize different websites' search feature
+    """
```
d1donlydfink: Update the class comment to reflect what this class actually does.
It will be worth spelling out wha the RAG acronym actually means.

neuro_san/coded_tools/website_search_global.py
Noravee: The classes were rename as suggested.

neuro_san/coded_tools/rag_global.py
Noravee: The class were renamed as suggested.

neuro_san/coded_tools/rag_global.py
Noravee: I've updated the class docstring to include more details and spelled out the RAG acronym.

FWIW, I did review the docstring earlier, but I see now that the description could’ve better captured the fact that this class performs RAG after using a website’s search—similar to how TavilySearch operates. Appreciate the nudge to clarify that more explicitly.

neuro_san/coded_tools/rag_global.py
```diff
@@ -0,0 +1,138 @@
...
+from langchain_community.vectorstores import InMemoryVectorStore
+from langchain_core.documents import Document
+from langchain_core.vectorstores.base import VectorStoreRetriever
+from langchain_openai import OpenAIEmbeddings
+from langchain_text_splitters import RecursiveCharacterTextSplitter
+
+from neuro_san.interfaces.coded_tool import CodedTool
+
+
+class RagGlobal(CodedTool):
```
Noravee: Change class name to match the convention.

neuro_san/coded_tools/rag_global.py
```diff
@@ -0,0 +1,138 @@
...
+    A CodedTool implementation that performs Retrieval-Augmented Generation (RAG)
+    using URLs obtained from a website search.
+
+    This class assumes that a prior search has been conducted to identify relevant URLs.
+    It then retrieves the content from those URLs and applies a language model to
+    generate responses based on the retrieved information.
+
+    This is useful in workflows where external search tools provide candidate links,
+    and RAG is applied afterward to synthesize a meaningful answer from the linked content.
+    """
```
Noravee: Modify the docstring to provide more explanation.

neuro_san/coded_tools/website_search_global.py
```diff
@@ -0,0 +1,100 @@
...
+from typing import Union
+
+import logging
+
+from duckduckgo_search import DDGS
+
+from neuro_san.interfaces.coded_tool import CodedTool
+
+
+class WebsiteSearchGlobal(CodedTool):
```
Noravee: Change class name to match the convention

Noravee:
### Changes
- Change class names to match the module name
- Modify docstring for RAG tool

### Test
- Run `website_rag` and `website_search`

ofrancon (approved):
lgtm

d1donlydfink (changes requested):
Not sure why we need "Global" in the title of these classes
There is nothing really global about them.

Noravee:
### Changes
- Rename `rag_global` to `rag` and `website_search` to `web_search` and modify `website_rag.hocon` accordingly

### Tests
- Run `prebuilt_search_tool`, `website_rag`, and `website_search` 

### Notes
- The module name must not match the name of an existing agent network; otherwise, Python cannot correctly locate the module. Therefore, `website_search` cannot be used because a network with that name already exists.

d1donlydfink (approved):

PR #157, opened 2025-04-16
UN-3086 neuro san anthropic models key error agent tool result

Noravee:
### Changes
- Modify LangChainRunContext.parse_tool_output to use AIMessage instead of AgentToolResultMessage if llm is Anthropic model.

### Tests
- Run website_search.hocon and prebuilt_search_tool.hocon with OpenAI and Anthropic models.


ofrancon (approved):
looks ok to me

neuro_san/internals/run_context/langchain/langchain_run_context.py
```diff
@@ -601,8 +602,16 @@ def parse_tool_output(self, tool_output: Dict[str, Any]) -> List[BaseMessage]:
...
+
+        # Anthropic models check message type with message.type and only allow
+        # human and ai messages, thus we have to parse tool output to AIMessage.
+        if isinstance(self.llm, ChatAnthropic):
+            tool_message = AIMessage(content=tool_result_dict.get("content"))
+        else:
+            # OpenAi and Ollama models do not check message type with message.type
+            # but check attribute of the message so we can use AgentToolResultMessage.
+            tool_message = AgentToolResultMessage(content=tool_result_dict.get("content"),
+                                                  tool_result_origin=tool_output.get("origin"))
```
d1donlydfink: Can you explain more here. What problem are you trying to solve?

AgentToolResultMessage is not something that gets into the chat history that goes to the llm, while AIMessage is.
Also, the AI message will be missing the origin of the tool returning the result, so chat streams that have (say) 3 tools returning results will not know where the result is coming from.

If the problem is that ToolResultMessage is somehow getting into the chat stream for anthropic models, this is probably not the place to make that repair.  It should be in the place that filters the chat messages.

Is the problem that Anthropic models are not getting any results from any tools?
If so, then perhaps the right thing to do is to put *both* AI messages and AgentToolResultMessages in the return list here for anthropic models. But if that is the case something is probably broken _within the langchain agent logic_ for how anthropic handles tool results.

The single AgentToolResultMessage that was reported here was strictly for neuro-san level reporting.

d1donlydfink (changes requested):
See questions in PR.

neuro_san/internals/run_context/langchain/langchain_run_context.py
Noravee: Here’s what I think is happening:

The `tool_message` in this context is expected to be a Langchain `ToolMessage`, as noted in the comment on lines 601–604:
```
# You might expect that this should be a ToolMessage, but making that
# kind of conversion at this point runs into problems with OpenAI models
# that process them.  So, to make things continue to work, report the
# content as an AI message - as if the bot came up with the answer itself.
```

However, Langchain checks for a `tool_call_id` in `ToolMessage`, and since we do not collect this ID, using a `ToolMessage` causes issues. Additionally, we need to capture the origin of the tool output — this is why `AgentToolResultMessage` is used, which subclasses `AIMessage` and adds a `tool_result_origin` field.

These AgentToolResultMessage instances are passed into the chat history via:
```
 async def submit_tool_outputs(self, run: Run, tool_outputs: List[Any]) -> Run:
        """
        :param run: The run handling the execution of the assistant
        :param tool_outputs: The tool outputs to submit
        :return: A potentially updated run handle
        """
        if tool_outputs is not None and len(tool_outputs) > 0:
            for tool_output in tool_outputs:
                tool_messages: List[BaseMessage] = self.parse_tool_output(tool_output)
                if tool_messages is not None:
                    for tool_message in tool_messages:
                        # Chat history is updated in write_message()
                        await self.journal.write_message(tool_message)
```
I confirmed that `AgentToolResultMessage` successfully reaches both ChatOpenAI and ChatAnthropic.

- For `ChatOpenAI` and `ChatOllama`, this works as expected because they use `isinstance(message, AIMessage)`, and `AgentToolResultMessage` is a subclass of `AIMessage`.

For `ChatAnthropic`, this becomes problematic because it uses `message.type`, which returns `agent_tool_result` for `AgentToolResultMessage`. Since that type is not recognized by Anthropic's handler, I fall back to using `AIMessage` for Anthropic models — though this means we lose `tool_result_origin` for those cases. On that note, it is possible do something like `AIMessage(content='content', additional_kwargs={'tool_result_origin': 'Agent_a'})`

If this isn't the right place to apply this workaround, or you’d prefer a different architectural approach, I’m happy to adjust.

d1donlydfink: Ah. I forgot about that inheritance from AIMessage.  It does indeed make a difference for reporting back to the client.  Let me poke around a bit to see if I can come up with something that preserves the origin.

d1donlydfink: I think maybe a better thing to do would be to do a conversion of the AgentToolResultMessage to an AI message in the place where OriginatingJournal.write_message() adds the message to the chat history here:
https://github.com/leaf-ai/neuro-san/blob/main/neuro_san/internals/journals/originating_journal.py#L68

We'd still want the original AgentToolResultMessage to be propagated back to the wrapped_journal later on so as to give the client the correct overall picture.

Such a conversion there should be fine for all llms with no loss of information to the client.

d1donlydfink (changes requested):
See PR comments for suggestions.

neuro_san/internals/journals/originating_journal.py
```diff
@@ -65,7 +67,27 @@ async def write_message(self, message: BaseMessage, origin: List[Dict[str, Any]]
...
+            #   to include additional metadata like `tool_result_origin` when supported.
+            #
+            # To avoid problem with any other LLMs, convert "AgentToolResultMessage" to "AIMessage"
+            # when appending it chathistory but allow it to be written in the journal as is to
+            # to maintain the information on tool origin.
+            if isinstance(message, AgentToolResultMessage):
+                ai_message = AIMessage(content=message.content)
+                self.chat_history.append(ai_message)
+            else:
+                self.chat_history.append(message)
```
Noravee, line 90: Convert `AgentToolResultMessage` to `AIMessage` before appending to chat history.

neuro_san/internals/run_context/langchain/langchain_run_context.py
Noravee: Done.

Noravee:
### Changes
- Revert langchain_run_context back 
- Modify OriginatingJournal.write_message to convert AgentToolResultMessage to AIMessage before appending to chat history

### Tests
- Run website_search.hocon and prebuilt_search_tool.hocon with OpenAI and Anthropic models.

d1donlydfink (approved):

PR #158, opened 2025-04-16
Consume internal dependencies from pypi and prepare for pypi publishing

donn-leaf:
Now that leaf-common and leaf-server-common live in pypi, consume those dependencies from pypi. In addition, add the necessary github actions to publish neuro-san to pypi.


.github/workflows/codeql.yml
donn-leaf, line 1: This matches leaf-common and leaf-server-common

d1donlydfink (approved):

d1donlydfink:
I have a PR in progress which will adjust the sample Dockerfiles

PR #164, opened 2025-04-21
UN-3112 Switch to different OpenAPI generator plug-in.

andreidenissov-cog:
This PR switches OpenAPI spec documents generator we use from "gnostic" tool to another one
being developed by "kollalabs" team in their own forked repo.
Difference between the two: "kollalabs" tool generates proper enumeration types, with string values and all,
so we don't have to use plain integer values for enumeration types.


neuro_san/api/scripts/do_generate.sh
```diff
@@ -146,6 +146,7 @@ done
 echo "generating OpenAPI specification for ${TOP_LEVEL}/${GENERATED_DIR}/agent.proto"
 # shellcheck disable=SC2086    # PROTO_PATH is compilation of cmd line args
 python -m grpc_tools.protoc ${PROTO_PATH} \
+    --plugin=protoc-gen-openapi=${HOME}/go/bin/protoc-gen-openapi-enums \
     --openapi_out="${TOP_LEVEL}"/${GENERATED_DIR} \
```
andreidenissov-cog, line 150: Directly specify plug-in executable we are going to use.

d1donlydfink (dismissed):
Cool. @dsargent will be pleased.

dsargent (approved):

dsargent:
Can you give a sample of how it will look?

dsargent:
Thanks, this will make the UI code (and anyone else who writes a client for Neuro-san) cleaner, more readable and safer.

andreidenissov-cog:
> Can you give a sample of how it will look?

  Yes. Take a look at neuro_san/api/grpc/agent_service.json (in this PR)
which is regenerated by new plug-in.

PR #168, opened 2025-04-22
UN-3080 Add "OPTIONS" handlers to support CORS requests

dsargent:
CORS requests from the browser were failing (I think) due to lack of `OPTIONS` support on these handlers.

Could only do limited "does it break everything?" testing because CORS isn't really a thing on `localhost`.

I'm a bit perplexed why the `list` endpoint handler works _without_ the `OPTIONS` support, but maybe it falls into what CORS considers a [simple request](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CORS#simple_requests) so is exempt from CORS 🤷🏻


swensel (approved):
LGTM. Thank you @dsargent

d1donlydfink (approved):
LGTM I guess but @andreidenissov-cog should have the last word.

neuro_san/http_sidecar/handlers/connectivity_handler.py
```diff
@@ -49,3 +49,9 @@ async def get(self, agent_name: str):
         finally:
             await self.flush()
             self.logger.info(metadata, "Finish GET %s/connectivity request", agent_name)
+
+    async def options(self):
+        """
+        Implementation of OPTIONS request handler for streaming chat API call.
+        """
+        await self.flush()
```
d1donlydfink, line 57: FWIW: This is the same thing that happens with streaming_chat()

d1donlydfink: Do these also need to happen in concierge_handler.py, health_check_handler.py and openapi_publish_handler.py?

dsargent: I stole it from `streaming_chat`. 

The only time we care about CORS is with browser-to-service interactions and the browser app doesn't hit those other endpoint you mentioned. I guess it might some day? Should the `OPTIONS` function be in the base handler?

dsargent: `concierge_handler.py` is the `/list` endpoint I mentioned. Somehow it's working fine without `OPTIONS` support - maybe due to the "simple request" rule I mentioned in the description.

PR #170, opened 2025-04-23
UN-3115 success_ratio in data-driven agent tests

d1donlydfink:
Tests can now specify a top level "success_ratio".  This is of the form "A/B" where B iterations are performed and at least A iterations need to pass in order to be successful. By default this ratio is "1/1".

Eventually this ratio can be used in a development scenario where B iterations are performed and we collect and analyze the reasons why the test failed however many times it did. That is for a later PR, but on its own this allows for the imperfections inherent in LLM-based agents to not necessarily fail when they are Good Enough for testing purposes.

### DIFF
--- a/tests/framework/driver/assert_capture.py
+++ b/tests/framework/driver/assert_capture.py
@@ -1,0 +1,200 @@
+# Copyright (C) 2023-2025 Cognizant Digital Business, Evolutionary AI.
+# All Rights Reserved.
+# Issued under the Academic Public License.
+#
+# You can be released from the terms, and requirements of the Academic Public
+# License by purchasing a commercial license.
+# Purchase of a commercial license is mandatory for any use of the
+# neuro-san SDK Software in commercial settings.
+#
+# END COPYRIGHT
+from typing import Any
+from typing import List
+
+from tests.framework.interfaces.assert_forwarder import AssertForwarder
+
+
+class AssertCapture(AssertForwarder):
+    """
+    AssertForwarder implementation that wraps another AssertForwarder and
+    captures any exceptions thrown from it.
+    """
+
+    def __init__(self, basis: AssertForwarder):
+        """
+        Constructor
+
+        :param basis: AssertForwarder
+        """
+        self.basis: AssertForwarder = basis
+        self.asserts: List[AssertionError] = []
+
+    def get_asserts(self) -> List[AssertionError]:
+        """
+        :return: The list of asserts captured from the basis
+        """
+        return self.asserts
+
+    # pylint: disable=invalid-name
+    def assertEqual(self, first: Any, second: Any, msg: str = None):
+        """
+        Assert that the first is equal to the second
+
+        :param first: First comparison element
+        :param second: Second comparison element
+        :param msg: optional string message
+        """
+        try:
+            self.basis.assertEqual(first, second, msg)
+        except AssertionError as exception:
+            self.asserts.append(exception)
+
+    # pylint: disable=invalid-name
+    def assertNotEqual(self, first: Any, second: Any, msg: str = None):
+        """
+        Assert that the first is not equal to the second
+
+        :param first: First comparison element
+        :param second: Second comparison element
+        :param msg: optional string message
+        """
+        try:
+            self.basis.assertNotEqual(first, second, msg)
+        except AssertionError as exception:
+            self.asserts.append(exception)
+
+    # pylint: disable=invalid-name
+    def assertTrue(self, expr: Any, msg: str = None):
+        """
+        Assert that the expression is true
+
+        :param expr: Expression to test
+        :param msg: optional string message
+        """
+        try:
+            self.basis.assertTrue(expr, msg)
+        except AssertionError as exception:
+            self.asserts.append(exception)
+
+    # pylint: disable=invalid-name
+    def assertFalse(self, expr: Any, msg: str = None):
+        """
+        Assert that the expression is false
+
+        :param expr: Expression to test
+        :param msg: optional string message
+        """
+        try:
+            self.basis.assertFalse(expr, msg)
+        except AssertionError as exception:
+            self.asserts.append(exception)
+
+    # pylint: disable=invalid-name
+    def assertIs(self, first: Any, second: Any, msg: str = None):
+        """
+        Assert that the first and second are the same object
+
+        :param first: First comparison element
+        :param second: Second comparison element
+        :param msg: optional string message
+        """
+        try:
+            self.basis.assertIs(first, second, msg)
+        except AssertionError as exception:
+            self.asserts.append(exception)
+
+    # pylint: disable=invalid-name
+    def assertIsNot(self, first: Any, second: Any, msg: str = None):
+        """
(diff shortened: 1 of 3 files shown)


tests/framework/driver/assert_capture.py
```diff
@@ -0,0 +1,200 @@
...
+from typing import List
+
+from tests.framework.interfaces.assert_forwarder import AssertForwarder
+
+
+class AssertCapture(AssertForwarder):
+    """
+    AssertForwarder implementation that wraps another AssertForwarder and
+    captures any exceptions thrown from it.
+    """
```
d1donlydfink, line 21: A class which captures AssertionErrors for later rumination.

tests/framework/driver/assert_capture.py
```diff
@@ -0,0 +1,200 @@
...
+        Assert that the first is equal to the second
+
+        :param first: First comparison element
+        :param second: Second comparison element
+        :param msg: optional string message
+        """
+        try:
+            self.basis.assertEqual(first, second, msg)
+        except AssertionError as exception:
+            self.asserts.append(exception)
```
d1donlydfink, line 50: All these methods are basically the same w/rt how the exception is caught from its basis and how it is sequestered for later use.

tests/framework/driver/data_driven_agent_test_driver.py
```diff
@@ -60,9 +61,75 @@ def one_test(self, hocon_file: str):
         """
         test_case: Dict[str, Any] = self.parse_hocon_test_case(hocon_file)
 
+        agent: str = test_case.get("agent")
+        self.asserts_basis.assertIsNotNone(agent)
+
+        # Get the success ration
+        success_ratio: str = test_case.get("success_ratio", "1/1")
+        self.asserts_basis.assertIn("/", success_ratio)
```
d1donlydfink, line 69: Add logic to detect the success ratio.

tests/framework/driver/data_driven_agent_test_driver.py
```diff
@@ -60,9 +61,75 @@ def one_test(self, hocon_file: str):
...
+
+        # Find the integer components of the success ratio
+        success_split: List[str] = success_ratio.split("/")
+        num_need_success: int = int(success_split[0])
+        num_iterations: int = int(success_split[-1])
+
+        # Put some bounds on the number of iterations
+        num_iterations = max(1, num_iterations)
+        num_need_success = min(num_need_success, num_iterations)
```
d1donlydfink, line 78: Parse the numerator and denominator and put some bounds on them

tests/framework/driver/data_driven_agent_test_driver.py
```diff
@@ -60,9 +61,75 @@ def one_test(self, hocon_file: str):
...
+
+        # Capture asserts for each iteration
+        iteration_asserts: List[AssertCapture] = []
+
+        # Loop through each iteration, capturing any asserts.
+        num_successful: int = 0
+        for index in range(num_iterations):
```
d1donlydfink, line 85: Iterate as per the denominator.
Conceivably we could parallelize this loop. That is an optimization for later.

tests/framework/driver/data_driven_agent_test_driver.py
```diff
@@ -60,9 +61,75 @@ def one_test(self, hocon_file: str):
...
+
+            num_successful += 1
+            if num_successful == num_need_success:
+                # Don't do more tests than we actually need to
+                break
+
+        # Don't bother reporting any asserts if we have met our success ratio.
+        # Return early to pass this test.
+        if num_successful >= num_need_success:
+            return
```
d1donlydfink, line 110: Do not fail when we are successful enough as per the success_ratio.

tests/framework/driver/data_driven_agent_test_driver.py
```diff
@@ -60,9 +61,75 @@ def one_test(self, hocon_file: str):
...
+        # Find the first assert that fails and use it to fail this test
+        for assert_capture in iteration_asserts:
+            asserts: List[AssertionError] = assert_capture.get_asserts()
+            if len(asserts) > 0:
+                one_assert: AssertionError = asserts[0]
+                message: str = f"""
+{num_successful} of {num_iterations} iterations on agent {agent} were successful.
+Need at least {num_need_success} to consider {hocon_file} test to be successful.
+"""
+                raise AssertionError(message) from one_assert
```
d1donlydfink, line 121: When we do fail, raise an exception explaining the failure w/rt the success ratio.
Also be sure to raise the first assert that failed as a single representative sample of how an agent fails.

tests/framework/driver/data_driven_agent_test_driver.py
```diff
@@ -60,9 +61,75 @@ def one_test(self, hocon_file: str):
...
+
+    def one_iteration(self, test_case: Dict[str, Any], asserts: AssertForwarder):
+        """
+        Perform a single iteration on the test case.
+
+        :param test_case: The dictionary describing the data-driven test case
+        :param asserts: The AssertForwarder to send asserts to.
+        """
```
d1donlydfink, line 129: Note that the asserts are now passed in instead of being taken from the member variable.
That member variable now sticks around as the basis for the AssertCapture.

tests/framework/driver/data_driven_agent_test_driver.py
```diff
@@ -94,7 +161,7 @@ def one_test(self, hocon_file: str):
             for interaction in interactions:
                 if isinstance(session, DirectAgentSession):
                     session.reset()
-                chat_context = self.interact(agent, session, interaction, chat_context)
+                chat_context = self.interact(agent, session, interaction, chat_context, asserts)
```
d1donlydfink, line 164: Pass along assert capture to lower-level functions.

tests/framework/evaluators/gist_agent_evaluator.py
```diff
@@ -35,10 +35,23 @@ def test_one(self, verify_value: Any, test_value: Any):
...
+
+        message: str = f"""
+text_sample unexpectedly {did_str} match acceptance criteria.
+
+text_sample:
+{test_value}
+
+acceptance_criteria:
+{verify_value}
+"""
```
d1donlydfink, line 55: Add extra information to the GistAgentEvaluator so there is more context as to how the test failed other than an uninformative False != True

tests/framework/evaluators/gist_agent_evaluator.py
```diff
@@ -35,10 +35,23 @@ def test_one(self, verify_value: Any, test_value: Any):
         pass_fail: bool = self.ask_llm(acceptance_criteria=str(verify_value),
                                        text_sample=str(test_value))
 
+        did_str: str = "didn't"
```
kxk302, line 43: A bit unclear why did_str is set s such. Maybe you can add a brief comment?

kxk302 (changes requested):

tests/framework/evaluators/gist_agent_evaluator.py
d1donlydfink: Fair!
Done.

PR #171, opened 2025-04-24
UN-2911 implement base_tool extension_cont

Noravee:
Follow up on #131 

### Changes
- Move `test_base_tool_factory` to `tests/internals/run_context`
- Remove the up-and-back path_to_basis in FileOfClass for `BaseToolInfoRestorer`

### Tests
- Run `prebuilt_search_tool` and `website_search`

### Notes
 - Exception messaging for when dynamically instantiating classes in `BaseToolFactory` is handled by
 ```python
def _check_invalid_args(self, method_class: Union[Type, MethodType], args: Dict[str, Any]):
        """
        Check for invalid arguments in class or method
        :param method_class: Class or method to check for the invalid arguments
        :param args: Arguments to check
        """
        class_args_set: Set[str] = set(signature(method_class).parameters.keys())
        args_set: Set[str] = set(args.keys())
        invalid_args: Set[str] = args_set - class_args_set

        if invalid_args:
            raise ValueError(
                f"Arguments {invalid_args} for '{method_class.__name__}' do not match any attributes "
                "of the class or any arguments of the method."
            )
```
for both nested arguments and tool instantiation.
- The above method is already in #131.

### CODE REVIEW

File: neuro_san/internals/run_context/langchain/base_tool_info_restorer.py [hunk 1]
```diff
@@ -43,7 +43,7 @@ def restore(self, file_reference: str = None):
 
         if file_reference is None or len(file_reference) == 0:
             # Read from the default
-            file_of_class = FileOfClass(__file__, "../../../../neuro_san/internals/run_context/langchain")
+            file_of_class = FileOfClass(__file__, ".")
             use_file = file_of_class.get_file_in_basis("base_tool_info.hocon")
```
Noravee, line 47: Since `base_tool_info_restorer.py` and `base_tool_info.hocon` are in the same folder, just look around.

d1donlydfink (approved):

PR #174, opened 2025-04-24
StreamingInputProcessor: update sly_data in the state

ofrancon:
In order to keep the "state" going from one interaction to another and continue conversations, the `sly_data` has to be updated.

### DIFF
--- a/neuro_san/client/agent_cli.py
+++ b/neuro_san/client/agent_cli.py
@@ -166,8 +166,8 @@ Some suggestions:
 
             print(f"\nResponse from {state.get('origin_str')}:")
             print(f"{state.get('last_chat_response')}")
-            if state.get("returned_sly_data") is not None:
-                pretty_sly: str = json.dumps(state.get('returned_sly_data'), indent=4, sort_keys=4)
+            if state.get("sly_data") is not None:
+                pretty_sly: str = json.dumps(state.get('sly_data'), indent=4, sort_keys=True)
                 print(f"Returned sly_data is: {pretty_sly}")
 
             if self.args.response_output_file is not None:
--- a/neuro_san/client/streaming_input_processor.py
+++ b/neuro_san/client/streaming_input_processor.py
@@ -89,7 +89,7 @@ class StreamingInputProcessor:
             # Update the state if there is something to update it with
             chat_context = self.processor.get_chat_context()
             last_chat_response = self.processor.get_answer()
-            returned_sly_data: Dict[str, Any] = self.processor.get_sly_data()
+            sly_data: Dict[str, Any] = self.processor.get_sly_data()
             origin_str = Origination.get_full_name_from_origin(self.processor.get_answer_origin())
 
         if origin_str is None or len(origin_str) == 0:
@@ -100,8 +100,7 @@ class StreamingInputProcessor:
             "num_input": num_input + 1,
             "last_chat_response": last_chat_response,
             "user_input": None,
-            "sly_data": None,
-            "returned_sly_data": returned_sly_data,
+            "sly_data": sly_data,
             "origin_str": origin_str
         }
         return_state.update(update)


neuro_san/client/streaming_input_processor.py
```diff
@@ -100,8 +100,7 @@ def process_once(self, state: Dict[str, Any]) -> Dict[str, Any]:
             "num_input": num_input + 1,
             "last_chat_response": last_chat_response,
             "user_input": None,
-            "sly_data": None,
-            "returned_sly_data": returned_sly_data,
+            "sly_data": sly_data,
```
ofrancon, line 112: Not sure why `sly_data` would have to be reset to `None`: if the client has set something there, don't we want to keep it for next message?
And if it's actually been updated by the server, don't we want to keep that update for next message?

neuro_san/client/agent_cli.py
```diff
@@ -166,8 +166,8 @@ def loop_until_done(self, user_input: str, sly_data: Dict[str, Any],
 
             print(f"\nResponse from {state.get('origin_str')}:")
             print(f"{state.get('last_chat_response')}")
-            if state.get("returned_sly_data") is not None:
-                pretty_sly: str = json.dumps(state.get('returned_sly_data'), indent=4, sort_keys=4)
+            if state.get("sly_data") is not None:
+                pretty_sly: str = json.dumps(state.get('sly_data'), indent=4, sort_keys=True)
```
ofrancon, line 170: Assuming `sly_data` is updated by the server now.
See change in `neuro_san/client/streaming_input_processor.py`

neuro_san/client/streaming_input_processor.py
```diff
@@ -89,7 +89,7 @@ def process_once(self, state: Dict[str, Any]) -> Dict[str, Any]:
             # Update the state if there is something to update it with
             chat_context = self.processor.get_chat_context()
             last_chat_response = self.processor.get_answer()
-            returned_sly_data: Dict[str, Any] = self.processor.get_sly_data()
+            sly_data: Dict[str, Any] = self.processor.get_sly_data()
```
d1donlydfink: sly_data coming back might not be the same thing as sly_data going to the agent.
You might want to update() the client's existing sly_data dictionary instead

neuro_san/client/streaming_input_processor.py
d1donlydfink: This is reasonable thinking, but ... (see comment below)

neuro_san/client/streaming_input_processor.py
ofrancon: Right, I'll update instead of replace, just in case the agent does something funky

neuro_san/client/streaming_input_processor.py
```diff
@@ -92,6 +94,13 @@ def process_once(self, state: Dict[str, Any]) -> Dict[str, Any]:
             returned_sly_data: Dict[str, Any] = self.processor.get_sly_data()
             origin_str = Origination.get_full_name_from_origin(self.processor.get_answer_origin())
 
+        # Update the sly_data if new sly_data was returned
+        if returned_sly_data is not None:
+            if sly_data is not None:
+                sly_data.update(returned_sly_data)
```
ofrancon, line 100: Updating the sly_data if we had some...

neuro_san/client/streaming_input_processor.py
```diff
@@ -92,6 +94,13 @@ def process_once(self, state: Dict[str, Any]) -> Dict[str, Any]:
...
+            else:
+                sly_data = returned_sly_data.copy()
```
ofrancon, line 102: ... or copying the returned sly_data dictionary if we didn't have sly_data before.

d1donlydfink (approved):

PR #175, opened 2025-04-24
Enable manual triggering for integration test

donn-leaf:
This PR puts in the necessary scheduling directive to allow us to manually trigger the integration test when needed.


andreidenissov-cog (approved):
LGTM.

PR #182, opened 2025-04-28
UN-2778 Dynamic agent endpoints

andreidenissov-cog:
This PR implements dynamic collection of agents supported by running neuro-san server.
Overall architecture looks like this:

- singleton instance of ServiceToolFactoryProvider keeps a dynamic table of agent names mapped to their AgentToolFactory instances. This is our "source of truth" representing actual state of the service;
- ServiceToolFactoryProvider provides methods to add and remove agent+AgentToolFactory pairs to/from the state;
- singleton instance of DynamicAgentRouter works as universal gRPC requests handler for any agent+method pair we may throw at the server. Again, it keeps an internal dynamic table of agent names which are allowed for serving requests;
- there are state listeners which can be registered with ServiceToolFactoryProvider to be notified about state changes, and instance of AgentServer works as such a listener (the only one for now);
- on state change notification, AgentServer calls DynamicAgentRouter methods to update its own routing table.
- for Http server, situation is different, because it runs in separate process space and cannot watch  ServiceToolFactoryProvider directly. So it queries host gRPC endpoint before executing any agent-related request for a set of available agents and routs request only if agent qualifies as "available". There are obviously performance implications for this way of doing things, but for streaming chat they are negligible, and simple requests are fast anyway.

Tested: by putting together simple file directory watchdog, so that agent.hocon files being added or removed there
result in compiled AgentToolFactory instances being added or removed to/from our ServiceToolFactoryProvider.
Service endpoints are then checked manually to be working or not working, both on gRPC and http APIs.

### DIFF
--- a/neuro_san/http_sidecar/handlers/base_request_handler.py
+++ b/neuro_san/http_sidecar/handlers/base_request_handler.py
@@ -141,13 +141,17 @@ class BaseRequestHandler(RequestHandler):
             data: Dict[str, Any] = {}
             grpc_session: ConciergeSession = self.get_concierge_grpc_session(metadata)
             agents_dict: Dict[str, Any] = grpc_session.list(data)
-            self.agents_updater.update_agents(agents_dict)
+            agents_list = agents_dict.get("agents", [])
+            agents_names: List[str] = []
+            for agent_dict in agents_list:
+                agents_names.append(agent_dict["agent_name"])
+
+            print(f">>>>>>>>>>>> AGENTS: {agents_names}")
+
+            self.agents_updater.update_agents(agents_names)
             return True
         except Exception as exc:  # pylint: disable=broad-exception-caught
             self.process_exception(exc)
-            return False
-        finally:
-            await self.flush()
             return False
 
     def extract_grpc_error_info(self, exc: grpc.aio.AioRpcError) -> Tuple[int, str, str]:
--- a/neuro_san/http_sidecar/handlers/base_request_handler.py
+++ b/neuro_san/http_sidecar/handlers/base_request_handler.py
@@ -27,6 +27,7 @@ from tornado.web import RequestHandler
 
 from neuro_san.http_sidecar.logging.http_logger import HttpLogger
 from neuro_san.http_sidecar.interfaces.agent_authorizer import AgentAuthorizer
+from neuro_san.http_sidecar.interfaces.agents_updater import AgentsUpdater
 from neuro_san.interfaces.async_agent_session import AsyncAgentSession
 from neuro_san.interfaces.concierge_session import ConciergeSession
 from neuro_san.session.async_grpc_service_agent_session import AsyncGrpcServiceAgentSession
@@ -55,6 +56,7 @@ class BaseRequestHandler(RequestHandler):
     # pylint: disable=attribute-defined-outside-init
     def initialize(self,
                    agent_policy: AgentAuthorizer,
+                   agents_updater: AgentsUpdater,
                    port: int,
                    forwarded_request_metadata: List[str],
                    openapi_service_spec_path: str):
@@ -62,12 +64,15 @@ class BaseRequestHandler(RequestHandler):
         This method is called by Tornado framework to allow
         injecting service-specific data into local handler context.
         :param agent_policy: abstract policy for agent requests
+        :param agents_updater: abstract policy for updating
+                               collection of agents being served
         :param port: gRPC service port.
         :param forwarded_request_metadata: request metadata to forward.
         :param openapi_service_spec_path: file path to OpenAPI service spec.
         """
 
         self.agent_policy = agent_policy
+        self.agents_updater = agents_updater
         self.port: int = port
         self.forwarded_request_metadata: List[str] = forwarded_request_metadata
         self.openapi_service_spec_path: str = openapi_service_spec_path
@@ -123,6 +128,27 @@ class BaseRequestHandler(RequestHandler):
                 port=self.port,
                 metadata=metadata)
         return grpc_session
+
+    async def update_agents(self, metadata: Dict[str, Any]) -> bool:
+        """
+        Update internal agents table by executing request
+        to underlying gRPC service.
+        :param metadata: metadata for request from caller context.
+        :return: True if update was successful
+                 False otherwise
+        """
+        try:
+            data: Dict[str, Any] = {}
+            grpc_session: ConciergeSession = self.get_concierge_grpc_session(metadata)
+            agents_dict: Dict[str, Any] = grpc_session.list(data)
+            self.agents_updater.update_agents(agents_dict)
+            return True
+        except Exception as exc:  # pylint: disable=broad-exception-caught
+            self.process_exception(exc)
+            return False
+        finally:
+            await self.flush()
+            return False
 
     def extract_grpc_error_info(self, exc: grpc.aio.AioRpcError) -> Tuple[int, str, str]:
         """
--- a/neuro_san/http_sidecar/handlers/base_request_handler.py
+++ b/neuro_san/http_sidecar/handlers/base_request_handler.py
@@ -54,6 +54,8 @@ class BaseRequestHandler(RequestHandler):
     request_id: int = 0
 
     # pylint: disable=attribute-defined-outside-init
+    # pylint: disable=too-many-arguments
+    # pylint: disable=too-many-positional-arguments
     def initialize(self,
                    agent_policy: AgentAuthorizer,
                    agents_updater: AgentsUpdater,
--- a/neuro_san/http_sidecar/handlers/base_request_handler.py
+++ b/neuro_san/http_sidecar/handlers/base_request_handler.py
@@ -147,9 +147,6 @@ class BaseRequestHandler(RequestHandler):
             agents_names: List[str] = []
             for agent_dict in agents_list:
                 agents_names.append(agent_dict["agent_name"])
-
-            print(f">>>>>>>>>>>> AGENTS: {agents_names}")
-
             self.agents_updater.update_agents(agents_names)
             return True
         except Exception as exc:  # pylint: disable=broad-exception-caught
--- a/neuro_san/http_sidecar/handlers/connectivity_handler.py
+++ b/neuro_san/http_sidecar/handlers/connectivity_handler.py
@@ -27,13 +27,17 @@ class ConnectivityHandler(BaseRequestHandler):
         """
         Implementation of GET request handler for "connectivity" API call.
         """
+        metadata: Dict[str, Any] = self.get_metadata()
+        update_done: bool = await self.update_agents(metadata)
+        if not update_done:
+            return
+
         if not self.agent_policy.allow(agent_name):
             self.set_status(404)
             self.logger.error({}, "error: Invalid request path %s", self.request.path)
             await self.flush()
             return
 
-        metadata: Dict[str, Any] = self.get_metadata()
         self.logger.info(metadata, "Start GET %s/connectivity", agent_name)
         try:
             data: Dict[str, Any] = {}
--- a/neuro_san/http_sidecar/http_sidecar.py
+++ b/neuro_san/http_sidecar/http_sidecar.py
@@ -109,7 +109,7 @@ class HttpSidecar(AgentAuthorizer, AgentsUpdater):
                 self.allowed_agents[agent_name] = True
             # All other agents are disabled:
             for agent_name, _ in self.allowed_agents.items():
-                if not agent_name in agents:
+                if agent_name not in agents:
                     self.allowed_agents[agent_name] = False
 
     def build_request_data(self) -> Dict[str, Any]:
--- a/neuro_san/http_sidecar/http_sidecar.py
+++ b/neuro_san/http_sidecar/http_sidecar.py
@@ -14,6 +14,7 @@ See class comment for details
 """
 
 import copy
+import threading
 from typing import Any, Dict, List
 
 from tornado.ioloop import IOLoop
@@ -23,6 +24,7 @@ from neuro_san.http_sidecar.logging.http_logger import HttpLogger
 from neuro_san.service.agent_server import DEFAULT_FORWARDED_REQUEST_METADATA
 
 from neuro_san.http_sidecar.interfaces.agent_authorizer import AgentAuthorizer
+from neuro_san.http_sidecar.interfaces.agents_updater import AgentsUpdater
 from neuro_san.http_sidecar.handlers.health_check_handler import HealthCheckHandler
 from neuro_san.http_sidecar.handlers.connectivity_handler import ConnectivityHandler
 from neuro_san.http_sidecar.handlers.function_handler import FunctionHandler
@@ -31,7 +33,7 @@ from neuro_san.http_sidecar.handlers.concierge_handler import ConciergeHandler
 from neuro_san.http_sidecar.handlers.openapi_publish_handler import OpenApiPublishHandler
 
 
-class HttpSidecar(AgentAuthorizer):
+class HttpSidecar(AgentAuthorizer, AgentsUpdater):
     """
     Class provides simple http endpoint for neuro-san API,
     working as a client to neuro-san gRPC service.
@@ -59,6 +61,7 @@ class HttpSidecar(AgentAuthorizer):
         self.openapi_service_spec_path: str = openapi_service_spec_path
         self.forwarded_request_metadata: List[str] = forwarded_request_metadata.split(" ")
         self.allowed_agents: Dict[str, bool] = {}
+        self.lock = threading.Lock()
 
     def __call__(self):
         """
@@ -98,6 +101,16 @@ class HttpSidecar(AgentAuthorizer):
     def allow(self, agent_name) -> bool:
         return self.allowed_agents.get(agent_name, False)
 
+    def update_agents(self, agents: Dict[str, Any]):
+        with self.lock:
+            # We assume all agents from "agents" dictionary are enabled:
+            for agent_name, _ in agents.items():
+                self.allowed_agents[agent_name] = True
+            # All other agents are disabled:
+            for agent_name, _ in self.allowed_agents.items():
+                if not agent_name in agents.keys():
+                    self.allowed_agents[agent_name] = False
+
     def build_request_data(self) -> Dict[str, Any]:
         """
         Build request data for Http handlers.
@@ -105,6 +118,7 @@ class HttpSidecar(AgentAuthorizer):
         """
         return {
             "agent_policy": self,
+            "agents_updater": self,
             "port": self.port,
             "forwarded_request_metadata": self.forwarded_request_metadata,
             "openapi_service_spec_path": self.openapi_service_spec_path
--- a/neuro_san/http_sidecar/http_sidecar.py
+++ b/neuro_san/http_sidecar/http_sidecar.py
@@ -61,13 +61,14 @@ class HttpSidecar(AgentAuthorizer, AgentsUpdater):
         self.openapi_service_spec_path: str = openapi_service_spec_path
         self.forwarded_request_metadata: List[str] = forwarded_request_metadata.split(" ")
         self.allowed_agents: Dict[str, bool] = {}
-        self.lock = threading.Lock()
+        self.lock = None
 
     def __call__(self):
         """
         Method to be called by a process running tornado HTTP server
         to actually start serving requests.
         """
+        self.lock = threading.Lock()
         self.logger = HttpLogger(self.forwarded_request_metadata)
 
         app = self.make_app()
@@ -101,14 +102,14 @@ class HttpSidecar(AgentAuthorizer, AgentsUpdater):
     def allow(self, agent_name) -> bool:
         return self.allowed_agents.get(agent_name, False)
 
-    def update_agents(self, agents: Dict[str, Any]):
+    def update_agents(self, agents: List[str]):
         with self.lock:
             # We assume all agents from "agents" dictionary are enabled:
-            for agent_name, _ in agents.items():
+            for agent_name in agents:
                 self.allowed_agents[agent_name] = True
             # All other agents are disabled:
             for agent_name, _ in self.allowed_agents.items():
-                if not agent_name in agents.keys():
+                if not agent_name in agents:
                     self.allowed_agents[agent_name] = False
 
     def build_request_data(self) -> Dict[str, Any]:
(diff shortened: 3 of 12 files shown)


neuro_san/http_sidecar/handlers/base_request_handler.py
```diff
@@ -124,6 +131,28 @@ def get_concierge_grpc_session(self, metadata: Dict[str, Any]) -> ConciergeSessi
...
         return grpc_session
 
+    async def update_agents(self, metadata: Dict[str, Any]) -> bool:
+        """
+        Update internal agents table by executing request
+        to underlying gRPC service.
+        :param metadata: metadata for request from caller context.
+        :return: True if update was successful
+                 False otherwise
+        """
```
andreidenissov-cog, line 141: Execute "concierge" gRPC request to the neuro-san host to get the current list of known (allowed) agents.

neuro_san/http_sidecar/handlers/connectivity_handler.py
```diff
@@ -27,13 +27,17 @@ async def get(self, agent_name: str):
         """
         Implementation of GET request handler for "connectivity" API call.
         """
+        metadata: Dict[str, Any] = self.get_metadata()
+        update_done: bool = await self.update_agents(metadata)
+        if not update_done:
+            return
+
```
andreidenissov-cog, line 34: Here and in two other handlers below:
before starting to execute this request, query the "host" side for current list of known agents,
before checking if we should route this request at all.

neuro_san/http_sidecar/handlers/connectivity_handler.py
```diff
@@ -27,13 +27,17 @@ async def get(self, agent_name: str):
...
         if not self.agent_policy.allow(agent_name):
             self.set_status(404)
             self.logger.error({}, "error: Invalid request path %s", self.request.path)
             await self.flush()
             return
 
-        metadata: Dict[str, Any] = self.get_metadata()
         self.logger.info(metadata, "Start GET %s/connectivity", agent_name)
```
andreidenissov-cog, line 41: This metadata generation is now moved up, to the very beginning of method code.

neuro_san/internals/tool_factories/service_tool_factory_provider.py
```diff
@@ -58,14 +73,32 @@ def add_agent_tool_registry(self, agent_name: str, registry: AgentToolFactory):
...
+        # Notify listeners about this state change:
+        # do it outside of internal lock
+        for listener in self.listeners:
             if is_new:
-                self.logger.info("BUILT tool registry for agent %s", agent_name)
+                listener.agent_added(agent_name)
+                self.logger.info("ADDED tool registry for agent %s", agent_name)
             else:
+                listener.agent_modified(agent_name)
                 self.logger.info("REPLACED tool registry for agent %s", agent_name)
```
andreidenissov-cog, line 80: Depending on whether agent registry being added is really new or replacing existing registry item in ServiceToolFactory, call appropriate listener method.

neuro_san/internals/tool_factories/service_tool_factory_provider.py
```diff
@@ -46,6 +48,19 @@ def get_instance(cls):
...
+        """
+        self.listeners.append(listener)
+
+    def remove_listener(self, listener: AgentStateListener):
+        """
+        Remove a state listener from registered set.
+        """
+        if listener in self.listeners:
+            self.listeners.remove(listener)
+
```
andreidenissov-cog, line 61: Methods to manage collection of listeners for ServiceToolFactoryProvider.

neuro_san/internals/tool_factories/service_tool_factory_provider.py
```diff
@@ -36,6 +37,7 @@ def __init__(self):
         self.agents_table: Dict[str, AgentToolFactory] = {}
         self.logger = logging.getLogger(self.__class__.__name__)
         self.lock = threading.Lock()
+        self.listeners: List[AgentStateListener] = []
```
andreidenissov-cog, line 39: Initial collection of listeners - empty.

neuro_san/internals/tool_factories/service_tool_factory_provider.py
```diff
@@ -58,14 +73,32 @@ def add_agent_tool_registry(self, agent_name: str, registry: AgentToolFactory):
...
+        Remove agent tool registry from the table
+        """
+        with self.lock:
+            self.agents_table.pop(agent_name, None)
+        # Notify listeners about this state change:
+        # do it outside of internal lock
+        for listener in self.listeners:
+            listener.agent_removed(agent_name)
+        self.logger.info("REMOVED tool registry for agent %s", agent_name)
+
```
andreidenissov-cog, line 94: We may also remove agent tool registry from our state.
If we do, notify all registered listeners about this.
Note that we do that outside of internal lock protecting internal table - 
idea is to never run clients code inside our internal locks,
because you have no guarantees as to what their latency or side effects could be.

neuro_san/service/agent_server.py
```diff
@@ -106,48 +110,82 @@ def setup_tool_factory_provider(self):
...
+            self.server_name_for_logs,
+            self.port, self.logger,
+            request_limit=self.request_limit,
+            max_workers=self.max_concurrent_requests,
+            max_concurrent_rpcs=None,
+            # Used for health checking. Probably needs agent-specific love.
+            protocol_services_by_name_values=values,
+            loop_sleep_seconds=5.0,
+            server_loop_callbacks=self.server_loop_callbacks)
+
```
andreidenissov-cog, line 165: Just formatting change.

neuro_san/session/direct_concierge_session.py
```diff
@@ -51,10 +51,10 @@ def list(self, request_dict: Dict[str, Any]) -> Dict[str, Any]:
...
-        manifest_registries: Dict[str, AgentToolRegistry] = manifest_restorer.restore()
-
-        agents_list = []
-        for agent_name, _ in manifest_registries.items():
+        tool_factory_provider: ServiceToolFactoryProvider = \
+            ServiceToolFactoryProvider.get_instance()
+        agents_names: List[str] = tool_factory_provider.get_agent_names()
+        agents_list: List[Dict[str, Any]] = []
+        for agent_name in agents_names:
             agents_list.append({"agent_name": agent_name, "description": ""})
```
andreidenissov-cog, line 59: For list of currently available agents, go to the "source of truth" - ServiceToolFactoryProvider own state.

neuro_san/service/agent_server.py
```diff
@@ -106,48 +110,82 @@ def setup_tool_factory_provider(self):
         for agent_name, tool_registry in self.tool_registries.items():
             tool_factory_provider.add_agent_tool_registry(agent_name, tool_registry)
 
+    def agent_added(self, agent_name: str):
+        """
```
andreidenissov-cog, line 116: AgentServer itself implements AgentStateListener interface, so whenever ServiceToolFactory state (internal agents table) is changing, we react to this by talking to DynamicAgentRouter to add or remove gRPC endpoint for corresponding agent.

neuro_san/service/agent_server.py
```diff
@@ -106,48 +110,82 @@ def setup_tool_factory_provider(self):
...
         # New-style service
-        security_cfg = None     # ... yet
+        self.security_cfg = None     # ... yet
 
-        self.setup_tool_factory_provider()
         tool_factory_provider: ServiceToolFactoryProvider = \
             ServiceToolFactoryProvider.get_instance()
+        # Add listener to handle adding per-agent gRPC services
+        # to our dynamic router:
+        tool_factory_provider.add_listener(self)
```
andreidenissov-cog, line 176: Register this here AgentServer as a listener to ServiceToolFactoryProvider instance.

neuro_san/http_sidecar/http_sidecar.py
```diff
@@ -98,13 +102,24 @@ def make_app(self):
...
+    def update_agents(self, agents: List[str]):
+        with self.lock:
+            # We assume all agents from "agents" dictionary are enabled:
+            for agent_name in agents:
+                self.allowed_agents[agent_name] = True
+            # All other agents are disabled:
+            for agent_name, _ in self.allowed_agents.items():
+                if agent_name not in agents:
+                    self.allowed_agents[agent_name] = False
+
```
andreidenissov-cog, line 114: Update our internal dynamic collection of known agents with actual set of agents coming from some external "source of truth". This internal collection determines which requests will be handled, and which "404" rejected.

neuro_san/http_sidecar/http_sidecar.py
```diff
@@ -98,13 +102,24 @@ def make_app(self):
...
     def build_request_data(self) -> Dict[str, Any]:
         """
         Build request data for Http handlers.
         :return: a dictionary with request data to be passed to a http handler.
         """
         return {
             "agent_policy": self,
+            "agents_updater": self,
             "port": self.port,
```
andreidenissov-cog, line 123: Pass ourselves as AgentsUpdater interface implementation
to all requests handlers.
Each request handler will query the host gRPC endpoint for current state of things.

neuro_san/service/agent_server.py
```diff
@@ -106,48 +110,82 @@ def setup_tool_factory_provider(self):
...
 
-        agent_names: List[str] = tool_factory_provider.get_agent_names()
-        for agent_name in agent_names:
-            service = AgentService(server_lifetime, security_cfg,
-                                   agent_name,
-                                   tool_factory_provider.get_agent_tool_factory_provider(agent_name),
-                                   self.server_logging)
-            self.services.append(service)
+        self.setup_tool_factory_provider()
```
andreidenissov-cog, line 178: This is where we do initial setup for SetupToolFactoryProvider singleton instance,
adding all agents from neuro-san manifest.
At the same time, through listener interface, we register corresponding gRPC endpoints
in DynamicAgentRouter instance.

neuro_san/service/agent_server.py
```diff
@@ -106,48 +110,82 @@ def setup_tool_factory_provider(self):
...
 
-            servicer_to_server = AgentServicerToServer(service, agent_name=agent_name)
-            servicer_to_server.add_rpc_handlers(server)
+        # Add DynamicAgentRouter instance as a generic RPC handler for our server:
+        service_router: DynamicAgentRouter = DynamicAgentRouter().get_instance()
+        server.add_generic_rpc_handlers((service_router,))
```
andreidenissov-cog, line 181: And DynamicAgentRouter instance is now our gRPC requests handler for AgentServer.

neuro_san/service/agent_servicer_to_server.py
```diff
@@ -30,20 +26,17 @@ class AgentServicerToServer:
...
         """
         Constructor
         """
         self.servicer: AgentServiceServicer = servicer
-        self.agent_name: str = agent_name
 
-    def add_rpc_handlers(self, server: Server):
+    def build_rpc_handlers(self):
         """
-        Adds the RpcMethodHandlers to the server
```
andreidenissov-cog, line 43: We don't add gRPC handlers here anymore, we only build collection of RpcMethodHandlers
to be given to DynamicAgentRouter.

neuro_san/service/dynamic_agent_router.py
```diff
@@ -0,0 +1,87 @@
...
+        Service incoming gRPC request.
+        :param handler_call_details: gRPC incoming call details.
+        :return: gRPC method handler for this call. This handler will be invoked
+            elsewhere by gRPC machinery.
+        """
+        full_method: str = "unknown"
+        try:
+            full_method = handler_call_details.method  # e.g. "/my.Service/SomeMethod"
+            _, service_name, method_name = full_method.split("/", 2)
+            with self.table_lock:
```
andreidenissov-cog, line 72: After we have parsed the full gRPC request path,
"service_name" is actually a name of some agent,
and we are looking up internal "agents_table" to see if method handlers are currently available for this agent,
and then look up and return requested method handler.

neuro_san/service/dynamic_agent_router.py
```diff
@@ -0,0 +1,87 @@
...
+        self.logger = logging.getLogger(self.__class__.__name__)
+
+    @classmethod
+    def get_instance(cls):
+        """
+        Get singleton instance of DynamicAgentRouter
+        """
+        if DynamicAgentRouter.instance is None:
+            DynamicAgentRouter.instance = DynamicAgentRouter()
+        return DynamicAgentRouter.instance
```
d1donlydfink: You probably don't need a singleton if the instance of this guy is held by the AgentServer - there is only ever one of those guys anyway.

neuro_san/service/dynamic_agent_router.py
```diff
@@ -0,0 +1,87 @@
...
+        try:
+            full_method = handler_call_details.method  # e.g. "/my.Service/SomeMethod"
+            _, service_name, method_name = full_method.split("/", 2)
+            with self.table_lock:
+                method_handlers = self.agents_table.get(service_name, None)
+            if method_handlers:
+                return method_handlers.get(method_name, None)
+        except Exception as exc:  # pylint: disable=broad-exception-caught
+            self.logger.error("Failed to execute %s - %s", full_method, str(exc))
+        return None  # Request not handled
```
d1donlydfink: Wow. Deep in the dirt.
It would be worth a reassuring comment as to how the None value here is treated by the gRPC infrastructure.
I would expect the caller to know how to send back HTTP error codes for requests not being handled by the server, for instance, but it's not obvious from this bag of tricks here.

neuro_san/service/agent_server.py
```diff
@@ -106,48 +110,82 @@ def setup_tool_factory_provider(self):
...
+        service_router.add_service(agent_service_name, agent_rpc_handlers)
+
+    def agent_modified(self, agent_name: str):
+        """
+        Agent is being modified in the service scope.
+        :param agent_name: name of an agent
+        """
+        # Endpoints configuration has not changed,
+        # so nothing to do here, actually.
+        _ = agent_name
```
d1donlydfink, line 139: Worth a comment as to how/where the modified agent file is handled.
Are we re-reading a hocon file on every request?
If so that could effect performance adversely (though going there is probably not for this PR), especially for simple requests that should take no time like function() and connectivity()

neuro_san/service/agent_servicer_to_server.py
d1donlydfink: Brave! 
I don't recall off the top of my head, but it's possible the "ServicerToServer" nomenclature means something very specific in gRPC land and if so with the interface change here it might warrant a rename.

neuro_san/internals/tool_factories/service_tool_factory_provider.py
```diff
@@ -46,6 +48,19 @@ def get_instance(cls):
             ServiceToolFactoryProvider.instance = ServiceToolFactoryProvider()
         return ServiceToolFactoryProvider.instance
```
d1donlydfink, line 47: Another case of a questionable singleton. 
The right ownership model could obviate the need for the singleton.
(This one is not for this PR)

neuro_san/internals/tool_factories/service_tool_factory_provider.py
```diff
@@ -46,6 +48,19 @@ def get_instance(cls):
...
+        self.listeners.append(listener)
+
+    def remove_listener(self, listener: AgentStateListener):
+        """
+        Remove a state listener from registered set.
+        """
+        if listener in self.listeners:
+            self.listeners.remove(listener)
+
     def build_agent_tool_registry(self, agent_name: str, config: Dict[str, Any]):
```
d1donlydfink: When/where does build_agent_tool_registry() get called now?
I think it might be removeable.

neuro_san/internals/tool_factories/service_tool_factory_provider.py
d1donlydfink: Is it really the case that if it's not is_new it's always modified?
Maybe a clarifying comment in add_agent_tool_registry() method comment as to when it is expected to be called.  My prepping on current code (not this PR) shows it's called from AgentServer and DirectAgentSessionFactory, but maybe there is a new way in given the new machinery?

neuro_san/internals/interfaces/agent_state_listener.py
```diff
@@ -0,0 +1,38 @@
...
+# Purchase of a commercial license is mandatory for any use of the
+# neuro-san SDK Software in commercial settings.
+#
+# END COPYRIGHT
+
+class AgentStateListener:
+    """
+    Abstract interface for publishing agent state changes -
+    when an agent is being added or removed from the service.
+    """
```
d1donlydfink, line 17: Cool.

neuro_san/internals/tool_factories/service_tool_factory_provider.py
d1donlydfink: This is actually kinda interesting. 
We have been thinking about being able to specify a few other factories via the agent hocon file (like LlmFactory and BaseToolFactory) to allow each agent to have their own definitions instead of a server-global definition.  This  could make *those* config files re-read when agents are updated too which I think would fall in line with expectations.  That is, right now those global guys are likely not on the filesystem-change radar.

For the time being, these types of things will fall off the agent-update burro.

neuro_san/http_sidecar/handlers/base_request_handler.py
```diff
@@ -53,21 +54,27 @@ class BaseRequestHandler(RequestHandler):
     request_id: int = 0
 
     # pylint: disable=attribute-defined-outside-init
+    # pylint: disable=too-many-arguments
+    # pylint: disable=too-many-positional-arguments
```
d1donlydfink, line 58: We really should raise the limbo stick here. I am tired of disabling both of these.
Also: attribute-defined-outside-init actually sounds like it's worth fixing soon. That's not a normal one.

neuro_san/http_sidecar/handlers/base_request_handler.py
```diff
@@ -124,6 +131,28 @@ def get_concierge_grpc_session(self, metadata: Dict[str, Any]) -> ConciergeSessi
...
+            agents_dict: Dict[str, Any] = grpc_session.list(data)
+            agents_list = agents_dict.get("agents", [])
+            agents_names: List[str] = []
+            for agent_dict in agents_list:
+                agents_names.append(agent_dict["agent_name"])
+            self.agents_updater.update_agents(agents_names)
+            return True
+        except Exception as exc:  # pylint: disable=broad-exception-caught
+            self.process_exception(exc)
+            return False
```
d1donlydfink, line 154: In the near future we will have agents that are available to serve that are need-to-know and not listed as part of the concierge service.  There is a ticket for this already.  Might be worth a comment that this coupling will need to be revisited when that happens.

That all said it's a good thing that there is a single source of truth in the gRPC services for now.
I see why you were asking about "Can we get rid of gRPC?" now.

We might be able to devise a way so that this update doesn't have to happen as often.
It's possible similar mechanisms might need to be in play when we get to other protocol compliance like MCP.

Also FWIW: We probably do not need metadata for this internal/implementation-detail query.

neuro_san/http_sidecar/handlers/connectivity_handler.py
d1donlydfink: Why would we not try to service the request at all simply because the update was not done?
Maybe worth a comment in the if/return clause above.

neuro_san/http_sidecar/http_sidecar.py
d1donlydfink: This isn't ideal because now every single request has to go through a lock.
Thinking (perhaps foolishly) out loud: Perhaps some kind of versioning mechanism could be put in place where requests can go about their business without a lock if they have the same version as what the single-source-of-truth has.

d1donlydfink (changes requested):
For now:
* See PR where clarifying comments could be illuminating
* Get rid of DynamicAgentRouter singleton
* Probably remove build_agent_tool_registry() from ServiceToolFactoryProvider
* Where does the filesystem watchdog live?

For future PRs:
* Brainstorm a way to not have a lock for every http request.  This will likely be beneficial for other protocols when we get there soon.
* Strive to remove ServiceToolFactoryProvider singleton.
* Consider the case of need-to-know agents not listed in the concierge service. Another thing coming soon.

neuro_san/http_sidecar/handlers/base_request_handler.py
andreidenissov-cog: attribute-defined-outside-init - yes, I don't like it either.
But this is Tornado quirk - we cannot (re)define our own constructor for RequestHandler sub-class,
at least GPT says so ("request handlers are managed internally, constructors take some special parameters, blah"). But we are provided with "initialize" method we can redefine.
So I don't expect this is easily fixable.

andreidenissov-cog:
> * Where does the filesystem watchdog live?
> 
  Right now, it lives in another hack-together branch intended for testing only.
This PR is only meant to extend internal machinery and prepare things for new feature(s) addition in next PRs,
when those features will be clearly understood.

neuro_san/http_sidecar/handlers/connectivity_handler.py
andreidenissov-cog: If this kind of update is not done, that means something is broken real bad inside our code.
In this case, we call "process_exception()" helper method inside and finish request as 500 "Internal server error".

neuro_san/internals/tool_factories/service_tool_factory_provider.py
andreidenissov-cog: > Is it really the case that if it's not is_new it's always modified?
   That's a caller responsibility to only call this method for existing agent if its definition (AgentToolFactory) has really changed. FWIW though, call with unchanged AgentToolFactory is actually a no-op right now.

neuro_san/http_sidecar/http_sidecar.py
andreidenissov-cog: Locking itself is not scary if: we know that code under lock will execute fast and will not involve any additional locking logic. That's the case here.
To be honest though, this here lock is not strictly necessary since our Tornado server runs a (single) event loop.
But: I expect this to change when we get rid of gRPC layer and will have to refactor our threads/event loops structure.

neuro_san/service/dynamic_agent_router.py
andreidenissov-cog: Yeah, got rid of it.

neuro_san/service/agent_server.py
andreidenissov-cog: Right now, modified agent is not handled anywhere yet - that's for new feature development.
On this level, we requite already built(compiled) AgentToolFactory, where it comes from is not a concern here.
One of possible options - registry directory watchdog will do this building before calling
ServiceToolFactoryProvider.add_agent_tool_registry()

d1donlydfink (approved):

PR #186, opened 2025-05-02
Modify trigger to avoid duplicate test run at PRs

donn-leaf:
This PR should take care of the duplicate test runs that occurred at a PR.


andreidenissov-cog (approved):
Looks OK to me.

PR #192, opened 2025-05-09
UN-3144 Potential help with timeout for long-lived chat connections.

andreidenissov-cog:
This PR is a pretty much speculative attempt to help with an issue reported by 1C group - 
some streaming chat requests not reaching neuro-san http server.
Idea (GPT suggested) that we may face a situation when long-running requests
lead to "keep-alive" server connections to be closed by a server, whereas client still trying to reuse them.
Change itself doesn't seem to affect client-initiated chats with neuro-san,
but timeout issue itself I hasn't been able to reproduce.


d1donlydfink (approved):
LGT AFAICT

PR #195, opened 2025-05-12
UN-3137 Separate gRPC neuro-san level into service and transport parts.

andreidenissov-cog:
This PR is a preparation for future steps towards neuro-san external protocols.
What we do here is separate the service logic itself - with metadata extraction,
logging setup and usage, API methods available from direct sessions,
- from transport level specifics.
Next: using this separation to implement http-based transport in-process,
and then based on that - moving towards other protocols, like A2A.


neuro_san/http_sidecar/handlers/base_request_handler.py
```diff
@@ -75,6 +78,7 @@ def initialize(self,
 
         self.agent_policy = agent_policy
         self.agents_updater = agents_updater
+        self.executor: ThreadPoolExecutor = executor
         self.port: int = port
```
andreidenissov-cog: Cleanup. This shouldn't be here. Not yet.

neuro_san/service/agent_main_loop.py
```diff
@@ -175,7 +175,7 @@ def loop_callback(self) -> bool:
         # Report back on service activity so the ServerLifetime that calls
         # this method can properly yield/sleep depending on how many requests
         # are in motion.
-        agent_services: List[AgentService] = self.server.get_services()
+        agent_services: List[GrpcAgentService] = self.server.get_services()
         for agent_service in agent_services:
             if agent_service.get_request_count() > 0:
                 return True
```
andreidenissov-cog, line 181: Switch import - GrpcAgentService is our GRPC implementation now.

neuro_san/service/agent_server.py
```diff
@@ -27,7 +27,7 @@
 from neuro_san.service.agent_server_logging import AgentServerLogging
 from neuro_san.service.agent_servicer_to_server import AgentServicerToServer
 from neuro_san.service.dynamic_agent_router import DynamicAgentRouter
-from neuro_san.service.agent_service import AgentService
+from neuro_san.service.grpc_agent_service import GrpcAgentService
 from neuro_san.service.concierge_service import ConciergeService
```
andreidenissov-cog, line 31: Switch to renamed GrpcAgentService.

neuro_san/service/agent_service.py
```diff
@@ -102,17 +94,18 @@ def get_request_count(self) -> int:
...
 
-    # pylint: disable=no-member
-    def Function(self, request: service_messages.FunctionRequest,
-                 context: grpc.ServicerContext) \
-            -> service_messages.FunctionResponse:
+    def function(self, request_dict: Dict[str, Any],
+                 request_metadata: Dict[str, Any],
+                 context: Any) \
+            -> Dict[str, Any]:
         """
```
andreidenissov-cog, line 101: For all our API methods: we switch to Dict-based parameters and results.
And rename method names to make pylons happy.

andreidenissov-cog: "context" here is still gRPC artifact, but only because I don't want to go down leaf-server-common code
and fix things there. This is for later.

neuro_san/service/grpc_agent_service.py
```diff
@@ -0,0 +1,150 @@
...
+                        their logging.
+        """
+        self.forwarder: GrpcMetadataForwarder = server_logging.get_forwarder()
+        self.service: AgentService =\
+            AgentService(request_logger,
+                         security_cfg,
+                         agent_name,
+                         tool_registry_provider,
+                         server_logging)
+
```
andreidenissov-cog, line 69: AgentService is our transport protocol-agnostic service logic, to which we will delegate actual service functions.

neuro_san/service/grpc_agent_service.py
```diff
@@ -0,0 +1,150 @@
...
+        return self.service.get_request_count()
+
+    # pylint: disable=no-member
+    def Function(self, request: service_messages.FunctionRequest,
+                 context: grpc.ServicerContext) \
+            -> service_messages.FunctionResponse:
+        """
+        Allows a client to get the outward-facing function for the agent
+        served by this service.
+
```
andreidenissov-cog, line 83: For all our API methods schema is the same:
convert parameters from protocol (gRPC) specific objects to Python dictionaries;
prepare metadata we will need;
delegate execution to AgentService instance;
convert resulting dictionary (or generator of dictionaries) to gRPC representation.

d1donlydfink (approved):
Interesting.
I sure hope you tested everything - like metadata logging.

PR #198, opened 2025-05-14
UN-3143 Fix broken GitHub smoke test

vince-leaf:
This PR should address the requirement of our test infrastructure to run in parallel for pytest.

### DIFF
--- a/.github/workflows/smoke.yml
+++ b/.github/workflows/smoke.yml
@@ -1,4 +1,4 @@
-name: Integration
+name: Smoke Tests
 
 on:
   schedule:
--- a/requirements-build.txt
+++ b/requirements-build.txt
@@ -8,6 +8,7 @@ timeout-decorator==0.5.0
 coverage==7.6.1
 pytest-cov==5.0.0
 parameterized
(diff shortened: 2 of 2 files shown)


.github/workflows/smoke.yml
```diff
@@ -1,4 +1,4 @@
-name: Integration
+name: Smoke Tests
```
vince-leaf, line 1: Corrected the test name.

requirements-build.txt
```diff
@@ -8,6 +8,7 @@ timeout-decorator==0.5.0
 coverage==7.6.1
 pytest-cov==5.0.0
 parameterized
+pytest-xdist
```
vince-leaf, line 11: Added support parallel run.

ofrancon (approved):
lgtm

d1donlydfink (changes requested):
I don't see the change to .github/workflows/integration.yml to not run the smoke tests.

d1donlydfink (approved):
Never mind. Don't need a not when we are running with a positive on marks.

PR #201, opened 2025-05-14
UN-3143 fixed Package 'netcat' has no installation candidate

vince-leaf:
This PR addresses the smoke test cases [Direct, HTTP, and GRPC] that must run successfully in the GitHub Actions job.
@donn-leaf will be the primary reviewer as this change is related to dealing with GitHub actions.
Also, I have added additional info to the readme regarding Docker server environment requirements.

The requirement for the smoke test will need to have server service started before the three types of test connections.

Summary:
- I find that having multiple "run:" blocks as one for starting service and the other run pytest is unsuccessful.
- To overcome this, I have confirmed and set both the start service and run pytest in the single long-run block.

Note:
- I exposed ports 30011 and 8080 but did not address the issue. Exposing the port will be needed when the server service is started in one container and the client is running in another container—in my case, one container with multiple "run:"
blocks for GitHub actions.
- But in GitHub Actions container jobs, especially when using container: under a job, you're running in a persistent container — the container itself does persist between steps, but the shell session used to start the background process does not guarantee or pytest launching the smoke test for each test cases do the right thing.
- Shortly, we will have the start server service to run a separate docker.


.github/workflows/smoke.yml
```diff
@@ -6,7 +6,7 @@ on:
   workflow_dispatch:      # enables manual triggering
 
 jobs:
-  test:
+  smoke-test:
```
vince-leaf, line 9: Updated as Job name.

.github/workflows/smoke.yml
```diff
@@ -17,58 +17,71 @@ jobs:
...
       - name: Show installed packages
         run: pip freeze
 
-      - name: Start agent server service
+      - name: Start agent server service AND run smoke test cases in parallel
+        shell: bash
+        env:
+          OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+          AGENT_TOOL_PATH: "./neuro_san/coded_tools"
+          PYTHONPATH: ${{ env.PYTHONPATH }}:.
```
vince-leaf, line 32: You need the following env set before starting the client that requires coded_tools.

.github/workflows/smoke.yml
```diff
@@ -17,58 +17,71 @@ jobs:
...
           echo "✅ Server process started with PID $(cat agent.pid)"
 
-      - name: Wait for ports 8080 and 30011 (fail on timeout)
-        run: |
-              for i in {1..30}; do
-                PORT_8080_READY=false
-                PORT_30011_READY=false
+          for i in {1..30}; do
+            PORT_8080_READY=false
+            PORT_30011_READY=false
```
vince-leaf, line 51: Confirmed the server started.

.github/workflows/smoke.yml
```diff
@@ -17,58 +17,71 @@ jobs:
...
+            fi
 
-                if [ "$PORT_8080_READY" = true ] && [ "$PORT_30011_READY" = true ]; then
-                  echo "✅ Both ports are ready"
-                  exit 0
-                fi
+            if [ "$PORT_8080_READY" = true ] && [ "$PORT_30011_READY" = true ]; then
+              echo "✅ Both ports are ready after awaiting $i seconds"
+              break
+            fi
```
vince-leaf, line 64: Confirmed the server port numbers

.github/workflows/smoke.yml
```diff
@@ -17,58 +17,71 @@ jobs:
...
-        env:
-          OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+          until curl -s http://localhost:8080/health > /dev/null; do
+            echo "Waiting for server health endpoint..."
+            sleep 1
+          done
+
+          echo "✅ Server is healthy and ready"
+
+          netstat -tuln | grep -E '8080|30011'
```
vince-leaf, line 82: Confirmed the ports are set to be listened to.

.github/workflows/smoke.yml
```diff
@@ -17,58 +17,71 @@ jobs:
...
+
+          pytest -s --verbose -m "smoke" --timer-top-n 100 -n auto
```
vince-leaf, line 84: Triggered the pytest test cases in the same run blocker as the start of service.

README.md
```diff
@@ -103,6 +103,8 @@ Option 2: Build and run the docker container for the hosting agent service:
 These build.sh / Dockerfile / run.sh scripts are intended to be portable so they can be used with
 your own projects' registries and coded_tools work.
 
+ℹ️ Ensure the required environment variables (OPENAI_API_KEY, AGENT_TOOL_PATH, and PYTHONPATH) are passed into the container — either by exporting them before running run.sh, or by configuring them inside the script
```
vince-leaf, line 106: Added readme

.gitignore
```diff
@@ -1,5 +1,6 @@
 # virtualenv folders
 venv/
+venv2/
```
donn-leaf, line 3: You can probably do something like venv*
Consider that in your next PR.

donn-leaf (approved):
Looks good. When you're ready, I can help you tackle converting the server start-up into a github actions "service".

PR #202, opened 2025-05-14
Fixes for tornado handlers.

andreidenissov-cog:
This PR implements several fixes to neuro-san Tornado request handlers.

- Make sure that handler.finish() is always called directly, not awaited, as recommended by Tornado docs;
- We add logging for possible tornado.iostream.StreamClosedError exception;
- Connections are back to being keep-alive by default, since this doesn't seem to be the source of issues we observe.

Tested by running local mini stress test:
3 clients each running 30x3 chat requests - around 1000 chat results in total. Server survives.

### DIFF
--- a/neuro_san/http_sidecar/handlers/base_request_handler.py
+++ b/neuro_san/http_sidecar/handlers/base_request_handler.py
@@ -23,6 +23,7 @@ import os
 
 import grpc
 
+import tornado
 from tornado.web import RequestHandler
 
 from neuro_san.http_sidecar.logging.http_logger import HttpLogger
@@ -79,6 +80,8 @@ class BaseRequestHandler(RequestHandler):
         self.forwarded_request_metadata: List[str] = forwarded_request_metadata
         self.openapi_service_spec_path: str = openapi_service_spec_path
         self.logger = HttpLogger(forwarded_request_metadata)
+        # Set default request_id for this request handler in case we will need it:
+        BaseRequestHandler.request_id += 1
 
         if os.environ.get("AGENT_ALLOW_CORS_HEADERS") is not None:
             self.set_header("Access-Control-Allow-Origin", "*")
@@ -99,7 +102,6 @@ class BaseRequestHandler(RequestHandler):
             elif item_name == "request_id":
                 # Generate unique id so we have some way to track this request:
                 result[item_name] = f"request-{BaseRequestHandler.request_id}"
-                BaseRequestHandler.request_id += 1
             else:
                 result[item_name] = "None"
         return result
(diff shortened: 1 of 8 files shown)


neuro_san/http_sidecar/handlers/base_request_handler.py
```diff
@@ -176,7 +178,7 @@ def process_exception(self, exc: Exception):
             # Handle invalid JSON input
             self.set_status(400)
             self.write({"error": "Invalid JSON format"})
-            self.logger.error({}, "error: Invalid JSON format")
+            self.logger.error(self.get_metadata(), "error: Invalid JSON format")
             return
```
andreidenissov-cog, line 182: We can actually put in request-specific metadata here.

neuro_san/http_sidecar/handlers/base_request_handler.py
```diff
@@ -203,14 +205,32 @@ def data_received(self, chunk):
     def prepare(self):
         self.logger.info(self.get_metadata(), f"[REQUEST RECEIVED] {self.request.method} {self.request.uri}")
 
-    def set_default_headers(self):
-        # Force to close this connection without trying to reuse it.
-        self.set_header("Connection", "close")
+    def do_finish(self):
```
andreidenissov-cog, line 208: No default header to reject keep-alive connections anymore.

neuro_san/http_sidecar/handlers/base_request_handler.py
```diff
@@ -203,14 +205,32 @@ def data_received(self, chunk):
...
+            return False
 
     async def options(self, *_args, **_kwargs):
         """
         Handles OPTIONS requests for CORS support
         """
         # No body needed. Tornado will return a 204 No Content by default
         self.set_status(http.HTTPStatus.NO_CONTENT)
-        await self.finish()
+        self.do_finish()
```
andreidenissov-cog, line 236: Here and below - doing wrapper for handler finish() instead of flush() makes more sense,
if this is not a streaming request.

neuro_san/session/http_service_agent_session.py
```diff
@@ -93,7 +93,6 @@ def _get_headers(self) -> Dict[str, Any]:
         headers: Dict[str, Any] = self.metadata
         if headers is None:
             headers = {}
-        headers["Connection"] = "close"
         return headers
```
andreidenissov-cog, line 96: Back to keep-alive connections for client session also.

neuro_san/http_sidecar/handlers/streaming_chat_handler.py
```diff
@@ -41,15 +41,19 @@ async def stream_out(self,
...
         async for sub_generator in generator:
             async for result_message in sub_generator:
                 result_dict: Dict[str, Any] = MessageToDict(result_message)
                 result_str: str = json.dumps(result_dict) + "\n"
                 self.write(result_str)
-                await self.flush()
+                flush_ok = await self.do_flush()
+                if not flush_ok:
+                    return sent_out
                 sent_out += 1
```
andreidenissov-cog, line 57: For stream, we must do flush() and bail out if connection is dropped.

ofrancon (approved):

PR #204, opened 2025-05-16
UN-3155 Polling agents registry updater.

andreidenissov-cog:
This PR implements another version of agents registry updater - the one that goes for direct polling of target registry directory instead of relying on OS-level file system events.
OS-level events work just fine for truly local setup,
but have problems if we deal with mapped storage volumes, especially if file system of different kinds are mixed.
Polling is much more reliable, but relatively CPU-intensive.

Tested: running neuro-san server in a Docker container, with registry mapped to host-side directory.
Changes in this directory are detected by the server with proper reaction.
Also: running medium stability test with another script periodically updating agents registry.

### DIFF
--- a/neuro_san/service/registries_watcher/periodic_updater/manifest_periodic_updater.py
+++ b/neuro_san/service/registries_watcher/periodic_updater/manifest_periodic_updater.py
@@ -15,7 +15,8 @@ import threading
 from typing import Dict
 
 from neuro_san.internals.tool_factories.service_tool_factory_provider import ServiceToolFactoryProvider
-from neuro_san.service.registries_watcher.periodic_updater.registry_observer import RegistryObserver
+from neuro_san.service.registries_watcher.periodic_updater.registry_event_observer import RegistryEventObserver
+from neuro_san.service.registries_watcher.periodic_updater.registry_polling_observer import RegistryPollingObserver
 from neuro_san.internals.graph.registry.agent_tool_registry import AgentToolRegistry
 from neuro_san.internals.graph.persistence.registry_manifest_restorer import RegistryManifestRestorer
 
@@ -25,6 +26,7 @@ class ManifestPeriodicUpdater:
     Class implementing periodic manifest directory updates
     by watching agent files and manifest file itself.
     """
+    use_polling: bool = True
 
     def __init__(self, manifest_path: str, update_period_seconds: int):
         """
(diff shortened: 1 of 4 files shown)


neuro_san/service/registries_watcher/periodic_updater/registry_event_observer.py
```diff
@@ -17,7 +17,7 @@
 from neuro_san.service.registries_watcher.periodic_updater.registry_change_handler import RegistryChangeHandler
 
 
-class RegistryObserver:
+class RegistryEventObserver:
     """
```
andreidenissov-cog, line 21: Class renaming.

neuro_san/service/registries_watcher/periodic_updater/registry_polling_observer.py
```diff
@@ -0,0 +1,46 @@
...
+import logging
+from typing import Tuple
+from pathlib import Path
+from watchdog.observers.polling import PollingObserver
+
+from neuro_san.service.registries_watcher.periodic_updater.registry_change_handler import RegistryChangeHandler
+
+
+class RegistryPollingObserver:
+    """
```
andreidenissov-cog, line 21: Polling observer.
Note that if our requested manifest update period is N seconds,
our polling cycle is computed to be shorter, around 1/4 of manifest update period,
to improve reaction to updates.

requirements.txt
```diff
@@ -6,8 +6,8 @@
 # requirements-build.txt
 
 # In-house dependencies
-leaf-common==1.2.21
-leaf-server-common==0.1.18
+leaf-common==1.2.22
+leaf-server-common==0.1.19
```
andreidenissov-cog, line 11: Update to our dependencies to get the latest fix in leaf-common.

neuro_san/service/registries_watcher/periodic_updater/manifest_periodic_updater.py
```diff
@@ -36,7 +38,11 @@ def __init__(self, manifest_path: str, update_period_seconds: int):
         self.update_period_seconds: int = update_period_seconds
         self.logger = logging.getLogger(self.__class__.__name__)
         self.updater = threading.Thread(target=self._run, daemon=True)
-        self.observer: RegistryObserver = RegistryObserver(self.manifest_path)
+        if self.use_polling:
+            poll_interval: int = self.compute_polling_interval(update_period_seconds)
+            self.observer = RegistryPollingObserver(self.manifest_path, poll_interval)
+        else:
+            self.observer = RegistryEventObserver(self.manifest_path)
         self.tool_factory: ServiceToolFactoryProvider = \
```
andreidenissov-cog, line 46: Select with observer we want to use.
For now, hard-coded to polling variant.

ofrancon: The idea is that we would poll for Docker volumes, but we would use event observer for local file systems?

ofrancon (approved):
lgtm

PR #206, opened 2025-05-22
Regenerate GRPC code to update GitHub org name

ofrancon:
Regenerated GRPC code to update GitHub org name


neuro_san/api/grpc/agent_pb2.py
ofrancon, line 1: The only real diff should be `leaf-ai` being replaced by `cognizant-ai-lab`

deepsaia (approved):
LGTM AFAICT.

PR #207, opened 2025-05-28
UN-3166 Add CORS support for healthcheck URL "/"

dsargent:
Previously there was no CORS support for the root (healthcheck). This is useful (for example) for the UI project to "test on connect" when the user enters a URL, so we can hit the healthcheck URL from the browser.

### DIFF
--- a/neuro_san/http_sidecar/handlers/health_check_handler.py
+++ b/neuro_san/http_sidecar/handlers/health_check_handler.py
@@ -1,4 +1,3 @@
-
 # Copyright (C) 2023-2025 Cognizant Digital Business, Evolutionary AI.
 # All Rights Reserved.
 # Issued under the Academic Public License.
@@ -12,6 +11,8 @@
 """
 See class comment for details
 """
+import http
+import os
 from typing import Any, Dict
 
 from tornado.web import RequestHandler
@@ -21,6 +22,17 @@ class HealthCheckHandler(RequestHandler):
     """
     Handler class for API endpoint health check.
     """
+
+    def initialize(self):
+        """
+        This method is called by Tornado framework to allow
+        injecting service-specific data into local handler context.
+        Here we use it to inject CORS headers if so configured.
+        """
+        if os.environ.get("AGENT_ALLOW_CORS_HEADERS") is not None:
+            self.set_header("Access-Control-Allow-Origin", "*")
+            self.set_header("Access-Control-Allow-Methods", "GET, OPTIONS")
+            self.set_header("Access-Control-Allow-Headers", "Content-Type, Transfer-Encoding")
 
     async def get(self):
         """
@@ -46,3 +58,11 @@ class HealthCheckHandler(RequestHandler):
         with no-op implementation.
         """
         return
+
+    async def options(self, *_args, **_kwargs):
+        """
+        Handles OPTIONS requests for CORS support
+        """
+        # No body needed. Just return a 204 No Content
+        self.set_status(http.HTTPStatus.NO_CONTENT)
+        self.finish()


neuro_san/http_sidecar/handlers/health_check_handler.py
dsargent, line 1: I thought of trying to refactor `BaseRequestHandler` for this since it already does the CORS thing but (1) that seemed messy for something as simple as this and (2) the CORS handling is slightly different here since we don't want to allow `POST`, only `GET` and of course `OPTIONS`.

Also the healthcheck handler doesn't currently inherit from `BaseRequestHandler` and it probably doesn't want to -- most of the stuff in there is not of interest.

Ultimately we want to tighten up these CORS rules too. Wildcards are too loose.

neuro_san/http_sidecar/handlers/health_check_handler.py
```diff
@@ -22,6 +23,17 @@ class HealthCheckHandler(RequestHandler):
     Handler class for API endpoint health check.
     """
 
+    def initialize(self):
+        """
+        This method is called by Tornado framework to allow
+        injecting service-specific data into local handler context.
+        Here we use it to inject CORS headers if so configured.
+        """
+        if os.environ.get("AGENT_ALLOW_CORS_HEADERS") is not None:
```
dsargent, line 32: We could consider reversing the default here so we allow CORS unless it's disabled, since in almost all cases we'd want it.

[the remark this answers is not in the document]
andreidenissov-cog: > Extract the environment variable name into a module-level constant (e.g. `CORS_ENV_VAR = "AGENT_ALLOW_CORS_HEADERS"`) to avoid typos and ease future updates.
   Umm, maybe? Need to define it for all handlers.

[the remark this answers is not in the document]
andreidenissov-cog: > Instead of enabling CORS on any presence of the variable, consider parsing it as a boolean (e.g. 'true'/'false') to avoid accidental enabling when the var is set but empty or misconfigured.
    I guess makes sense, but again, should define module-level check to avoid copy-pasting in at least 2 places.

swensel (approved):
LGTM, thank you @dsargent

neuro_san/http_sidecar/handlers/health_check_handler.py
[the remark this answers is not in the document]
dsargent: Meh...just some prissy suggestions by Copilot. I enable it routinely because it's free to do so and it occasionally unearths an interesting nugget, but mostly the S/N ratio is poor.

neuro_san/http_sidecar/handlers/health_check_handler.py
```diff
@@ -46,3 +58,11 @@ def data_received(self, chunk):
...
         """
         return
+
+    async def options(self, *_args, **_kwargs):
+        """
+        Handles OPTIONS requests for CORS support
+        """
+        # No body needed. Just return a 204 No Content
+        self.set_status(http.HTTPStatus.NO_CONTENT)
+        self.finish()
```
d1donlydfink, line 68: Feels like there is some derivation from BaseRequestHandler that could happen here so that everything has the same options() policy, but:
a) I get you are trying to get something done and this is probably just fine for that
b) I don't have the best sense of whether my hypothesis above is a truly good idea or not, so I will leave that assessment to @andreidenissov-cog

d1donlydfink (approved):
LGTM, but @andreidenissov-cog should have the last word here.

Some question as to whether HealthCheckHandler should ultimately derive from BaseRequestHandler or not.

neuro_san/http_sidecar/handlers/health_check_handler.py
dsargent: I had the same doubts -- see my other comment. It felt like overkill introducing an intermediate layer in the hierarchy "just for this" but maybe at some point if it becomes a bigger use case?

PR #208, opened 2025-05-29
UN-3137 Add AsyncAgentService module.

andreidenissov-cog:
This PR is the first in a series of PRs leading to implementing http service stack completely independent of gRPC machinery.
In this PR we add new AsyncAgentService module implementing async neuro-san API. Internally, it uses already existing 
AsyncDirectAgentSession and generally follows code pattern of AgentService module.
Differences:
1. API is async and intended to be used in running event loop (namely, Tornado server event loop)
2. For chat results, we directly pass up async generator provided by neuro-san internals, without any sync/async conversions
3. We use different kind of logger - that requires request-specific metadata to be provided for every 
logger call, because we cannot rely on thread-local context to store this data.
   Added EventLoopLogger interface abstracts required API, and existing HttpLogger implements it.
   Note that AsyncAgentService is not yet used anywhere, I want to add it to the codebase as a building block to manage overall amount of changes.

   Was tested in a separate hacked-together branch running agents chat requests.

### DIFF
--- a/neuro_san/service/async_agent_service.py
+++ b/neuro_san/service/async_agent_service.py
@@ -12,17 +12,14 @@
 
 from typing import Any
 from typing import Dict
-from typing import Iterator
 from typing import Generator
 
-import copy
 import json
 import uuid
 
 from leaf_common.asyncio.asyncio_executor import AsyncioExecutor
 
 from leaf_server_common.server.atomic_counter import AtomicCounter
-from leaf_server_common.server.request_logger import RequestLogger
 
 from neuro_san.interfaces.event_loop_logger import EventLoopLogger
 from neuro_san.internals.interfaces.agent_tool_factory_provider import AgentToolFactoryProvider
@@ -110,7 +107,8 @@ class AsyncAgentService:
         }
         metadata.update(request_metadata)
         if do_log:
-            self.request_logger.info(metadata,
+            self.request_logger.info(
+                metadata,
                 "Received a %s request for %s",
                 f"{self.agent_name}.Function", log_marker)
 
@@ -125,7 +123,8 @@ class AsyncAgentService:
         response_dict = await session.function(request_dict)
 
         if do_log:
-            self.request_logger.info(metadata,
+            self.request_logger.info(
+                metadata,
                 "Done with %s request for %s",
                 f"{self.agent_name}.Function", log_marker)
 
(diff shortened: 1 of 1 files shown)


neuro_san/interfaces/event_loop_logger.py
```diff
@@ -0,0 +1,56 @@
...
+#
+# END COPYRIGHT
+"""
+See class comment for details
+"""
+from typing import Any
+from typing import Dict
+
+
+class EventLoopLogger:
```
d1donlydfink, line 19: The things we put in the top-level interfaces are ones that we expect external python clients to use.
We don't expect people to implement EventLoopLogger themselves, so it probably doesn't belong here.
Maybe this belongs under http_sidecar or services somehow?

Can move in your next PR.

neuro_san/service/async_agent_service.py
```diff
@@ -0,0 +1,253 @@
...
+DO_NOT_LOG_REQUESTS = [
+]
+
+
+# pylint: disable=too-many-instance-attributes
+class AsyncAgentService:
+    """
+    A base implementation of the Neuro-San Async Agent Service,
+    independent of target transport protocol.
+    """
```
d1donlydfink, line 47: Interesting indeed!  I take it this is largely a cut/paste of what is in agent_service.py today.
Reasonable as a step forward to not interfere with the guts of the other guy.
Hopefully that other guy and this guy will share some common abstractions so as not to duplicate policy code.
Or is your intent to remove the other guy entirely?

d1donlydfink (approved):
Some ?s in PR about further intent, but nothing is a deal breaker here.

As you undertake your journey, please be sure the end product will uphold the tenet
 that the existing server guts for grpc has, which is that each request has its own async event loop so that if one request decides to block synchronously because of bad/insensitive CodedTool or langchain tool coding, it does not impede any other request.  That is a server with one event loop to rule them all is a road to subtle performance 10-car pile-up.

neuro_san/interfaces/event_loop_logger.py
andreidenissov-cog: OK good point. It has actually been under http_sidecar at first, but then dependencies graph looks wrong.
I'll move it somewhere.

neuro_san/service/async_agent_service.py
andreidenissov-cog: Yeah, that's something to figure out.
And the main question is: what kind of service protocols we foresee to be used by us going forward?
So far all the shiny new things are http-based.

PR #209, opened 2025-05-29
UN-3120 no key messaging

d1donlydfink:
There were a couple of modes of failure that could happen when OPENAI_API_KEY or ANTHROPIC_API_KEY were not defined that were reasonably described as cryptic.

This PR consolidates the error detection and reporting for those cases so as to put out the same bit of helpful error text when those situations do arise.

It's likely I haven't gotten all cases, so if you come across a situation where the new messaging should be applied, give me a holler.

### DIFF
--- a/neuro_san/internals/run_context/langchain/api_key_error_check.py
+++ b/neuro_san/internals/run_context/langchain/api_key_error_check.py
@@ -16,7 +16,8 @@ from typing import List
 
 # Dictionary with provider key env var -> strings to look for
 API_KEY_EXCEPTIONS: Dict[str, List] = {
-    "OPENAI_API_KEY": ["OPENAI_API_KEY", "Incorrect API key provided"]
+    "OPENAI_API_KEY": ["OPENAI_API_KEY", "Incorrect API key provided"],
+    "ANTHROPIC_API_KEY": ["ANTHROPIC_API_KEY", "anthropic_api_key", "invalid x-api-key", "credit balance"],
 }
 
 
@@ -54,7 +55,9 @@ server or run-time enviroment in order to use this agent network.
 Some things to try:
 1) Double check that your value for {api_key} is set correctly
 2) If you do not have a value for {api_key}, visit the LLM provider's website to get one.
-3) Sometimes these errors happen because of firewall blockages to the site that hosts the LLM.
+3) It's possible that your credit balance on your account with the LLM provider is too low
+   to make the request.  Check that.
+4) Sometimes these errors happen because of firewall blockages to the site that hosts the LLM.
    Try checking that you can reach the regular UI for the LLM from a web browser
    on the same machine making this request.
 """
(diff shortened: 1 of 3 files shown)


neuro_san/internals/run_context/langchain/api_key_error_check.py
```diff
@@ -0,0 +1,65 @@
...
+API_KEY_EXCEPTIONS: Dict[str, List] = {
+    "OPENAI_API_KEY": ["OPENAI_API_KEY", "Incorrect API key provided"],
+    "ANTHROPIC_API_KEY": ["ANTHROPIC_API_KEY", "anthropic_api_key", "invalid x-api-key", "credit balance"],
+}
+
+
+class ApiKeyErrorCheck:
+    """
+    Class for common policy when checking for API key errors for various LLM providers.
+    """
```
d1donlydfink, line 27: New class for common error checking from multiple exception points.

neuro_san/internals/run_context/langchain/default_llm_factory.py
```diff
@@ -251,6 +255,17 @@ def create_base_chat_model(self, config: Dict[str, Any],
...
+
+            # Catch some common wrong or missing API key errors in a single place
+            # with some verbose error messaging.
+            except (OpenAIError, ValidationError) as exception:
+                # Will re-raise but with the right exception text it will
+                # also provide some more helpful failure text.
+                message: str = ApiKeyErrorCheck.check_for_api_key_exception(exception)
+                if message is not None:
+                    raise ValueError(message) from exception
+                found_exception = exception
```
d1donlydfink, line 267: Raise the value error with nice verbiage provided by the new class if the exception corresponds to something to do with an API_KEY error.

If not, process the exception as it would have happened normally.

neuro_san/internals/run_context/langchain/api_key_error_check.py
```diff
@@ -0,0 +1,65 @@
...
+
+Some things to try:
+1) Double check that your value for {api_key} is set correctly
+2) If you do not have a value for {api_key}, visit the LLM provider's website to get one.
+3) It's possible that your credit balance on your account with the LLM provider is too low
+   to make the request.  Check that.
+4) Sometimes these errors happen because of firewall blockages to the site that hosts the LLM.
+   Try checking that you can reach the regular UI for the LLM from a web browser
+   on the same machine making this request.
+"""
```
d1donlydfink, line 63: New verbiage for API_KEY problems.

dsargent: Could also suggest to try a different wifi/network -- since a lot of our colleagues' issues are due to corporate network restrictions.

andreidenissov-cog (approved):
Looks good.

Noravee (approved):
lgtm.

ofrancon (approved):
lgtm

PR #211, opened 2025-05-30
UN-3137 Switch to Tornado http server in-process as a separate thread.

andreidenissov-cog:
This PR moves Tornado http server from separate process to separate thread in main neuro-san process.
The move will allow http service stack to access neuro-san internals directly, without relying on local gRPC endpoint.
Note that switch away from local gRPC is not done in this PR - except for Concierge service query.
Switching chat request to using AsyncAgentService is somewhat tricky and warrants I think a separate PR.
As for simple Function and Connectivity requests - right now it's not worth it to mix/match direct and gRPC sessions.

Tested: stability test running concurrent local clients:
2 grpc + 2 http each doing 20x3 one-shot chats.

### DIFF
--- a/neuro_san/http_sidecar/handlers/base_request_handler.py
+++ b/neuro_san/http_sidecar/handlers/base_request_handler.py
@@ -30,9 +30,7 @@ from neuro_san.http_sidecar.logging.http_logger import HttpLogger
 from neuro_san.http_sidecar.interfaces.agent_authorizer import AgentAuthorizer
 from neuro_san.http_sidecar.interfaces.agents_updater import AgentsUpdater
 from neuro_san.interfaces.async_agent_session import AsyncAgentSession
-from neuro_san.interfaces.concierge_session import ConciergeSession
 from neuro_san.session.async_grpc_service_agent_session import AsyncGrpcServiceAgentSession
-from neuro_san.session.grpc_concierge_session import GrpcConciergeSession
 
 
 class BaseRequestHandler(RequestHandler):
--- a/neuro_san/http_sidecar/handlers/base_request_handler.py
+++ b/neuro_san/http_sidecar/handlers/base_request_handler.py
@@ -119,15 +119,16 @@ class BaseRequestHandler(RequestHandler):
                 agent_name=agent_name)
         return grpc_session
 
-    async def update_agents(self) -> bool:
+    async def update_agents(self, metadata: Dict[str, Any]) -> bool:
         """
         Update internal agents table by executing request
         to underlying gRPC service.
+        :param metadata: metadata to be used for logging if necessary.
         :return: True if update was successful
                  False otherwise
         """
         try:
-            self.agents_updater.update_agents()
+            self.agents_updater.update_agents(metadata=metadata)
             return True
         except Exception as exc:  # pylint: disable=broad-exception-caught
             self.process_exception(exc)
--- a/neuro_san/http_sidecar/handlers/base_request_handler.py
+++ b/neuro_san/http_sidecar/handlers/base_request_handler.py
@@ -121,35 +121,15 @@ class BaseRequestHandler(RequestHandler):
                 agent_name=agent_name)
         return grpc_session
 
-    def get_concierge_grpc_session(self, metadata: Dict[str, Any]) -> ConciergeSession:
-        """
-        Build gRPC session to talk to "concierge" service
-        :return: ConciergeSession to use
-        """
-        grpc_session: ConciergeSession = \
-            GrpcConciergeSession(
-                host="localhost",
-                port=self.port,
-                metadata=metadata)
-        return grpc_session
-
-    async def update_agents(self, metadata: Dict[str, Any]) -> bool:
+    async def update_agents(self) -> bool:
         """
         Update internal agents table by executing request
         to underlying gRPC service.
-        :param metadata: metadata for request from caller context.
         :return: True if update was successful
                  False otherwise
         """
         try:
-            data: Dict[str, Any] = {}
-            grpc_session: ConciergeSession = self.get_concierge_grpc_session(metadata)
-            agents_dict: Dict[str, Any] = grpc_session.list(data)
-            agents_list = agents_dict.get("agents", [])
-            agents_names: List[str] = []
-            for agent_dict in agents_list:
-                agents_names.append(agent_dict["agent_name"])
-            self.agents_updater.update_agents(agents_names)
+            self.agents_updater.update_agents()
             return True
         except Exception as exc:  # pylint: disable=broad-exception-caught
             self.process_exception(exc)
--- a/neuro_san/http_sidecar/handlers/concierge_handler.py
+++ b/neuro_san/http_sidecar/handlers/concierge_handler.py
@@ -16,6 +16,7 @@ from typing import Any, Dict
 
 from neuro_san.http_sidecar.handlers.base_request_handler import BaseRequestHandler
 from neuro_san.interfaces.concierge_session import ConciergeSession
+from neuro_san.session.direct_concierge_session import DirectConciergeSession
 
 
 class ConciergeHandler(BaseRequestHandler):
@@ -31,8 +32,8 @@ class ConciergeHandler(BaseRequestHandler):
         self.logger.info(metadata, "Start GET /api/v1/list")
         try:
             data: Dict[str, Any] = {}
-            grpc_session: ConciergeSession = self.get_concierge_grpc_session(metadata)
-            result_dict: Dict[str, Any] = grpc_session.list(data)
+            session: ConciergeSession = DirectConciergeSession(metadata=metadata)
+            result_dict: Dict[str, Any] = session.list(data)
 
             # Return gRPC response to the HTTP client
             self.set_header("Content-Type", "application/json")
--- a/neuro_san/http_sidecar/handlers/connectivity_handler.py
+++ b/neuro_san/http_sidecar/handlers/connectivity_handler.py
@@ -28,7 +28,7 @@ class ConnectivityHandler(BaseRequestHandler):
         Implementation of GET request handler for "connectivity" API call.
         """
         metadata: Dict[str, Any] = self.get_metadata()
-        update_done: bool = await self.update_agents()
+        update_done: bool = await self.update_agents(metadata=metadata)
         if not update_done:
             return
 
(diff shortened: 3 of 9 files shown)


neuro_san/http_sidecar/handlers/connectivity_handler.py
```diff
@@ -28,7 +28,7 @@ async def get(self, agent_name: str):
         Implementation of GET request handler for "connectivity" API call.
         """
         metadata: Dict[str, Any] = self.get_metadata()
-        update_done: bool = await self.update_agents(metadata)
+        update_done: bool = await self.update_agents()
         if not update_done:
             return
```
andreidenissov-cog, line 33: Changed method signature.

neuro_san/http_sidecar/handlers/function_handler.py
```diff
@@ -28,7 +28,7 @@ async def get(self, agent_name: str):
         Implementation of GET request handler for "function" API call.
         """
         metadata: Dict[str, Any] = self.get_metadata()
-        update_done: bool = await self.update_agents(metadata)
+        update_done: bool = await self.update_agents()
         if not update_done:
             return
```
andreidenissov-cog, line 33: Changed method signature.

neuro_san/http_sidecar/handlers/function_handler.py
```diff
@@ -28,7 +28,7 @@ async def get(self, agent_name: str):
         Implementation of GET request handler for "function" API call.
         """
         metadata: Dict[str, Any] = self.get_metadata()
-        update_done: bool = await self.update_agents(metadata)
+        update_done: bool = await self.update_agents()
         if not update_done:
```
andreidenissov-cog, line 32: Changed method signature.

neuro_san/http_sidecar/handlers/streaming_chat_handler.py
```diff
@@ -63,7 +63,7 @@ async def post(self, agent_name: str):
         """
 
         metadata: Dict[str, Any] = self.get_metadata()
-        update_done: bool = await self.update_agents(metadata)
+        update_done: bool = await self.update_agents()
         if not update_done:
```
andreidenissov-cog, line 67: Changed method signature.

neuro_san/http_sidecar/interfaces/agents_updater.py
```diff
@@ -21,9 +20,9 @@ class AgentsUpdater:
     being served.
     """
 
-    def update_agents(self, agents: List[str]):
+    def update_agents(self):
         """
-        :param agents: list of agents names which should be served currently.
+        Update list of agents for which serving is allowed.
         :return: nothing
```
andreidenissov-cog, line 29: Don't need list of agents - we'll query it internally.

neuro_san/http_sidecar/handlers/concierge_handler.py
```diff
@@ -31,8 +32,8 @@ def get(self):
         self.logger.info(metadata, "Start GET /api/v1/list")
         try:
             data: Dict[str, Any] = {}
-            grpc_session: ConciergeSession = self.get_concierge_grpc_session(metadata)
-            result_dict: Dict[str, Any] = grpc_session.list(data)
+            session: ConciergeSession = DirectConciergeSession(metadata=metadata)
+            result_dict: Dict[str, Any] = session.list(data)
```
andreidenissov-cog, line 37: Switch to DirectConciergeSession - we are in the same process space now.

neuro_san/http_sidecar/handlers/base_request_handler.py
```diff
@@ -121,35 +119,15 @@ def get_agent_grpc_session(self,
...
-            data: Dict[str, Any] = {}
-            grpc_session: ConciergeSession = self.get_concierge_grpc_session(metadata)
-            agents_dict: Dict[str, Any] = grpc_session.list(data)
-            agents_list = agents_dict.get("agents", [])
-            agents_names: List[str] = []
-            for agent_dict in agents_list:
-                agents_names.append(agent_dict["agent_name"])
-            self.agents_updater.update_agents(agents_names)
+            self.agents_updater.update_agents()
             return True
```
andreidenissov-cog, line 132: Simplify things: just call agents_updater.update_agents() to get the latest state of agents allowed for service.

neuro_san/service/agent_server.py
```diff
@@ -91,9 +92,9 @@ def __init__(self, port: int,
         self.server_lifetime = None
         self.security_cfg = None
         self.services: List[GrpcAgentService] = []
-
         self.service_router: DynamicAgentRouter = DynamicAgentRouter()
-
+        # Event to notify that we have started serving
+        self.notify_started: threading.Event = threading.Event()
         self.logger.info("tool_registries found: %s", str(list(self.tool_registries.keys())))
```
andreidenissov-cog, line 98: Add an event to sync gRPC server start and http server thread start.

neuro_san/service/agent_server.py
```diff
@@ -186,4 +193,5 @@ def serve(self):
             concierge_service,
             server)
 
+        self.notify_started.set()
         self.server_lifetime.run()
```
andreidenissov-cog, line 197: At this point, gRPC server internal state is initialized, and initial agents repository is constructed.
So tell (in this case, http server) it's OK to start running.

neuro_san/service/agent_main_loop.py
```diff
@@ -154,19 +154,15 @@ def main_loop(self):
...
-        http_server_process.start()
-
-        try:
-            self.server.serve()
-        finally:
-            http_server_process.terminate()
-            http_server_process.join()
+        http_server_thread = threading.Thread(target=http_sidecar, daemon=True)
+        http_server_thread.start()
+
```
andreidenissov-cog, line 164: Our http server thread is "daemon" - it will be terminated by Python when "main" server thread is done.

andreidenissov-cog: On the start, http_server_thread will wait for gRPC server to signal that it's ready.

neuro_san/http_sidecar/http_sidecar.py
```diff
@@ -102,9 +110,16 @@ def make_app(self):
...
-    def update_agents(self, agents: List[str]):
+    def update_agents(self):
+        data: Dict[str, Any] = {}
+        session: ConciergeSession = DirectConciergeSession(metadata={})
+        agents_dict: Dict[str, List[Dict[str, str]]] = session.list(data)
+        agents_list: List[Dict[str, str]] = agents_dict["agents"]
+        agents: List[str] = []
+        for agent_dict in agents_list:
+            agents.append(agent_dict["agent_name"])
         with self.lock:
```
andreidenissov-cog, line 129: We can do updating "allowed" agents right here, and using  DirectConciergeSession too - this is now available,
since we share process space.

neuro_san/http_sidecar/http_sidecar.py
```diff
@@ -65,20 +69,24 @@ def __init__(self, port: int, http_port: int,
...
         self.lock = threading.Lock()
         self.logger = HttpLogger(self.forwarded_request_metadata)
-
         app = self.make_app()
+
+        event_set = self.start_event.wait(timeout=self.TIMEOUT_TO_START_SECONDS)
+        if not event_set:
+            self.logger.error({}, "Timeout (%d sec) waiting for HTTP server to start", self.TIMEOUT_TO_START_SECONDS)
+            return
+
```
andreidenissov-cog, line 85: Wait for "go" signal which will be set by gRPC server and corresponding machinery.

neuro_san/http_sidecar/http_sidecar.py
```diff
@@ -38,25 +39,28 @@ class HttpSidecar(AgentAuthorizer, AgentsUpdater):
...
-    # pylint: disable=too-many-arguments, too-many-positional-arguments
     # pylint: disable=too-many-instance-attributes
-    def __init__(self, port: int, http_port: int,
-                 agents: Dict[str, Any],
+    # pylint: disable=too-many-arguments, too-many-positional-arguments
+
+    TIMEOUT_TO_START_SECONDS: int = 10
+
+    def __init__(self, start_event: threading.Event,
+                 port: int, http_port: int,
```
andreidenissov-cog, line 48: Adding external event as a new parameter.

neuro_san/http_sidecar/handlers/base_request_handler.py
```diff
@@ -121,35 +119,15 @@ def get_agent_grpc_session(self,
...
-        """
-        grpc_session: ConciergeSession = \
-            GrpcConciergeSession(
-                host="localhost",
-                port=self.port,
-                metadata=metadata)
-        return grpc_session
-
-    async def update_agents(self, metadata: Dict[str, Any]) -> bool:
+    async def update_agents(self) -> bool:
```
d1donlydfink: How does the metadata get in here then?

neuro_san/http_sidecar/http_sidecar.py
d1donlydfink: You should move this PR comment into the code.
Not obvious otherwise.

neuro_san/service/agent_main_loop.py
d1donlydfink: Does this thread ever get a signal to stop/join, like when the leaf_server_common infrastructure says
"OK, we reached our AGENT_REQUEST_LIMIT, and now we are going to shut down cleanly for pod rotation". ... which is useful for devops-y testing of the behavior of pods going down and coming back up.

neuro_san/http_sidecar/http_sidecar.py
andreidenissov-cog: OK done.

neuro_san/http_sidecar/handlers/base_request_handler.py
andreidenissov-cog: Well, we don't _really_ need metadata here, since we are going directly to Concierge logic, which is straightforward.
But OK, added it back to keep things uniform.

d1donlydfink: FWIW: CodedTools can use loggers however they want and they expect their metadata to be set up already in the logging infrastructure. 1C guys in particular are dependent on this.

d1donlydfink (approved):

PR #212, opened 2025-05-30
UN-3171 Add sly_data_schema to FunctionResponse

d1donlydfink:
Idea from a conversation with Paul, this ended up being a lot easier than I thought it would be.

Allowing for sly_data_schema to be sent in the Function() response means that a generic client like
MAAUI can prompt for user-specific information in a generic way.  This should increase the utility of what other
people can demo for us in that for specific cases we don't have to provide deploy-time environment variables 
for special accounts.

Worth noting that there is no checking for the required fields of the schema on the server side.

### DIFF
--- a/docs/agent_hocon_reference.md
+++ b/docs/agent_hocon_reference.md
@@ -358,7 +358,7 @@ specific information the agent needs as input arguments over the private sly_dat
 channel when it is called.  The sly_data itself is generally considered to be private information
 that does not belong in the chat stream, for example: credential information.
 
-The sly_data schema specification here has the same format as the [parameters](#parameters)
+The sly_data_schema specification here has the same format as the [parameters](#parameters)
 schema definition above.  Ideally there should be one [properties](#properties) entry per
 sly_data dictionary input key, and any absolutely necessary keys should be listed in the [required](#required)
 list.
@@ -370,7 +370,7 @@ information before sending any chat input.
 The front-man is the only agent node that ever needs to specify this aspect of the [function](#function)
 definition, as sly_data itself is already visible to all other internal agents of the network.
 
-Example networks that advertise their sly_data schema:
+Example networks that advertise their sly_data_schema:
 - [math_guy.hocon](../neuro_san/registries/math_guy.hocon)
 
 
--- a/docs/agent_hocon_reference.md
+++ b/docs/agent_hocon_reference.md
@@ -40,6 +40,7 @@ Sub-keys to those dictionaries will be described in the next-level down heading 
       - [type](#type)
       - [properties](#properties)
       - [required](#required)
+    - [sly_data](#sly_data)
   - [instructions](#instructions)
   - [command](#command)
   - [tools (agents)](#tools-agents)
@@ -51,11 +52,11 @@ Sub-keys to those dictionaries will be described in the next-level down heading 
   - [allow](#allow)
     - [connectivity](#connectivity)
     - [to_downstream](#to_downstream)
-      - [sly_data](#sly_data)
-    - [from_downstream](#from_downstream)
       - [sly_data](#sly_data-1)
-    - [to_upstream](#to_upstream)
+    - [from_downstream](#from_downstream)
       - [sly_data](#sly_data-2)
+    - [to_upstream](#to_upstream)
+      - [sly_data](#sly_data-3)
   - [display_as](#display_as)
   - [max_message_history](#max_message_history)
   - [error_formatter](#error_formatter-1)
@@ -350,6 +351,29 @@ This is an optional list of string keys in the [properties](#properties) diction
 to be required whenever an upstream agent calls the one being described.
 Note that it's possible to specify a default value for any property that is not listed as required.
 
+#### sly_data
+
+The optional [JSON Schema](https://json-schema.org) dictionary describing what
+specific information the agent needs as input arguments over the private sly_data dictionary
+channel when it is called.  The sly_data itself is generally considered to be private information
+that does not belong in the chat stream, i.e. credential information.
+
+The sly_data schema specification here has the same format as the [parameters](#parameters)
+schema definition above.  Ideally there should be one [properties](#properties) entry per
+sly_data dictionary key, and any absolutely necessary keys should be listed in the [required](#required)
+list.
+
+Note that it is not strictly necessary to advertise to the outside world the sly_data that
+your agent network requires, but doing so does allow generic clients to prompt for this extra
+information before sending any chat input.
+
+The front-man is the only agent node that ever needs to specify this aspect of the [function](#function)
+definition, as sly_data is already visible to all other internal agents of the network.
+
+Example networks that advertise their sly_data schema:
+- [math_guy.hocon](../neuro_san/registries/math_guy.hocon)
+
+
 ### instructions
 
 When included, the single (often very long) string value here tells an LLM-enabled agent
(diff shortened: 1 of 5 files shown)


docs/agent_hocon_reference.md
```diff
@@ -51,11 +52,11 @@ Sub-keys to those dictionaries will be described in the next-level down heading
...
     - [connectivity](#connectivity)
     - [to_downstream](#to_downstream)
-      - [sly_data](#sly_data)
-    - [from_downstream](#from_downstream)
       - [sly_data](#sly_data-1)
-    - [to_upstream](#to_upstream)
+    - [from_downstream](#from_downstream)
       - [sly_data](#sly_data-2)
+    - [to_upstream](#to_upstream)
+      - [sly_data](#sly_data-3)
```
d1donlydfink: Confusing diff, but the addition of the extra sly_data key above prompts reordering down below.

docs/agent_hocon_reference.md
```diff
@@ -350,6 +351,29 @@ This is an optional list of string keys in the [properties](#properties) diction
...
+
+Note that it is not strictly necessary to advertise to the outside world the sly_data that
+your agent network requires, but doing so does allow generic clients to prompt for this extra
+information before sending any chat input.
+
+The front-man is the only agent node that ever needs to specify this aspect of the [function](#function)
+definition, as sly_data is already visible to all other internal agents of the network.
+
+Example networks that advertise their sly_data schema:
+- [math_guy.hocon](../neuro_san/registries/math_guy.hocon)
```
d1donlydfink, line 374: Add a blurb in the docs about specifying sly_data schema.

neuro_san/api/grpc/agent.proto
```diff
@@ -74,10 +74,17 @@ message Function {
     // Outward-facing description of what the agent does.
     string description = 1 [json_name="description"];
 
-    // Optional map of parameters that the agent needs in order to work.
+    // Optional map of parameters passed in via the natural-language chat text channel
+    // that the agent needs in order to work.
     // This is really a pydantic/OpenAI function description, which is
     // a bit too complex to specify directly in protobuf.
     google.protobuf.Struct parameters = 2 [json_name="parameters"];
```
d1donlydfink, line 81: Emphasize that parameters is what is need in the chat/text channel.

neuro_san/api/grpc/agent.proto
```diff
@@ -74,10 +74,17 @@ message Function {
...
+
+    // Optional map of parameters passed in via the sly_data dictionary private data channel
+    // that the agent needs in order to work.
+    // This is really a pydantic/OpenAI function description, which is
+    // a bit too complex to specify directly in protobuf.
+    google.protobuf.Struct sly_data = 3 [json_name="sly_data"];
```
d1donlydfink: Add sly_data schema to the Function spec returned.

Interesting to note that there are no session implementation changes required as we just pick up the function dictionary from the agent spec and send it over.  Here we simply define the valid fields.

neuro_san/registries/math_guy.hocon
```diff
@@ -54,6 +54,24 @@ on the secret numbers.
...
+                        "x": {
+                            "type": "float",
+                            "description": "The first operand for the arithmetic operation"
+                        },
+                        "y": {
+                            "type": "float",
+                            "description": "The second operand for the arithmetic operation"
+                        }
+                    },
+                    "required": ["x", "y"]
```
d1donlydfink, line 74: Add an example of sly_data_schema to math_guy.

kxk302 (approved):
LGTM

hshahrzad (approved):
Looks good to me.

docs/agent_hocon_reference.md
```diff
@@ -350,6 +351,29 @@ This is an optional list of string keys in the [properties](#properties) diction
 to be required whenever an upstream agent calls the one being described.
 Note that it's possible to specify a default value for any property that is not listed as required.
 
+#### sly_data
+
+The optional [JSON Schema](https://json-schema.org) dictionary describing what
+specific information the agent needs as input arguments over the private sly_data dictionary
+channel when it is called.  The sly_data itself is generally considered to be private information
+that does not belong in the chat stream, i.e. credential information.
+
```
andreidenissov-cog, line 360: "for example" credential information?
Because right now it reads sly_data == credentials

d1donlydfink: Good suggestion. Adopted.

andreidenissov-cog (dismissed):
LGTM.

vince-leaf (approved):
LGTM

neuro_san/registries/math_guy.hocon
dsargent: So this means if we were to call the http API, asking for Function info on this agent, for `sly_data` we'd get something back like:

```json
  "x": {
    "type": "float",
    "description": "The first operand for the arithmetic operation"
  },
  "y": {
    "type": "float",
    "description": "The second operand for the arithmetic operation"
  }
```

?

That seems workable. I like that it has types.

neuro_san/api/grpc/agent.proto
dsargent: Maybe `sly_data_schema` since it's the schema for the sly_data, not the actual data? Could get confused with the actual sly_data otherwise.

dsargent (dismissed):
LGTM, one suggestion + one question

neuro_san/api/grpc/agent.proto
ofrancon: Good point - `sly_data` is confusing. `sly_data_schema` or `expected_sly_data_schema` would be better

neuro_san/registries/math_guy.hocon
```diff
@@ -54,6 +54,24 @@ on the secret numbers.
                         }
                     },
                     "required": ["operator"]
+                },
+
+                # You can also optionally specify the schema needed for any sly_data.
+                # This is to allow generic clients to prompt for private-data inputs
+                # that do not belong in the chat stream.
+                "sly_data": {
```
ofrancon: `expected_sly_data` or `expected_sly_data_schema` would make things less confusing. Otherwise I'd consider this is actual sly data values that are passed to the agent as "constants".

ofrancon (approved):
Suggesting to use `expected_sly_data` or `expected_sly_data_schema` in `agent.proto` to make things less confusing.

neuro_san/api/grpc/agent.proto
dsargent: I would say one or the other -- `expected` or `schema` since they mean the same thing. Not both.

Noravee (approved):
lgtm.

neuro_san/registries/math_guy.hocon
d1donlydfink: You would actually get a little more than that:

```
{"type": "object", "properties": {"x": {"type": "float", "description": "The first operand for the arithmetic operation"}, "y": {"type": "float", "description": "The second operand for the arithmetic operation"}}, "required": ["x", "y"]}```

This follows the JSON schema spec.  Note also the `required` list, some parameters might not be in there.

neuro_san/api/grpc/agent.proto
d1donlydfink: I think I will go with sly_data_schema.
Thanks for the critique

d1donlydfink: Done.

neuro_san/registries/math_guy.hocon
d1donlydfink: Changed to sly_data_schema

dsargent (approved):

PR #215, opened 2025-05-31
UN-3137 Switch to janus sync/async queue in SessionInvocationContext

andreidenissov-cog:
This PR switches async queues we use from "standard" asyncio.Queue to third-party implementation.
Why is this needed?
For http-only service stack, we need to stream results of chat request from request-specific event loop (running in a request-specific separate thread) to "main" Tornado server event loop. As directly as possible, with no extra sync/async translations along the way.
Turns out asyncio.Queue is not suitable for this, because it cannot work with event loop other than the one it was created in and bound to.
janus.Queue is designed specifically to bridge this inter-loops gap. What is good is that it is backward compatible with current neuro-san setup - everything still works.

Tested: by usual stability test.

### DIFF
--- a/neuro_san/http_sidecar/handlers/base_request_handler.py
+++ b/neuro_san/http_sidecar/handlers/base_request_handler.py
@@ -119,15 +119,16 @@ class BaseRequestHandler(RequestHandler):
                 agent_name=agent_name)
         return grpc_session
 
-    async def update_agents(self) -> bool:
+    async def update_agents(self, metadata: Dict[str, Any]) -> bool:
         """
         Update internal agents table by executing request
         to underlying gRPC service.
+        :param metadata: metadata to be used for logging if necessary.
         :return: True if update was successful
                  False otherwise
         """
         try:
-            self.agents_updater.update_agents()
+            self.agents_updater.update_agents(metadata=metadata)
             return True
         except Exception as exc:  # pylint: disable=broad-exception-caught
             self.process_exception(exc)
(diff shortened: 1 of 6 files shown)


neuro_san/session/session_invocation_context.py
```diff
@@ -121,6 +121,8 @@ def close(self):
         if self.asyncio_executor is not None:
             self.asyncio_executor.shutdown()
             self.asyncio_executor = None
+        if self.queue is not None:
+            self.queue.close()
```
andreidenissov-cog, line 126: janus queue has to be closed explicitly (they say).

requirements.txt
```diff
@@ -53,6 +53,9 @@ httpx==0.27.2
 # HTTP server part
 tornado>=6.4.2
 
+# Sync/async queue
+janus >= 2.0.0
+
```
andreidenissov-cog, line 58: Add janus requirement.

neuro_san/internals/chat/async_collating_queue.py
```diff
@@ -60,7 +60,7 @@ async def __anext__(self) -> Dict[str, Any]:
                 Will throw StopAsyncIteration when the final item is detected
                 via the is_final_item() method..
         """
-        message = await self.queue.get()
+        message = await self.queue.async_q.get()
         if self.is_final_item(message):
```
andreidenissov-cog, line 64: This is how janus queue is special: for getting things _from_ a queue,
we use async get from async "side" of this queue.

neuro_san/internals/chat/async_collating_queue.py
```diff
@@ -72,7 +72,7 @@ async def put(self, item: Any):
 
         :param item: The item to put on the queue.
         """
-        await self.queue.put(item)
+        self.queue.sync_q.put(item)
```
andreidenissov-cog, line 76: And for putting things to a queue, we use sync put from sync "side" of this queue.
We can afford here to use sync operation in request-specific event loop,
because queue "put" is fast and "almost guaranteed" to succeed.

d1donlydfink (approved):
Cool. I was not aware of janus stuff before.

PR #219, opened 2025-06-03
Fix health check request handler.

andreidenissov-cog:
This PR fixes health check request handler.
Since we are defining our custom "default" Tornado request logger,
several things have to be added to any request handler,
namely self.logger and self.get_metadata() method.


d1donlydfink (approved):
LGTM AFAICT

PR #229, opened 2025-06-07
UN-3189 Fix non-consecutive system message and maybe save some tokens too

d1donlydfink:
Note that the diff is vs. @vince-leaf 's test case branch.

* Fixing the root cause of the problem has to do with not sending extra messages when we are already sending them in chat history.  This could save everyone tokens and potentially time as well - though time part is unclear.
* Some other tweaks for dealing with messages from Claude
* Make some of the fixes I recommended in Vince's PR here like:
    * rename accounting.py -> accountant.py to have the file name match the class inside. Rename a test to reflect this change as well.
    * Remove http and grpc test cases for anthropic.  We do not need to test all LLM Providers against all connections.
* Some extra fixes here:
    * Import the most recent instructions for music_nerd_pro from neuro-san-studio. This helps reduce the temptation to call tools twice (though not always)
    * Remove the smoke_llm_anthropic pytest label from vince's tests and use plain-old smoke instead. Without this, they are taken for unit tests and checkin tests never finished because of the server-ness.

Now, with all this going in, this does indeed fix the original problem of the "Non-consecutive system message".
However, the test still seems to like to call the accountant more than once and likely would not pass any check on the value under Anthropic Claude Haiku. (OpenAI is still fine tho). Some ideas here:
* @ofrancon has some mojo for using sly_data here. Maybe switch all the music nerd pro tests for all LLMs to that?
* Per @ofrancon, It's possible a beefier Claude model like sonnet or opus will do better at greater cost.  
Will leave all this to @vince-leaf


neuro_san/internals/chat/chat_history_message_processor.py
```diff
@@ -115,6 +116,9 @@ def escape_message(self, chat_message_dict: Dict[str, Any]) -> Dict[str, Any]:
         transformed: Dict[str, Any] = copy(chat_message_dict)
         text: str = transformed.get("text")
 
+        if text is None:
+            return None
```
d1donlydfink, line 120: Claude has a tendency to produce messages with no text value in them.  Don't include those in the history.

neuro_san/internals/chat/chat_history_message_processor.py
```diff
@@ -95,7 +95,8 @@ def process_message(self, chat_message_dict: Dict[str, Any], message_type: ChatM
             # Transform the message with properly escaped text.
             transformed_message_dict = self.escape_message(chat_message_dict)
 
-        self.message_history.append(transformed_message_dict)
+        if transformed_message_dict is not None:
+            self.message_history.append(transformed_message_dict)
```
d1donlydfink, line 99: Claude has a tendency to produce messages with no text value in them. Don't include those in the history.

neuro_san/internals/messages/message_utils.py
```diff
@@ -185,7 +185,10 @@ def convert_to_chat_message(message: BaseMessage, origin: List[Dict[str, Any]] =
         # If it is a list, "text" is a key of a dictionary which is the first element of
         # the list. For more details: https://python.langchain.com/docs/integrations/chat/anthropic/#content-blocks
         if src == "content" and isinstance(value, list):
-            value = value[0].get("text", "")
+            if len(value) > 0:
+                value = value[0].get("text", "")
+            else:
+                value = None
```
d1donlydfink, line 191: I have gotten empty lists before from Claude as the value.

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
```diff
@@ -370,10 +369,10 @@ async def _create_prompt_template(self, instructions: str, assignments: str) ->
         # Assemble the prompt message list
         message_list: List[Tuple[str, str]] = []
 
+        system_message = SystemMessage(instructions)
         if len(self.chat_history) == 0:
             # We have not had any chat history just yet, so build it from scratch
             # Add to our own chat history which is updated in write_message()
-            system_message = SystemMessage(instructions)
```
d1donlydfink, line 376: Move this up top for use by both if and else blocks.

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
```diff
@@ -383,22 +382,10 @@ async def _create_prompt_template(self, instructions: str, assignments: str) ->
...
-                    message_tuple = ("system", instructions)
-                else:
-                    message_tuple = convert_to_message_tuple(base_message)
-
-                if message_tuple is not None:
-                    message_list.append(message_tuple)
+            # Chat history already contains what we need and it is
+            # sent as part of the first placeholder message.
+            # However, we always want ot use the most current instructions.
+            self.chat_history[0] = system_message
```
d1donlydfink, line 388: What we had been doing before is constructing a message list based on the chat history.
But then just after this we included the chat history as a parameter, which basically could send every message again to the LLM.

This was the source of the Claude "non-consecutive system message" problem.

Fixing this could save us tokens and time, so very good find overall!

We still want to ensure that our system instructions are what the agent wants and not what is sent
over by the user, so revise the chat history for those cases for insurance (just like we did before).

neuro_san/registries/music_nerd_pro.hocon
```diff
@@ -54,21 +54,7 @@ You’re equal parts playlist curator, music historian, and pop culture mythbust
...
-        - "First, give your full music answer as plain text."
-
-        - "Then, on a separate line, return a valid JSON object with:"  
-
-            - "the key "running_cost" (with double quotes), and:
-
-            - "the value as a number with exactly one decimal place, without double quotes."
-        
-        example: "running_cost": 1.0
+Return your answer and the running cost in a JSON message.
```
d1donlydfink, line 57: Copy the newer verbiage from neuro-san-studio. More consistent results.

neuro_san/registries/music_nerd_pro.hocon
```diff
@@ -90,7 +76,7 @@ to the API to get the updated running cost. If no running cost it known, pass 0.
                     "required": ["running_cost"]
                 }
             },
-            "class": "accounting.Accountant",
+            "class": "accountant.Accountant"
```
d1donlydfink, line 79: Rename the source file to be consistent with the name of the class within.

neuro_san/registries/music_nerd_pro_llm_anthropic.hocon
```diff
@@ -26,7 +26,7 @@
         # he is the only one with no parameters in his function definition,
         # and therefore he needs to talk to the outside world to get things rolling.
         {
-            "name": "MusicNerdPro_LLM_Anthropic",
+            "name": "MusicNerdPro",
```
d1donlydfink, line 29: FWIW: Name of front-man does not necessarily have to reflect that of the hocon file.
This will be easier for any future copy/paste.

tests/fixtures/music_nerd_pro_llm_anthropic/combination_responses_with_history_direct.hocon
```diff
@@ -14,9 +14,6 @@
     # Describes what agent to test against.
     "agent": "music_nerd_pro_llm_anthropic",
 
-    # Connect to the agent via a server
-    "connections": ["direct"],
```
d1donlydfink, line 18: Don't need to specify the connection for direct.

tests/fixtures/music_nerd_pro_llm_anthropic/combination_responses_with_history_grpc.hocon
d1donlydfink, line 1: Don't need a grpc version as we don't need to test all combinations of LLM Provider vs all combinations of connection.

tests/neuro_san/hocon_tests/smoke/music_nerd_pro/test_smoke_test_hocons_llm_anthropic.py
```diff
@@ -34,12 +34,10 @@ class TestSmokeTestHocons(TestCase):
         # Ideally more basic functionality will come first.
         # Barring that, try to stick to alphabetical order.
         "music_nerd_pro_llm_anthropic/combination_responses_with_history_direct.hocon",
-        "music_nerd_pro_llm_anthropic/combination_responses_with_history_grpc.hocon",
-        "music_nerd_pro_llm_anthropic/combination_responses_with_history_http.hocon",
```
d1donlydfink, line 38: Remove these tests. Don't need to test llm provider vs connection type.

tests/neuro_san/hocon_tests/smoke/music_nerd_pro/test_smoke_test_hocons_llm_anthropic.py
```diff
@@ -34,12 +34,10 @@ class TestSmokeTestHocons(TestCase):
...
 
         # List more hocon files as they become available here.
     ]))
-    @pytest.mark.smoke_llm_anthropic
+    @pytest.mark.smoke
```
d1donlydfink, line 40: Rename the pytest mark to be smoke so main GitHub Action does not spin on trying to do smoke tests as unit tests.

tests/neuro_san/coded_tools/music_nerd_pro/test_accountant.py
```diff
@@ -1,11 +1,11 @@
 from unittest import TestCase
 
-from neuro_san.coded_tools.music_nerd_pro.accounting import Accountant
+from neuro_san.coded_tools.music_nerd_pro.accountant import Accountant
 
 
-class TestAccounting(TestCase):
+class TestAccountant(TestCase):
```
d1donlydfink, line 6: Rename to reflect class name that is tested

ofrancon (approved):
lgtm, thanks!

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
Noravee: We can save more tokens by ensuring the chat history does not include the latest human message, since it's already passed separately as input.

Noravee (approved):
LGTM.

neuro_san/registries/music_nerd_pro_llm_anthropic.hocon
vince-leaf: I started out using this "MusicNerdPro," but there were some mismatches due to the test case always using the OPENAI key. So, I went with the name above to make the assessment easier.

vince-leaf (approved):
LGTM: I will deal with this one test case class name on my side.

PR #232, opened 2025-06-10
UN-3017 Convert outgoing chat messages to expected format.

andreidenissov-cog:
This PR converts all ChatMessageType enum values in outgoing chat responses to strings
as expected by UI clients.
Will work also for agent_cli client, though it can already handle ChatMessageType int values if they do come from a server.
Tested by observing chat response messages being sent to output
with ChatMessageType values represented as strings.

### DIFF
--- a/neuro_san/service/chat_message_converter.py
+++ b/neuro_san/service/chat_message_converter.py
@@ -1,0 +1,49 @@
+
+# Copyright (C) 2023-2025 Cognizant Digital Business, Evolutionary AI.
+# All Rights Reserved.
+# Issued under the Academic Public License.
+#
+# You can be released from the terms, and requirements of the Academic Public
+# License by purchasing a commercial license.
+# Purchase of a commercial license is mandatory for any use of the
+# neuro-san SDK Software in commercial settings.
+#
+# END COPYRIGHT
+from typing import Any
+from typing import Dict
+from neuro_san.internals.messages.chat_message_type import ChatMessageType
+
+
+class ChatMessageConverter:
+    """
+    Helper class to prepare chat response messages
+    for external clients consumption.
+    """
+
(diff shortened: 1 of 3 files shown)


neuro_san/service/chat_message_converter.py
```diff
@@ -0,0 +1,49 @@
...
+        message_dict: Dict[str, Any] = response_dict.get('response', None)
+        if message_dict is not None:
+            ChatMessageConverter.convert_message(message_dict)
+
+    @classmethod
+    def convert_message(cls, message_dict: Dict[str, Any]):
+        """
+        Convert chat message to a format expected by external clients:
+        :param message_dict: chat message to process
+        """
```
andreidenissov-cog, line 47: We go down ChatMessage structure recursively to handle all ChatMessageType instances.

neuro_san/service/chat_message_converter.py
```diff
@@ -0,0 +1,49 @@
...
+# Purchase of a commercial license is mandatory for any use of the
+# neuro-san SDK Software in commercial settings.
+#
+# END COPYRIGHT
+from typing import Any
+from typing import Dict
+from neuro_san.internals.messages.chat_message_type import ChatMessageType
+
+
+class ChatMessageConverter:
```
d1donlydfink: Derive from DictionaryConverter.
That guy has better semantics where it doesn't touch the original dictionary which might actually be beneficial in these cases.

d1donlydfink (changes requested):
Derive from DictionaryConverter.
That guy has better semantics where it doesn't touch the original dictionary which might actually be beneficial in these cases.

Also: How did you test?  Http and Grpc and Direct?

d1donlydfink (approved):

PR #237, opened 2025-06-12
Put in a delay to help force output buffer flush.

andreidenissov-cog:
This is basically a hack to help Tornado async streaming force out output http buffers
and ensure that client will get "one response per http chunk".
That seems to work, at least until UI client knows how to handle "multiple responses per chunk",
which would be much more reliable.
In case of emergency: play with delay duration.

Tested with MAA UI against local neuro-san server:
music-nerd-pro, esp-assistant, web-search. 
No amnesia noted.

BTW: putting in time.sleep(0) doesn't seem to help.

### DIFF
--- a/neuro_san/http_sidecar/handlers/base_request_handler.py
+++ b/neuro_san/http_sidecar/handlers/base_request_handler.py
@@ -20,9 +20,9 @@ from typing import Tuple
 
 import json
 import os
+import asyncio
 
 import grpc
-import asyncio
 
 import tornado
 from tornado.web import RequestHandler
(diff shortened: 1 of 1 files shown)


neuro_san/http_sidecar/handlers/base_request_handler.py
```diff
@@ -191,6 +192,7 @@ async def do_flush(self) -> bool:
         """
         try:
             await self.flush()
+            await asyncio.sleep(0.3)
```
d1donlydfink, line 207: What's missing is a comment as to why we are doing this here at this place in the code and why 0.3 and not 0.0 or 0.1.
Your future self could easily want to remove that without appropriate words of caution describing what is at stake.  Some one-liners deserve a paragraph when they are this subtle.

andreidenissov-cog: Added comment in the code.

d1donlydfink (approved):

PR #243, opened 2025-06-13
Use pathlib for agent_cli. Theoretically helps with windows.

d1donlydfink:
Only tackling client for now. Server could very well be more involved.
We are a linux shop with no windows to test on.
There is always the sample container under neuro_san/deploy for windows folks desparate for a server.

### DIFF
--- a/neuro_san/client/thinking_file_message_processor.py
+++ b/neuro_san/client/thinking_file_message_processor.py
@@ -32,6 +32,14 @@ class ThinkingFileMessageProcessor(MessageProcessor):
     def __init__(self, thinking_file: str, thinking_dir: str):
         """
         Constructor
+
+        :param thinking_file: A string representing the path to a single file where
+                              all agent output is combined.  We no longer recommmend this
+                              now that we have ...
+        :param thinking_dir: A string representing the path to a single directory where
+                             each agent in the network gets its own history file according
+                             to its agent network origin name.  This is much easier to
+                             debug as you do not have to tease apart output from interacting agents.
         """
         self.thinking_file: Path = None
         if thinking_file is not None:


neuro_san/client/agent_cli.py
```diff
@@ -88,7 +89,8 @@ def chat(self):
             #           recognize pathlib as a valid library with which to resolve these kinds
             #           of issues.  Furthermore, this is a client command line tool that is never
             #           used inside servers which just happens to be part of a library offering.
-            with open(self.args.first_prompt_file, 'r', encoding="utf-8") as prompt_file:
+            prompt: Path = Path(self.args.first_prompt_file)
+            with prompt.open('r', encoding="utf-8") as prompt_file:
```
d1donlydfink, line 93: Use path lib wherever possible. Theoretically this is safer on windows.
Only tackling client for now.

neuro_san/client/thinking_file_message_processor.py
```diff
@@ -32,8 +33,13 @@ def __init__(self, thinking_file: str, thinking_dir: str):
         """
         Constructor
         """
-        self.thinking_file: str = thinking_file
```
ofrancon, line 35: If you could document the constructor to describe what `thinking_file` and `thinking_dir` are that would be useful too.

ofrancon (approved):
Should do no harm

neuro_san/client/thinking_file_message_processor.py
d1donlydfink: Added comment per your suggestion.

neuro_san/client/thinking_file_message_processor.py
```diff
@@ -31,9 +32,22 @@ class ThinkingFileMessageProcessor(MessageProcessor):
...
         """
         Constructor
+
+        :param thinking_file: A string representing the path to a single file where
+                              all agent output is combined.  We no longer recommmend this
+                              now that we have ...
+        :param thinking_dir: A string representing the path to a single directory where
+                             each agent in the network gets its own history file according
+                             to its agent network origin name.  This is much easier to
+                             debug as you do not have to tease apart output from interacting agents.
```
d1donlydfink, line 42: Add comment per @ofrancon 's suggestion.

ofrancon (approved):
Thanks for the docsting

PR #244, opened 2025-06-13
UN-3237 Fix dynamic manifest updating for http service path.

andreidenissov-cog:
This PR fixes a problem with dynamic manifest/agent .hocon files updates
when using http service stack.
Problem surfaced now because previously we were always going through gRPC service, which works.
Now direct http route ran into this issue.

Fix tested by: modifying hello_world.hocon file for "function" request,
enabling/disabling hello_world + music_nerd agents in manifest.hocon
and observing results from CLI client requests.
gRPC route continues to work.

### DIFF
--- a/neuro_san/http_sidecar/http_sidecar.py
+++ b/neuro_san/http_sidecar/http_sidecar.py
@@ -160,7 +160,8 @@ class HttpSidecar(AgentAuthorizer, AgentsUpdater):
                 if self.allowed_agents.get(agent_name, None) is None:
                     self.add_agent(agent_name)
             # All other agents are disabled:
-            for agent_name, _ in self.allowed_agents.items():
+            allowed_set = set(self.allowed_agents.keys())
+            for agent_name in allowed_set:
                 if agent_name not in agents:
                     self.remove_agent(agent_name)
 
--- a/neuro_san/internals/tool_factories/service_tool_factory_provider.py
+++ b/neuro_san/internals/tool_factories/service_tool_factory_provider.py
@@ -84,17 +84,32 @@ class ServiceToolFactoryProvider(ToolFactoryProvider):
         Replace agents registries with "registries" collection.
         Previous state could be empty.
         """
-        prev_agents: Dict[str, AgentToolFactory] = {}
-        with self.lock:
-            prev_agents = self.agents_table
-            self.agents_table = {}
-        # Notify listeners that previous set of agents is removed
-        for agent_name, _ in prev_agents.items():
-            for listener in self.listeners:
-                listener.agent_removed(agent_name)
-        # Add the new agent+registry pairs
-        for agent_name, tool_registry in registries.items():
-            self.add_agent_tool_registry(agent_name, tool_registry)
+        current_agents = set(self.agents_table.keys())
+        new_agents = set(registries.keys())
+        # Remove agents which are not in the new collection:
+        agents_to_remove = current_agents - new_agents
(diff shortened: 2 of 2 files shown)


neuro_san/http_sidecar/http_sidecar.py
```diff
@@ -160,7 +160,8 @@ def update_agents(self, metadata: Dict[str, Any]):
                 if self.allowed_agents.get(agent_name, None) is None:
                     self.add_agent(agent_name)
             # All other agents are disabled:
-            for agent_name, _ in self.allowed_agents.items():
+            allowed_set = set(self.allowed_agents.keys())
+            for agent_name in allowed_set:
                 if agent_name not in agents:
                     self.remove_agent(agent_name)
```
andreidenissov-cog, line 166: Do not remove elements from a dictionary while iterating it.

neuro_san/internals/tool_factories/service_tool_factory_provider.py
```diff
@@ -84,17 +84,15 @@ def setup_tool_registries(self, registries: Dict[str, AgentToolFactory]):
...
-            self.add_agent_tool_registry(agent_name, tool_registry)
+        current_agents = set(self.agents_table.keys())
+        new_agents = set(registries.keys())
+        # Remove agents which are not in the new collection:
+        agents_to_remove = current_agents - new_agents
+        for agent_name in agents_to_remove:
+            self.remove_agent_tool_registry(agent_name)
+        # Now add (or possibly replace) agents from new collection:
+        for agent_name in new_agents:
+            self.add_agent_tool_registry(agent_name, registries[agent_name])
```
andreidenissov-cog, line 96: That's a rather unexpected one: we should not change a value of self.agents_table (_pointer_ to dictionary)
since this value already went downstream into instances of SingleAgentToolFactoryProvider.
We should only change the contents of this dictionary, this way all instances of SingleAgentToolFactoryProvider
will still access agents table we have just now modified, and so will see the latest agents definitions.

neuro_san/internals/tool_factories/service_tool_factory_provider.py
```diff
@@ -84,17 +84,15 @@ def setup_tool_registries(self, registries: Dict[str, AgentToolFactory]):
         Replace agents registries with "registries" collection.
         Previous state could be empty.
         """
-        prev_agents: Dict[str, AgentToolFactory] = {}
-        with self.lock:
-            prev_agents = self.agents_table
-            self.agents_table = {}
-        # Notify listeners that previous set of agents is removed
```
andreidenissov-cog, line 91: This line is what caused the problem - assignment to self.agents_table

neuro_san/internals/tool_factories/service_tool_factory_provider.py
```diff
@@ -84,17 +84,15 @@ def setup_tool_registries(self, registries: Dict[str, AgentToolFactory]):
...
-        for agent_name, _ in prev_agents.items():
-            for listener in self.listeners:
-                listener.agent_removed(agent_name)
```
d1donlydfink, line 94: Previously there was a call to listener.agent_removed().
Does that now happen in remove_agent_tool_registry()?

d1donlydfink: Ah. Looking at code not in PR, yes it does.

d1donlydfink (approved):
After some questions over slack, LGTM

PR #245, opened 2025-06-16
UN-3230 Provide github actions support for non-default llm providers

donn-leaf:


.github/workflows/integration.yml
```diff
@@ -25,7 +25,7 @@ jobs:
         run: pip freeze
 
 
-      - name: Run pytest (excluding integration tests)
+      - name: Run pytest integration tests
```
donn-leaf, line 28: Just cleaned up some old verbiage.

.github/workflows/smoke.yml
```diff
@@ -28,6 +28,7 @@ jobs:
         shell: bash
         env:
           OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+          ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}
```
donn-leaf, line 31: Here we add the anthropic key to the env. But notice, in conftest.py (below) there is logic that conditionally _removes_ the unneeded key based on the label of the test. In this way we don't accidentally use the default key.

.github/workflows/tests.yml
```diff
@@ -42,6 +42,7 @@ jobs:
         run: pytest --verbose -m "not integration and not smoke and not e2e" --timer-top-n 10
         env:
           OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+          ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}
```
donn-leaf: Same comment as above. Note in this case we don't have anything in "tests.yml" that is using anthropic (yet). This line is future-proofing.

pytest.ini
```diff
@@ -4,6 +4,8 @@
 markers =
     integration: Mark a test as an integration test.  These generally take > 30 seconds to complete.
     smoke: Mark a test as an smoke test. These often exercise the full stack
+    non_default_llm_provider: tests that exercise llm providers other than our default Open AI
```
donn-leaf, line 7: This is important to know about. Say you want to add a new provider, you'll need to add that here, and in your marks within the actual test(s).

tests/conftest.py
```diff
@@ -0,0 +1,24 @@
...
+    """Ensure only the appropriate LLM provider keys are available for the test being run."""
+
+    is_non_default = request.node.get_closest_marker("non_default_llm_provider")
+    is_anthropic = request.node.get_closest_marker("anthropic")
+
+    if is_non_default:
+        # For any non-default provider: clear OPENAI key to prevent accidental use
+        monkeypatch.delenv("OPENAI_API_KEY", raising=False)
+
+        if is_anthropic:
```
donn-leaf, line 16: As we add more non-default llm providers, we'll need to update this logic.

tests/conftest.py
donn-leaf, line 1: This file provides logic that sets env vars according to which llm provider is used in various tests. If there is no specific llm provide mark provided, open ai is assumed as the default. When non-default provider is used, the appropriate env var is set.

tests/neuro_san/hocon_tests/smoke/music_nerd_pro/test_smoke_test_hocons_llm_anthropic.py
```diff
@@ -39,6 +39,8 @@ class TestSmokeTestHocons(TestCase):
     ]))
     @pytest.mark.timeout(30)  # 30 seconds for this test
     @pytest.mark.smoke
+    @pytest.mark.non_default_llm_provider
+    @pytest.mark.anthropic
```
donn-leaf, line 43: This is important to know. Add these marks to your tests for non-default-llm-providers.

.github/workflows/tests.yml
d1donlydfink: If we are adding the new marker, this raises the question as to when we want the marker to run.
Right now it will run with every checkin because the above pytest line lacks the "not non_default_llm_provider" line.  

It probably makes sense to have this as part of the integration or smoke test, but definitely not the every-checkin tests.   

My 2 cents: As for the other 2 kinds of tests, feels good enough for me to have these non_default_llm_provider tests run once a night, which puts them in the integration test category.

So that all said, maybe we don't need a new pytest designation, but maybe we just need to agree on the correct one out of all the existing designations.

tests/neuro_san/hocon_tests/smoke/music_nerd_pro/test_smoke_test_hocons_llm_anthropic.py
vince-leaf: @donn-leaf, I was thinking of a similar one when I started organizing the smoke test by model name, but @d1donlydfink switched back.

I have it:
@pytest.mark.smoke
@pytest.mark.smoke-anthropic

vince-leaf: Another thing is if removing "smoke," will we not only trigger a set of smoke tests?

vince-leaf (dismissed):
LGTM

.github/workflows/tests.yml
donn-leaf: I'm not sure I follow. Currently the tests.yml that defines which tests run at checkins is "not integration and not smoke and not e2e". The smoke test is the only place where we have any non-default-llm-provider marks. So as it stands now, when tests.yml runs, no non-default-llm-provider tests are run. If at some point a developer thought it is worth adding a non-default-llm-provider to the standard tests, they could do so by adding the mark to the new test. 

That said, I can always remove the anthropic key from the tests.yml env so any attempt to run non-default should fail.

d1donlydfink (approved):

tests/neuro_san/hocon_tests/smoke/music_nerd_pro/test_smoke_test_hocons_llm_anthropic.py
d1donlydfink: It's worth noting that individual tests in the list are likely each marked with all the same annotations

PR #246, opened 2025-06-16
UN-3254 Add comments about > 1 response per HTTP line

d1donlydfink:


neuro_san/api/grpc/agent.proto
```diff
@@ -49,6 +49,14 @@ service AgentService {
...
     //     copy this whole-cloth to the chat_context of your next StreamingChat
     //     request to continue the conversation.
+    //  3) It is important to note that since this is a streaming API, for HTTP clients:
+    //      a) Any single response will always be on the same line. That is,
+    //         responses will not be broken up across multiple lines in an
+    //         HTTP response.
+    //      b) We cannot yet guarantee that there will be only one streamed response
+    //         per HTTP response line. That is, it is possible for more than one
+    //         response *might* come on a single line if they come quickly enough,
+    //         though this is not the empirically observed norm.
```
d1donlydfink, line 59: Add some more comments to be explicit about some low-level HTTP concerns that @dsargent hast had to grapple with.

neuro_san/api/grpc/agent_pb2_grpc.py
```diff
@@ -69,6 +69,14 @@ def StreamingChat(self, request, context):
...
         copy this whole-cloth to the chat_context of your next StreamingChat
         request to continue the conversation.
+        3) It is important to note that since this is a streaming API, for HTTP clients:
+        a) Any single response will always be on the same line. That is,
+        responses will not be broken up across multiple lines in an
+        HTTP response.
+        b) We cannot yet guarantee that there will be only one streamed response
+        per HTTP response line. That is, it is possible for more than one
+        response *might* come on a single line if they come quickly enough,
+        though this is not the empirically observed norm.
```
d1donlydfink, line 79: Re-run of the do_generate.sh script.

neuro_san/api/grpc/agent_service.json
```diff
@@ -127,7 +127,7 @@
...
+        "description": "Most important semantics of the streaming:\n  1) The \"answer\" to a query of any agent network is in the *last* streamed\n     AGENT_FRAMEWORK message.\n  2) To RESTfully continue your conversation with the agent network:\n     The very last AGENT_FRAMEWORK message before the stream closes will\n     have its chat_context field filled in with a structure. You can\n     copy this whole-cloth to the chat_context of your next StreamingChat\n     request to continue the conversation.\n  3) It is important to note that since this is a streaming API, for HTTP clients:\n      a) Any single response will always be on the same line. That is,\n         responses will not be broken up across multiple lines in an\n         HTTP response.\n      b) We cannot yet guarantee that there will be only one streamed response\n         per HTTP response line. That is, it is possible for more than one\n         response *might* come on a single line if they come quickly enough,\n         though this is not the empirically observed norm.",
```
d1donlydfink, line 130: Comments get added to the OpenAPI JSON with the do_generate.sh script.

swensel (approved):
Thank you @d1donlydfink

neuro_san/api/grpc/agent.proto
dsargent: I appreciate the intent, though it may be confusing to say this _here_ because protobuf is a _binary_ standard, not text-based, so there won't be newlines. Those get added by other layers (in our case, the web server layer).

I think the most you can say based on the protobuf is that, in line with the protobuf spec, a single "message" may contain from 0 to n `ChatResponse` items.

Other implementations might choose to implement this another way. For example, if you had a gRPC implementation, you wouldn't get newlines but repeated structs.

d1donlydfink: These comments are the basis for the OpenAPI spec. Hence I call out HTTP specifically.

PR #247, opened 2025-06-16
UN-3225 Add library version(s) to health-check

d1donlydfink:
With this change the health-check would now return:
```
{
    "service": "neuro-san agents",
    "status": "healthy",
    "versions": {
        "neuro-san": "0.5.36"
    }
}
```

An open question is do we want to report more than just the neuro-san version?
Can go there if there is demand.

### DIFF
--- a/neuro_san/http_sidecar/handlers/health_check_handler.py
+++ b/neuro_san/http_sidecar/handlers/health_check_handler.py
@@ -11,9 +11,15 @@
 """
 See class comment for details
 """
+from typing import Any
+from typing import Dict
+from typing import List
+
 import http
 import os
-from typing import Any, Dict, List
+
+from importlib.metadata import version as library_version
+from importlib.metadata import PackageNotFoundError
 
 from tornado.web import RequestHandler
 from neuro_san.http_sidecar.logging.http_logger import HttpLogger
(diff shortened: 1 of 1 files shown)


neuro_san/http_sidecar/handlers/health_check_handler.py
```diff
@@ -42,10 +48,26 @@ async def get(self):
         Implementation of GET request handler for API health check.
         """
 
+        # Allow for a list of libraries to report versioning.
+        # For now we simply report one, but we might want to allow extension here
+        # at some point.
+        versioned_libs: List[str] = ["neuro-san"]
```
d1donlydfink, line 54: Prepare for > 1 library whose version we want to report.
It's possible that we could add entries to this via an env var, but wanted to gauge interest before going there.

neuro_san/http_sidecar/handlers/health_check_handler.py
```diff
@@ -42,10 +48,26 @@ async def get(self):
...
+        versions: Dict[str, Any] = {}
+
+        for lib in versioned_libs:
+            version: str = None
+            try:
+                version = str(library_version(lib))
+            except PackageNotFoundError:
+                version = "unknown"
+            versions[lib] = version
```
d1donlydfink, line 63: Check pip install version for version string.

neuro_san/http_sidecar/handlers/health_check_handler.py
```diff
@@ -42,10 +48,26 @@ async def get(self):
...
+
         try:
-            result_dict: Dict[str, Any] = \
-                {"service": "neuro-san agents",
-                 "status": "healthy"}
+            result_dict: Dict[str, Any] = {
+                "service": "neuro-san agents",
+                "status": "healthy",
+                "versions": versions
+            }
```
d1donlydfink, line 70: Report versions in the health-check response.

neuro_san/http_sidecar/handlers/health_check_handler.py
```diff
@@ -42,10 +48,26 @@ async def get(self):
...
+        # Allow for a list of libraries to report versioning.
+        # For now we simply report one, but we might want to allow extension here
+        # at some point.
+        versioned_libs: List[str] = ["neuro-san"]
+        versions: Dict[str, Any] = {}
+
+        for lib in versioned_libs:
+            version: str = None
+            try:
+                version = str(library_version(lib))
```
dsargent, line 60: Reflecting on the library version to get it -- nice 👌🏻

donn-leaf (approved):
Righteous

d1donlydfink:
@dsargent asks:
> Can we get this for neuro-san-studio too?

Ah. Here is the demand I was looking for... kinda.

Here's the thing, if you are building the container from within the repo,
you are not pip-installing what you are looking for, and therefore you do not get the version number, at least from pip.

I think previously this information tended to come in from an env var set for the deployment.
I can go there next.

PR #250, opened 2025-06-17
UN-3259 add http test case to unit test

vince-leaf:
This PR added an HTTP test case to our neuro-san repo unit test. The unit test `tests` GitHub Action pipeline gets triggered on each build.

Switching this unit test to run all cases parallel has reduced time to about 50%.
Before:
====================== 20 passed, 4 deselected in 25.96s =======================
After:
============================= 20 passed in 13.89s ==============================

Our smoke test remains to be triggered once a day. The smoke test covered OPEN-AI [GRPC, Direct]; ANTHROPIC [HTTP]

### DIFF
--- a/.github/workflows/smoke.yml
+++ b/.github/workflows/smoke.yml
@@ -28,6 +28,7 @@ jobs:
         shell: bash
         env:
           OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+          ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}
           AGENT_TOOL_PATH: "./neuro_san/coded_tools"
           PYTHONPATH: ${{ env.PYTHONPATH }}:.
         run: |
--- a/tests/neuro_san/hocon_tests/smoke/music_nerd_pro/test_smoke_test_hocons_llm_anthropic.py
+++ b/tests/neuro_san/hocon_tests/smoke/music_nerd_pro/test_smoke_test_hocons_llm_anthropic.py
@@ -39,6 +39,8 @@ class TestSmokeTestHocons(TestCase):
     ]))
     @pytest.mark.timeout(30)  # 30 seconds for this test
     @pytest.mark.smoke
+    @pytest.mark.non_default_llm_provider
+    @pytest.mark.anthropic
     def test_hocon(self, test_name: str, test_hocon: str):
         """
         Test method for a single parameterized test case specified by a hocon file.
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -25,7 +25,7 @@ jobs:
         run: pip freeze
 
 
-      - name: Run pytest (excluding integration tests)
+      - name: Run pytest integration tests
         run: pytest -s --verbose -m "integration" --timer-top-n 100
         env:
           OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
--- a/pytest.ini
+++ b/pytest.ini
@@ -4,6 +4,8 @@
 markers =
     integration: Mark a test as an integration test.  These generally take > 30 seconds to complete.
     smoke: Mark a test as an smoke test. These often exercise the full stack
+    non_default_llm_provider: tests that exercise llm providers other than our default Open AI
+    anthropic: tests that specifically use anthropic as the llm provider
 
 filterwarnings =
     # Ignore warnings about protobuf 4 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,0 +1,24 @@
+import os
+import pytest
+
+
+@pytest.fixture(autouse=True)
+def configure_llm_provider_keys(request, monkeypatch):
+    """Ensure only the appropriate LLM provider keys are available for the test being run."""
+
+    is_non_default = request.node.get_closest_marker("non_default_llm_provider")
+    is_anthropic = request.node.get_closest_marker("anthropic")
+
+    if is_non_default:
+        # For any non-default provider: clear OPENAI key to prevent accidental use
+        monkeypatch.delenv("OPENAI_API_KEY", raising=False)
+
+        if is_anthropic:
+            if not os.getenv("ANTHROPIC_API_KEY"):
+                pytest.skip("Missing ANTHROPIC_API_KEY for test marked 'anthropic'")
+        else:
+            pytest.skip("Unknown non-default provider; test requires explicit key handling.")
+    else:
+        # Default case: assume OpenAI is used
+        if not os.getenv("OPENAI_API_KEY"):
+            pytest.skip("Missing OPENAI_API_KEY for default LLM test.")


.github/workflows/tests.yml
```diff
@@ -38,8 +38,59 @@ jobs:
...
         run: flake8
 
-      - name: Run pytest (excluding integration tests)
-        run: pytest --verbose -m "not integration and not smoke and not e2e" --timer-top-n 10
+      - name: Run pytest with start of server (excluding integration tests & Smoke tests)
+        shell: bash
+        run: |
+          apt-get update && apt-get install -y netcat-openbsd procps curl net-tools
+
+          nohup python -m neuro_san.service.agent_main_loop > agent.log 2>&1 &
```
vince-leaf: Start the server service using nohup.

.github/workflows/tests.yml
```diff
@@ -38,8 +38,59 @@ jobs:
...
+          echo $! > agent.pid
+          sleep 2
+
+          if ! ps -p $(cat agent.pid) > /dev/null; then
+            echo "❌ Server process failed to start"
```
vince-leaf: Check service started with PID

.github/workflows/tests.yml
```diff
@@ -38,8 +38,59 @@ jobs:
...
+            echo "----- agent.log -----"
+            cat agent.log
+            exit 1
+          fi
+
+          echo "✅ Server process started with PID $(cat agent.pid)"
+
+          for i in {1..30}; do
+            PORT_8080_READY=false
+            PORT_30011_READY=false
```
vince-leaf: Check and wait for Server service ports to be ready.

.github/workflows/tests.yml
```diff
@@ -38,8 +38,59 @@ jobs:
...
+          fi
+
+          until curl -s http://localhost:8080/health > /dev/null; do
+            echo "Waiting for server health endpoint..."
+            sleep 1
+          done
+
+          echo "✅ Server is healthy and ready"
+
+          netstat -tuln | grep -E '8080|30011'
```
vince-leaf: Make sure the ports are listed.

.github/workflows/tests.yml
```diff
@@ -38,8 +38,59 @@ jobs:
...
+          pytest --verbose -m "not integration and not smoke" --timer-top-n 10 -n auto
```
vince-leaf, line 45: Run the unit test within the same docker run and test cases in parallel.

tests/neuro_san/hocon_tests/smoke/music_nerd_pro/test_smoke_test_hocons.py
```diff
@@ -35,7 +35,6 @@ class TestSmokeTestHocons(TestCase):
         # Barring that, try to stick to alphabetical order.
         "music_nerd_pro/combination_responses_with_history_direct.hocon",
         "music_nerd_pro/combination_responses_with_history_grpc.hocon",
-        "music_nerd_pro/combination_responses_with_history_http.hocon",
```
vince-leaf, line 38: Removed HTTP test case from smoke test.

tests/neuro_san/hocon_tests/smoke/music_nerd_pro/test_smoke_test_hocons_llm_anthropic.py
```diff
@@ -33,7 +33,7 @@ class TestSmokeTestHocons(TestCase):
         # These can be in any order.
         # Ideally more basic functionality will come first.
         # Barring that, try to stick to alphabetical order.
-        "music_nerd_pro_llm_anthropic/combination_responses_with_history_direct.hocon",
+        "music_nerd_pro_llm_anthropic/combination_responses_with_history_http.hocon",
```
vince-leaf, line 36: Switched Anthropic test case to HTTP.

tests/neuro_san/zzz_hocons/test_integration_test_hocons.py
```diff
@@ -37,6 +37,7 @@ class TestIntegrationTestHocons(TestCase):
 
         # List more hocon files as they become available here.
     ]))
+    @pytest.mark.timeout(10)  # 10 seconds for this test
```
vince-leaf: Timeout was added for each of the integration test cases.

tests/neuro_san/zzz_hocons/test_unit_test_hocons.py
```diff
@@ -34,9 +36,11 @@ class TestUnitTestHocons(TestCase):
         "math_guy/basic_sly_data.hocon",
         "math_guy/forwarded_sly_data.hocon",
         "music_nerd/beatles_with_history.hocon",
+        "music_nerd_pro/combination_responses_with_history_http.hocon",
```
vince-leaf, line 39: Added HTTP test case to unit test.

tests/neuro_san/zzz_hocons/test_unit_test_hocons.py
```diff
@@ -34,9 +36,11 @@ class TestUnitTestHocons(TestCase):
...
 
         # List more hocon files as they become available here.
     ]))
+    @pytest.mark.timeout(10)  # 10 seconds for this test
```
vince-leaf, line 43: Added timeout to each of unit test cases

tests/neuro_san/zzz_hocons/test_unit_test_hocons.py
```diff
@@ -34,9 +36,11 @@ class TestUnitTestHocons(TestCase):
         "math_guy/basic_sly_data.hocon",
         "math_guy/forwarded_sly_data.hocon",
         "music_nerd/beatles_with_history.hocon",
+        "music_nerd_pro/combination_responses_with_history_http.hocon",
 
         # List more hocon files as they become available here.
     ]))
+    @pytest.mark.timeout(30)  # 30 seconds for this test
```
vince-leaf, line 43: Added timeout to unit test cases

.github/workflows/tests.yml
```diff
@@ -38,8 +38,59 @@ jobs:
...
+            echo "❌ Timeout: One or both ports failed to open after $i seconds"
+            exit 1
+          fi
+
+          until curl -s http://localhost:8080/health > /dev/null; do
+            echo "Waiting for server health endpoint..."
+            sleep 1
+          done
+
+          echo "✅ Server is healthy and ready"
```
d1donlydfink: Vince can you put this in a script please?
This looks like it's duplicated between tests.yml and smoke.yml now.

Also: What if these tests here end up running at the same time as smoke tests?
Won't servers collide?

If so, make we keep the server-oriented tests to smoke.

d1donlydfink (changes requested):
* Think through w/ @donn-leaf  what happens if tests.yml and smoke.yml servers are running at the same time occupying the same port on the same server. Can that happen?
* Regardless: put common server stuff in a script

.github/workflows/tests.yml
vince-leaf: > Vince can you put this in a script please?

Yes, I can do that.

> Also: What if these tests here end up running at the same time as smoke tests?
> Won't servers collide?

I assume not, as it runs a separate container as a different pipeline, right @donn-leaf?

vince-leaf:
> * Think through w/ @donn-leaf  what happens if tests.yml and smoke.yml servers are running at the same time occupying the same port on the same server. Can that happen?
> * Regardless: put common server stuff in a script

Yes, I can put all the validation of the service server started into a script. However, having the common server trigger from the script would not likely work. I have spent hours getting the server service started for the agent_cli to see the ports.

.github/workflows/tests.yml
vince-leaf: @d1donlydfink , any suggestion on where to put the server service start script?

.github/workflows/smoke.yml
```diff
@@ -31,56 +31,7 @@ jobs:
...
-          until curl -s http://localhost:8080/health > /dev/null; do
-            echo "Waiting for server health endpoint..."
-            sleep 1
-          done
-
-          echo "✅ Server is healthy and ready"
-
-          netstat -tuln | grep -E '8080|30011'
-
+          neuro_san/scripts/server_start.sh
```
vince-leaf: @d1donlydfink, Smoke test is now set to use server_start script.

neuro_san/scripts/server_start.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
+  
+# Script to be run to start server before any of test case client that require services like HTTP or GRPC
+
+apt-get update && apt-get install -y netcat-openbsd procps curl net-tools
```
vince-leaf: Moved all the start server services into script

tests/fixtures/music_nerd_pro_llm_anthropic/combination_responses_with_history_http.hocon
```diff
@@ -0,0 +1,50 @@
...
+#
+# END COPYRIGHT
+
+# This file defines everything necessary for a data-driven test.
+{
+    # Describes what agent to test against.
+    "agent": "music_nerd_pro_llm_anthropic",
+
+    # Connect to the agent via a server
+    "connections": ["http"],
```
vince-leaf, line 18: Added new HTTP test Anthropic

tests/neuro_san/hocon_tests/smoke/music_nerd_pro/test_smoke_test_hocons.py
```diff
@@ -35,11 +35,10 @@ class TestSmokeTestHocons(TestCase):
...
 
         # List more hocon files as they become available here.
     ]))
-    @pytest.mark.timeout(10)  # 10 seconds for this test
+    @pytest.mark.timeout(30)  # 30 seconds for this test
```
vince-leaf, line 41: Bumped timeout out

.github/workflows/tests.yml
```diff
@@ -38,8 +38,11 @@ jobs:
       - name: Run flake8
         run: flake8
 
-      - name: Run pytest (excluding integration tests)
-        run: pytest --verbose -m "not integration and not smoke and not e2e" --timer-top-n 10
+      - name: Run pytest with start of server (excluding integration tests & Smoke tests)
+        shell: bash
+        run: |
+          neuro_san/scripts/server_start.sh
```
vince-leaf: @d1donlydfink , The test is now set to use the server_start script

vince-leaf:
> * Think through w/ @donn-leaf  what happens if tests.yml and smoke.yml servers are running at the same time occupying the same port on the same server. Can that happen?
> * Regardless: put common server stuff in a script

I saw both tests.yml and smoke.yml triggered and run at the same time without any issues.

d1donlydfink (changes requested):
* Move server_start.sh under build_scripts.  It is indeed a build_script.  There is no need for a new directory
* Can we not use Anthropic for a unit test case that everyone will need to run?  Stick to a single provider -- OpenAI for main stream test cases
* Did integration tests successfully run with a timeout of 10 seconds? Probably not.

tests/neuro_san/zzz_hocons/test_integration_test_hocons.py
d1donlydfink: If these pass in 10 seconds I would be surprised.
The whole reason these are integration tests is because they take longer than usual.
Did these actually pass when you ran them with this timeout?

tests/fixtures/music_nerd_pro_llm_anthropic/combination_responses_with_history_http.hocon
```diff
@@ -0,0 +1,50 @@
...
+# License by purchasing a commercial license.
+# Purchase of a commercial license is mandatory for any use of the
+# neuro-san SDK Software in commercial settings.
+#
+# END COPYRIGHT
+
+# This file defines everything necessary for a data-driven test.
+{
+    # Describes what agent to test against.
+    "agent": "music_nerd_pro_llm_anthropic",
```
d1donlydfink, line 15: Can we not use Anthropic for a unit test case that everyone will need to run?

d1donlydfink (approved):

PR #259, opened 2025-06-18
UN-3253 Default infinite timeout for streaming_chat() over HTTP

d1donlydfink:
@Noravee was reporting that using http sessions for external agents was timing out and causing agent failures.
This is different behavior than what we had seen with gRPC sessions.

I was able to chase it down to the grpc bowels of hell to see that in grpc-land we had specifically been passing in a None timeout (meaning infinite) by default for streaming_chat() for the gRPC case.

This was not the case for the HTTP session. This PR should make that parity now.

### DIFF
--- a/neuro_san/session/abstract_http_service_agent_session.py
+++ b/neuro_san/session/abstract_http_service_agent_session.py
@@ -57,7 +57,6 @@ class AbstractHttpServiceAgentSession(AgentSessionConstants):
         :param agent_name: The name of the agent to talk to
         """
         _ = umbrella_timeout
-        _ = streaming_timeout_in_seconds
 
         self.security_cfg: Dict[str, Any] = security_cfg
         self.use_host: str = "localhost"
@@ -72,7 +71,8 @@ class AbstractHttpServiceAgentSession(AgentSessionConstants):
             raise ValueError("agent_name is None")
         self.agent_name: str = agent_name
 
-        self.timeout_in_seconds = timeout_in_seconds
+        self.timeout_in_seconds: int = timeout_in_seconds
+        self.streaming_timeout_in_seconds: int = streaming_timeout_in_seconds
         self.metadata: Dict[str, str] = metadata
 
     def get_request_path(self, method: str) -> str:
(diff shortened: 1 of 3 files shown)


neuro_san/session/abstract_http_service_agent_session.py
```diff
@@ -72,7 +71,8 @@ def __init__(self, host: str = None,
             raise ValueError("agent_name is None")
         self.agent_name: str = agent_name
 
-        self.timeout_in_seconds = timeout_in_seconds
+        self.timeout_in_seconds: int = timeout_in_seconds
+        self.streaming_timeout_in_seconds: int = streaming_timeout_in_seconds
```
d1donlydfink, line 75: Keep what was passed in for streaming_timeout_in_seconds

neuro_san/session/async_http_service_agent_session.py
```diff
@@ -40,8 +40,7 @@ async def function(self, request_dict: Dict[str, Any]) -> Dict[str, Any]:
         result_dict: Dict[str, Any] = None
         try:
             async with ClientSession(headers=self.get_headers(),
-                                     conn_timeout=self.timeout_in_seconds,
-                                     read_timeout=self.timeout_in_seconds,
+                                     timeout=self.timeout_in_seconds
```
d1donlydfink: Latest code says that independent conn_timeout and read_timeout is deprecated in favor of a single timeout arg for aiohttp.ClientSession.  For UnaryUnary methods, use the timeout_in_seconds.  Here and for connectivity().

The sync version already only had a single timeout arg.

neuro_san/session/async_http_service_agent_session.py
```diff
@@ -90,8 +88,7 @@ async def streaming_chat(self, request_dict: Dict[str, Any]) -> Generator[Dict[s
         path: str = self.get_request_path("streaming_chat")
         try:
             async with ClientSession(headers=self.get_headers(),
-                                     conn_timeout=self.timeout_in_seconds,
-                                     read_timeout=self.timeout_in_seconds,
+                                     timeout=self.streaming_timeout_in_seconds
```
d1donlydfink: For streaming in grpc land we had been passing in a None for an infinite timeout.
We had not been matching that here.  Use the separate streaming_timeout_in_seconds which is by default None.
Here and in the sync version.

Noravee (approved):
LGTM.

PR #261, opened 2025-06-19
Allow newer versions of urllib3 for dependabot

d1donlydfink:
A dependabot warning is out for earlier versions of urllib3.
While we do use it in ExternalAgentParsing, it's only for basic parsing.
No reason we be capping the version here.


Noravee (approved):
LGTM

PR #264, opened 2025-06-23
Change environment variable "OPENAI_API_KEY" to "AZURE_OPENAI_API_KEY"

Noravee:
### Changes
- The `openai_api_key` argument of `AzureChatOpenAI` now falls back to `AZURE_OPENAI_API_KEY` instead of `OPENAI_API_KEY`.

- Added `AZURE_OPENAI_API_KEY`, `AZURE_OPENAI_ENDPOINT`, `OPENAI_API_VERSION`, and `deployment_name` to the lookup table in `api_key_error_check.py`.

- Updated `ApiKeyErrorCheck.check_for_api_key_exception()` to support returning multiple keys, since some keys may trigger identical error messages.

### Tests
- Ran agent network with `AzureOpenAI`


neuro_san/internals/run_context/langchain/llms/standard_langchain_llm_factory.py
```diff
@@ -112,7 +112,7 @@ def create_base_chat_model(self, config: Dict[str, Any],
                             model_name=model_name,
                             temperature=config.get("temperature"),
                             openai_api_key=self.get_value_or_env(config, "openai_api_key",
-                                                                 "OPENAI_API_KEY"),
+                                                                 "AZURE_OPENAI_API_KEY"),
```
Noravee: Per [LangChain's documentation](https://python.langchain.com/api_reference/_modules/langchain_openai/chat_models/azure.html#AzureChatOpenAI), support for `OPENAI_API_KEY` in `AzureChatOpenAI` will be removed to prevent potential conflicts.

neuro_san/internals/run_context/langchain/util/api_key_error_check.py
```diff
@@ -36,26 +45,26 @@ def check_for_api_key_exception(exception: Exception) -> str:
...
 
-        # Search for strings in the exception message
-        found = False
+        # Collect all keys that have any associated string in the exception message
+        # since there could be multiple keys with the exact same message.
         for api_key, string_list in API_KEY_EXCEPTIONS.items():
             for find_string in string_list:
                 if find_string in exception_message:
-                    found = True
+                    matched_keys.append(api_key)
```
Noravee, line 55: Collect all keys whose associated strings are found in the exception message, since multiple keys may share the same error pattern.

neuro_san/internals/run_context/langchain/util/api_key_error_check.py
```diff
@@ -36,26 +45,26 @@ def check_for_api_key_exception(exception: Exception) -> str:
...
+                    # No need to check the remaining strings for this key
                     break
-            if found:
-                break
 
-        if found:
+        if matched_keys:
+            keys_str = ", ".join(matched_keys)
```
Noravee, line 60: Join the list of matching keys into a comma-separated string.

neuro_san/internals/run_context/langchain/llms/standard_langchain_llm_factory.py
d1donlydfink: OK, however correct this may be, you can still screw existing installations with the draconian switch.
Better for backwards compatibility would be something like this:
```
Just above constructor:

openai_api_key: str = self.get_value_or_env(config, "openai_api_key", "AZURE_OPENAI_API_KEY")
if openai_api_key is None:
    openai_api_key = self.get_value_or_env(config, "openai_api_key", "OPENAI_API_KEY")

...

In the args for the constructor:

    ...
    openai_api_key=openai_api_key,
    ...
```

This would allow existing installations to not break if they were reliant on the OPENAI_API_KEY as we had it before (as well as previous versions of langchain-openai).

Always think through what your actions will mean for other people using your software.
Master this and you will see nothing but approvals in my PR reviews.

d1donlydfink (changes requested):
* Allow for backwards compatibility

neuro_san/internals/run_context/langchain/llms/standard_langchain_llm_factory.py
Noravee: Thanks for the thoughtful feedback — that makes a lot of sense.

You're right; I was too quick to align with the upcoming change in LangChain without fully considering its impact on existing installations. Your suggested fallback approach strikes a good balance between compatibility and forward-looking changes. I'll go ahead and update the implementation to use this two-step resolution for `openai_api_key`.

I’ll also make sure to think more critically about backwards compatibility in future changes. Appreciate the reminder and the clear example — very helpful.

neuro_san/internals/run_context/langchain/llms/standard_langchain_llm_factory.py
```diff
@@ -108,11 +108,13 @@ def create_base_chat_model(self, config: Dict[str, Any],
                     "include_usage": True
                 }
             }
+            openai_api_key: str = self.get_value_or_env(config, "openai_api_key", "AZURE_OPENAI_API_KEY")
+            if openai_api_key is None:
+                openai_api_key = self.get_value_or_env(config, "openai_api_key", "OPENAI_API_KEY")
```
Noravee, line 113: Add backwards compatibility for `OPENAI_API_KEY`.

d1donlydfink (approved):

PR #265, opened 2025-06-23
UN-3042 implement toolbox extension via agent network hocon

Noravee:
### Changes
- Check for key `agent_toolbox_info_file` from hocon file in `ToolboxFactory.load()`.
- Add argument `agent_toolbox_info_file` in the `ContextTypeToolboxFactory.load()`.
- Add `get_agent_toolbox_info_file()` method in `AgentNetwork`.
- Load toolbox from hocon in `ConnectivityReporter.report_network_connectivity()`.

### Tests
- Ran agent network with tool in toolbox defined in hocon 

### Notes
- The documentation will be updated in the next PR.

### DIFF
--- a/neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
+++ b/neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
@@ -83,8 +83,6 @@ class DefaultLlmFactory(ContextTypeLlmFactory, LangChainLlmFactory):
     def load(self):
         """
         Loads the LLM information from hocon files.
-
-        :param agent_llm_info_file: User-specified llm info from agent network hocon
         """
         restorer = LlmInfoRestorer()
         self.llm_infos = restorer.restore()


neuro_san/internals/run_context/langchain/toolbox/toolbox_factory.py
```diff
@@ -78,15 +78,21 @@ def __init__(self):
...
         restorer = ToolboxInfoRestorer()
         self.toolbox_infos = restorer.restore()
 
         # Mix in user-specified toolbox info, if available.
-        toolbox_info_file: str = os.getenv("AGENT_TOOLBOX_INFO_FILE")
+        # First check "agent_toolbox_info_file" key from agent network hocon.
+        # If that is unavailable, fallback to env variable.
+        toolbox_info_file: str = agent_toolbox_info_file
+        if not agent_toolbox_info_file:
+            toolbox_info_file = os.getenv("AGENT_TOOLBOX_INFO_FILE")
```
Noravee, line 99: Check `agent_toolbox_info_file` key from agent network hocon if that is unavailable, fallback to env variable, `AGENT_TOOLBOX_INFO_FILE`.

neuro_san/internals/graph/registry/agent_network.py
```diff
@@ -134,6 +134,12 @@ def get_agent_llm_info_file(self) -> str:
         """
         return self.config.get("agent_llm_info_file")
 
+    def get_agent_toolbox_info_file(self) -> str:
+        """
+        :return: The absolute path of agent toolbox info file for toolbox extension.
+        """
+        return self.config.get("agent_toolbox_info_file")
```
Noravee: Add a method in `AgentNetwork` to get `agent_toolbox_info_file` from agent network hocon.

neuro_san/internals/chat/connectivity_reporter.py
```diff
@@ -82,7 +85,7 @@ def report_network_connectivity(self) -> List[Dict[str, Any]]:
         """
         # Load the toolbox factory once
         if self.toolbox_factory is not None:
-            self.toolbox_factory.load()
+            self.toolbox_factory.load(agent_toolbox_info_file)
```
Noravee: Load toolbox from hocon in `ConnectivityReporter.report_network_connectivity()`.

neuro_san/internals/graph/registry/agent_network.py
d1donlydfink: A couple of things here:
* Why do we need "agent" in the key name? Isn't that redundant? Perhaps better as "toolbox_info_file"
* I'm not a fan of slowly adding methods onto AgentNetwork that get one thing from the config. There's no really policy in them.  It's instead already possible to do a get_config() on it and access that dictionary directly for the key you want.  Queue up a refactor for yourself for this and "agent_llm_info_file" above.
* You should have already added an entry to the agent hocon docs when you added a new key. You can do that when you change the name to "toolbox_info_file".

FWIW:
* Consider that prefixing things with "agent" is largely redundant within this codebase, especially so when the stem is self-describing enough.  The general exceptions here are things that users will use within their own integrations where an agent_ or AGENT_ prefix helps distinguish from other similar items they might encounter.  That all said, there is plenty of history and I have strayed myself in that direction, but we don't have to add to the the problem with new stuff.
*  As for the "agent" part of "agent_llm_info_file", that ship has already sailed.

neuro_san/internals/interfaces/context_type_toolbox_factory.py
```diff
@@ -41,10 +41,12 @@ class ContextTypeToolboxFactory:
         - "parameters":  Information on arguments of the tool.
     """
 
-    def load(self):
+    def load(self, agent_toolbox_info_file: str):
```
d1donlydfink: OK. Red Flags should go up whenever you are changing an interface.  You should be asking yourself "Is this interface change really necessary?".  In this case the answer is "probably not".

How to approach this depends on the answer to this question:
Is ToolboxFactory extensible by the user?

1) If it is extensible by the user, then you need to think of the effect of this change on the existing user,
which is: It's a breaking change, and we do not want to screw anyone's existing installation up if we can help it.

A slightly more palatable way to do this would be to have this argument be optional with a default value of None.  Users would still have to absorb the method signature change, but the optionality of the argument would mean that they could continue to ignore what just came in in their own implementation.

But I'm guessing this isn't really the case, so I offer this next option:

2) If it's not extensible by the user, the best way to do this is to pass this information to implementations' constructors, also ideally as an optional parameter whose default value is None.

This allows the main interface for load() to not have to change at all, and by my assessment many fewer files in this PR would have to change as a result.

Even if it were extensible by the user, I think this would still be the best way forward.

**Hard-Won Info-Nugget**: The pattern of making Simple interfaces more complex by adding parameters to the interface methods is an Easy trap to fall into. The extra plumbing gets the information where it needs to go. But the way to _keep_ the interfaces Simple is _almost always_ to resist modifying the interface methods and instead pass extra info through implementation's constructors.   This conserves wider utility in existing code via the interfaces because they simply do not change.

neuro_san/internals/chat/connectivity_reporter.py
d1donlydfink: Given the principle of not changing the main interface (see ContextTypeToolboxFactory comments below),
This file would not have to change at all if the information here were coming into the ToolboxFactory contstructor because the signature for load() would stay intact.

neuro_san/service/agent_service.py
```diff
@@ -82,11 +82,13 @@ def __init__(self,
 
         self.llm_factory: ContextTypeLlmFactory = MasterLlmFactory.create_llm_factory()
         self.toolbox_factory: ContextTypeToolboxFactory = MasterToolboxFactory.create_toolbox_factory()
-        # Load once and include "agent_llm_info_file" from agent network hocon to llm factory..
+        # Load once and include "agent_llm_info_file" and "agent_toolbox_info_file" from agent network hocon
+        # to llm factory and toolbox factory, respectively.
         agent_network: AgentNetwork = self.agent_network_provider.get_agent_network()
-        agent_llm_info_file = agent_network.get_agent_llm_info_file()
+        agent_llm_info_file: str = agent_network.get_agent_llm_info_file()
+        agent_toolbox_info_file: str = agent_network.get_agent_toolbox_info_file()
```
d1donlydfink: Given the principle of not changing the main interface (see ContextTypeToolboxFactory comments below),
This file would not have to change at all if the information here were coming into the ToolboxFactory contstructor because the signature for load() would stay intact.

neuro_san/service/async_agent_service.py
```diff
@@ -77,11 +77,13 @@ def __init__(self,
 
         self.llm_factory: ContextTypeLlmFactory = MasterLlmFactory.create_llm_factory()
         self.toolbox_factory: ContextTypeToolboxFactory = MasterToolboxFactory.create_toolbox_factory()
-        # Load once and include "agent_llm_info_file" from agent network hocon to llm factory..
+        # Load once and include "agent_llm_info_file" and "agent_toolbox_info_file" from agent network hocon
+        # to llm factory and toolbox factory, respectively.
         agent_network: AgentNetwork = self.agent_network_provider.get_agent_network()
-        agent_llm_info_file = agent_network.get_agent_llm_info_file()
+        agent_llm_info_file: str = agent_network.get_agent_llm_info_file()
+        agent_toolbox_info_file: str = agent_network.get_agent_toolbox_info_file()
```
d1donlydfink: Given the principle of not changing the main interface (see ContextTypeToolboxFactory comments below),
This file would not have to change at all if the information here were coming into the ToolboxFactory contstructor because the load() signature would stay intact.

neuro_san/session/async_direct_agent_session.py
```diff
@@ -104,7 +104,8 @@ async def connectivity(self, request_dict: Dict[str, Any]) -> Dict[str, Any]:
         }
 
         reporter = ConnectivityReporter(self.agent_network)
-        connectivity_info: List[Dict[str, Any]] = reporter.report_network_connectivity()
+        agent_toolbox_info_file: str = self.agent_network.get_agent_toolbox_info_file()
+        connectivity_info: List[Dict[str, Any]] = reporter.report_network_connectivity(agent_toolbox_info_file)
```
d1donlydfink: Given the principle of not changing the main interface (see ContextTypeToolboxFactory comments below),
This file would not have to change at all if the information here were coming into the ToolboxFactory contstructor because the signature for load() would stay intact.

neuro_san/session/direct_agent_session.py
```diff
@@ -111,7 +111,8 @@ def connectivity(self, request_dict: Dict[str, Any]) -> Dict[str, Any]:
         }
 
         reporter = ConnectivityReporter(self.agent_network)
-        connectivity_info: List[Dict[str, Any]] = reporter.report_network_connectivity()
+        agent_toolbox_info_file: str = self.agent_network.get_agent_toolbox_info_file()
```
d1donlydfink: Given the principle of not changing the main interface (see ContextTypeToolboxFactory comments below),
This file would not have to change at all if the information here were coming into the ToolboxFactory contstructor because the signature for load() would stay intact.

tests/neuro_san/internals/chat/test_connectivity_reporter.py
```diff
@@ -50,8 +50,9 @@ def test_hello_world(self):
         """
         agent_network: AgentNetwork = self.get_sample_registry("hello_world.hocon")
         reporter = ConnectivityReporter(agent_network)
+        agent_toolbox_info_file: str = agent_network.get_agent_toolbox_info_file()
 
-        messages: List[Dict[str, Any]] = reporter.report_network_connectivity()
+        messages: List[Dict[str, Any]] = reporter.report_network_connectivity(agent_toolbox_info_file)
```
d1donlydfink: Given the principle of not changing the main interface (see ContextTypeToolboxFactory comments below),
This file would not have to change at all if the information here were coming into the ToolboxFactory contstructor because the signature for load() would stay intact.

neuro_san/internals/run_context/langchain/toolbox/toolbox_factory.py
```diff
@@ -78,15 +78,21 @@ def __init__(self):
         self.overlayer = DictionaryOverlay()
         self.toolbox_info_file: str = None
```
d1donlydfink: Changing the constructor of ToolboxFactory to accept the toolbox_info_file as a parameter instead of plumbing it through load() solves a lot of problems:
* So many files would not have to change
* The config from which you are getting the value is already available to MasterToolboxFactory, so when ToolboxFactory is constructed in there, you have all the information you need at your disposal.
* In fact, if you access the config passed in there, you would also not need to add the get_agent_toolbox_info_file() method to the AgentNetwork.

d1donlydfink (commented):
OK. Red Flags should go up whenever you are changing an interface. You should be asking yourself "Is this interface change really necessary?". In this case the answer is "probably not".

How to approach this depends on the answer to this question:
Is ToolboxFactory extensible by the user?

If it is extensible by the user, then you need to think of the effect of this change on the existing user,
which is: It's a breaking change, and we do not want to screw anyone's existing installation up if we can help it.
A slightly more palatable way to do this would be to have this argument be optional with a default value of None. Users would still have to absorb the method signature change, but the optionality of the argument would mean that they could continue to ignore what just came in in their own implementation.

But I'm guessing this isn't really the case, so I offer this next option:

If it's not extensible by the user, the best way to do this is to pass this information to implementations' constructors, also ideally as an optional parameter whose default value is None.
This allows the main interface for load() to not have to change at all, and by my assessment many fewer files in this PR would have to change as a result.


Changing the constructor of ToolboxFactory to accept the toolbox_info_file as a parameter instead of plumbing it through load() solves a lot of problems:
* So many files would not have to change
* The config from which you are getting the value is already available to MasterToolboxFactory, so when ToolboxFactory is constructed in there, you have all the information you need at your disposal.
* In fact, if you access the config passed in there, you would also not need to add the get_agent_toolbox_info_file() method to the AgentNetwork.

d1donlydfink (changes requested):
OK. Red Flags should go up whenever you are changing an interface. You should be asking yourself "Is this interface change really necessary?". In this case the answer is "probably not".

How to approach this depends on the answer to this question:
Is ToolboxFactory extensible by the user?

If it is extensible by the user, then you need to think of the effect of this change on the existing user,
which is: It's a breaking change, and we do not want to screw anyone's existing installation up if we can help it.
A slightly more palatable way to do this would be to have this argument be optional with a default value of None. Users would still have to absorb the method signature change, but the optionality of the argument would mean that they could continue to ignore what just came in in their own implementation.

But I'm guessing this isn't really the case, so I offer this next option:

If it's not extensible by the user, the best way to do this is to pass this information to implementations' constructors, also ideally as an optional parameter whose default value is None.
This allows the main interface for load() to not have to change at all, and by my assessment many fewer files in this PR would have to change as a result.

Changing the constructor of ToolboxFactory to accept the toolbox_info_file as a parameter instead of plumbing it through load() solves a lot of problems:

So many files would not have to change
The config from which you are getting the value is already available to MasterToolboxFactory, so when ToolboxFactory is constructed in there, you have all the information you need at your disposal.
In fact, if you access the config passed in there, you would also not need to add the get_agent_toolbox_info_file() method to the AgentNetwork.

neuro_san/internals/graph/registry/agent_network.py
Noravee: - The key is changed to `toolbox_info_file`.
- Remove the method to get `toolbox_info_file` and `agent_llm_info_file`.
- Add documentation on the key `toolbox_info_file`.

neuro_san/internals/chat/connectivity_reporter.py
Noravee: The `load()` method has been reverted to have no argument.

neuro_san/internals/interfaces/context_type_toolbox_factory.py
Noravee: The load() method has been reverted to have no argument.

neuro_san/service/async_agent_service.py
Noravee: The load() method has been reverted to have no argument.

neuro_san/service/agent_service.py
Noravee: The load() method has been reverted to have no argument.

neuro_san/session/async_direct_agent_session.py
Noravee: The load() method has been reverted to have no argument.

neuro_san/session/direct_agent_session.py
Noravee: The load() method has been reverted to have no argument.

tests/neuro_san/internals/chat/test_connectivity_reporter.py
Noravee: The load() method has been reverted to have no argument.

neuro_san/internals/run_context/langchain/toolbox/toolbox_factory.py
Noravee: - The `config` is now in the constructor.
- `get_agent_toolbox_info_file()` was removed.

Noravee:
Thank you for your suggestion — I’ve made changes based on your feedback.

## Changes

- Reverted `ToolboxFactory.load()` to take no arguments.
- Added `config` as an optional constructor argument (default: `None`).
- Passed `config` from `MasterToolboxFactory`.
- Removed `get_agent_toolbox_info_file()` from `AgentNetwork`.
- Updated `DefaultLlmFactory` to follow the same pattern.
- Documented the `toolbox_info_file` key.
- Added **Google Gemini** to the LLM table in `agent_hocon_reference.md`.
- Replaced `OPENAI_API_KEY` with `AZURE_OPENAI_API_KEY` in the same table.

## Tests

- Verified functionality by running an agent network with LLM and toolbox extensions via the agent HOCON file.

neuro_san/internals/run_context/langchain/toolbox/toolbox_factory.py
```diff
@@ -70,13 +70,17 @@ class ToolboxFactory(ContextTypeToolboxFactory):
         "neuro_san/internals/run_context/langchain/toolbox/toolbox_info.hocon"
     """
 
-    def __init__(self):
+    def __init__(self, config: Optional[Dict[str, Any]] = None):
         """
         Constructor
+
```
Noravee, line 76: `config` as optional argument in the constructor of `ToolboxFactory`.

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
```diff
@@ -62,28 +63,34 @@ class DefaultLlmFactory(ContextTypeLlmFactory, LangChainLlmFactory):
                                     the model description in this class.
     """
 
-    def __init__(self):
+    def __init__(self, config: Optional[Dict[str, Any]] = None):
         """
         Constructor
```
Noravee, line 68: `config` as optional argument in the constructor of `DefaultLlmFactory`.

neuro_san/internals/run_context/factory/master_toolbox_factory.py
```diff
@@ -38,9 +38,9 @@ def create_toolbox_factory(config: Dict[str, Any] = None) -> ContextTypeToolboxF
         if context_type.startswith("openai"):
             toolbox_factory = None
         elif context_type.startswith("langchain"):
-            toolbox_factory = ToolboxFactory()
+            toolbox_factory = ToolboxFactory(config)
         else:
             # Default case
-            toolbox_factory = ToolboxFactory()
+            toolbox_factory = ToolboxFactory(config)
```
Noravee, line 44: Insert `config` when instantiate the factory in `MasterToolboxFactory`.

neuro_san/internals/run_context/factory/master_llm_factory.py
```diff
@@ -37,10 +37,10 @@ def create_llm_factory(config: Dict[str, Any] = None) -> ContextTypeLlmFactory:
         if context_type.startswith("openai"):
             llm_factory = None
         elif context_type.startswith("langchain"):
-            llm_factory = DefaultLlmFactory()
+            llm_factory = DefaultLlmFactory(config)
         else:
             # Default case
-            llm_factory = DefaultLlmFactory()
+            llm_factory = DefaultLlmFactory(config)
```
Noravee, line 43: Insert `config` when instantiate the factory in `MasterLlmFactory`.

neuro_san/internals/interfaces/context_type_llm_factory.py
```diff
@@ -41,7 +41,7 @@ class ContextTypeLlmFactory:
                                     the model description in this class.
     """
 
-    def load(self, agent_llm_info_file: str):
+    def load(self):
```
Noravee, line 44: Revert `load()` to have no arguments in `ContextTypeLlmFactory`.

neuro_san/internals/graph/registry/agent_network.py
```diff
@@ -128,12 +128,6 @@ def find_front_man(self) -> str:
         front_man = front_men[0]
         return front_man
 
-    def get_agent_llm_info_file(self) -> str:
-        """
```
Noravee, line 157: Remove `get_agent_llm_info_file()` in `AgentNetwork`.

docs/toolbox_info_hocon_reference.md
```diff
@@ -50,106 +52,135 @@ Each top-level key in the `toolbox_info.hocon` file represents a usable tool nam
...
 To add or override tools in the system, you can supply your own `toolbox_info.hocon` file and specify its path via the environment variable:
 
 ```bash
 AGENT_TOOLBOX_INFO_FILE=/path/to/your/toolbox_info.hocon
 ```
+
 This allows you to customize the set of available tools without modifying the built-in configuration.
+
+### toolbox_info_file key in specific agent hocon files
+
```
Noravee, line 183: Add documentation.

docs/agent_hocon_reference.md
```diff
@@ -166,7 +172,8 @@ to use LLMs from various providers.
 | LLM Provider  | API Key environment variable  |
 |:--------------|:------------------------------|
 | Anthropic     | ANTHROPIC_API_KEY             |
-| Azure OpenAI  | OPENAI_API_KEY                |
+| Azure OpenAI  | AZURE_OPENAI_API_KEY          |
+| Google Gemini | GOOGLE_API_KEY                |
```
d1donlydfink, line 176: Nice!

docs/agent_hocon_reference.md
```diff
@@ -464,17 +471,19 @@ agent parallelism and performance problems at scale.
...
 An optional string that refers to a predefined tool listed in a toolbox configuration file.
 Currently supported tool types include:
+
 - langchain's base tools
 - coded tools.
 
 The default toolbox configuration is located at [toolbox_info.hocon](../neuro_san/internals/run_context/langchain/toolbox/toolbox_info.hocon).
 
-To use your own tools, create a custom toolbox .hocon file and point to it by setting the AGENT_TOOLBOX_INFO_FILE environment variable.
+To use your own tools, create a custom toolbox .hocon file and point to it by setting the toolbox_info_file key in agent network hocon or by setting the AGENT_TOOLBOX_INFO_FILE environment variable.
```
d1donlydfink: A link to the new toolbox_info_file key here would be beneficial

docs/agent_hocon_reference.md
```diff
@@ -139,6 +140,11 @@ The agent_llm_info_file key allows you to specify a custom HOCON file that exten
 
 For more information on selecting and customizing models, see the [model_name](#model_name) section below.
 
+### toolbox_info_file
+
+The agent_toolbox_info_file key lets you define a custom HOCON file that adds to the default set of tools available to agents within a neuro-san network. This is particularly helpful when you have tools shared across multiple agent networks.
```
d1donlydfink: This is no longer the agent_toolbox_info_file key.
Always be your first reviewer!

docs/agent_hocon_reference.md
```diff
@@ -139,6 +140,11 @@ The agent_llm_info_file key allows you to specify a custom HOCON file that exten
...
+
+For further details, refer to the [toolbox](#toolbox) section below.
```
d1donlydfink, line 147: Perhaps also worth providing a link to the toolbox_info_hocon_reference.md file as well.

docs/toolbox_info_hocon_reference.md
```diff
@@ -2,26 +2,28 @@
 
 This document describes the specifications for the `toolbox_info.hocon` file used in the **neuro-san** system. This file allows you to extend or override the default tools shipped with the `neuro-san` library.
 
-The **neuro-san** system uses the HOCON (Human-Optimized Config Object Notation) format for data-driven configuration. HOCON is similar to JSON but includes enhancements such as comments and more concise syntax. You can explore the HOCON specification [here](https://github.com/lightbend/config/blob/main/HOCON.md) for further details.
+The **neuro-san** system uses the HOCON (Human-Optimized Config Object Notation) format for data-driven configuration. HOCON is similar to JSON but includes enhancements such as comments and more concise syntax. You can explore the full [HOCON specification on GitHub](https://github.com/lightbend/config/blob/main/HOCON.md) for further details.
```
d1donlydfink, line 5: Nice addition

docs/toolbox_info_hocon_reference.md
```diff
@@ -2,26 +2,28 @@
...
+      - [base_tool_info_url](#base_tool_info_url-optional)
+      - [display_as](#display_as-optional)
+    - [Coded Tools](#coded-tools)
+      - [class](#class-1)
+      - [description](#description)
+      - [parameters](#parameters-optional)
+      - [display_as](#display_as-optional-1)
+  - [Extending Toolbox Info](#extending-toolbox-info)
+    - [AGENT_TOOLBOX_INFO_FILE environment variable](#agent_toolbox_info_file-environment-variable)
+    - [toolbox_info_file key in specific agent hocon files](#toolbox_info_file-key-in-specific-agent-hocon-files)
```
d1donlydfink, line 26: 4 spaces is standard for indents. Unless a tool did this? Maybe it's possible to set the tool to do 4 spaces for indents.

docs/toolbox_info_hocon_reference.md
```diff
@@ -50,106 +52,135 @@ Each top-level key in the `toolbox_info.hocon` file represents a usable tool nam
...
+        "args": {
+            "k": 3
         }
     }
-    ```
-    This instantiates `BingSearchResults(api_wrapper=BingSearchAPIWrapper(k=3))`
+}
+```
+
+This instantiates `BingSearchResults(api_wrapper=BingSearchAPIWrapper(k=3))`
```
d1donlydfink, line 99: Do we still want to include BingSearchResults as a primary example if it's going away?

neuro_san/internals/graph/registry/agent_network.py
d1donlydfink: Nice how this leads to removal of code.

neuro_san/internals/run_context/factory/master_llm_factory.py
d1donlydfink: This is nice. I think we can do this without any kerfuffle because we have people derive from LangChainLlmFactory and that interface is not changing here. 😅

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
```diff
@@ -62,28 +63,34 @@ class DefaultLlmFactory(ContextTypeLlmFactory, LangChainLlmFactory):
...
+
+        :param config: The config dictionary which may or may not contain
+                       keys for the context_type and agent_llm_info_file
         """
         self.llm_infos: Dict[str, Any] = {}
         self.overlayer = DictionaryOverlay()
         self.llm_factories: List[LangChainLlmFactory] = [
             StandardLangChainLlmFactory()
         ]
+        self.config: Dict[str, Any] = config
```
d1donlydfink: Better: Do not store the entire config here, but take just what you need from the constructor.
This way there is guarding from some misguided future temptation to modify the config itself.

d1donlydfink (changes requested):
A couple of comments about the docs:
* add links
* indent using 4 spaces 

With respect to config in the Factory constructors:
* Pass the config as an arg, but take what you need from it in the constructor.  This forces a narrower scope of the class using it.  Or take what you need from the config before you call the constructor with a targeted arg.

docs/agent_hocon_reference.md
Noravee: A link has been added.

docs/agent_hocon_reference.md
Noravee: You're right — thanks for catching that. I’ve corrected the typo. Will make sure to double-check more carefully next time.

docs/agent_hocon_reference.md
Noravee: A link has been added.

docs/toolbox_info_hocon_reference.md
Noravee: the spaces have been reverted back to 4 spaces.

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
Noravee: Extract only the needed value from config.

docs/toolbox_info_hocon_reference.md
Noravee: New example will be added in a separate PR.

docs/agent_hocon_reference.md
```diff
@@ -471,13 +479,18 @@ Currently supported tool types include:
 
 The default toolbox configuration is located at [toolbox_info.hocon](../neuro_san/internals/run_context/langchain/toolbox/toolbox_info.hocon).
 
-To use your own tools, create a custom toolbox .hocon file and point to it by setting the AGENT_TOOLBOX_INFO_FILE environment variable.
+To use your own tools, create a custom toolbox `.hocon` file and reference it by either:
+
+- Setting the `toolbox_info_file` key in the agent network `.hocon` file, or  
+- Defining the `AGENT_TOOLBOX_INFO_FILE` environment variable.
+
+For more details on tool extension, see the [Toolbox Extension Guide](./toolbox_info_hocon_reference.md#extending-toolbox-info).
```
Noravee, line 487: Add link for tool extension.

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
```diff
@@ -62,17 +63,24 @@ class DefaultLlmFactory(ContextTypeLlmFactory, LangChainLlmFactory):
...
+        :param config: The config dictionary which may or may not contain
+                       keys for the context_type and agent_llm_info_file
         """
         self.llm_infos: Dict[str, Any] = {}
         self.overlayer = DictionaryOverlay()
         self.llm_factories: List[LangChainLlmFactory] = [
             StandardLangChainLlmFactory()
         ]
+        if config:
+            self.llm_info_file: str = config.get("agent_llm_info_file")
```
Noravee, line 79: Take only value from needed value from config in the constructor.

neuro_san/internals/run_context/langchain/toolbox/toolbox_factory.py
```diff
@@ -70,13 +70,19 @@ class ToolboxFactory(ContextTypeToolboxFactory):
...
+        :param config: The config dictionary which may or may not contain
+                       keys for the context_type and toolbox_info_file
         """
         self.toolbox_infos: Dict[str, Any] = {}
         self.overlayer = DictionaryOverlay()
-        self.toolbox_info_file: str = None
+        if config:
+            self.toolbox_info_file: str = config.get("toolbox_info_file")
```
Noravee, line 83: Take only value from needed value from config in the constructor.

PR #267, opened 2025-06-25
UN-3273 Fix 'Filename too long' by using uuids as filenames when that occurs

d1donlydfink:
@babakatwork is the primary reviewer

* When we get an OSError which is a "Filename too long" error, try again to write the filename as a uuid.
* Since it's possible for filenames to now not match the agent's full origin string they belong to, we add this as a preface to any newly created file.

### DIFF
--- a/neuro_san/client/thinking_file_message_processor.py
+++ b/neuro_san/client/thinking_file_message_processor.py
@@ -15,6 +15,7 @@ from typing import Dict
 from typing import List
 
 import json
+import uuid
 
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


neuro_san/client/thinking_file_message_processor.py
```diff
@@ -89,31 +90,57 @@ def write_message(self, response: Dict[str, Any], origin_str: str):
...
-            # Maybe add some context to where message is coming from if not using thinking_dir
-            if not self.thinking_dir:
-                use_origin += f" from {origin_str}"
-
-            # Maybe add some context as to where the tool result came from if we have info for that.
-            tool_result_origin: List[Dict[str, Any]] = response.get("tool_result_origin")
-            if tool_result_origin is not None:
-                last_origin_only: List[Dict[str, Any]] = [tool_result_origin[-1]]
-                origin_str = Origination.get_full_name_from_origin(last_origin_only)
-                use_origin += f" (result from {origin_str})"
```
d1donlydfink, line 114: Determination of how to put together use_origin is moved to _determine_origin_reporting() below.

neuro_san/client/thinking_file_message_processor.py
```diff
@@ -89,31 +90,57 @@ def write_message(self, response: Dict[str, Any], origin_str: str):
...
+            if how_to_open_file == "w":
+                # New file, preface it with an origin log.
+                thinking.write(f"Agent: {origin_str}\n")
```
d1donlydfink, line 138: If we are first writing the file, always preface any content with a line about where the information is coming from. This ends up being useful even if the file is appropriately named.

neuro_san/client/thinking_file_message_processor.py
```diff
@@ -89,31 +90,57 @@ def write_message(self, response: Dict[str, Any], origin_str: str):
...
+                # Retry with a uuid as file name.
+                # If this fails, there's no helping ya.
+                new_uuid: str = str(uuid.uuid4())
+                self._write_to_file(new_uuid, origin_str, message_type_str, use_origin, text)
+            else:
+                raise os_error
+
+    def _write_to_file(self, origin_filename: str, origin_str: str,
+                       message_type_str: str, use_origin: str, text: str):
+
```
d1donlydfink, line 124: Break the process of writing to the file out so it can potentially be repeated if an exception comes around.

neuro_san/client/thinking_file_message_processor.py
```diff
@@ -89,31 +90,57 @@ def write_message(self, response: Dict[str, Any], origin_str: str):
...
         filename: Path = self.thinking_file
         if self.thinking_dir:
-            if origin_str is None or len(origin_str) == 0:
+            if origin_filename is None or len(origin_filename) == 0:
                 return
-            filename = Path(self.thinking_dir, origin_str)
+            filename = Path(self.thinking_dir, origin_filename)
```
d1donlydfink, line 129: The filename bit is now separate from the origin_str itself.

neuro_san/client/thinking_file_message_processor.py
```diff
@@ -89,31 +93,68 @@ def write_message(self, response: Dict[str, Any], origin_str: str):
             # There is no real text, but there is a structure. JSON-ify it.
             text = json.dumps(structure, indent=4, sort_keys=True)
 
+        # Figure out how we are going to report the origin given the message.
+        use_origin: str = self._determine_origin_reporting(response, origin_str)
```
d1donlydfink, line 97: Break out the policy of how we are going to report the origin as separate method.

neuro_san/client/thinking_file_message_processor.py
```diff
@@ -89,31 +93,68 @@ def write_message(self, response: Dict[str, Any], origin_str: str):
...
+
+        # Determine the filename to use given the origin_str.
+        # Previously we might have used a uuid as a filename, but by default
+        # we want the filenames to match the full origin_str.
+        origin_filename: str = self.origins.get(origin_str, origin_str)
```
d1donlydfink, line 102: See if we have already used a different filename given the origin we have.

neuro_san/client/thinking_file_message_processor.py
```diff
@@ -89,31 +93,68 @@ def write_message(self, response: Dict[str, Any], origin_str: str):
...
+                # Retry with a uuid as file name.
+                # If this fails, there's no helping ya.
+                origin_filename = str(uuid.uuid4())
+                self._write_to_file(origin_filename, origin_str, message_type_str, use_origin, text)
+
+                # Squirell that uuid away so results continue to go to the same
+                # file over and over again.
+                self.origins[origin_str] = origin_filename
+            else:
+                raise os_error
```
d1donlydfink, line 120: Do normal write to file but if the write fails because of Filename too long, try again with a uuid as filename.

PR #269, opened 2025-06-25
Removed the verbose section from documantation. 

kxk302:
And, fixed a bunch of lint issues

### DIFF
--- a/docs/agent_hocon_reference.md
+++ b/docs/agent_hocon_reference.md
@@ -78,7 +78,7 @@ A commondefs dictionary where keys are strings to be found within braces within 
 and the values of the dictionary are to replace those keys anywhere throughout the dictionary where string
 values are to be found.  Example:
 
-```
+```json
 {
     "commondefs": {
         "replacement_strings": {
@@ -104,7 +104,7 @@ of the dictionary are to replace those keys anywhere throughout the dictionary w
 whenever an exact match of the string value is found.
 values are to be found.  Example:
 
-```
+```json
 {
     "commondefs": {
         "replacement_values": {
@@ -123,7 +123,7 @@ values are to be found.  Example:
 
 This results in a final interpretation where the "function" value is:
 
-```
+```json
 "function": {
     "key1": "my_value",
     "key2": 1.0
@@ -138,7 +138,6 @@ your string values within your replacement_values and things will work out as yo
 The agent_llm_info_file key allows you to specify a custom HOCON file that extends the default list of available LLMs used by agents in a neuro-san network. This is especially useful if you're using models that are not included in the default configuration (e.g., newly released models or organization-specific endpoints).
 
 For more information on selecting and customizing models, see the [model_name](#model_name) section below.
-
 
 ### llm_config
 
@@ -172,7 +171,7 @@ to use LLMs from various providers.
 | OpenAI        | OPENAI_API_KEY                |
 
 Note: _We strongly recommend to **not** set secrets as values within any source file._
-These files tend to creep into source control repos, and it is __very__ bad practice
+These files tend to creep into source control repos, and it is **very** bad practice
 to expose secrets by checking them in.
 
 If your favorite model, or new hotness is not listed in default_llm_info.hocon,
@@ -225,7 +224,7 @@ As long as a parameter is a scalar listed in the args section for your LLM's cla
 file, you can set that parameter in any llm_config within its own technical limits however you like.
 
 Note: _We strongly recommend to **not** set secrets as values within any source file, including hocon files._
-These files tend to creep into source control repos, and it is __very__ bad practice
+These files tend to creep into source control repos, and it is **very** bad practice
 to expose secrets by checking them in.
 
 #### verbose
@@ -260,7 +259,7 @@ which contains the following keys:
 #### error_fragments
 
 A list of strings where if any one of the strings appears in agent output,
-it is considered an error and reported as such per the [error_formatter](#error-formatter).
+it is considered an error and reported as such per the [error_formatter](#error_formatter).
 
 ### tools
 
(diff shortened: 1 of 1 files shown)


docs/agent_hocon_reference.md
```diff
@@ -28,7 +28,6 @@ Sub-keys to those dictionaries will be described in the next-level down heading
     - [max_execution_seconds](#max_execution_seconds)
     - [temperature](#temperature)
     - [Other LLM-specific Parameters](#other-llm-specific-parameters)
-    - [verbose](#verbose)
```
kxk302: Removed from TOC

docs/agent_hocon_reference.md
```diff
@@ -45,22 +44,19 @@ Sub-keys to those dictionaries will be described in the next-level down heading
   - [command](#command)
   - [tools (agents)](#tools-agents)
     - [External Agents](#external-agents)
-  - [llm_config](#llm_config-1)
```
kxk302: These must have unique names or else the doc navigation could get confused.

docs/agent_hocon_reference.md
```diff
@@ -45,22 +44,19 @@ Sub-keys to those dictionaries will be described in the next-level down heading
...
+  - [Per agent llm_config](#per-agent-llm_config)
   - [class](#class)
   - [toolbox](#toolbox)
   - [args](#args)
   - [allow](#allow)
     - [connectivity](#connectivity)
     - [to_downstream](#to_downstream)
-      - [sly_data](#sly_data)
```
kxk302: Removed redundant (ambiguous) headers

docs/agent_hocon_reference.md
```diff
@@ -45,22 +44,19 @@ Sub-keys to those dictionaries will be described in the next-level down heading
...
     - [from_downstream](#from_downstream)
-      - [sly_data](#sly_data-1)
     - [to_upstream](#to_upstream)
-      - [sly_data](#sly_data-2)
   - [display_as](#display_as)
   - [max_message_history](#max_message_history)
-  - [error_formatter](#error_formatter-1)
```
kxk302: Unique header names

docs/agent_hocon_reference.md
```diff
@@ -78,7 +74,7 @@ A commondefs dictionary where keys are strings to be found within braces within
 and the values of the dictionary are to replace those keys anywhere throughout the dictionary where string
 values are to be found.  Example:
 
-```
```
kxk302, line 81: Specifying the language of what is being formatted

docs/agent_hocon_reference.md
```diff
@@ -225,22 +220,9 @@ As long as a parameter is a scalar listed in the args section for your LLM's cla
 file, you can set that parameter in any llm_config within its own technical limits however you like.
 
 Note: _We strongly recommend to **not** set secrets as values within any source file, including hocon files._
-These files tend to creep into source control repos, and it is __very__ bad practice
+These files tend to creep into source control repos, and it is **very** bad practice
 to expose secrets by checking them in.
 
-#### verbose
```
kxk302, line 231: Removed verbose section

docs/agent_hocon_reference.md
d1donlydfink: I'm not sure the answer is to just remove these.  They are related, but they are definitely not redundant, as they have specific sections in the docs below with slightly different semantics between them all that need to be made clear to any who care enough to use them.
The links were generated by a package and when this is rendered by GitHub (at least) they link to what is expected.

docs/agent_hocon_reference.md
d1donlydfink: Same as above.
These were created by a tool.  GitHub Renders these and the links perform just fine.
I don't think we should change this.

docs/agent_hocon_reference.md
d1donlydfink: I think there is no harm in keeping verbose in the docs.
It's not like we are actually removing or changing the functionality described.

What is activated is largely extra debugging that is available in langchain,
and I actually think it's better that we are up front about all the keys we support in the docs - better than keeping it secret and someone goes and digs in the source and starts asking about it in a backchannel.

While IMHO its functionality obsolete given the client capabilities we have now, I've gotten feedback (from 1C guys for instance) that they actually want to turn this on for their own server logging. 🤷 To each their own.

This all said, I would still agree that removing the verbose key from regular use in our example hocons
is still a good decision because on first impression it's just a bunch of noise that newbies need to sort through.

docs/agent_hocon_reference.md
```diff
@@ -412,19 +394,19 @@ This allows common agent network definitions to be used as functions for other l
 
 Furthermore, it is also possible to reference agents on other neuro-san _servers_ by using a URL as a tool reference.
 
-Example: "http://localhost:8080/math_guy"
+Example: [http://localhost:8080/math_guy](http://localhost:8080/math_guy)
 
 This enables entire ecosystems of agent webs.
 
-### llm_config
+### Per agent llm_config
```
d1donlydfink: What tool are you using that is motivating this change?
I don't like that it's inconsistent with all the other keys?

docs/agent_hocon_reference.md
d1donlydfink: No they don't have to have unique names in the square brackets.  The links need to have unique names so they link to unique content, and they do.  GitHub renders them and links them all just fine.

If the problem is because of a linter tool, maybe we need to ignore this warning when running the tool.

docs/agent_hocon_reference.md
d1donlydfink: Unnecessary.  See comment in verbose below.

docs/agent_hocon_reference.md
kxk302: I just removed the "sly_data" header -- the text remains under "to_downstream".

docs/agent_hocon_reference.md
kxk302: markdownlinter-cli

kxk302:
> * don't need to remove verbose from docs
> * Shouldn't need to do "per agent" everything just to keep links happy.  Links seem to be happy when rendered in GitHub.  What tool is motivating the change?

I can add overrides to suppress those warnings.

docs/agent_hocon_reference.md
```diff
@@ -565,6 +569,7 @@ Dictionary which specifies security policy for information going back to any cal
 
 This has no effect on any information flowing between agents internal to the network.
 
+<!-- markdownlint-disable-next-line MD024 -->
```
d1donlydfink, line 572: Good to know!

PR #270, opened 2025-06-26
UN-3276 support for user defined llm in hocon file with class key

Noravee:
## Changes

- Add `UserSpecifiedLangchainLlmFactory`

### Tests

- Ran agent network with `gpt-4.1-mini`, which is not in the default llm info.


neuro_san/internals/run_context/langchain/llms/user_specified_langchain_llm_factory.py
```diff
@@ -0,0 +1,86 @@
...
+from langchain_core.language_models.base import BaseLanguageModel
+from langchain_openai.chat_models.azure import AzureChatOpenAI
+from langchain_openai.chat_models.base import ChatOpenAI
+
+from neuro_san.internals.run_context.langchain.llms.langchain_llm_factory import LangChainLlmFactory
+
+
+class UserSpecifiedLangChainLlmFactory(LangChainLlmFactory):
+    """
+    A factory for constructing LLMs based on user-specified configurations provided under the "llm_config"
```
Noravee: Add `UserSpecifiedLangChainLlmFactory` that uses arguments from given by the user.

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
```diff
@@ -168,6 +171,14 @@ def create_full_llm_config(self, config: Dict[str, Any]) -> Dict[str, Any]:
         :param config: The llm_config from the user
         :return: The fully specified config with defaults filled in.
         """
+
+        self.llm_class = config.get("class")
+        if self.llm_class:
+            # If config has "class", it is a user-specified llm so return config as is,
+            # and replace "StandardLangChainLlmFactory" with "UserSpecifiedLangChainLlmFactory".
+            self.llm_factories[0] = UserSpecifiedLangChainLlmFactory()
```
Noravee: If `llm_config` has `class`, use `UserSpecifiedLangChainLlmFactory`.

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
```diff
@@ -495,7 +494,7 @@ async def ainvoke(self, agent_executor: AgentExecutor, inputs: Dict[str, Any], i
         while return_dict is None and retries > 0:
             try:
                 return_dict: Dict[str, Any] = await agent_executor.ainvoke(inputs, invoke_config)
-            except (APIError, BadRequestError, AuthenticationError, ChatGoogleGenerativeAIError) as api_error:
+            except (openai.APIError, anthropic.APIError, ChatGoogleGenerativeAIError) as api_error:
```
Noravee: Use `APIError` for anthropic as well.

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
```diff
@@ -509,6 +508,11 @@ async def ainvoke(self, agent_executor: AgentExecutor, inputs: Dict[str, Any], i
                 retries = retries - 1
                 exception = key_error
                 backtrace = traceback.format_exc()
+            except TypeError as type_error:
```
Noravee: Add TypeError for non-existing arguments user may pass.

neuro_san/internals/run_context/langchain/util/api_key_error_check.py
```diff
@@ -22,12 +22,10 @@
...
-    "AZURE_OPENAI_API_KEY": ["Error code: 401", "invalid subscription key", "wrong API endpoint", "Connection error"],
-    "AZURE_OPENAI_ENDPOINT": ["validation error", "base_url", "azure_endpoint", "AZURE_OPENAI_ENDPOINT",
-                              "Connection error"],
-    "OPENAI_API_VERSION": ["validation error", "api_version", "OPENAI_API_VERSION", "Error code: 404",
-                           "Resource not found"],
-    "deployment_name": ["Error code: 404", "Resource not found", "API deployment for this resource does not exist"],
+    "AZURE_OPENAI_API_KEY": ["invalid subscription key", "wrong API endpoint"],
+    "AZURE_OPENAI_ENDPOINT": ["base_url", "azure_endpoint", "AZURE_OPENAI_ENDPOINT"],
+    "OPENAI_API_VERSION": ["api_version", "OPENAI_API_VERSION"],
+    "deployment_name": ["API deployment for this resource does not exist"],
```
Noravee: Remove common messages like "Error code: 404".

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
```diff
@@ -23,9 +23,8 @@
 from logging import Logger
 from logging import getLogger
 
-from openai import APIError
-from anthropic import BadRequestError
-from anthropic import AuthenticationError
+import openai
+import anthropic
```
d1donlydfink: This is not incorrect, but it's technically better to not "import the world" of a single package.
This can cause a whole lot more Python to be loaded when the file is first accessed causing problems like:
* trigger of extra imports/dependencies that aren't needed (not the case here)
* a lot more time to compile with the file is first loaded.  Specifically TensorFlow had huge problems in this regard and spent many years and a lot of bizarre effort trying to get their import times down for performance.  This is all because they set the expectations early that people should "just" import tensorflow, which dragged in the world.   We don't need to contribute to that when we know exactly what we want from a package.

So, please, please import only what you need, especially from 3rd part libs.

Now that all said, there is some value to what you are doing in the lines below with
openai.APIError, etc.  It becomes clearer as to where the error is from.  This is precisely what the
Python "as" modifier with imports is for:

```
from openai import APIError as OpenAI_APIError
...
except OpenAI_APIError
```
... could work better to those ends.

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
```diff
@@ -509,6 +508,11 @@ async def ainvoke(self, agent_executor: AgentExecutor, inputs: Dict[str, Any], i
                 retries = retries - 1
                 exception = key_error
                 backtrace = traceback.format_exc()
+            except TypeError as type_error:
+                self.logger.warning("retrying from TypeError")
+                retries = retries - 1
+                exception = type_error
+                backtrace = traceback.format_exc()
```
d1donlydfink: Seems like you should explain yourself here.

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
```diff
@@ -71,6 +73,7 @@ def __init__(self):
         self.llm_factories: List[LangChainLlmFactory] = [
             StandardLangChainLlmFactory()
         ]
+        self.llm_class: str = None
```
d1donlydfink: Seems like this does not need to be a member variable for the class.

Use a local inside load()

Always keep the scope as small as possible.

neuro_san/internals/run_context/langchain/llms/user_specified_langchain_llm_factory.py
```diff
@@ -0,0 +1,86 @@
...
+        """
+        # Construct the LLM
+        llm: BaseLanguageModel = None
+        chat_class: str = config.get("class")
+        if chat_class is not None:
+            chat_class = chat_class.lower()
+
+        # Take "class" out of config and add "callback".
+        config.pop("class")
+        config["callbacks"] = callbacks
```
d1donlydfink: Here you are modifying an input argument, which will modify the dictionary for the rest of the system until it shuts down.  This would violate the read-only-ness of the expectations of any of our configs.
Solution:  make a shallow copy.copy() before modifying the input arg config.

neuro_san/internals/run_context/langchain/llms/user_specified_langchain_llm_factory.py
```diff
@@ -0,0 +1,86 @@
...
+            llm = AzureChatOpenAI(**config)
+        elif chat_class == "anthropic":
+            llm = ChatAnthropic(**config)
+        elif chat_class == "ollama":
+            llm = ChatOllama(**config)
+        elif chat_class == "nvidia":
+            llm = ChatNVIDIA(**config)
+        elif chat_class == "gemini":
+            llm = ChatGoogleGenerativeAI(**config)
+        else:
```
d1donlydfink: Why not use the the existing mechanisms to instantiate the BaseLanguageModel at this point?
It seems you are losing a whole lot by going immediately to the constructor here, like:
* All the env var stuff that is already handled in StandardLangChainLlmFactory
* Any user extensions to unknown providers that were added on by way of a llm_info.hocon factory specification.

d1donlydfink (changes requested):
A lot to address in here.

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
Noravee: Thank you for your suggestiion. Now the import is only on the error.

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
Noravee: This was for cases that the user-specified args are invalid, but it has been removed. The check takes place before class instantiation instead.

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
Noravee: the instance variable `llm_class` was removed.

neuro_san/internals/run_context/langchain/llms/user_specified_langchain_llm_factory.py
Noravee: Thank you for your suggestions. It has been implemented.

llm_extension/groq_langchain_llm_factory.py
```diff
@@ -0,0 +1,58 @@
...
+
+from langchain_core.callbacks.base import BaseCallbackHandler
+from langchain_core.language_models.base import BaseLanguageModel
+from langchain_groq import ChatGroq
+
+from neuro_san.internals.run_context.langchain.llms.langchain_llm_factory import LangChainLlmFactory
+
+
+class GroqLangChainLlmFactory(LangChainLlmFactory):
+    """
```
Noravee: This is for testing only. It will not be in the final PR.

llm_extension/llm_info.hocon
```diff
@@ -0,0 +1,26 @@
...
+
+# The schema specifications for this file are documented here:
+# https://github.com/cognizant-ai-lab/neuro-san/blob/main/docs/llm_info_hocon_reference.md
+
+{
+
+
+    "classes": {
+        "factories": [ "llm_extension.groq_langchain_llm_factory.GroqLangChainLlmFactory" ]
+        "groq": {
```
Noravee: This is for testing only. It will not be in the final PR.

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
```diff
@@ -23,9 +23,8 @@
 from logging import Logger
 from logging import getLogger
 
-from openai import APIError
-from anthropic import BadRequestError
-from anthropic import AuthenticationError
+from openai import APIError as OpenAI_APIError
+from anthropic import APIError as Anthropic_APIError
```
Noravee, line 27: Only import `APIError`.

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
```diff
@@ -155,6 +145,38 @@ def resolve_one_llm_factory(self, llm_factory_class_name: str, llm_info_file: st
                              "must be of type LangChainLlmFactory")
         return llm_factory
 
+    def _resolve_class_from_path(
+            self,
+            class_path: str,
+            expected_base: Type,
+            source_file: str,
+            description: str
+    ) -> Type:
```
Noravee: Since there are 2 methods that need to resolve class: `resolve_one_llm_factory()` and `reate_base_chat_model()`,    refactor by creating a new method `_resolve_class_from_path()`.

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
```diff
@@ -169,13 +191,19 @@ def create_llm(self, config: Dict[str, Any], callbacks: List[BaseCallbackHandler
...
 
     def create_full_llm_config(self, config: Dict[str, Any]) -> Dict[str, Any]:
         """
         :param config: The llm_config from the user
         :return: The fully specified config with defaults filled in.
         """
+
+        if config.get("class"):
+            # If config has "class", it is a user-specified llm so return config as is,
+            return config
```
Noravee: If config has `class` it is a user-specified llm so return config as is.

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
```diff
@@ -279,6 +307,27 @@ def create_base_chat_model(self, config: Dict[str, Any],
                 # Let the next model have a crack
                 found_exception = exception
 
+        # Try resolving via 'class' in config if factories failed
+        class_path = config.get("class")
+        if found_exception is not None and class_path:
```
Noravee: Try resolving via `class` in config if factories failed

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
```diff
@@ -279,6 +307,27 @@ def create_base_chat_model(self, config: Dict[str, Any],
...
+                class_path=class_path,
+                expected_base=BaseLanguageModel,
+                source_file="agent network hocon file",
+                description="llm_config"
+            )
+
+            # copy the config, take 'class' out, and unpack into llm constructor
+            user_config = config.copy()
+            user_config.pop("class")
+            ToolboxFactory.check_invalid_args(llm_class, user_config)
```
Noravee: Using `check_invalid_args()` from `ToolboxFactory` to check for invalid arguments.

neuro_san/internals/run_context/langchain/llms/standard_langchain_llm_factory.py
```diff
@@ -70,7 +70,10 @@ def create_base_chat_model(self, config: Dict[str, Any],
         if chat_class is not None:
             chat_class = chat_class.lower()
 
-        model_name: str = config.get("model_name")
+        # Check for key "model_name", "model", and "model_id" to use as model name
+        # If the config is from default_llm_info, this is always "model_name"
+        # but with user-specified config, it is possible to have the other keys will be specifed instead.
+        model_name: str = config.get("model_name") or config.get("model") or config.get("model_id")
```
Noravee, line 76: Check for key `model_name`, `model`, and `model_id` to use as model name. If the config is from default_llm_info, this is always `model_name` but with user-specified config, it is possible to have the other keys will be specifed instead.

neuro_san/internals/run_context/langchain/toolbox/toolbox_factory.py
```diff
@@ -210,16 +211,24 @@ def _resolve_class(self, class_path: str) -> Type[BaseTool]:
         except AttributeError as exception:
             raise ValueError(f"Class {class_path} not found in PYTHONPATH") from exception
 
-    def _check_invalid_args(self, method_class: Union[Type, MethodType], args: Dict[str, Any]):
+    @staticmethod
+    def check_invalid_args(method_class: Union[Type, MethodType], args: Dict[str, Any]):
```
Noravee: Change `check_invalid_args` from an internal method to a static method.

neuro_san/internals/run_context/langchain/toolbox/toolbox_factory.py
```diff
@@ -210,16 +211,24 @@ def _resolve_class(self, class_path: str) -> Type[BaseTool]:
...
         """
         Check for invalid arguments in class or method
         :param method_class: Class or method to check for the invalid arguments
         :param args: Arguments to check
         """
-        class_args_set: Set[str] = set(signature(method_class).parameters.keys())
+        pydantic_args: Set[str] = set()
+        # Check for if it is a class that extends pydantic BaseModel
+        if isclass(method_class) and issubclass(method_class, BaseModel):
```
Noravee: Check if the class extends `BaseModel`. This is necessary since pydantic allows arguments to have `alias`.

neuro_san/registries/manifest.hocon
```diff
@@ -47,7 +47,9 @@
     # Agents having to do with test infrastructure
     "gist.hocon": true,
     "assess_failure.hocon": true,
-
+    "test_new_model_default_class.hocon": true,
+    "test_new_class.hocon": true,
+    "test_new_model_extended_class.hocon": true
```
Noravee: This is for testing only. It will not be in the final PR.

neuro_san/registries/test_new_class.hocon
```diff
@@ -0,0 +1,64 @@
...
+# END COPYRIGHT
+
+# The schema specifications for this file are documented here:
+# https://github.com/cognizant-ai-lab/neuro-san/blob/main/docs/agent_hocon_reference.md
+
+{
+    "llm_config": {
+        "class": "langchain_groq.chat_models.ChatGroq",
+        "model_name": "deepseek-r1-distill-llama-70b",
+        "temperature": 0.5
```
Noravee: This is for testing only. It will not be in the final PR.

neuro_san/registries/test_new_model_default_class.hocon
```diff
@@ -0,0 +1,64 @@
...
+# END COPYRIGHT
+
+# The schema specifications for this file are documented here:
+# https://github.com/cognizant-ai-lab/neuro-san/blob/main/docs/agent_hocon_reference.md
+
+{
+    "llm_config": {
+        "class": "openai",
+        "model_name": "gpt-4.1-mini",
+        "temperature": 0.5
```
Noravee: This is for testing only. It will not be in the final PR.

neuro_san/registries/test_new_model_extended_class.hocon
```diff
@@ -0,0 +1,65 @@
...
+
+# The schema specifications for this file are documented here:
+# https://github.com/cognizant-ai-lab/neuro-san/blob/main/docs/agent_hocon_reference.md
+
+{
+    "agent_llm_info_file": "llm_extension/llm_info.hocon"
+    "llm_config": {
+        "class": "groq",
+        "model_name": "deepseek-r1-distill-llama-70b",
+        "temperature": 0.5,
```
Noravee: This is for testing only. It will not be in the final PR.

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
```diff
@@ -495,7 +494,7 @@ async def ainvoke(self, agent_executor: AgentExecutor, inputs: Dict[str, Any], i
         while return_dict is None and retries > 0:
             try:
                 return_dict: Dict[str, Any] = await agent_executor.ainvoke(inputs, invoke_config)
-            except (APIError, BadRequestError, AuthenticationError, ChatGoogleGenerativeAIError) as api_error:
+            except (OpenAI_APIError, Anthropic_APIError, ChatGoogleGenerativeAIError) as api_error:
```
d1donlydfink, line 497: Any reason we are getting rid of BadRequestError?

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
d1donlydfink: Don't you want this to be something like `llm is None and found_exception is None` instead of `found_exception is not None`?

I ask because it seems like something not None if found_exception could easily be a misrepresented key for a model that it was reasonable to expect to come from the llm_factory.

Also: Might make some sense to break out what is inside this if clause starting at line 312 into its own method.
That way the logic inside can remain the same and be tested separately while tinkering with the criteria to invoke it separately.

In all, this is getting better because we are letting the existing LlmFactories have a crack at things before we go off trying to resolve classes separately.

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
```diff
@@ -279,6 +307,27 @@ def create_base_chat_model(self, config: Dict[str, Any],
...
+            # Resolve the 'class'
+            llm_class = self._resolve_class_from_path(
+                class_path=class_path,
+                expected_base=BaseLanguageModel,
+                source_file="agent network hocon file",
+                description="llm_config"
+            )
+
+            # copy the config, take 'class' out, and unpack into llm constructor
+            user_config = config.copy()
```
d1donlydfink: Very good you are doing the shallow copy.

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
```diff
@@ -279,6 +307,27 @@ def create_base_chat_model(self, config: Dict[str, Any],
...
                 found_exception = exception
 
+        # Try resolving via 'class' in config if factories failed
+        class_path = config.get("class")
+        if found_exception is not None and class_path:
+            if not isinstance(class_path, str):
+                raise ValueError("'class' in llm_config must be a string")
+
+            # Resolve the 'class'
+            llm_class = self._resolve_class_from_path(
```
d1donlydfink: Use type hints for the new variable so we know what to expect when reading this. Most likely this is:
`llm_class: Type[BaseLanguageModel] = self._resolve_class_from_path(...`

neuro_san/internals/run_context/langchain/llms/standard_langchain_llm_factory.py
d1donlydfink: Good to have model_name as having precedence.

neuro_san/internals/run_context/langchain/toolbox/toolbox_factory.py
```diff
@@ -152,7 +153,7 @@ def create_tool_from_toolbox(
             self._get_from_api_wrapper_method(tool_class) or tool_class
 
         # Validate and instantiate
-        self._check_invalid_args(callable_obj, final_args)
+        ToolboxFactory.check_invalid_args(callable_obj, final_args)
```
d1donlydfink: You should still be able to use `self.` even though this is a static method call.  Looks cleaner IMHO.

neuro_san/internals/run_context/langchain/toolbox/toolbox_factory.py
```diff
@@ -210,16 +211,24 @@ def _resolve_class(self, class_path: str) -> Type[BaseTool]:
...
         :param args: Arguments to check
         """
-        class_args_set: Set[str] = set(signature(method_class).parameters.keys())
+        pydantic_args: Set[str] = set()
+        # Check for if it is a class that extends pydantic BaseModel
+        if isclass(method_class) and issubclass(method_class, BaseModel):
+            # Include field names as args
+            pydantic_args = set(method_class.model_fields.keys())
+        # Combine the arguments
+        class_args_set: Set[str] = set(signature(method_class).parameters.keys()).union(pydantic_args)
```
d1donlydfink: This is pretty squirrelly. Maybe some more code comments here.

neuro_san/internals/run_context/langchain/util/api_key_error_check.py
d1donlydfink: Not entirely sure you should remove the 401 and 404 messages.
These are indications that there is a firewall problem, and the messaging below encompasses that,
which happens to be a very common problem for Cogs because default CorpSec policies.
If we can give people instructions here then that staves off many more questions we no longer have to answer.

neuro_san/internals/run_context/langchain/llms/user_specified_langchain_llm_factory.py
Noravee: shallow copy is used.

neuro_san/internals/run_context/langchain/util/api_key_error_check.py
Noravee: The 401 and 404 messages are added back.

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
Noravee: The `llm is None` and `found_exception is None` cannot happen simultaneously. The if clause is changed to `if llm is None and found_exception is not None and class_path` and the logic is refactored into `create_base_chat_model_from_user_class()` method.

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
Noravee: The Anthropic_APIError catches both BadRequestError and AuthenticationError.

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
Noravee: Thanks for catching it.

neuro_san/internals/run_context/langchain/toolbox/toolbox_factory.py
Noravee: The `check_invalid_args()` method is refactored into another util class `ArgumentValidator`.

neuro_san/internals/run_context/langchain/toolbox/toolbox_factory.py
Noravee: This code snippet was modified and refactored into `ArgumentValidator` class.

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
```diff
@@ -279,11 +307,47 @@ def create_base_chat_model(self, config: Dict[str, Any],
                 # Let the next model have a crack
                 found_exception = exception
 
+        # Try resolving via 'class' in config if factories failed
+        class_path: str = config.get("class")
+        if llm is None and found_exception is not None and class_path:
```
Noravee, line 293: The conditions to instantiate using `class` in `llm_config` in agent network hocon are
- Failed instantiation from the llm factories (thus `llm is None`)
- Consequently, `found_exception is not None`
- There is `class` in the config.

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
```diff
@@ -279,11 +307,47 @@ def create_base_chat_model(self, config: Dict[str, Any],
...
+            found_exception = None
+
         if found_exception is not None:
             raise found_exception
 
         return llm
 
+    def create_base_chat_model_from_user_class(self, class_path: str, config: Dict[str, Any]) -> BaseLanguageModel:
+        """
+        Create a BaseLanguageModel from the user-specified langchain model class.
```
Noravee, line 309: Refactor the logic of instantiating using "class" and user-defined args to `create_base_chat_model_from_user_class()` method.

neuro_san/internals/run_context/langchain/util/argument_validator.py
```diff
@@ -0,0 +1,81 @@
...
+from typing import Union
+from types import MethodType
+from inspect import isclass
+from inspect import signature
+from pydantic import BaseModel
+
+
+class ArgumentValidator:
+    """
+    A utility class for inspecting method and class arguments, particularly useful for
```
Noravee, line 25: Refactor checking for invalid args to a utility class `ArgumentValidator`.

tests/neuro_san/internals/run_context/langchain/util/test_argument_validator.py
```diff
@@ -0,0 +1,88 @@
...
+
+def sample_function(name: str, city: str):
+    """Used for method signature testing"""
+    print(name, city)
+
+
+# ------------------------- Tests -------------------------------
+
+class TestArgumentValidator:
+    """Tests for the ArgumentValidator class and its static validation utilities."""
```
Noravee, line 39: Add test for `ArgumentValidator`.

tests/neuro_san/internals/run_context/langchain/toolbox/test_toolbox_factory.py
```diff
@@ -18,6 +18,12 @@
 
 from neuro_san.internals.run_context.langchain.toolbox.toolbox_factory import ToolboxFactory
 
+RESOLVER_PATH = "leaf_common.config.resolver.Resolver.resolve_class_in_module"
+VALIDATIOR_PATH = (
```
Noravee, line 22: Modify test for toolbox factory to reflect the refactoring.

Noravee:
- Refactor the instantiation of LLM from `class` in `DefaultLlmFactory`
- Refactor the check for invalid args into a util class `ArgumentValidator`
- Create test for `ArgumentValidator`

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
```diff
@@ -176,6 +202,28 @@ def create_full_llm_config(self, config: Dict[str, Any]) -> Dict[str, Any]:
...
+            #
+            # In the standard factory, LLM classes are instantiated like:
+            #   ChatOpenAI(arg_1=config.get("arg_1"), arg_2=config.get("arg_2"))
+            # If a required argument like "arg_2" is missing in the config, config.get("arg_2") returns None,
+            # which may raise an error during instantiation if the argument has no default.
+            #
+            # To prevent this, we first fetch the default arguments for the given class from llm_info,
+            # then merge them with the user-provided config. This ensures all expected arguments are present,
+            # and the user’s config values take precedence over the defaults.
+            config_from_class_in_llm_info: Dict[str, Any] = self.get_chat_class_args(class_from_llm_config)
```
Noravee, line 176: Combine user-specified config with the one in `class` in llm_info.

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
```diff
@@ -227,8 +275,9 @@ def get_chat_class_args(self, chat_class_name: str, use_model_name: str) -> Dict
         chat_classes: Dict[str, Any] = self.llm_infos.get("classes")
         chat_class: Dict[str, Any] = chat_classes.get(chat_class_name)
         if chat_class is None:
-            raise ValueError(f"llm info entry for {use_model_name} uses a 'class' of {chat_class_name} "
-                             "which is not defined in the 'classes' table.")
+            # If chat_class_name is not in "classes" in llm_info,
+            # it could be a user-specified langchain model class
+            return {}
```
Noravee, line 239: Return an empty dictionary instead of raising error if cannot find class in `classes` of llm info.

llm_extension/groq_langchain_llm_factory.py
d1donlydfink: In the spirit of robust tests + limited library dependencies, you could put this class and extra hocon file under the test/ directory and the extra langchain_groq dependency in requirements-build.txt as long as your intent is to add some kind of regularly run test that uses it (unit test if < 15 secs, integration test if longer)

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
```diff
@@ -155,7 +145,43 @@ def resolve_one_llm_factory(self, llm_factory_class_name: str, llm_info_file: st
...
+            source_file: str,
+            description: str
+    ) -> Type:
+
+        parts: List[str] = class_path.split(".")
+        if len(parts) <= 2:
+            raise ValueError(
+                f"Value for '{description}' in {source_file} must be of the form "
+                "<package>.<module>.<ClassName>"
+            )
```
d1donlydfink: FWIW: I have a ResolverUtil brewing in another PR which would likely help here.  
See https://github.com/cognizant-ai-lab/neuro-san/pull/285 UN-3317 First stab at UsageLogger interface
We can integrate after mine goes in.

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
```diff
@@ -215,7 +263,7 @@ def create_full_llm_config(self, config: Dict[str, Any]) -> Dict[str, Any]:
 
         return full_config
 
-    def get_chat_class_args(self, chat_class_name: str, use_model_name: str) -> Dict[str, Any]:
+    def get_chat_class_args(self, chat_class_name: str, use_model_name: Optional[str] = None) -> Dict[str, Any]:
```
d1donlydfink: FWIW: We haven't really been using Optional in our typing unless its completely necessary.
The default arg of None communicates that pretty clearly.

I think the only time we would really need Optional[] is if we were sending this through something very rigorous like Pydantic for serialization and/or deriving from an interface that specifically had Optional as part of it.

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
d1donlydfink: Where does the new equivalent of the above former exception live now?

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
```diff
@@ -242,9 +291,10 @@ def get_chat_class_args(self, chat_class_name: str, use_model_name: str) -> Dict
         return args
 
     def create_base_chat_model(self, config: Dict[str, Any],
-                               callbacks: List[BaseCallbackHandler] = None) -> BaseLanguageModel:
+                               callbacks: Optional[List[BaseCallbackHandler]] = None) -> BaseLanguageModel:
```
d1donlydfink: Another optional Optional.

neuro_san/internals/run_context/langchain/util/argument_validator.py
```diff
@@ -0,0 +1,81 @@
...
+        if invalid_args:
+            raise ValueError(
+                f"Arguments {invalid_args} for '{method_class.__name__}' do not match any attributes "
+                "of the class or any arguments of the method."
+            )
+
+    @staticmethod
+    def get_base_model_args(base_model_class: Type[BaseModel]) -> Set[str]:
+        """
+        Extract all field names and aliases from a Pydantic BaseModel class.
```
d1donlydfink, line 64: It's good that you call out the BaseModel class as coming from Pydantic.
I always get confused thinking it's an LLM, but it's not, it's lower level than that.

neuro_san/internals/run_context/langchain/util/argument_validator.py
d1donlydfink: This is very nicely abstracted.

d1donlydfink (commented):
This is looking really good... good enough at least to take out of Draft mode.
A couple more little things and we are there I think.

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
Noravee: I’ve added the exception back—`ValueError` will now be raised if `chat_class` is None and `use_model_name` is provided. If `use_model_name` is not set, we assume the class may be a user-defined LangChain model and allow it to proceed.

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
Noravee: ResolverUtil is implemented in the llm factory.

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
Noravee: I removed `Optional` type hint.

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
Noravee: I removed `Optional` type hint.

neuro_san/internals/utils/resolver_util.py
```diff
@@ -37,6 +37,35 @@ def create_instance(class_name: str, class_name_source: str, type_of_class: Type
...
+            instance = class_reference()
+        except TypeError as exception:
+            raise ValueError(f"Class '{class_name}' from {class_name_source} "
+                             "must have a no-args constructor") from exception
+
+        return instance
+
+    @staticmethod
+    def create_class(class_name: str, class_name_source: str, type_of_class: Type) -> Type:
+        """
```
Noravee, line 56: I refactor `create_instance()` to `create_class()` method because I need we need constructor with args for user-defined llm.

neuro_san/internals/utils/resolver_util.py
```diff
@@ -59,16 +88,10 @@ def create_instance(class_name: str, class_name_source: str, type_of_class: Type
...
-            instance = class_reference()
-        except TypeError as exception:
-            raise ValueError(f"Class '{class_name}' from {class_name_source} "
-                             "must have a no-args constructor") from exception
-
         # Make sure it is the correct type
-        if not isinstance(instance, type_of_class):
-            raise ValueError(f"Class '{class_name}' from {class_name_source} "
-                             "must be of type {type_of_class.__name__}")
+        if not issubclass(class_reference, type_of_class):
```
Noravee, line 92: Move check correct type to `create_class()`.

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
```diff
@@ -279,11 +288,54 @@ def create_base_chat_model(self, config: Dict[str, Any],
...
+        :param callbacks: A list of BaseCallbackHandlers to add to the chat model.
+
+        :return: A BaseLanguageModel
+        """
+
+        if not isinstance(class_path, str):
+            raise ValueError("'class' in llm_config must be a string")
+
+        # Resolve the 'class'
+        llm_class: Type[BaseLanguageModel] = ResolverUtil.create_class(
```
Noravee, line 322: Use `ResolveUtil.create_class` for the llm since we need to instantiate it with args.

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
```diff
@@ -279,11 +288,54 @@ def create_base_chat_model(self, config: Dict[str, Any],
...
+        # Copy the config, take 'class' out, and add callbacks
+        # Then unpack into llm constructor
+        user_config: Dict[str, Any] = config.copy()
+        user_config.pop("class")
+        user_config["callbacks"] = callbacks
+
+        # Check for invalid args and throw error if found
+        ArgumentValidator.check_invalid_args(llm_class, user_config)
+
+        return llm_class(**user_config)
```
Noravee, line 337: LLM instantiation with args.

Noravee (commented):
## Changes
- Refactor `ResolverUtil.create_instance()` to `ResolverUtil.create_class()`
- Use `ResolverUtil` in `DefaultLlmFactory`.

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
```diff
@@ -123,39 +123,20 @@ def resolve_one_llm_factory(self, llm_factory_class_name: str, llm_info_file: st
...
-        except TypeError as exception:
-            raise ValueError(f"Class {llm_factory_class_name} in {llm_info_file} "
-                             "must have a no-args constructor") from exception
-
-        # Make sure its the correct type
-        if not isinstance(llm_factory, LangChainLlmFactory):
-            raise ValueError(f"Class {llm_factory_class_name} in {llm_info_file} "
-                             "must be of type LangChainLlmFactory")
+        # Resolve and instantiate the factory class
+        llm_factory = ResolverUtil.create_instance(
```
Noravee, line 127: Use ResolverUtil.create_instance for resolve and instantiation of the factory class.

d1donlydfink: Ooo. Nice.

neuro_san/internals/utils/resolver_util.py
d1donlydfink: Makes sense

tests/neuro_san/internals/run_context/langchain/toolbox/test_toolbox_factory.py
```diff
@@ -43,8 +49,7 @@ def test_create_toolbox_returns_single_base_tool(self, factory):
         # Mock user-provided arguments
         user_args = {"param2": "user_value", "param3": "extra_value"}
 
-        with patch("leaf_common.config.resolver.Resolver.resolve_class_in_module") as mock_resolver, \
-                patch.object(factory, "_check_invalid_args") as mock_check_invalid:
+        with patch(RESOLVER_PATH) as mock_resolver, patch(VALIDATIOR_PATH) as mock_check_invalid:
```
d1donlydfink, line 52: Ooo. fancy!

d1donlydfink (approved):
This looks great Nor.  Overall things actually look a lot simpler as this has evolved over time.  Nice work!

PR #271, opened 2025-06-27
UN-2818 json in response

d1donlydfink:


neuro_san/api/grpc/agent.proto
```diff
@@ -93,6 +93,10 @@ message Function {
     // this is really a pydantic/OpenAI function description, which is
     // a bit too complex to specify directly in protobuf.
     google.protobuf.Struct sly_data_schema = 3 [json_name="sly_data_schema"];
+
+    // Optional list of formats that could be parsed from the text of the messages
+    // into the structure field of a ChatMessage.
+    repeated string structure_formats = 4 [json_name="structure_formats"];
```
d1donlydfink: @dsargent : Proposed protocol change for Function response to list the formats that are being parsed for structure in the messages.

This is deliberately separate from stuff that we might parse into the mime_data field.

neuro_san/internals/chat/data_driven_chat_session.py
```diff
@@ -172,12 +174,16 @@ async def streaming_chat(self, user_input: str,
...
+        front_man_extractor = DictionaryExtractor(front_man_spec)
+
         # Get the formats we should parse from the final answer from the config for the network.
         # As of 6/24/25, this is an unadvertised experimental feature.
-        parse_formats: Union[str, List[str]] = self.registry.get_config().get("parse_formats")
+        structure_formats: Union[str, List[str]] = front_man_extractor.get("function.structure_formats")
 
         # Find "the answer" and have that be the content of the last message we send
-        answer_processor = AnswerMessageProcessor(parse_formats=parse_formats)
+        answer_processor = AnswerMessageProcessor(structure_formats=structure_formats)
```
d1donlydfink, line 184: Be specific that we are parsing structure_formats for now.

neuro_san/internals/chat/data_driven_chat_session.py
```diff
@@ -186,7 +192,6 @@ async def streaming_chat(self, user_input: str,
         structure: Dict[str, Any] = answer_processor.get_structure()
 
         # Send back sly_data as the front-man permits
-        front_man_spec: Dict[str, Any] = self.front_man.get_agent_tool_spec()
```
d1donlydfink, line 189: This moves up to happen sooner.

neuro_san/internals/parsers/structure/first_available_structure_parser.py
```diff
@@ -24,14 +24,14 @@ class FirstAvailableStructureParser(StructureParser):
     and does structure parsing on the first one that matches for the text.
     """
 
-    def __init__(self, parse_formats: List[str]):
+    def __init__(self, structure_formats: List[str]):
```
d1donlydfink, line 27: Rename for better scoping - we are only doing structure stuff.

neuro_san/internals/parsers/structure/json_structure_parser.py
```diff
@@ -47,6 +46,10 @@ def parse_structure(self, content: str) -> Dict[str, Any]:
 
         for start_delim, end_delim in delimiters.items():
             if start_delim in content:
+
+                # Note: This code assumes we only have one delimited JSON structure to parse
+                #       within the content.
```
d1donlydfink, line 52: Add comment as to scope of parsing.

neuro_san/internals/parsers/structure/json_structure_parser.py
```diff
@@ -80,7 +83,11 @@ def parse_structure(self, content: str) -> Dict[str, Any]:
         # Attempt parsing the structure from the meat
         structure: Dict[str, Any] = None
         try:
-            structure = parse_json_markdown(meat)
+            structure = loads(meat)
+            if not isinstance(structure, Dict):
+                # json_repair seems to sometimes return an empty string if there is nothing
+                # for it to grab onto.
+                structure = None
```
d1donlydfink, line 91: Use loads() from json_repair package.
Its stated intent is to attack json parsing issues that LLMs create. Sounds like a winner.

neuro_san/message_processing/answer_message_processor.py
```diff
@@ -108,9 +108,9 @@ def process_message(self, chat_message_dict: Dict[str, Any], message_type: ChatM
         if structure is not None:
             self.structure = structure
 
-        if self.structure is None and self.answer is not None:
+        if structure is None and text is not None:
```
d1donlydfink, line 111: Don't look at the member variables, as those accumulate state from the whole of the message stream.
Instead, look at the output of the current message and decide what to do based on that.
This makes for more consistent structure output on the client side.

neuro_san/test/driver/data_driven_agent_test_driver.py
```diff
@@ -40,7 +40,7 @@ class DataDrivenAgentTestDriver:
     specified as a hocon file.
     """
 
-    TEST_KEYS: List[str] = ["text", "sly_data"]
+    TEST_KEYS: List[str] = ["text", "structure", "sly_data"]
```
d1donlydfink, line 43: Look for structure keys in the test case hocon.

neuro_san/test/evaluators/abstract_agent_evaluator.py
```diff
@@ -51,6 +51,8 @@ def evaluate(self, processor: BasicMessageProcessor, test_key: str, verify_for:
         test_dict: Dict[str, Any] = {}
         if first_component == "text":
             test_dict[first_component] = processor.get_answer()
+        elif first_component == "structure":
+            test_dict[first_component] = processor.get_structure()
```
d1donlydfink, line 55: Get the structure from the message processor when looking for structure keys in the test case hocon.

requirements.txt
```diff
@@ -68,3 +68,6 @@ timedinput>=0.1.1
 # to work out-of-the-box.
 # This is not used in any part of neuro-san core operations.
 duckduckgo_search>=7.3.0
+
+# Structure parsing
+json-repair>=0.47.3,<1.0
```
d1donlydfink, line 73: Include the json_repair library as part of the requirements now.
Nice and small and lightweight.  Seems well-constructed.

tests/fixtures/music_nerd_pro/combination_responses_with_history_direct.hocon
```diff
@@ -26,13 +26,20 @@
...
+                    "answer": {
+                        # Keywords says we are going to look for exact matches for each
+                        # element in a list of strings.  All elements need to show up
+                        # in order to pass.
+                        "keywords": "Beatles"
+                    },
+                    "running_cost": {
+                        "value": 3.0
+                    }
+                }
```
d1donlydfink, line 42: Change the music_nerd_pro test cases to look for stuff in the returned structure, since with this change we are specifically asking music_nerd_pro to return a pre-parsed JSON structure.

docs/test_case_hocon_reference.md
```diff
@@ -309,6 +310,63 @@ there is a single [value](#valuenot_value) check against the number 19481.0.
...
+```
+
+In this response part, what is being tested is the structure dictionary returned with "the answer".
+That dictionary is expected to have two keys called "answer" and "running_cost". Each
+have their own tests against single values.  Here, the "answer" key test looks for a single
+[keyword](#keywordsnot_keywords) in a string, and the "running_cost" looks for a single specific
+number [value](#valuenot_value).
+
+Note that for structure, as for sly_data, it's possible to return nested dictionaries.
+To test the key/value pairs inside nested dictionaries, simply nest your test dictionaries.
```
d1donlydfink, line 368: Add some docs as to how to add test cases for structure values.

neuro_san/internals/parsers/structure/json_structure_parser.py
```diff
@@ -15,8 +15,7 @@
 from typing import List
 
 from json.decoder import JSONDecodeError
-
-from langchain_core.utils.json import parse_json_markdown
+from json_repair import loads
```
d1donlydfink, line 18: Use new json_repair package recommended by Darren.

Also FWIW: These StructureParsers are separate from the client-facing MessageProcessors because at some point we might want to turn them on for inter-agent communication. But for now, it's the client-facing communication that is most important.

neuro_san/api/grpc/agent.proto
d1donlydfink: The more I think about this the more I think it's not necessary to add here reporting this information back to the client.

1) We are already giving information with each ChatMessage sent over the stream as to whether or not there is structure information in it.  All a client needs/should do is check if that field is populated and use the information as a primary source of results when it is there.  (We are regularly combining text + structure with the get_compiled_answer() method on the Python MessageProcessors)
2) I don't think the client should really care much about what format was used internally to produce the structure. Implementation detail.

If there is agreement here about the non-necessity of communicating this information via Function(), I think I would move where they key is placed in the hocon.

@dsargent : Thoughts?  What would you do differently with this information coming from Function()

d1donlydfink: @Noravee Any insight into whether map and/or a2a has any such dilemma about parsing structures from text output?

dsargent: I don't think most clients will care about this field. Clients will know what they're interested in, and as long as `structure` is guaranteed to be parseable JSON, they can parse it then inspect it to see if it's an error, "successful result" block, or whatever they expect to see.

The issue for me and the UI client has always been the "Superman problem": conflating free-flow text with potential JSON. Once we have a guaranteed JSON field, that problem goes away.

dsargent: We could consider having a separate `error` field though, as that is a common enough case to be useful. Like having `stderr` and `stdout` as separate streams. The `error` field would either be missing/blank, or contain a JSON error block if there is one. `structure` would be reserved for JSON, as today (and hopefully allow `repeated` soon...then we'd have to rename it to `structures`?)

dsargent: FWIW, here's what A2A uses for arbitrary JSON data.

image

neuro_san/internals/parsers/structure/json_structure_parser.py
dsargent: s/recommended/found/ -- I've never used it and no idea if it's any good; we use a similarly-named package on the UI side, which has worked fine for us. But caveat emptor on the Python one.

neuro_san/api/grpc/agent.proto
Noravee: Every message in A2A consists of a list of "parts", where each part explicitly declares its type using a required field called kind.

Supported kinds include:

"text" for plain human-readable text,

"data" for structured JSON objects,

"file" for file attachments.

For example, if an agent replies:

“Here’s the name and price of the product: { "name": "product_name", "price": 5 }”

An A2A-compliant agent would **not send that entire thing as one text block**. Instead, it would send a message containing **two separate parts**:

```
"parts": [
  {
    "kind": "text",
    "text": "Here’s the name and price of the product:"
  },
  {
    "kind": "data",
    "data": {
      "name": "rubber band",
      "price": 5
    }
  }
]
```

d1donlydfink: @dsargent : Your feedback confirms my suspicion, so I will remove this from the protocol.

As for Errors, I think this is a matter of looking for keys in the return structure as long as this structure parsing is turned on (which we can do for OF once this makes its way over to neuro-ai.).   If people are using our JSON error formatter, there will at least be an "error" key when there is a problem.   

The rub is getting this to be super-generic.  If an agent developer decides that their agent can return their own "error" key as well with different semantics, we can't really help that; that's the agent developer using a tool in their own way however they want. 🤷   

@Noravee : This is very much along the lines of what we are doing, except we cannot currently have > 1 structure in the output.  That is perhaps the protocol change to make, although not with this PR; That would also require fundamental changes to the JsonStructureParser.

neuro_san/internals/parsers/structure/json_structure_parser.py
d1donlydfink: so far I think it is quite good.  Has handled all kinds of slightly-not-json-y text thrown at it nicely.

docs/agent_hocon_reference.md
```diff
@@ -632,3 +633,27 @@ Same as top-level [error_formatter above](#error_formatter), except at single-ag
...
+Supported values are:
+
+- "json"    Looks for JSON in the messages from the LLM and extracts
+
+Currently, the front-man is the only agent node that ever needs to specify this aspect of the [function](#function)
+definition.
+
+Example networks that parse their structure_formats:
+
+- [music_nerd_pro.hocon](../neuro_san/registries/music_nerd_pro.hocon)
```
d1donlydfink, line 659: Add some docs about structure_formats.

FWIW: I have verified that it would be possible to include "structure_formats" within inner-agent definitions and not screw any internals up.  This bodes well for using structure parsing in inter-agent communications in the future.

neuro_san/client/thinking_file_message_processor.py
```diff
@@ -89,9 +89,14 @@ def write_message(self, response: Dict[str, Any], origin_str: str):
...
 
+        if text is None:
+            text = ""
+
         if structure is not None:
             # There is no real text, but there is a structure. JSON-ify it.
-            text = json.dumps(structure, indent=4, sort_keys=True)
+            if len(text) > 0:
+                text += "\n"
+            text += f"```json\n{json.dumps(structure, indent=4, sort_keys=True)}\n```"
```
d1donlydfink, line 99: Make it clear in the thinking files that what came back was JSON from the structure field in a consistent way.

neuro_san/internals/chat/data_driven_chat_session.py
```diff
@@ -240,3 +243,24 @@ def prepare_chat_context(self, chat_message_history: List[Dict[str, Any]]) -> Di
         }
 
         return chat_context
+
+    def create_outgoing_message_processor(self) -> MessageProcessor:
+        """
+        :return: A MessageProcessor that filters messages outgoing to the client.
+                How this works is based on settings on the front man.
+                Can be None.
+        """
```
d1donlydfink, line 252: Not entirely happy about the location here, but it gets things going.

Basically look to the front man as to whether or not we should be attempting to look for structure in any and all messages going back to the client and if so, set up a MessageProcessor that will tweak all the messages for external consumption in the DirectAgentSessions.

This might move when some consolidation between sync/async DirectAgentSession happens.

neuro_san/message_processing/answer_message_processor.py
```diff
@@ -27,11 +27,11 @@ class AnswerMessageProcessor(MessageProcessor):
     of the chat session.
     """
 
-    def __init__(self, parse_formats: Union[str, List[str]] = None):
+    def __init__(self, structure_formats: Union[str, List[str]] = None):
```
d1donlydfink, line 30: Similar rename as above.

neuro_san/message_processing/structure_message_processor.py
```diff
@@ -0,0 +1,81 @@
...
+    import FirstAvailableStructureParser
+from neuro_san.message_processing.message_processor import MessageProcessor
+
+
+class StructureMessageProcessor(MessageProcessor):
+    """
+    Implementation of the MessageProcessor that looks for
+    structure information in chat text and extracts structure based on
+    the format list that was passed into the constructor.
+    """
```
d1donlydfink, line 28: Class to swap out JSON in text w/ parsed structure if at all desired.

neuro_san/session/async_direct_agent_session.py
```diff
@@ -170,6 +175,10 @@ async def streaming_chat(self, request_dict: Dict[str, Any]) -> Generator[Dict[s
             response_dict: Dict[str, Any] = copy(template_response_dict)
             if message_filter.allow(message):
                 # We expect the message to be a dictionary form of chat.ChatMessage
+                if message_processor is not None:
+                    message_type: ChatMessageType = message.get("type")
+                    # Can modify message
+                    message_processor.process_message(message, message_type)
```
d1donlydfink, line 181: If the front man says we want to parse out structure formats do that for every message.

neuro_san/api/grpc/agent.proto
d1donlydfink: This was removed from the protocol.

neuro_san/registries/music_nerd_pro.hocon
```diff
@@ -57,14 +57,17 @@ You’re Music Nerd Pro, the go-to brain for all things rock, pop, and everythin
...
 For each question you receive, call your Accountant agent to calculate the running cost. Otherwise you won't get paid!
 Then answer with a JSON message that has two keys:
 1. An "answer" key whose value has the answer to the question
 2. A "running_cost" key whose value has the running cost computed by the Accountant agent.
 """,
-            "tools": ["Accountant"]
+            "tools": ["Accountant"],
+
+            # Parse any JSON in responses into the structure field of the ChatMessage.
+            "structure_formats": "json"
```
d1donlydfink, line 70: Change music_nerd_pro to always return a parsed JSON structure.

This allows us to test the test infrastructure as well.

neuro_san/session/async_direct_agent_session.py
```diff
@@ -170,6 +175,10 @@ async def streaming_chat(self, request_dict: Dict[str, Any]) -> Generator[Dict[s
             response_dict: Dict[str, Any] = copy(template_response_dict)
             if message_filter.allow(message):
                 # We expect the message to be a dictionary form of chat.ChatMessage
+                if message_processor is not None:
+                    message_type: ChatMessageType = message.get("type")
+                    # Can modify message
+                    await message_processor.async_process_message(message, message_type)
```
d1donlydfink, line 181: If the front man says we want to parse out structure formats do that for every inter-agent message as well going back to the client.

swensel (approved):
LGTM AFAICT

hshahrzad (approved):
Looks good to me.

PR #279, opened 2025-07-01
UN-3152 Add /healthz endpoint for health check.

andreidenissov-cog:
Simple PR to add standard /healthz health check endpoint to http service.

### DIFF
--- a/neuro_san/http_sidecar/handlers/health_check_handler.py
+++ b/neuro_san/http_sidecar/handlers/health_check_handler.py
@@ -43,18 +43,17 @@ class HealthCheckHandler(RequestHandler):
             self.set_header("Access-Control-Allow-Methods", "GET, OPTIONS")
             self.set_header("Access-Control-Allow-Headers", "Content-Type, Transfer-Encoding")
 
-        self.versions: Dict[str, Any] = self.determine_versions()
-
     async def get(self):
         """
         Implementation of GET request handler for API health check.
         """
 
         try:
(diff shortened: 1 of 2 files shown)


neuro_san/http_sidecar/http_sidecar.py
```diff
@@ -126,6 +126,7 @@ def make_app(self, requests_limit: int, logger: EventLoopLogger):
         }
         handlers = []
         handlers.append(("/", HealthCheckHandler, health_request_data))
+        handlers.append(("/healthz", HealthCheckHandler, health_request_data))
         handlers.append(("/api/v1/list", ConciergeHandler, request_data))
```
andreidenissov-cog, line 130: Add a new endpoint. Same handler as for "/" route.

neuro_san/http_sidecar/handlers/health_check_handler.py
```diff
@@ -43,18 +43,17 @@ def initialize(self, forwarded_request_metadata: List[str]):
...
         """
 
         try:
+            versions: Dict[str, Any] = self.determine_versions()
             result_dict: Dict[str, Any] = {
                 "service": "neuro-san agents",
                 "status": "healthy",
-                "versions": self.versions
+                "versions": versions
             }
```
andreidenissov-cog, line 57: Move logic for determining current versions under general try-catch block,
so we don't have some unhandled exception (not really likely, but) just flying out.
Perf impact: should be none, since "initialize" is called for every new incoming request anyway.

PR #281, opened 2025-07-01
UN-3305 Fix message processing on grpc when text is None

d1donlydfink:
@vince-leaf found a problem with the new structure stuff checked in earlier today where the client side was returning None for an answer for grpc only.

What is happening is:
* server infrastructure populates text as non-None empty string (len==0). (This allows direct and http to be fine)
* grpc marshaling turns an empty string into a None value over the wire
* Client side couldn't deal with None text value even though there was happy stuff in structure.

### DIFF
--- a/neuro_san/internals/filters/answer_message_filter.py
+++ b/neuro_san/internals/filters/answer_message_filter.py
@@ -40,9 +40,10 @@ class AnswerMessageFilter(MessageFilter):
             # whose origin length is the only one of length 1.
             return False
 
-        text = chat_message_dict.get("text")
-        if text is None:
-            # Final answers need to be text (for now).
+        text: str = chat_message_dict.get("text")
+        structure: Dict[str, Any] = chat_message_dict.get("structure")
+        if text is None and structure is None:
+            # Final answers need to be text or structure.
             # There might be more options in the future.
             return False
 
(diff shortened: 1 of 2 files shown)


neuro_san/internals/filters/answer_message_filter.py
```diff
@@ -40,9 +40,10 @@ def allow_message(self, chat_message_dict: Dict[str, Any], message_type: ChatMes
             # whose origin length is the only one of length 1.
             return False
 
-        text = chat_message_dict.get("text")
-        if text is None:
-            # Final answers need to be text (for now).
+        text: str = chat_message_dict.get("text")
+        structure: Dict[str, Any] = chat_message_dict.get("structure")
+        if text is None and structure is None:
+            # Final answers need to be text or structure.
```
d1donlydfink, line 46: At least one of text or structure has to be non-None to consider the message to be valid.

neuro_san/message_processing/basic_message_processor.py
```diff
@@ -88,6 +88,8 @@ def get_compiled_answer(self) -> str:
         structure: Dict[str, Any] = self.get_structure()
         if structure is not None:
             string_struct: str = json.dumps(structure, indent=4, sort_keys=True)
+            if compiled is None:
+                compiled = ""
```
d1donlydfink, line 92: If a None string comes over the wire for the message text, start out with an empty string so the subsequent additions to it are fine.

vince-leaf (commented):
LGTM

PR #285, opened 2025-07-03
UN-3317 First stab at UsageLogger interface

d1donlydfink:
Preliminary PR for coming to agreement on what the contract will be for usage logging.

@VenkatesanMohanram , @ShivjithAkshaiR : The 2 files to look at for the contract are:
* neuro_san/interfaces/usage_logger.py  - for the definition of the UsageLogger interface that Venkat and team will implement in their own separate repo.
* neuro_san/deploy/Dockerfile - for the definition of the environment variables that control the basic server behavior to invoke their implementation.

... these are what you two should concentrate on.

These two files represent a first stab at satisfying the following priorities:
* **Requirements** (must-haves):
    * The neuro-san base library must not drag in unnecessary dependencies or setup for minimal library usage. This discourages adoption.
    * User adoption of any Usage Logger implementation is completely optional
    * Configuration of any Usage Logger implementation is ultimately traced back to environment variables that a dev or dev-ops engineer can tweak at container build and/or deployment time.
    *  Deployment of any Usage Logger implementation will be a result of combining pip installation of artifacts from:
        * this neuro-san repo 
        * the repo containing the Usage Logger implementation
    * All Usage Logger implementations need to be able to work inside of an asynchronous server environment
    * The neuro-san client-server protocol will not change in order to satisfy the needs of any UsageLogger.
    * The results of any UsageLogger implementation should be able to at least distinguish token usage patterns for any single user of a multi-user system across multiple requests without users having to specify their own LLM API keys.
* **Goals** (nice-to-haves):
    * A single Usage Logger implementation that is concrete enough to be able to drop as-is into any neuro-san deployment when desired.
    * Eventually ship Venkat's team's Usage Logger implementation as a separate open-source library that neuro-san deployments could choose to use if they wish.
    * Venkat's Team's implementation is exemplary enough to show other people how to adapt/create a UsageLogger for their own purposes
* **Non-Goals** (specifically-not-gunna-haves):
    * We will not be checking in source for any concrete Usage Logger implementation into this repo, save for a very simple debugging version that requires no new dependencies.

I include these because it has been my observation that disagreements about implementation almost always come down to misalignment and misclassification of such statements that define a feature.

My apologies to you both. I probably should have cut a PR and a stab at the above priorities a good while ago.
Doing so could have given you better direction sooner - always a good thing, but alas I have lapsed.
It's been crazy though. 🤷 
   
---

As for the rest:
* Create a bare-bones DebugUsageLogger to test infrastructure without more sophisticated implementation
* Create a WrappedUsageLogger that prepares information from the services to send to any UsageLogger
* Create a UsageLoggerFactory which reads the class env var to make the right UsageLogger instance
* Move some common code that leads up to using the class Resolver into a ResolverUtils method. This could eventually make its way to leaf-common
* Plumb the token accounting dictionary to the UsageLogger inside the AgentService and AsyncAgentService 

Still need to do some testing with stuff as-is, but existing client/server tests are passing, so that's a good sign I didn't screw things up too badly.


neuro_san/deploy/Dockerfile
```diff
@@ -204,5 +204,15 @@ ENV AGENT_MANIFEST_UPDATE_PERIOD_SECONDS=0
...
+# A reference to a Python class that can be used to log per-user request information.
+# This class must have a no-args constructor and implement the
+# neuro_san.interfaces.usage_logger.UsageLogger interface.
+# When not set, this defaults to a null implementation.
+ENV AGENT_USAGE_LOGGER=""
+
+# A space-delimited list of http metadata request keys to forward to the AGENT_USAGE_LOGGER
+# described above. When not set, this defaults to the value provided by
+# AGENT_FORWARDED_REQUEST_METADATA
+ENV AGENT_USAGE_LOGGER_METADATA=""
```
d1donlydfink, line 216: These are the environment variables that control the generic invocation of the UsageLoggers.
Any given UsageLogger implementation might have more config, but at the least that should be fed
by its own environment variables, as this level will have no knowledge of those implementations.

What is not shown in this PR is that the lines above already have a definition for AGENT_FORWARDED_REQUEST_METADATA which is defined to be "user_id request_id",
which implies that if this last env var is not set, those will be the keys that come to the UsageLogger via the request_metadata dictionary (see below).

Venkat and Shivjith, please comment.

neuro_san/interfaces/usage_logger.py
```diff
@@ -0,0 +1,50 @@
...
+
+class UsageLogger:
+    """
+    Interface for logging per-user usage statistics.
+
+    The idea here is that employing an implementation of one of these
+    for any given Neuro SAN server is completely optional.
+    """
+
+    async def log_usage(self, token_dict: Dict[str, Any], request_metadata: Dict[str, Any]):
```
d1donlydfink, line 25: This is the main entry point for any invocation of a UsageLogger.

neuro_san/interfaces/usage_logger.py
```diff
@@ -0,0 +1,50 @@
...
+                                    can come from model costs from libraries instead of directly from
+                                    providers.
+                    "total_tokens" - Total tokens used for the request.
+
+                More keys can appear, but should not be counted on.
+                The ones listed above contain potentially salient information for usage logging purposes.
+
+        :param request_metadata: A dictionary of filtered request metadata whose keys contain
+                identifying information for the usage log.
+        """
```
d1donlydfink, line 49: Description of the contract.

Worth noting that the existing implementation will send a token_dict whose initial key will be "all" and the accompanying dictionary will be all tokens for all LLM usage for the single user's request.

But as I was thinking about it today we might eventually need to send stats broken up by LLM provider as well,
because we do indeed have the ability to send part of a user's request to OpenAI (say) and another part to Anthropic (say).  We will have to get our internal shit together to make that kind of fine-grained assessment,
but that can come later.

neuro_san/internals/utils/resolver_util.py
```diff
@@ -0,0 +1,74 @@
...
+        Goes through some standard checking with standard exceptions thrown
+        to make sure the class is set up the way we want it.
+
+        :param class_name: The fully qualified class name to instantiate
+        :param class_name_source: A string description of where we are getting the value of
+                    class_name, so that exceptions can be more instructive.
+        :param type_of_class: The type that the instance must be in order to pass muster.
+        :return: An instance of the class referred to by class_name if everything is successful.
+                Can return None if class_name is a None or empty string.
+        """
```
d1donlydfink, line 37: This helper method contains an abstraction of commonly repeated code to instantiate a class of a particular type.
Eventually we might use this in other places (thinking DefaultLlmFactory here, from where this was largely lifted), but making such changes is out of scope for this PR.

neuro_san/service/generic/agent_service.py
```diff
@@ -245,6 +247,12 @@ def streaming_chat(self, request_dict: Dict[str, Any],
         request_reporting: Dict[str, Any] = invocation_context.get_request_reporting()
         invocation_context.close()
 
+        # Maybe report token accounting to a UsageLogger
+        token_dict: Dict[str, Any] = request_reporting.get("token_accounting")
+        if token_dict is not None:
+            usage_logger: WrappedUsageLogger = UsageLoggerFactory.create_usage_logger()
+            usage_logger.synchronous_log_usage(token_dict, request_metadata)
```
d1donlydfink, line 254: Have the synchronous service employ the UsageLogger via the new infrastructure.

neuro_san/service/generic/async_agent_service.py
```diff
@@ -241,6 +243,12 @@ async def streaming_chat(self, request_dict: Dict[str, Any],
         request_reporting: Dict[str, Any] = invocation_context.get_request_reporting()
         invocation_context.close()
 
+        # Maybe report token accounting to a UsageLogger
+        token_dict: Dict[str, Any] = request_reporting.get("token_accounting")
+        if token_dict is not None:
+            usage_logger: WrappedUsageLogger = UsageLoggerFactory.create_usage_logger()
+            await usage_logger.log_usage(token_dict, request_metadata)
```
d1donlydfink, line 250: Have the asynchronous service employ the UsageLogger via the new infrastructure.

neuro_san/service/usage/debug_usage_logger.py
```diff
@@ -0,0 +1,49 @@
...
+                                    can come from model costs from libraries instead of directly from
+                                    providers.
+                    "total_tokens" - Total tokens used for the request.
+
+                More keys can appear, but should not be counted on.
+                The ones listed above contain potentially salient information for usage logging purposes.
+
+        :param request_metadata: A dictionary of filtered request metadata whose keys contain
+                identifying information for the usage log.
+        """
```
d1donlydfink, line 49: Does nothing for now. Can add prints or something later.

neuro_san/service/usage/usage_logger_factory.py
```diff
@@ -0,0 +1,40 @@
...
+        Reads the server environment variables to create a UsageLogger instance.
+
+        :return: A WrappedUsageLogger that wraps the class referred to by the
+                AGENT_USAGE_LOGGER env var.  Can throw an exception
+                if there are problems creating the class referenced by the env var.
+        """
+        usage_logger_class_name: str = environ.get("AGENT_USAGE_LOGGER")
+        usage_logger: UsageLogger = ResolverUtil.create_instance(usage_logger_class_name,
+                                                                 "AGENT_USAGE_LOGGER env var",
+                                                                 UsageLogger)
```
d1donlydfink, line 38: Use the environment variable defined in the example Dockerfile to potentially create and use a UsageLogger instance.  Wrap it in a class that does preparation for it.

neuro_san/service/usage/wrapped_usage_logger.py
```diff
@@ -0,0 +1,130 @@
...
+
+from neuro_san.interfaces.usage_logger import UsageLogger
+
+
+class WrappedUsageLogger(UsageLogger):
+    """
+    Implementation of the UsageLogger interface that wraps another UsageLogger
+    and whose log_usage method makes sure sloppier inputs adhere better to the
+    UsageLogger contract.
+    """
```
d1donlydfink, line 28: Class for easier server access to any given UsageLogger implementation.
Also handles the case gracefully when there is no UsageLogger desired.

neuro_san/service/usage/wrapped_usage_logger.py
```diff
@@ -0,0 +1,130 @@
...
+
+                If the requested keys in the metadata are not there, they will also not appear
+                in the returned minimized dictionary.
+        """
+        minimized: Dict[str, Any] = {}
+
+        # Try getting the value from the more specific env var before falling back to the
+        # other env var.
+        keys_string: str = environ.get("AGENT_USAGE_LOGGER_METADATA",
+                                       environ.get("AGENT_FORWARDED_REQUEST_METADATA"))
```
d1donlydfink, line 116: Check the environment variables to make a need-to-know version of the request metadata dictionary that comes along with every request.

neuro_san/service/usage/wrapped_usage_logger.py
```diff
@@ -0,0 +1,130 @@
...
+        :return: A token dictionary compliant to the UsageLogger interface
+        """
+        compliant: Dict[str, Any] = token_dict
+
+        if "total_tokens" in token_dict.keys():
+            # We have a raw token dictionary without any attribution to the LLM.
+            # Make the dictionary compliant
+            compliant = {
+                "all": token_dict
+            }
```
d1donlydfink, line 96: For now, we create a single "all" class for the token dictionaries before we hand it off to any UsageLogger implementation.  This allows for some future-proofing when we get the infrastructure to report token usage between separate LLM providers.

neuro_san/deploy/Dockerfile
ShivjithAkshaiR: Hi @d1donlydfink ,

Thanks for the detailed breakdown. 
We’ve reviewed the `UsageLogger` interface and `AGENT_USAGE_LOGGER` env var mechanism.
We confirm this fits our usage logging use case: it gives us token stats (`token_dict`) and filtered metadata (`request_metadata`), and we can plug in our implementation via env vars without touching Neuro-SAN core.

Really appreciate how clearly, you've laid out the intent and fallback behavior.
From what we understand, AGENT_USAGE_LOGGER_METADATA provides a clean override mechanism for selecting which incoming HTTP metadata fields get routed to the UsageLogger, while defaulting to the values specified by AGENT_FORWARDED_REQUEST_METADATA (user_id, request_id) if not set. This aligns well with Neuro SAN’s overall design keeping the core system modular and config-driven while enabling targeted extensions like usage logging through well-scoped, opt-in layers.

We’re currently doing a deeper review to ensure our logger implementation integrates cleanly within the async service context and behaves as expected when the metadata routing logic is driven entirely through env vars. The separation between request_metadata and sly_data also gives us clarity in how we handle secure and observable fields distinctly, which is great from a logging and auditability perspective.

Things are looking good so far.
Also let us know if there’s a timeline, you'd like us to follow for the integration.

Thanks again for the clarity and the thoughtful structure around this!

BR// Venkat and Shiv

neuro_san/interfaces/usage_logger.py
ShivjithAkshaiR: Got it, the contract and entry point are clear. Thanks!

neuro_san/interfaces/usage_logger.py
ShivjithAkshaiR: Noted! We'll proceed with the current approach using "all" as the primary key in token_dict, aggregating total token usage per request. The idea of breaking down usage by LLM provider makes sense, especially as Neuro SAN evolves to support hybrid routing across providers like OpenAI and Anthropic. We’ll design our logger to keep that extensibility in mind for future updates.

neuro_san/service/usage/wrapped_usage_logger.py
ShivjithAkshaiR: The current approach makes sense and the future-proofing for per-provider token reporting is appreciated. All good on our end.

neuro_san/interfaces/usage_logger.py
d1donlydfink: @Noravee : I think you know the most about LangSmith here.  If you have any thoughts about integration points there, please chime in.  (@VenkatesanMohanram and @ShivjithAkshaiR please chime in if you have juicy tidbits as well)

My initial concerns with LangSmith as a basis are:
* async environment
* multiple LLM providers used within a single agent network
* multiple users

neuro_san/service/usage/wrapped_usage_logger.py
```diff
@@ -0,0 +1,130 @@
...
+        minimal_metadata: Dict[str, Any] = self.minimize_metadata(request_metadata)
+
+        await self.wrapped.log_usage(compliant_token_dict, minimal_metadata)
+
+    def synchronous_log_usage(self, token_dict: Dict[str, Any], request_metadata: Dict[str, Any]):
+        """
+        Logs the token usage for external capture.
+        See comments for log_usage() above.
+        """
+        run(self.log_usage(token_dict, request_metadata))
```
d1donlydfink, line 82: It's possible this isn't quite the right thing to do from our async server infrastructure.
@andreidenissov-cog : Thoughts?

neuro_san/deploy/Dockerfile
d1donlydfink: Thanks for the timely feedback, gents!

neuro_san/service/usage/usage_logger_factory.py
d1donlydfink: FWIW: If we did without the UsageLogger type check, then it's conceivable that any usage logger implementation would not have to depend on neuro-san library itself.  Unclear if this is desirable at this point from @ShivjithAkshaiR and @VenkatesanMohanram 's point of view.  

Another option is to break out these kind of interfaces into their own lighter repo. That's more work, though.
We'll see what needs arise.

neuro_san/service/usage/wrapped_usage_logger.py
andreidenissov-cog: Umm, I think this _should_ be OK since for sync agent service we are supposed to be running
in "just" separate Python/gRPC thread outside of any event loop.
And our AsyncExecutor event loop is already done and closed.
So aside from perf impact possibly (new event loop for every logging call) it should work.

neuro_san/interfaces/usage_logger.py
Noravee: LangSmith supports async environments and works well with multiple LLM providers within a single agent network. For multi-user setups, you just need to manage separate API keys per user. You can enable it without writing extra code—just set LANGSMITH_TRACING=true and provide your LANGSMITH_API_KEY.

d1donlydfink: Ah. That "'just' manage separate API keys per user" is something we do not do.  I knew there was a catch.
Thanks for explaining.

neuro_san/service/usage/wrapped_usage_logger.py
d1donlydfink: OK. We can optimize that if/when the need arises.

PR #286, opened 2025-07-07
UN-3327 - Update neuro-san http server to allow the 'user_id' param.

swensel:
# Description

Update neuro-san http server to allow the 'user_id' param.

Tested this by pointing my local MAA UI with https://github.com/cognizant-ai-lab/neuro-ui/pull/66 to a local Neuro-San instance with this PR.

### DIFF
--- a/neuro_san/service/http/handlers/base_request_handler.py
+++ b/neuro_san/service/http/handlers/base_request_handler.py
@@ -83,7 +83,7 @@ class BaseRequestHandler(RequestHandler):
         if os.environ.get("AGENT_ALLOW_CORS_HEADERS") is not None:
             self.set_header("Access-Control-Allow-Origin", "*")
             self.set_header("Access-Control-Allow-Methods", "GET, POST, OPTIONS")
-            self.set_header("Access-Control-Allow-Headers", "Content-Type, Transfer-Encoding")
+            self.set_header("Access-Control-Allow-Headers", "Content-Type, Transfer-Encoding, User_id")
 
     def get_metadata(self) -> Dict[str, Any]:
         """


neuro_san/service/http/handlers/base_request_handler.py
```diff
@@ -83,7 +83,7 @@ def initialize(self,
         if os.environ.get("AGENT_ALLOW_CORS_HEADERS") is not None:
             self.set_header("Access-Control-Allow-Origin", "*")
             self.set_header("Access-Control-Allow-Methods", "GET, POST, OPTIONS")
-            self.set_header("Access-Control-Allow-Headers", "Content-Type, Transfer-Encoding")
+            self.set_header("Access-Control-Allow-Headers", "Content-Type, Transfer-Encoding, User_id")
```
dsargent, line 86: `User_id` is fine but hyphens are more idiomatic than underscores in http headers (look at all the standard header names: `Accept-Encoding`, `Cache-Control`, `X-Foo-Bar` etc.)

Both are allowed according to the [RFC](https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.6) but hyphens are more thematic.

"But..but...that's what the http server requires". Yes, I know. Underscores are idiomatic _for Python_.

dsargent (approved):

neuro_san/service/http/handlers/base_request_handler.py
swensel: @d1donlydfink Do you have a preference about this?

d1donlydfink: Ah. Interesting.

This is fine for now to check in as-is.

@andreidenissov-cog: we should really be getting this list from our env var(s):
AGENT_FORWARDED_REQUEST_METADATA  (we've had for a while now)
and soon to be AGENT_USAGE_LOGGER_METADATA from https://github.com/cognizant-ai-lab/neuro-san/pull/285

I will file a ticket.

d1donlydfink: Filed: https://leaf-ai.atlassian.net/browse/UN-3330

d1donlydfink: Also: Does upper/lower case matter for these keys?

d1donlydfink (approved):

neuro_san/service/http/handlers/base_request_handler.py
swensel: Darren [pointed out in my other PR](https://github.com/cognizant-ai-lab/neuro-ui/pull/66#issuecomment-3037646303) that per the RFC the http header names are case-insensitive.

dsargent: The web server mustn't care about case (per the RFC) but that doesn't mean Python won't care.

PR #288, opened 2025-07-08
Un-3330 Set metadata keys from env variables.

andreidenissov-cog:
This PR extends setup for server requests metadata keys by using both:
--forwarded_request_metadata CLI parameter (env var AGENT_FORWARDED_REQUEST_METADATA) and
--usage_logger_metadata CLI parameter (env var AGENT_USAGE_LOGGER_METADATA).
Final metadata keys set is a union of both collections.
If CORS headers are enabled for http server,
they are specified also as those combined metadata keys.

Tested by manual debug output of final metadata keys for various combinations of env variables.

### DIFF
--- a/neuro_san/service/http/handlers/base_request_handler.py
+++ b/neuro_san/service/http/handlers/base_request_handler.py
@@ -38,17 +38,6 @@ class BaseRequestHandler(RequestHandler):
     Provides logic to inject neuro-san service specific data
     into local handler context.
     """
-    grpc_to_http = {
-        grpc.StatusCode.INVALID_ARGUMENT: 400,
-        grpc.StatusCode.UNAUTHENTICATED: 401,
-        grpc.StatusCode.PERMISSION_DENIED: 403,
-        grpc.StatusCode.NOT_FOUND: 404,
-        grpc.StatusCode.ALREADY_EXISTS: 409,
-        grpc.StatusCode.INTERNAL: 500,
-        grpc.StatusCode.UNAVAILABLE: 503,
-        grpc.StatusCode.DEADLINE_EXCEEDED: 504
-    }
-
     request_id: int = 0
 
     # pylint: disable=attribute-defined-outside-init
@@ -83,7 +72,12 @@ class BaseRequestHandler(RequestHandler):
         if os.environ.get("AGENT_ALLOW_CORS_HEADERS") is not None:
             self.set_header("Access-Control-Allow-Origin", "*")
             self.set_header("Access-Control-Allow-Methods", "GET, POST, OPTIONS")
-            self.set_header("Access-Control-Allow-Headers", "Content-Type, Transfer-Encoding, User_id")
+            headers: str = "Content-Type, Transfer-Encoding"
+            metadata_headers: str = ", ".join(forwarded_request_metadata)
+            if len(metadata_headers) > 0:
+                headers += f", {metadata_headers}"
+            # Set all allowed headers:
+            self.set_header("Access-Control-Allow-Headers", headers)
 
     def get_metadata(self) -> Dict[str, Any]:
         """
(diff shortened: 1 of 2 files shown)


neuro_san/service/http/handlers/base_request_handler.py
```diff
@@ -65,7 +64,12 @@ def initialize(self,
         if os.environ.get("AGENT_ALLOW_CORS_HEADERS") is not None:
             self.set_header("Access-Control-Allow-Origin", "*")
             self.set_header("Access-Control-Allow-Methods", "GET, POST, OPTIONS")
-            self.set_header("Access-Control-Allow-Headers", "Content-Type, Transfer-Encoding, User_id")
+            headers: str = "Content-Type, Transfer-Encoding"
+            metadata_headers: str = ", ".join(forwarded_request_metadata)
+            if len(metadata_headers) > 0:
+                headers += f", {metadata_headers}"
+            # Set all allowed headers:
+            self.set_header("Access-Control-Allow-Headers", headers)
```
andreidenissov-cog, line 73: Generate CORS headers list from "forwarded_request_metadata" collection passed in.

neuro_san/service/main_loop/server_main_loop.py
```diff
@@ -55,7 +55,8 @@ def __init__(self):
         self.server_name_for_logs: str = DEFAULT_SERVER_NAME_FOR_LOGS
         self.max_concurrent_requests: int = DEFAULT_MAX_CONCURRENT_REQUESTS
         self.request_limit: int = DEFAULT_REQUEST_LIMIT
-        self.forwarded_request_metadata: int = GrpcAgentServer.DEFAULT_FORWARDED_REQUEST_METADATA
+        self.forwarded_request_metadata: str = GrpcAgentServer.DEFAULT_FORWARDED_REQUEST_METADATA
+        self.usage_logger_metadata: str = ""
         self.service_openapi_spec_file: str = self._get_default_openapi_spec_path()
```
andreidenissov-cog, line 60: Default for "usage_logger_metadata" is an empty string.

neuro_san/service/main_loop/server_main_loop.py
```diff
@@ -92,6 +93,10 @@ def parse_args(self):
                                                        self.forwarded_request_metadata),
                                 help="Space-delimited list of http request metadata keys to forward "
                                      "to logs/other requests")
+        arg_parser.add_argument("--usage_logger_metadata", type=str,
+                                default=os.environ.get("AGENT_USAGE_LOGGER_METADATA", ""),
+                                help="Space-delimited list of http request metadata keys to forward "
+                                     "to models usage statistics logger")
         arg_parser.add_argument("--openapi_service_spec_path", type=str,
```
andreidenissov-cog, line 100: Additional CLI server parameter.

neuro_san/service/main_loop/server_main_loop.py
```diff
@@ -113,6 +118,9 @@ def parse_args(self):
         self.max_concurrent_requests = args.max_concurrent_requests
         self.request_limit = args.request_limit
         self.forwarded_request_metadata = args.forwarded_request_metadata
+        self.usage_logger_metadata = args.usage_logger_metadata
+        if len(self.usage_logger_metadata) == 0:
+            self.usage_logger_metadata = self.forwarded_request_metadata
         self.service_openapi_spec_file = args.openapi_service_spec_path
```
andreidenissov-cog, line 126: If absent, usage_logger_metadata is set equal to forwarded_request_metadata.

neuro_san/service/main_loop/server_main_loop.py
```diff
@@ -136,14 +144,20 @@ def main_loop(self):
         """
         self.parse_args()
 
+        # Construct forwarded metadata list as a union of
+        # self.forwarded_request_metadata and self.usage_logger_metadata
+        metadata_set = set(self.forwarded_request_metadata.split())
+        metadata_set = metadata_set | set(self.usage_logger_metadata.split())
+        metadata_str: str = " ".join(sorted(metadata_set))
+
```
andreidenissov-cog, line 154: Combine both metadata lists into final one to be used.

neuro_san/service/main_loop/server_main_loop.py
d1donlydfink: This can also be None when coming from the env var, you should check for that.

neuro_san/service/http/handlers/base_request_handler.py
d1donlydfink: Probably good to watch out for forwarded_request_metadata being None, which would not work in that join()

neuro_san/service/main_loop/server_main_loop.py
d1donlydfink: Beware of None.

PR #290, opened 2025-07-09
UN-3310 Faster server startup.

andreidenissov-cog:
Main idea of this PR is switch to "lazy" service instances instantiation.
Right now, we always create service instances - 2 of them in fact for each agent defined in server manifest;
one for gRPC, one for http, even if say we don't use gRPC for this particular server invocation,
and only use a single agent for chat.
Each service creation involves reading at least 2 files, which adds up.
Changing to service instantiation on demand makes server start up much faster.
There is an initial lag of course for loading all Python classes, but nothing to be done for that I guess.

### DIFF
--- a/neuro_san/service/generic/agent_service_provider.py
+++ b/neuro_san/service/generic/agent_service_provider.py
@@ -1,0 +1,81 @@
+
+# Copyright (C) 2023-2025 Cognizant Digital Business, Evolutionary AI.
+# All Rights Reserved.
+# Issued under the Academic Public License.
+#
+# You can be released from the terms, and requirements of the Academic Public
+# License by purchasing a commercial license.
+# Purchase of a commercial license is mandatory for any use of the
+# neuro-san SDK Software in commercial settings.
+#
+# END COPYRIGHT
+
+from typing import Any
+from typing import Dict
+from threading import Lock
+import copy
+
+from leaf_server_common.server.request_logger import RequestLogger
+
+from neuro_san.service.generic.agent_service import AgentService
+from neuro_san.internals.interfaces.agent_network_provider import AgentNetworkProvider
+from neuro_san.service.generic.agent_server_logging import AgentServerLogging
+
+
+class AgentServiceProvider:
+    """
+    Class providing lazy construction of AgentService instance
+    with given constructor parameters.
+    """
+
+    # pylint: disable=too-many-arguments,too-many-positional-arguments
+    def __init__(self,
+                 request_logger: RequestLogger,
+                 security_cfg: Dict[str, Any],
+                 agent_name: str,
+                 agent_network_provider: AgentNetworkProvider,
+                 server_logging: AgentServerLogging):
+        """
+        Constructor.
+        :param request_logger: The instance of the RequestLogger that helps
+                    keep track of stats
+        :param security_cfg: A dictionary of parameters used to
+                        secure the TLS and the authentication of the gRPC
+                        connection.  Supplying this implies use of a secure
+                        GRPC Channel.  If None, uses insecure channel.
+        :param agent_name: The agent name for the service
+        :param agent_network_provider: The AgentNetworkProvider to use for the session.
+        :param server_logging: An AgentServerLogging instance initialized so that
+                        spawned asynchronous threads can also properly initialize
+                        their logging.
+        """
+        self.request_logger = request_logger
+        self.security_cfg = copy.deepcopy(security_cfg)
+        self.server_logging: AgentServerLogging = server_logging
+        self.agent_network_provider: AgentNetworkProvider = agent_network_provider
+        self.agent_name: str = agent_name
+        self.lock: Lock = Lock()
+        self.service_instance: AgentService = None
+
+    def get_service(self) -> AgentService:
+        """
+        Get service instance.
+        Create it if it is not instantiated yet.
+        """
+        if self.service_instance is None:
+            with self.lock:
+                if self.service_instance is None:
+                    self.service_instance = AgentService(
+                        self.request_logger,
+                        self.security_cfg,
+                        self.agent_name,
+                        self.agent_network_provider,
+                        self.server_logging)
+        return self.service_instance
+
+    def service_created(self) -> bool:
+        """
+        Return True if service instance has already been instantiated;
+               False otherwise
+        """
+        return self.service_instance is not None
--- a/neuro_san/service/generic/async_agent_service_provider.py
+++ b/neuro_san/service/generic/async_agent_service_provider.py
@@ -1,0 +1,80 @@
+
+# Copyright (C) 2023-2025 Cognizant Digital Business, Evolutionary AI.
+# All Rights Reserved.
+# Issued under the Academic Public License.
+#
+# You can be released from the terms, and requirements of the Academic Public
+# License by purchasing a commercial license.
+# Purchase of a commercial license is mandatory for any use of the
+# neuro-san SDK Software in commercial settings.
+#
+# END COPYRIGHT
+
+from typing import Any
+from typing import Dict
+from threading import Lock
+import copy
+
+from neuro_san.service.generic.async_agent_service import AsyncAgentService
+from neuro_san.internals.interfaces.agent_network_provider import AgentNetworkProvider
+from neuro_san.service.interfaces.event_loop_logger import EventLoopLogger
+from neuro_san.service.generic.agent_server_logging import AgentServerLogging
+
+
+class AsyncAgentServiceProvider:
+    """
+    Class providing lazy construction of AsyncAgentService instance
+    with given constructor parameters.
+    """
+
+    # pylint: disable=too-many-arguments,too-many-positional-arguments
+    def __init__(self,
+                 request_logger: EventLoopLogger,
+                 security_cfg: Dict[str, Any],
+                 agent_name: str,
+                 agent_network_provider: AgentNetworkProvider,
+                 server_logging: AgentServerLogging):
+        """
+        Constructor.
+        :param request_logger: The instance of the EventLoopLogger that helps
+                        log information from running event loop
+        :param security_cfg: A dictionary of parameters used to
+                        secure the TLS and the authentication of the gRPC
+                        connection.  Supplying this implies use of a secure
+                        GRPC Channel.  If None, uses insecure channel.
+        :param agent_name: The agent name for the service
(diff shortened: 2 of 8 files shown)


neuro_san/service/http/server/http_sidecar.py
```diff
@@ -181,9 +181,14 @@ def add_agent(self, agent_name: str):
...
-            AsyncAgentService(self.logger, None, agent_name, agent_network_provider, agent_server_logging)
-        self.allowed_agents[agent_name] = agent_service
+        agent_service_provider: AsyncAgentServiceProvider = \
+            AsyncAgentServiceProvider(
+                self.logger,
+                None,
+                agent_name,
+                agent_network_provider,
+                agent_server_logging)
+        self.allowed_agents[agent_name] = agent_service_provider
```
andreidenissov-cog, line 192: Instead of AsyncAgentService instances, "allowed_agents" map now contains AsyncAgentServiceProvider instances for lazy instantiation of AsyncAgentService.

neuro_san/service/generic/agent_service_provider.py
```diff
@@ -0,0 +1,81 @@
...
+        :param server_logging: An AgentServerLogging instance initialized so that
+                        spawned asynchronous threads can also properly initialize
+                        their logging.
+        """
+        self.request_logger = request_logger
+        self.security_cfg = copy.deepcopy(security_cfg)
+        self.server_logging: AgentServerLogging = server_logging
+        self.agent_network_provider: AgentNetworkProvider = agent_network_provider
+        self.agent_name: str = agent_name
+        self.lock: Lock = Lock()
```
andreidenissov-cog, line 57: Save constructor parameters for later - when we will need to instantiate AgentService object on the first use.

neuro_san/service/generic/agent_service_provider.py
```diff
@@ -0,0 +1,81 @@
...
+        """
+        if self.service_instance is None:
+            with self.lock:
+                if self.service_instance is None:
+                    self.service_instance = AgentService(
+                        self.request_logger,
+                        self.security_cfg,
+                        self.agent_name,
+                        self.agent_network_provider,
+                        self.server_logging)
```
andreidenissov-cog, line 73: Standard instance creation pattern for multi-threaded environment.

neuro_san/service/generic/async_agent_service_provider.py
```diff
@@ -0,0 +1,80 @@
...
+# Purchase of a commercial license is mandatory for any use of the
+# neuro-san SDK Software in commercial settings.
+#
+# END COPYRIGHT
+
+from typing import Any
+from typing import Dict
+from threading import Lock
+import copy
+
```
andreidenissov-cog, line 17: This class more or less duplicates AgentServiceProvider logic, but for AsyncAgentService instances.
Maybe we could cook up something more generic and elegant Pythonic style for both classes,
but not in this PR.

neuro_san/service/grpc/grpc_agent_service.py
```diff
@@ -60,18 +61,23 @@ def __init__(self,
...
     def get_request_count(self) -> int:
         """
         :return: The number of currently active requests
         """
-        return self.service.get_request_count()
+        if not self.service_provider.service_created():
+            # Service is not yet instantiated - it has no requests
+            return 0
+        service: AgentService = self.service_provider.get_service()
+        return service.get_request_count()
```
andreidenissov-cog, line 80: That's perf optimization: if service instance has not been instantiated, then its request count is obviously 0.
Otherwise with current gRPC server logic, we'll always instantiate all agent services,
even when not using gRPC service at all.

neuro_san/service/grpc/grpc_agent_service.py
```diff
@@ -60,18 +61,23 @@ def __init__(self,
...
-                         agent_name,
-                         agent_network_provider,
-                         server_logging)
+        self.service_provider: AgentServiceProvider =\
+            AgentServiceProvider(
+                request_logger,
+                security_cfg,
+                agent_name,
+                agent_network_provider,
+                server_logging)
```
andreidenissov-cog, line 71: Create "lazy" service provider - instantiate actual AgentService only when needed.

neuro_san/service/grpc/grpc_agent_service.py
```diff
@@ -89,7 +95,9 @@ def Function(self, request: service_messages.FunctionRequest,
 
         # Get our args in order to pass to grpc-free session level
         request_dict: Dict[str, Any] = MessageToDict(request)
-        response_dict: Dict[str, Any] = self.service.function(request_dict, request_metadata, context)
+        service: AgentService = self.service_provider.get_service()
+        response_dict: Dict[str, Any] =\
+            service.function(request_dict, request_metadata, context)
```
andreidenissov-cog, line 101: Here and below - get AgentService from AgentServiceProvider for actual request processing.

neuro_san/service/http/handlers/base_request_handler.py
```diff
@@ -97,6 +99,26 @@ def get_metadata(self) -> Dict[str, Any]:
...
+            return None
+
+        service_provider: AsyncAgentServiceProvider = self.agent_policy.allow(agent_name)
+        if service_provider is None:
+            self.set_status(404)
+            self.logger.error(metadata, "error: Invalid request path %s", self.request.path)
+            self.do_finish()
+            return None
+        return service_provider.get_service()
+
```
andreidenissov-cog, line 121: Factor out common logic for getting AsyncAgentService for request processing.
Will be used in all (3) neuro-san API requests.

neuro_san/service/http/handlers/base_request_handler.py
```diff
@@ -144,7 +166,7 @@ def prepare(self):
             self.do_finish()
             return
 
-        self.logger.info(self.get_metadata(), f"[REQUEST RECEIVED] {self.request.method} {self.request.uri}")
+        self.logger.debug(self.get_metadata(), f"[REQUEST RECEIVED] {self.request.method} {self.request.uri}")
```
andreidenissov-cog, line 170: Less chatty.

neuro_san/service/http/handlers/connectivity_handler.py
```diff
@@ -29,15 +29,8 @@ async def get(self, agent_name: str):
         Implementation of GET request handler for "connectivity" API call.
         """
         metadata: Dict[str, Any] = self.get_metadata()
-        update_done: bool = await self.update_agents(metadata=metadata)
-        if not update_done:
-            return
-
-        service: AsyncAgentService = self.agent_policy.allow(agent_name)
+        service: AsyncAgentService = await self.get_service(agent_name, metadata)
         if service is None:
```
andreidenissov-cog, line 33: Here and below - use common function from base_request_handler module.

neuro_san/service/http/handlers/function_handler.py
```diff
@@ -29,15 +29,8 @@ async def get(self, agent_name: str):
...
-        if not update_done:
-            return
-
-        service: AsyncAgentService = self.agent_policy.allow(agent_name)
+        service: AsyncAgentService = await self.get_service(agent_name, metadata)
         if service is None:
-            self.set_status(404)
-            self.logger.error({}, "error: Invalid request path %s", self.request.path)
-            self.do_finish()
             return
```
andreidenissov-cog, line 34: Use common function from base_request_handler module.

neuro_san/service/http/handlers/streaming_chat_handler.py
```diff
@@ -58,15 +58,8 @@ async def post(self, agent_name: str):
...
-        if not update_done:
-            return
-
-        service: AsyncAgentService = self.agent_policy.allow(agent_name)
+        service: AsyncAgentService = await self.get_service(agent_name, metadata)
         if service is None:
-            self.set_status(404)
-            self.logger.error({}, "error: Invalid request path %s", self.request.path)
-            self.do_finish()
             return
```
andreidenissov-cog, line 64: Use common function from base_request_handler module.

neuro_san/service/http/server/http_sidecar.py
```diff
@@ -107,13 +107,13 @@ def __call__(self, other_server: AgentServer):
...
-        app.listen(self.http_port)
-        self.logger.info({}, "HTTP server is running on port %d", self.http_port)
-        self.logger.info({}, "HTTP server is shutting down after %d requests", self.requests_limit)
         # Construct initial "allowed" list of agents:
         # no metadata to use here yet.
         self.update_agents(metadata={})
         self.logger.debug({}, "Serving agents: %s", repr(self.allowed_agents.keys()))
+        app.listen(self.http_port)
+        self.logger.info({}, "HTTP server is running on port %d", self.http_port)
+        self.logger.info({}, "HTTP server is shutting down after %d requests", self.requests_limit)
```
andreidenissov-cog, line 117: Start actually listening after all preparations are done.
Log server start at this point.

neuro_san/service/generic/agent_service_provider.py
```diff
@@ -0,0 +1,81 @@
...
+        return self.service_instance
+
+    def service_created(self) -> bool:
```
d1donlydfink, line 76: Nit: Since this is a boolean, perhaps a better name would be is_service_created()
Here and below in the async version.

neuro_san/service/generic/async_agent_service_provider.py
```diff
@@ -0,0 +1,80 @@
...
+        """
+        if self.service_instance is None:
+            with self.lock:
+                if self.service_instance is None:
+                    self.service_instance = AsyncAgentService(
+                        self.request_logger,
+                        self.security_cfg,
+                        self.agent_name,
+                        self.agent_network_provider,
+                        self.server_logging)
```
d1donlydfink, line 72: For next PR consideration: Since all of this is the same except for the type of service_instance, maybe there is a path to abstraction here.
However well functioning all the async stuff is, someday Python will clean up this copy-paste cancer on the language.

neuro_san/service/http/handlers/connectivity_handler.py
d1donlydfink: Nice that this consolidates all that other status/error/do_finish business.

d1donlydfink (approved):
Please consider the nits I mention below for a future PR.

It's conceivable we could get some complaints about initial start up time of individual service. (Thinking: 1C) Cross that bridge when we get there.

PR #292, opened 2025-07-09
Adding env var for deployment_name

kxk302:


neuro_san/internals/run_context/langchain/llms/standard_langchain_llm_factory.py
```diff
@@ -142,7 +142,8 @@ def create_base_chat_model(self, config: Dict[str, Any],
                             # Azure-specific
                             azure_endpoint=self.get_value_or_env(config, "azure_endpoint",
                                                                  "AZURE_OPENAI_ENDPOINT"),
-                            deployment_name=config.get("deployment_name"),
+                            deployment_name=self.get_value_or_env(config, "deployment_name",
```
kxk302, line 145: First get the value for "deployment_name" from config file (Previous behavior). If not available, get the value from the new env variable

neuro_san/internals/run_context/langchain/util/api_key_error_check.py
```diff
@@ -27,7 +27,7 @@
                               "Connection error"],
     "OPENAI_API_VERSION": ["validation error", "api_version", "OPENAI_API_VERSION", "Error code: 404",
                            "Resource not found"],
-    "deployment_name": ["Error code: 404", "Resource not found", "API deployment for this resource does not exist"],
+    "AZURE_OPENAI_DEPLOYMENT_NAME": ["Error code: 404", "Resource not found", "API deployment for this resource does not exist"],
```
kxk302: We now have an env var for this.

neuro_san/internals/run_context/langchain/util/api_key_error_check.py
```diff
@@ -27,7 +27,7 @@
                               "Connection error"],
     "OPENAI_API_VERSION": ["validation error", "api_version", "OPENAI_API_VERSION", "Error code: 404",
```
kxk302, line 28: @d1donlydfink there is another ticket (https://leaf-ai.atlassian.net/browse/UN-3322) to add a documentation link to the error message in this file on how to set API keys etc for Azure. I've looked at the way the error message is constructed and can't think of a non-hacky way to add this feature. Any thoughts on how to do this?

d1donlydfink: If I understand correctly, you could break out a separate dictionary with extra text that gets included based on the env var key here that was found.  If there is no entry in such a dictionary, then no extra text is added.
Sound reasonable?

d1donlydfink (dismissed):

PR #295, opened 2025-07-09
Add JC as a contact for commercial license

d1donlydfink:


LICENSE.txt
```diff
@@ -2,6 +2,7 @@ License
 The following license governs the use of neuro-san SDK Software in academic
 and educational environments. Commercial use requires a commercial license from
 Cognizant Technology Solutions Corp, www.cognizant.com.
+Contact jc.noubeyo@cognizant.com to enquire about commercial licensing.
```
d1donlydfink: Right at the top, right where it talks about commercial licensing in the first place.
I feel like adding this bit does not actually intrude on any existing legal biz.

deepsaia: Instead of direct email, would it be reasonable to rather have a DL that could be used everywhere?

jcyn2004: I agree that using a DL rather than an email would be more effective. In case we use an email, let's pick a professional one such as jc.noubeyo@cognizant.com.

ofrancon: It should be a DL indeed

d1donlydfink: Completely reasonable. Will ask @sidcog  about it.

LICENSE.txt
```diff
@@ -2,6 +2,7 @@ License
 The following license governs the use of neuro-san SDK Software in academic
 and educational environments. Commercial use requires a commercial license from
 Cognizant Technology Solutions Corp, www.cognizant.com.
+Contact NeuroSanLicensing@cognizant.com to enquire about commercial licensing.
```
d1donlydfink, line 5: Now using the NeuroSanLicensing distribution list.

ofrancon (approved):
lgtm

sidcog (approved):
The address works.

PR #298, opened 2025-07-10
Add notion of timestamps to AgentNetworkStorage

d1donlydfink:
* Add timestamp to know when AgentNetworkStorage last changed
* Use the timestamp(s) in HttpSidecar to do less work on each request.  Grpc has different mechanism entirely (listener)

This should make each and every HTTP request faster, helping us to scale up in terms of number of agents.

@andreidenissov-cog : Please look for opportunities for race conditions.
I think I am good, but you might see something I do not.

Tested math_guy_passthrough over http with mods with AGENT_MANIFEST_UPDATE_PERIOD_SECONDS=5

### DIFF
--- a/neuro_san/internals/network_providers/agent_network_storage.py
+++ b/neuro_san/internals/network_providers/agent_network_storage.py
@@ -9,10 +9,12 @@
 #
 # END COPYRIGHT
 
-import logging
-import threading
 from typing import Dict
 from typing import List
+
+import logging
+import threading
+import time
 
 from neuro_san.internals.graph.registry.agent_network import AgentNetwork
 from neuro_san.internals.interfaces.agent_network_provider import AgentNetworkProvider
@@ -33,6 +35,7 @@ class AgentNetworkStorage:
         self.logger = logging.getLogger(self.__class__.__name__)
         self.lock = threading.Lock()
         self.listeners: List[AgentStateListener] = []
+        self.last_modified: float = time.time()
 
     def add_listener(self, listener: AgentStateListener):
         """
@@ -57,6 +60,8 @@ class AgentNetworkStorage:
         with self.lock:
             is_new = self.agents_table.get(agent_name, None) is None
             self.agents_table[agent_name] = agent_network
+            self.last_modified = time.time()
+
         # Notify listeners about this state change:
         # do it outside of internal lock
         for listener in self.listeners:
@@ -89,6 +94,8 @@ class AgentNetworkStorage:
         """
         with self.lock:
             self.agents_table.pop(agent_name, None)
+            self.last_modified = time.time()
+
         # Notify listeners about this state change:
         # do it outside of internal lock
         for listener in self.listeners:
@@ -109,3 +116,9 @@ class AgentNetworkStorage:
         with self.lock:
             # Create static snapshot of agents names collection
             return list(self.agents_table.keys())
+
+    def get_last_modified(self) -> float:
+        """
+        :return: A float timestamp from time.time() as to when the instance was last modified.
+        """
+        return self.last_modified
--- a/neuro_san/internals/network_providers/agent_network_storage.py
+++ b/neuro_san/internals/network_providers/agent_network_storage.py
@@ -58,7 +58,7 @@ class AgentNetworkStorage:
         """
         is_new: bool = False
         with self.lock:
-            is_new = self.agents_table.get(agent_name, None) is None
+            is_new = self.agents_table.get(agent_name) is None
             self.agents_table[agent_name] = agent_network
             self.last_modified = time.time()
 
--- a/neuro_san/service/http/server/http_sidecar.py
+++ b/neuro_san/service/http/server/http_sidecar.py
@@ -19,10 +19,10 @@ from typing import List
 
 import random
 import threading
+import time
 
 from tornado.ioloop import IOLoop
 
-from neuro_san.interfaces.concierge_session import ConciergeSession
 from neuro_san.internals.network_providers.agent_network_storage import AgentNetworkStorage
 from neuro_san.internals.network_providers.single_agent_network_provider import SingleAgentNetworkProvider
 from neuro_san.service.generic.agent_server_logging import AgentServerLogging
@@ -39,7 +39,6 @@ from neuro_san.service.http.logging.http_logger import HttpLogger
 from neuro_san.service.http.server.http_server_app import HttpServerApp
 from neuro_san.service.interfaces.agent_server import AgentServer
 from neuro_san.service.interfaces.event_loop_logger import EventLoopLogger
-from neuro_san.session.direct_concierge_session import DirectConciergeSession
 
 
 class HttpSidecar(AgentAuthorizer, AgentsUpdater):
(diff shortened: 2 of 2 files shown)


neuro_san/internals/network_providers/agent_network_storage.py
```diff
@@ -33,6 +35,7 @@ def __init__(self):
         self.logger = logging.getLogger(self.__class__.__name__)
         self.lock = threading.Lock()
         self.listeners: List[AgentStateListener] = []
+        self.last_modified: float = time.time()
```
d1donlydfink, line 38: Add a timestamp as to when the AgentNetworkStorage was last modified.

neuro_san/internals/network_providers/agent_network_storage.py
```diff
@@ -57,6 +60,8 @@ def add_agent_network(self, agent_name: str, agent_network: AgentNetwork):
         with self.lock:
             is_new = self.agents_table.get(agent_name, None) is None
             self.agents_table[agent_name] = agent_network
+            self.last_modified = time.time()
```
d1donlydfink, line 63: Update the timestamp during add_agent_network()

Worth noting that in this class there is no modify().
To do that clients call this add() method again.

neuro_san/internals/network_providers/agent_network_storage.py
```diff
@@ -89,6 +94,8 @@ def remove_agent_network(self, agent_name: str):
         """
         with self.lock:
             self.agents_table.pop(agent_name, None)
+            self.last_modified = time.time()
```
d1donlydfink, line 97: Update the timestamp during remove_agent_network()

neuro_san/internals/network_providers/agent_network_storage.py
```diff
@@ -55,8 +58,10 @@ def add_agent_network(self, agent_name: str, agent_network: AgentNetwork):
         """
         is_new: bool = False
         with self.lock:
-            is_new = self.agents_table.get(agent_name, None) is None
+            is_new = self.agents_table.get(agent_name) is None
```
d1donlydfink, line 61: None is already the default value for get()

neuro_san/internals/network_providers/agent_network_storage.py
```diff
@@ -109,3 +116,9 @@ def get_agent_names(self) -> List[str]:
         with self.lock:
             # Create static snapshot of agents names collection
             return list(self.agents_table.keys())
+
+    def get_last_modified(self) -> float:
+        """
+        :return: A float timestamp from time.time() as to when the instance was last modified.
+        """
+        return self.last_modified
```
d1donlydfink, line 124: Allow caller access.

neuro_san/service/http/server/http_sidecar.py
```diff
@@ -90,6 +89,7 @@ def __init__(self, start_event: threading.Event,
         self.openapi_service_spec_path: str = openapi_service_spec_path
         self.forwarded_request_metadata: List[str] = forwarded_request_metadata.split(" ")
         self.allowed_agents: Dict[str, AsyncAgentServiceProvider] = {}
+        self.allowed_last_modified: float = 0
```
d1donlydfink, line 92: Set to 0 as its possible Sidecar can come up before AgentNetworkStorage-s.

neuro_san/service/http/server/http_sidecar.py
```diff
@@ -144,34 +144,64 @@ def make_app(self, requests_limit: int, logger: EventLoopLogger):
         return HttpServerApp(handlers, requests_limit, logger)
 
     def allow(self, agent_name) -> AsyncAgentServiceProvider:
-        return self.allowed_agents.get(agent_name, None)
+        return self.allowed_agents.get(agent_name)
```
d1donlydfink, line 147: Default for dictionary get() is already None

neuro_san/service/http/server/http_sidecar.py
```diff
@@ -144,34 +144,64 @@ def make_app(self, requests_limit: int, logger: EventLoopLogger):
...
+
+    def is_needing_allowed_update(self) -> bool:
+        """
+        Quick low-impact check to see if the allowed list of agents need updating.
+
+        :return: True if the allowed_agents list needs updating. False otherwise
+        """
```
d1donlydfink, line 154: Quick low-impact check to see if the allowed list of agents need updating.

Theoretically we shouldn't need to take a lock if all we are doing is checking a single read-only scalar value.

neuro_san/service/http/server/http_sidecar.py
```diff
@@ -144,34 +144,64 @@ def make_app(self, requests_limit: int, logger: EventLoopLogger):
...
         :param metadata: metadata to be used for logging if necessary.
         :return: nothing
         """
-        data: Dict[str, Any] = {}
+        _ = metadata
+
+        # First do a quick check against network storage timestamps to see if
+        # we need to anything more heavyweight.
+        if not self.is_needing_allowed_update():
+            return
```
d1donlydfink, line 177: Do the new quick check first and return early if no news is good news.
Should make every request snappier.

neuro_san/service/http/server/http_sidecar.py
```diff
@@ -144,34 +144,64 @@ def make_app(self, requests_limit: int, logger: EventLoopLogger):
...
+            new_allowed_names.extend(agent_names)
 
-        public_storage: AgentNetworkStorage = self.network_storage_dict.get("public")
-        # Why do we need the Concierge if we already have access to public_storage?
-        session: ConciergeSession = DirectConciergeSession(public_storage, metadata=metadata)
-        agents_dict: Dict[str, List[Dict[str, str]]] = session.list(data)
-        agents_list: List[Dict[str, str]] = agents_dict["agents"]
-        agents: List[str] = []
-        for agent_dict in agents_list:
-            agents.append(agent_dict["agent_name"])
```
d1donlydfink, line 164: Get rid of call to DirectConciergeSession.  A lot of packing/unpacking for no real reason.
Replaced by smaller green block above.

neuro_san/service/http/server/http_sidecar.py
```diff
@@ -144,34 +144,64 @@ def make_app(self, requests_limit: int, logger: EventLoopLogger):
...
+        # we need to anything more heavyweight.
+        if not self.is_needing_allowed_update():
+            return
+
+        # Keep a list of the new allowed agents.
+        new_allowed_names: List[str] = []
+        for network_storage in self.network_storage_dict.values():
+            # Add to the list of agents from the storage as the new allowed set.
+            agent_names: List[str] = network_storage.get_agent_names()
+            new_allowed_names.extend(agent_names)
```
d1donlydfink, line 184: Instead of using a ConciergeSession, just ask the network_storage directly for the agent names.
Much simpler and quicker for something that is done on every request.

neuro_san/service/http/server/http_sidecar.py
```diff
@@ -144,34 +144,64 @@ def make_app(self, requests_limit: int, logger: EventLoopLogger):
...
             # All other agents are disabled:
             allowed_set = set(self.allowed_agents.keys())
             for agent_name in allowed_set:
-                if agent_name not in agents:
+                if agent_name not in new_allowed_names:
                     self.remove_agent(agent_name)
+                    modified = True
+
+            if modified:
+                self.allowed_last_modified = time.time()
```
d1donlydfink, line 204: Only set the timestamp if something changed.
Theoretically this should always happen, but it's not actually obvious that it does.

PR #303, opened 2025-07-11
Add class in llm_config in agent_hocon_reference.md

Noravee:
- Add `class` in `llm_config` and explain about using llm that is not in the default llm info
- Move `error_formatter`, `error_fragments`, `max_execution_seconds`, `max_iterations`, and `verbose` to global scope
- Add `max_execution_seconds`, `max_iterations`, and `verbose` in agent scope


docs/agent_hocon_reference.md
```diff
@@ -239,27 +219,84 @@ Note: _We strongly recommend to **not** set secrets as values within any source
 These files tend to creep into source control repos, and it is **very** bad practice
 to expose secrets by checking them in.
 
-#### verbose
+#### class
+
+You can use the `class` key in two ways:
+
+#### 1. For supported providers (predefined in `default_llm_info.hocon`)
```
d1donlydfink: Shouldn't 1 and 2 have another "#" for being a sub-heading under class?
That is: Class has 4x#  1. and 2. each have 5x#

If we did that, then these would show up in the Table of Contents after a new automated pass, though.
So maybe just no ##'s at all for extra heading and find a way to do ordered list with next in bold maybe.

Noravee: Using five # symbols makes the font very small.
I will use bold text for them then.

docs/agent_hocon_reference.md
```diff
@@ -239,27 +219,89 @@ Note: _We strongly recommend to **not** set secrets as values within any source
 These files tend to creep into source control repos, and it is **very** bad practice
 to expose secrets by checking them in.
 
-#### verbose
+#### class
+
+You can use the `class` key in two ways:
+
+**1. For supported providers (predefined in `default_llm_info.hocon`)**
```
Noravee, line 226: Use bold as suggested.

docs/agent_hocon_reference.md
```diff
@@ -239,27 +219,89 @@ Note: _We strongly recommend to **not** set secrets as values within any source
...
+| NVidia        | nvidiea       |
+| Ollma         | ollama        |
+| OpenAI        | openai        |
+
+You may only provide parameters that are explicitly defined for that provider's class under the
+`classes.<class>.args` section of  
+[`default_llm_info.hocon`](../neuro_san/internals/run_context/langchain/llms/default_llm_info.hocon).  
+Unsupported parameters will be ignored
+
+**2. For custom providers (not in `default_llm_info.hocon`)**
```
Noravee, line 244: Use bold as suggested.

d1donlydfink (approved):

PR #306, opened 2025-07-12
Fix async close() issue

d1donlydfink:
I think I might have found the source of the async close() issue we've been seeing for so long.
@andreidenissov-cog is primary reviewer.

### DIFF
--- a/neuro_san/service/generic/async_agent_service.py
+++ b/neuro_san/service/generic/async_agent_service.py
@@ -208,7 +208,9 @@ class AsyncAgentService:
 
         # Prepare
         factory = ExternalAgentSessionFactory(use_direct=False)
-        invocation_context = SessionInvocationContext(factory, self.llm_factory, self.toolbox_factory, metadata)
+        invocation_context = SessionInvocationContext(factory, self.llm_factory,
+                                                      self.toolbox_factory, metadata,
+                                                      originates_from_async=True)
         invocation_context.start()
 
         # Set up logging inside async thread
--- a/neuro_san/service/usage/wrapped_usage_logger.py
+++ b/neuro_san/service/usage/wrapped_usage_logger.py
@@ -118,7 +118,7 @@ class WrappedUsageLogger(UsageLogger):
         if keys_string is None or len(keys_string) == 0:
             return minimized
 
-        keys: List[str] = keys_string.split[" "]
+        keys: List[str] = keys_string.split(" ")
         for key in keys:
             if key is None or len(key) == 0:
                 # Skip any empty key split from the list. Allows for multi-spaces.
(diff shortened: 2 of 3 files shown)


neuro_san/service/generic/async_agent_service.py
```diff
@@ -208,7 +208,9 @@ async def streaming_chat(self, request_dict: Dict[str, Any],
 
         # Prepare
         factory = ExternalAgentSessionFactory(use_direct=False)
-        invocation_context = SessionInvocationContext(factory, self.llm_factory, self.toolbox_factory, metadata)
+        invocation_context = SessionInvocationContext(factory, self.llm_factory,
+                                                      self.toolbox_factory, metadata,
+                                                      originates_from_async=True)
```
d1donlydfink, line 213: Tell the SessionInvocationContext that we are originating from an asynchronous source.

neuro_san/session/session_invocation_context.py
```diff
@@ -38,7 +38,8 @@ class SessionInvocationContext(InvocationContext):
     def __init__(self, async_session_factory: AsyncAgentSessionFactory,
                  llm_factory: ContextTypeLlmFactory,
                  toolbox_factory: ContextTypeToolboxFactory = None,
-                 metadata: Dict[str, str] = None):
+                 metadata: Dict[str, str] = None,
+                 originates_from_async: bool = False):
```
d1donlydfink, line 43: Take the new optional boolean from the constructor.

neuro_san/session/session_invocation_context.py
```diff
@@ -119,7 +124,8 @@ def close(self):
         Release resources owned by this context
         """
         if self.asyncio_executor is not None:
-            self.asyncio_executor.shutdown()
+            if not self.originates_from_async:
+                self.asyncio_executor.shutdown()
```
d1donlydfink, line 129: Do not shutdown the executor if we are already coming from an asynchronous source.
This seems to make the close() problem go away on HTTP requests.
Kind of explains why it was only seen after the 1st request and only in HTTP.

neuro_san/service/usage/wrapped_usage_logger.py
```diff
@@ -118,7 +118,7 @@ def minimize_metadata(self, request_metadata: Dict[str, Any]) -> Dict[str, Any]:
         if keys_string is None or len(keys_string) == 0:
             return minimized
 
-        keys: List[str] = keys_string.split[" "]
+        keys: List[str] = keys_string.split(" ")
```
d1donlydfink, line 121: Also fix syntax error in WrappedUsageLogger

d1donlydfink:
@andreidenissov-cog has correctly pointed out that by going down the path of this PR we will start to leak resources.
Namely: threads and event loops (at least).

We have since found that upgrading to the latest dependencies in a new venv seems to solve the problem.
At this point we are not sure which dependency is key to making this work, but hey it works.

PR #312, opened 2025-07-15
UN-3352 Refactor watcher timeouts

d1donlydfink:
@andreidenissov-cog is the primary reviewer.

This is a continuation of the previous refactoring except now I have a better idea of how to abstract the periodicity of each StorageUpdater.

* Allow each StorageUpdater to have a notion of its own periodicity with new get_update_period_in_seconds() and needs_updating() methods
* Create an AbstractStorageUpdater which has common needs_updating() policy w/rt its own periodicity
* Move the computation of the polling interval to RegistryStorageUpdater. He is also an AbstractStorageUpdater now.
* Have the StorageWatcher's loop compute the greatest common divisor of all of its StorageUpdaters to set its own periodicity.

Tested: math_guy_passthrough with manifest updates on and off.

### DIFF
--- a/docs/agent_hocon_reference.md
+++ b/docs/agent_hocon_reference.md
@@ -20,7 +20,7 @@ Sub-keys to those dictionaries will be described in the next-level down heading 
     - [commondefs](#commondefs)
         - [replacement_strings](#replacement_strings)
         - [replacement_values](#replacement_values)
-    - [agent_llm_info_file](#agent_llm_info_file)
+    - [llm_info_file](#llm_info_file)
     - [toolbox_info_file](#toolbox_info_file)
     - [llm_config](#llm_config)
         - [model_name](#model_name)
@@ -141,11 +141,19 @@ your string values within your replacement_values and things will work out as yo
 
 ### agent_llm_info_file
 
-The `agent_llm_info_file` key allows you to specify a custom HOCON file that extends the default list of available LLMs used
-by agents in a neuro-san network. This is especially useful if you're using models that are not included in the default
-configuration (e.g., newly released models or organization-specific endpoints).
+Deprecated since version 0.5.46: Use `llm_info_file` instead. It will not be removed until `neuro-san==0.6.0`.
 
-For more information on selecting and customizing models, see the [model_name](#model_name) section below.
+### llm_info_file
+
+The `llm_info_file` key allows you to specify a custom HOCON file that extends the default list of available LLMs used
+by agents in a neuro-san network. This is especially useful if you're using models or providers that are not included in
+the default configuration (e.g., newly released models or organization-specific endpoints).
+
+For more information on selecting and customizing models, see the [model_name](#model_name) and [class](#class) section below.
+
+### agent_toolbox_info_file
+
+Deprecated since version 0.5.46: Use `toolbox_info_file` instead. It will not be removed until `neuro-san==0.6.0`.
 
 ### toolbox_info_file
 
@@ -193,8 +201,8 @@ to expose secrets by checking them in.
 If your favorite model, or new hotness is not listed in `default_llm_info.hocon`,
 you can still use it by specifying the [class](#class) key directly, or by extending the list in one of two ways:
 
-(1) Set the absolute path to your extension HOCON file using the [agent_llm_info_file](#agent_llm_info_file)
-key in the agent network HOCON file.
+(1) Set the absolute path to your extension HOCON file using the or
+[llm_info_file](#llm_info_file) keys in the agent network HOCON file.
 
 (2) Set the extension HOCON file to the environment variable
 [AGENT_LLM_INFO_FILE](./llm_info_hocon_reference.md#AGENT_LLM_INFO_FILE-environment-variable).
@@ -526,9 +534,9 @@ Currently supported tool types include:
 
 The default toolbox configuration is located at [toolbox_info.hocon](../neuro_san/internals/run_context/langchain/toolbox/toolbox_info.hocon).
 
-To use your own tools, create a custom toolbox `.hocon` file and reference it by either:
+To use your own tools, create a custom toolbox `.hocon` file and reference it in one of the following ways:
 
-- Setting the `toolbox_info_file` key in the agent network `.hocon` file, or
+- Setting the `toolbox_info_file` key in the agent network `.hocon` file.
 - Defining the `AGENT_TOOLBOX_INFO_FILE` environment variable.
 
 For more details on tool extension, see the [Toolbox Extension Guide](./toolbox_info_hocon_reference.md#extending-toolbox-info).
--- a/docs/llm_info_hocon_reference.md
+++ b/docs/llm_info_hocon_reference.md
@@ -36,7 +36,7 @@ Sub-keys to those dictionaries will be described in the next-level down heading 
         - [default_config](#default_config)
     - [Extending LLM Info Specifications](#extending-llm-info-specifications)
         - [AGENT_LLM_INFO_FILE environment variable](#agent_llm_info_file-environment-variable)
-        - [agent_llm_info_file key in specific agent hocon files](#agent_llm_info_file-key-in-specific-agent-hocon-files)
+        - [llm_info_file key in specific agent hocon files](#llm_info_file-keys-in-agent-network-hocon)
 
 <!--TOC-->
 
@@ -50,34 +50,34 @@ The default file used with the system is called
 ### Model Name Keys
 
 Top-level keys in the file correspond to newly defined usable names for models an the agent network's
-[llm_config](./agent_hocon_reference.md#model-name).
+[llm_config](./agent_hocon_reference.md#model_name).
 
 The value for any model name key is a dictionary describing the model itself, which the next few headings
 will describe.
 
-#### class
+#### `class`
 
-The class is a string descriptor that refers to an entry in the [classes](#classes) table below.
+The `class` is a string descriptor that refers to an entry in the [`classes`](#classes) table below.
 Nominally each class corresponds to a single LLM provider that has its own instance of langchain's
 BaseLanguageModel, and therefore its own class to construct upon initiating a new LLM-powered agent
 for use in the agent network.
 
 Most often, a single BaseLanguageModel class will support multiple models.
 
-#### model_info_url
+#### `model_info_url`
 
 A URL string that points to the information page for details on the given model.
 
 The idea here is that agent network developers can check this URL as a reference when
 LLM performance does not match expectations.
 
-#### modalities
+#### `modalities`
 
 A dictionary that describes the I/O capabilities that the LLM "speaks".
 
 The keys to this dictionary are described immediately below.
 
-##### input
+##### `input`
 
 A list of strings describing the input modalities of the LLM.
 Some common input modalities include:
@@ -90,7 +90,7 @@ Typically, an LLM will at least have a "text" modality but might have more than 
 Note it is not possible to simply list a new input type and have the capability
 magically manifest itself.  This is a matter of how the LLM was trained.
 
-##### output
+##### `output`
 
 A list of strings describing the output modalities of the LLM.
 Some common output modalities include:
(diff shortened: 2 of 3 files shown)


neuro_san/service/watcher/interfaces/abstract_storage_updater.py
```diff
@@ -0,0 +1,68 @@
...
+# END COPYRIGHT
+
+from neuro_san.service.watcher.interfaces.storage_updater import StorageUpdater
+
+
+class AbstractStorageUpdater(StorageUpdater):
+    """
+    Abstract base class for StorageUpdater implementations for common policy
+    about checking for when it is needed to do update_storage().
+    """
```
d1donlydfink, line 20: Houses common needs_updating() logic.

neuro_san/service/watcher/interfaces/storage_updater.py
```diff
@@ -18,14 +18,29 @@ class StorageUpdater(Startable):
     Interface for specific updating jobs that the Watcher performs.
     """
 
-    def update_storage(self):
+    def start(self):
```
d1donlydfink, line 21: Rearranged the methods to reflect the order in which they tend to get called.
Nothing exciting about these two other than that.

neuro_san/service/watcher/interfaces/storage_updater.py
```diff
@@ -18,14 +18,29 @@ class StorageUpdater(Startable):
...
         """
-        Perform an update
+        Perform start up.
         """
         raise NotImplementedError
 
-    def start(self):
+    def get_update_period_in_seconds(self) -> int:
```
d1donlydfink, line 27: Standardized method for getting the timing from the StorageUpdater.
This allows us to make some higher-level looping decisions in StorageWatcher (later on).

neuro_san/service/watcher/interfaces/storage_updater.py
```diff
@@ -18,14 +18,29 @@ class StorageUpdater(Startable):
...
         """
-        Perform start up.
+        :return: An int describing how long this updater ideally wants to go between
+                calls to update_storage().
+        """
+        raise NotImplementedError
+
+    def needs_updating(self, time_now_in_seconds: float) -> bool:
```
d1donlydfink, line 34: A way for StorageWatcher to ask if updating is necessary before actually updating.

neuro_san/service/watcher/main_loop/storage_watcher.py
```diff
@@ -41,53 +45,57 @@ def __init__(self,
...
+            sleep(self.update_period_in_seconds)
 
-    def compute_polling_interval(self, update_period_seconds: int) -> int:
-        """
-        Compute polling interval for polling observer
-        given requested manifest update period
-        """
-        if update_period_seconds <= 5:
-            return 1
-        return int(round(update_period_seconds / 4))
```
d1donlydfink, line 83: compute_polling_interval() moves to RegistryStorageUpdater where it always belonged.

neuro_san/service/watcher/main_loop/storage_watcher.py
```diff
@@ -41,53 +45,57 @@ def __init__(self,
...
+            time_now_in_seconds: float = time()
+            for storage_updater in self.storage_updaters:
+                if storage_updater.needs_updating(time_now_in_seconds):
+                    storage_updater.update_storage()
```
d1donlydfink, line 118: Note that this part of the loop comes after the sleep() above.

Now when we loop through each StorageUpdater, we ask if it needs updating before doing the update.
The default implementation of this guy will see if the timing squares with its own expectations.

neuro_san/service/watcher/registries/registry_storage_updater.py
```diff
@@ -31,22 +31,40 @@ class RegistryStorageUpdater(StorageUpdater):
     use_polling: bool = True
 
     def __init__(self, network_storage_dict: Dict[str, AgentNetworkStorage],
-                 manifest_path: str,
-                 poll_interval: int):
+                 update_period_in_seconds: int,
+                 manifest_path: str):
```
d1donlydfink, line 35: Rearrange the args.

neuro_san/service/watcher/registries/registry_storage_updater.py
```diff
@@ -16,13 +16,13 @@
...
 from neuro_san.internals.network_providers.agent_network_storage import AgentNetworkStorage
-from neuro_san.service.watcher.interfaces.storage_updater import StorageUpdater
+from neuro_san.service.watcher.interfaces.abstract_storage_updater import AbstractStorageUpdater
 from neuro_san.service.watcher.registries.event_registry_observer import EventRegistryObserver
 from neuro_san.service.watcher.registries.polling_registry_observer import PollingRegistryObserver
 from neuro_san.service.watcher.registries.registry_observer import RegistryObserver
 
 
-class RegistryStorageUpdater(StorageUpdater):
+class RegistryStorageUpdater(AbstractStorageUpdater):
```
d1donlydfink, line 25: Derive from abstract class. Let the period timing aspect of needs_updating() come here for free.

neuro_san/service/watcher/registries/registry_storage_updater.py
```diff
@@ -31,22 +31,40 @@ class RegistryStorageUpdater(StorageUpdater):
...
         self.manifest_path: str = manifest_path
 
         self.observer: RegistryObserver = None
         if self.use_polling:
+            poll_interval: int = self.compute_polling_interval()
             self.observer = PollingRegistryObserver(self.manifest_path, poll_interval)
         else:
             self.observer = EventRegistryObserver(self.manifest_path)
 
+    def compute_polling_interval(self) -> int:
```
d1donlydfink, line 59: This method moves here from StorageWatcher.

neuro_san/service/watcher/main_loop/storage_watcher.py
```diff
@@ -32,7 +36,7 @@ class StorageWatcher(Startable):
     def __init__(self,
                  network_storage_dict: Dict[str, AgentNetworkStorage],
                  manifest_path: str,
-                 update_period_seconds: int,
+                 manifest_update_period_in_seconds: int,
```
d1donlydfink, line 38: Be specific about this value being used for the RegistryStorageUpdater.
Let us no longer assume that this value has anything to do with the update period that is used in this class.

neuro_san/service/watcher/main_loop/storage_watcher.py
```diff
@@ -41,53 +45,57 @@ def __init__(self,
...
+        :param manifest_update_period_in_seconds: update period in seconds
         :param server_status: server status to register the state of updater
         """
         self.network_storage_dict: Dict[str, AgentNetworkStorage] = network_storage_dict
         self.manifest_path: str = manifest_path
-        self.update_period_seconds: int = update_period_seconds
-        self.logger = logging.getLogger(self.__class__.__name__)
-        self.updater_thread = threading.Thread(target=self._run, daemon=True)
+        self.logger: Logger = getLogger(self.__class__.__name__)
+        self.updater_thread = Thread(target=self._run, daemon=True)
```
d1donlydfink, line 52: Anality. 🤷

neuro_san/service/watcher/main_loop/storage_watcher.py
```diff
@@ -41,55 +44,99 @@ def __init__(self,
...
 
-        poll_interval: int = self.compute_polling_interval(update_period_seconds)
         self.storage_updaters: List[StorageUpdater] = [
-            RegistryStorageUpdater(network_storage_dict, manifest_path, poll_interval)
+            RegistryStorageUpdater(network_storage_dict, manifest_update_period_in_seconds, manifest_path),
```
d1donlydfink, line 55: Rearrange the order of args here. The idea is that the common args will be up-front and the add-on args will be at the end. Future implementations of StorageUpdater will likely all need the network_storage_dict and the update_period_in_seconds.

neuro_san/service/watcher/main_loop/storage_watcher.py
```diff
@@ -41,55 +44,99 @@ def __init__(self,
...
+        for storage_updater in storage_updaters:
+            update_period: int = storage_updater.get_update_period_in_seconds()
+            if update_period > 0:
+                # Anybody whose update period is 0 doesn't count.
+                update_periods.append(update_period)
+
+        # If we have nobody, then we shouldn't even run.
+        # Luckily greatest_common_divisor() returns 0 for an empty list.
+        update_period_in_seconds: int = greatest_common_divisor(*update_periods)
+        return update_period_in_seconds
```
d1donlydfink, line 83: Have this top-level notion of update_period_in_seconds be a greatest common denominator of all the StorageUpdaters. This should allow timing to happen for updates however each one expects it, while minimizing the busy-waiting of the loop.

Example: If we have two StorageUpdaters and one has period 6 secs and the other period 8 secs,
the greatest common divisor of the two will be 2 secs and that's what will be reported here with this log message.

neuro_san/service/watcher/main_loop/storage_watcher.py
```diff
@@ -41,55 +44,99 @@ def __init__(self,
...
+        """
+        Start running periodic StorageUpdaters.
+        """
+        self.logger.info("Starting StorageWatcher with %d seconds period",
+                         self.update_period_in_seconds)
+
+        for storage_updater in self.storage_updaters:
+            storage_updater.start()
+
+        self.updater_thread.start()
```
d1donlydfink, line 95: Move start() up here so reading this file flows like its lifecycle.

neuro_san/service/watcher/main_loop/storage_watcher.py
```diff
@@ -41,55 +44,99 @@ def __init__(self,
...
-        if update_period_seconds <= 5:
-            return 1
-        return int(round(update_period_seconds / 4))
+            # See how long it took to update everyone.
+            elapsed_time_in_seconds: float = time() - time_now_in_seconds
 
-    def start(self):
+            # sleep() for just the right amount of time for next time around.
+            # If it took longer than one cycle, then yield the cpu at least.
+            sleep_for_seconds = max(self.update_period_in_seconds - elapsed_time_in_seconds, 0)
```
d1donlydfink, line 125: Compute how much to sleep next time around dynamically just in case some of these updaters take a little long.

neuro_san/service/watcher/main_loop/storage_watcher.py
```diff
@@ -41,55 +44,99 @@ def __init__(self,
...
+
+    def stop(self):
```
d1donlydfink, line 127: Add a stop method that goes in reverse order of start()

neuro_san/service/watcher/main_loop/storage_watcher.py
```diff
@@ -41,55 +44,99 @@ def __init__(self,
                     AgentNetworkStorage instance which keeps all the AgentNetwork instances
                     of a particular grouping.
         :param manifest_path: file path to server manifest file
-        :param update_period_seconds: update period in seconds
+        :param manifest_update_period_in_seconds: update period in seconds
         :param server_status: server status to register the state of updater
         """
         self.network_storage_dict: Dict[str, AgentNetworkStorage] = network_storage_dict
-        self.manifest_path: str = manifest_path
```
d1donlydfink, line 48: manifest path no longer needs to live on this guy. He just gets passed to the RegistryStorageUpdater

neuro_san/service/watcher/main_loop/storage_watcher.py
```diff
@@ -41,55 +44,99 @@ def __init__(self,
...
-        self.update_period_seconds: int = update_period_seconds
```
d1donlydfink, line 49: update_period_seconds gets initialized lower down in the constructor.

andreidenissov-cog:
What I don't like about this approach: "event loop" schema with sequential run of each updater means that updaters begin to depend on each other w.r.t. timing.
Exaggerated example: say we have some big and lazy updater which will run for 20 secs.
We have another one high-frequency fast updater which has to run every 2 secs for 1 sec or whatever.
There will be periods when fast updater does not run for 20 secs, and this maybe a problem.
Since we will not (hopefully) use hundreds of updaters, I would rather go with a setup:
each updater in its own thread managing its own timing requirements.

PR #314, opened 2025-07-15
Bump langchain-openai version - We believe this might address the "Event Loop is closed" problem.

d1donlydfink:


requirements.txt
```diff
@@ -38,7 +38,7 @@ langchain>=0.3.15,<0.4
 langchain-anthropic>=0.3.11,<0.4
 langchain-community>=0.3.19,<0.4
 langchain-google-genai>=2.0.11,<3.0
-langchain-openai>=0.2.5,<0.4
+langchain-openai>=0.3.28,<0.4
```
d1donlydfink, line 41: Use latest langchain-openai, which by some binary search of dependencies vs server tests, seems to be the guy to remove the "Event Loop is closed" problem. 🤷

vince-leaf (commented):
LGTM

PR #315, opened 2025-07-15
Assert that the container is not None before trying in/notIn tests

d1donlydfink:
Better handle a problem that @vince-leaf uncovered

### DIFF
--- a/neuro_san/test/unittest/unit_test_assert_forwarder.py
+++ b/neuro_san/test/unittest/unit_test_assert_forwarder.py
@@ -112,6 +112,7 @@ class UnitTestAssertForwarder(AssertForwarder):
         :param container: Container comparison element
         :param msg: optional string message
         """
+        self.assertIsNotNone(container, msg=msg)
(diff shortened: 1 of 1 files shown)


neuro_san/test/unittest/unit_test_assert_forwarder.py
```diff
@@ -112,6 +112,7 @@ def assertIn(self, member: Any, container: Any, msg: str = None):
         :param container: Container comparison element
         :param msg: optional string message
         """
+        self.assertIsNotNone(container, msg=msg)
```
d1donlydfink, line 115: Be sure the container is not None before trying assertIn()

neuro_san/test/unittest/unit_test_assert_forwarder.py
```diff
@@ -122,6 +123,7 @@ def assertNotIn(self, member: Any, container: Any, msg: str = None):
...
```
d1donlydfink, line 126: Be sure the container is not None before trying assertNotIn()

PR #319, opened 2025-07-17
UN-3360 Fix the HttpServer logger crash.

andreidenissov-cog:
This PR fixes a crash in neuro-san http server when we try to log an error in incoming request.
(Error code 404 for example).
Along the way, I added some extra logging logic for "standard" Tornado ErrorHandler to display client metadata together with other request parameters. Helpful for debugging the source of a problem.

### DIFF
--- a/neuro_san/service/http/handlers/base_request_handler.py
+++ b/neuro_san/service/http/handlers/base_request_handler.py
@@ -83,9 +83,20 @@ class BaseRequestHandler(RequestHandler):
         from incoming request.
         :return: dictionary of user request metadata; possibly empty
         """
-        headers: Dict[str, Any] = self.request.headers
+        return BaseRequestHandler.get_request_metadata(self.request, self.forwarded_request_metadata)
+
+    @classmethod
+    def get_request_metadata(cls, request, forwarded_request_metadata: List[str]) -> Dict[str, Any]:
+        """
+        Extract user metadata defined by forwarded_request_metadata list
+        from incoming request.
+        :param request: incoming http request
+        :param forwarded_request_metadata: list of metadata keys
+        :return: dictionary of user request metadata; possibly empty
+        """
+        headers: Dict[str, Any] = request.headers
         result: Dict[str, Any] = {}
-        for item_name in self.forwarded_request_metadata:
+        for item_name in forwarded_request_metadata:
             if item_name in headers.keys():
                 result[item_name] = headers[item_name]
             elif item_name == "request_id":
--- a/neuro_san/service/http/server/http_server.py
+++ b/neuro_san/service/http/server/http_server.py
@@ -144,7 +144,7 @@ class HttpServer(AgentAuthorizer, AgentStateListener):
         handlers.append((r"/api/v1/([^/]+)/connectivity", ConnectivityHandler, request_initialize_data))
         handlers.append((r"/api/v1/([^/]+)/streaming_chat", StreamingChatHandler, request_initialize_data))
 
-        return HttpServerApp(handlers, requests_limit, logger)
+        return HttpServerApp(handlers, requests_limit, logger, self.forwarded_request_metadata)
 
     def allow(self, agent_name) -> AsyncAgentServiceProvider:
         return self.allowed_agents.get(agent_name, None)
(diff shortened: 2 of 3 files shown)


neuro_san/service/http/handlers/base_request_handler.py
```diff
@@ -83,9 +83,20 @@ def get_metadata(self) -> Dict[str, Any]:
         from incoming request.
         :return: dictionary of user request metadata; possibly empty
         """
-        headers: Dict[str, Any] = self.request.headers
+        return BaseRequestHandler.get_request_metadata(self.request, self.forwarded_request_metadata)
+
+    @classmethod
+    def get_request_metadata(cls, request, forwarded_request_metadata: List[str]) -> Dict[str, Any]:
+        """
+        Extract user metadata defined by forwarded_request_metadata list
```
andreidenissov-cog, line 91: Add a class-level method we can use outside of RequestHandlers class hierarchy.

neuro_san/service/http/server/http_server.py
```diff
@@ -144,7 +144,7 @@ def make_app(self, requests_limit: int, logger: EventLoopLogger):
         handlers.append((r"/api/v1/([^/]+)/connectivity", ConnectivityHandler, request_initialize_data))
         handlers.append((r"/api/v1/([^/]+)/streaming_chat", StreamingChatHandler, request_initialize_data))
 
-        return HttpServerApp(handlers, requests_limit, logger)
+        return HttpServerApp(handlers, requests_limit, logger, self.forwarded_request_metadata)
```
andreidenissov-cog, line 148: Pass list of metadata keys to HttpServerApp - will be used in extended request logger.

neuro_san/service/http/server/http_server_app.py
```diff
@@ -137,10 +147,23 @@ def get_stats(self) -> str:
...
-                            status, request.method, request.uri, request.remote_ip, duration)
+        if isinstance(handler, BaseRequestHandler):
+            request = handler.request
+            metadata: Dict[str, Any] = handler.get_metadata()
+            status = handler.get_status()
+            duration = 1000 * request.request_time()  # in milliseconds
+            # handler.logger is our custom HttpLogger
+            handler.logger.info(metadata, "%d %s %s (%s) %.2fms",
+                                status, request.method, request.uri, request.remote_ip, duration)
+        elif isinstance(handler, ErrorHandler):
```
andreidenissov-cog, line 158: Use our custom logging logic for BaseRequestHandler (and all our subclassed handlers) only.

neuro_san/service/http/server/http_server_app.py
```diff
@@ -137,10 +147,23 @@ def get_stats(self) -> str:
...
+            request = handler.request
+            metadata: Dict[str, Any] =\
+                BaseRequestHandler.get_request_metadata(request, self.forwarded_request_metadata)
+            status = handler.get_status()
+            duration = 1000 * request.request_time()  # in milliseconds
+            # handler.logger is our custom HttpLogger
+            self.logger.error(metadata, "%d %s %s (%s) %.2fms",
+                              status, request.method, request.uri, request.remote_ip, duration)
```
andreidenissov-cog, line 166: Special twist for "standard" Tornado error handler - use our own logger with metadata extracted from incoming request. This allows to identify source of problem requests better, at least map them to client user_id.

neuro_san/service/http/handlers/base_request_handler.py
```diff
@@ -83,9 +83,20 @@ def get_metadata(self) -> Dict[str, Any]:
         from incoming request.
         :return: dictionary of user request metadata; possibly empty
         """
-        headers: Dict[str, Any] = self.request.headers
+        return BaseRequestHandler.get_request_metadata(self.request, self.forwarded_request_metadata)
```
d1donlydfink, line 86: You should be able to use "self" instead of BaseRequestHandler if you were to make it a static method.  Would read better anyway.

neuro_san/service/http/server/http_server.py
d1donlydfink: Could this also help the 1C guys intermittent lack of request metadata in their logs?

neuro_san/service/http/server/http_server_app.py
d1donlydfink: Seems like the 2 if blocks could be consolidated except for how the metadata is obtained to avoid copy/paste.
Doesn't have to be now.

PR #321, opened 2025-07-18
UN-3365 Bumped up the langchain-ollama version to 0.3.4

vince-leaf:
We must bump this lanchain-ollama version from 0.2.3 to 0.3.4 to support turn-on and off reasoning/thinking.


Noravee (approved):
LGTM

PR #324, opened 2025-07-22
Put user_id in metadata from USER env var during tests by default

d1donlydfink:
This will help identify who is running tests against a server when that happens.

This helps shore up the cases coming from our software where user_id might not be set.

### DIFF
--- a/neuro_san/test/driver/data_driven_agent_test_driver.py
+++ b/neuro_san/test/driver/data_driven_agent_test_driver.py
@@ -17,6 +17,7 @@ from typing import Union
 
 from copy import copy
 from datetime import datetime
+from os import environ
 
 from leaf_common.parsers.dictionary_extractor import DictionaryExtractor
 from leaf_common.persistence.easy.easy_hocon_persistence import EasyHoconPersistence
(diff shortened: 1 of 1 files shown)


neuro_san/test/driver/data_driven_agent_test_driver.py
```diff
@@ -159,8 +160,14 @@ def one_iteration(self, test_case: Dict[str, Any], asserts: AssertForwarder, tim
...
         # Collect other session information
         use_direct: bool = test_case.get("use_direct", False)
-        metadata: Dict[str, Any] = test_case.get("metadata", None)
         timeout_in_seconds: float = test_case.get("timeout_in_seconds", None)
+        metadata: Dict[str, Any] = test_case.get("metadata", None)
+        if metadata is None:
+            # Use a default from the user's environment to at least let
+            # a server know who is doing the querying.
+            metadata = {
+                "user_id": environ.get("USER")
```
vince-leaf, line 169: "user_id":  environ.get("USER", "unknown_user")
Do we need a fallback value?
Otherwise, LGTM

vince-leaf (approved):
added a comment and LGTM.

neuro_san/test/driver/data_driven_agent_test_driver.py
d1donlydfink: Every os has this set for any user account. I'd actually rather it be None. Something would be really wrong if it weren't set.

PR #326, opened 2025-07-23
UN-3383 Neuro-San: Remove common API error keywords from AzureOpenAI error detection

Noravee:
### Problem

Some API errors originating from OpenAI are incorrectly flagged as AzureOpenAI errors. This misclassification happens because certain common error keywords are included in the AzureOpenAI error keyword list.

### Solution

Remove the following overly generic error keywords from the AzureOpenAI error detection list to prevent false positives:

- `Connection error`

- `Error code: 401`

- `Error code: 404`

- `Resource not found`

- `Validation error`

### DIFF
--- a/neuro_san/coded_tools/web_search.py
+++ b/neuro_san/coded_tools/web_search.py
@@ -15,7 +15,7 @@ from typing import Union
 
 import logging
 
-from duckduckgo_search import DDGS
+from ddgs import DDGS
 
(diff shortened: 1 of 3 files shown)


d1donlydfink (approved):
OK, but if start getting "Errors are cryptic" complaints (like I anticipate in a meeting on Thursday) I'm sending them to you ;)

It's possible we might eventually feed in model_name as an input to this for finer-grained control

PR #328, opened 2025-07-24
Un-3217 Implement pooling for AsyncExecutors.

andreidenissov-cog:
This PR implements a very simple pooling for AsyncExecutors we use to run per-request service session.
The main goal here is to compensate for (unknown to us) httpx connections pooling policy, which results in transport connection being closed at some unpredictable moment in time, after our request is already processed and underlying AsyncExecutor is shutdown.
With pooling, we basically keep all our AsyncExecutors alive and running, with their event loops active and available for incoming "close this connection" task. There is also some performance advantage in not creating AsyncExecutor from scratch for each request , with thread creation/starting and all.
We can still specify "backward compatibility" mode for the pool with pass-thru creation and shutdown for each AsyncExecutor.

Tested: usual stress tests for both OpenAI and AzureOpenAI models with no exceptions observed.


neuro_san/service/generic/agent_service.py
```diff
@@ -42,6 +43,9 @@
 DO_NOT_LOG_REQUESTS = [
 ]
 
+# No limit on number of concurrently running executors
+MAX_CONCURRENT_EXECUTORS = 0
```
d1donlydfink: Why does this need to be defined twice?

Maybe this should move to AsyncioExecutorPool and be the default for the constructor?

neuro_san/service/generic/async_agent_service.py
```diff
@@ -41,6 +42,9 @@
 DO_NOT_LOG_REQUESTS = [
 ]
 
+# No limit on number of concurrently running executors
+MAX_CONCURRENT_EXECUTORS = 0
```
d1donlydfink: Why does this need to be defined twice?

d1donlydfink (changes requested):
Very interesting approach!

In a way there is a leak because if there is a burst of requests, there will be 1 executor per request
and that number will never go down. But I can see how this will address the "Event loop is closed problem".
Maybe there is a way to slowly shut them down? (not for this PR)

* Single definition for MAX_CONCURRENT_EXECUTORS please
* Move AsyncioExecutorPool to internals.utils to avoid the tangles you've just created.

neuro_san/service/generic/agent_service.py
```diff
@@ -86,6 +90,7 @@ def __init__(self,
         config: Dict[str, Any] = agent_network.get_config()
         self.llm_factory: ContextTypeLlmFactory = MasterLlmFactory.create_llm_factory(config)
         self.toolbox_factory: ContextTypeToolboxFactory = MasterToolboxFactory.create_toolbox_factory(config)
+        self.async_executors_pool: AsyncioExecutorPool = AsyncioExecutorPool(MAX_CONCURRENT_EXECUTORS)
```
d1donlydfink: Should this pool be owned by the server and not the service?

neuro_san/service/generic/async_agent_service.py
```diff
@@ -81,6 +85,7 @@ def __init__(self,
         config: Dict[str, Any] = agent_network.get_config()
         self.llm_factory: ContextTypeLlmFactory = MasterLlmFactory.create_llm_factory(config)
         self.toolbox_factory: ContextTypeToolboxFactory = MasterToolboxFactory.create_toolbox_factory(config)
+        self.async_executors_pool: AsyncioExecutorPool = AsyncioExecutorPool(MAX_CONCURRENT_EXECUTORS)
```
d1donlydfink: Should this pool be owned by the server and not the service?

neuro_san/client/direct_agent_session_factory.py
```diff
@@ -24,6 +24,7 @@
 from neuro_san.internals.graph.persistence.registry_manifest_restorer import RegistryManifestRestorer
 from neuro_san.internals.interfaces.agent_network_provider import AgentNetworkProvider
 from neuro_san.internals.network_providers.agent_network_storage import AgentNetworkStorage
+from neuro_san.service.generic.asyncio_executor_pool import AsyncioExecutorPool
```
d1donlydfink: Cannot import stuff from neuro_san.service in the client area.  Huge reach-around tangle
and will be a problem should we decide to break service stuff out to separate repo.

I think what you want to do is put this in internals.utils for now.
It could easily make its way down to leaf-common.asyncio.

neuro_san/session/session_invocation_context.py
```diff
@@ -17,6 +17,7 @@
 from leaf_common.asyncio.asyncio_executor import AsyncioExecutor
 from leaf_server_common.logging.logging_setup import setup_extra_logging_fields
 
+from neuro_san.service.generic.asyncio_executor_pool import AsyncioExecutorPool
```
d1donlydfink: Cannot include neuro_san.service here either.

Move it to internals.utils.

neuro_san/service/generic/asyncio_executor_pool.py
```diff
@@ -0,0 +1,67 @@
...
+    def __init__(self, max_concurrent: int, reuse_mode: bool = True):
+        """
+        Constructor.
+        :param max_concurrent: maximum allowed number of concurrently running AsyncioExecutors.
+        :param reuse_mode: True, if requested executor instances
+                                 are taken from pool of available ones (pool mode);
+                           False, if requested executor instances are created new
+                                 and shutdown on return (backward compatible mode)
+        """
+        self.max_concurrent = max_concurrent
```
d1donlydfink: Where is self.max_concurrent ever used?

neuro_san/service/generic/agent_service.py
```diff
@@ -42,6 +43,9 @@
 DO_NOT_LOG_REQUESTS = [
 ]
 
+# No limit on number of concurrently running executors
+MAX_CONCURRENT_EXECUTORS = 0
+
```
andreidenissov-cog: For now, we don't limit number of AsyncExecutors which can be created in the pool.

neuro_san/service/generic/agent_service.py
```diff
@@ -86,6 +90,7 @@ def __init__(self,
         config: Dict[str, Any] = agent_network.get_config()
         self.llm_factory: ContextTypeLlmFactory = MasterLlmFactory.create_llm_factory(config)
         self.toolbox_factory: ContextTypeToolboxFactory = MasterToolboxFactory.create_toolbox_factory(config)
+        self.async_executors_pool: AsyncioExecutorPool = AsyncioExecutorPool(MAX_CONCURRENT_EXECUTORS)
         # Load once
```
andreidenissov-cog, line 92: Create an instance of AsyncioExecutorPool for this service.

neuro_san/service/generic/agent_service.py
```diff
@@ -214,7 +219,12 @@ def streaming_chat(self, request_dict: Dict[str, Any],
...
         # Prepare
         factory = ExternalAgentSessionFactory(use_direct=False)
-        invocation_context = SessionInvocationContext(factory, self.llm_factory, self.toolbox_factory, metadata)
+        invocation_context = SessionInvocationContext(
+            factory,
+            self.async_executors_pool,
+            self.llm_factory,
+            self.toolbox_factory,
+            metadata)
         invocation_context.start()
```
andreidenissov-cog, line 226: SessionInvocationContext now takes AsyncioExecutorPool as a parameter.

neuro_san/session/session_invocation_context.py
```diff
@@ -44,15 +48,18 @@ def __init__(self, async_session_factory: AsyncAgentSessionFactory,
...
                          the header. Default is None. Preferred format is a
                          dictionary of string keys to string values.
         """
 
         self.async_session_factory: AsyncAgentSessionFactory = async_session_factory
+        self.async_executors_pool: AsyncioExecutorPool = async_executors_pool
         # Get an async executor to run all tasks for this session instance:
-        self.asyncio_executor: AsyncioExecutor = AsyncioExecutor()
+        self.asyncio_executor: AsyncioExecutor = self.async_executors_pool.get_executor()
         self.origination: Origination = Origination()
```
andreidenissov-cog, line 63: Instead of creating AsynioExecutor instance directly here, request it from the pool.

neuro_san/session/session_invocation_context.py
```diff
@@ -119,7 +127,7 @@ def close(self):
         Release resources owned by this context
         """
         if self.asyncio_executor is not None:
-            self.asyncio_executor.shutdown()
+            self.async_executors_pool.return_executor(self.asyncio_executor)
             self.asyncio_executor = None
```
andreidenissov-cog, line 131: And then we are done, return our AsyncioExecutor back to the pool.

neuro_san/service/generic/async_agent_service.py
```diff
@@ -41,6 +42,9 @@
 DO_NOT_LOG_REQUESTS = [
 ]
 
+# No limit on number of concurrently running executors
+MAX_CONCURRENT_EXECUTORS = 0
+
```
andreidenissov-cog: For now, we don't limit number of AsyncExecutors which can be created in the pool.

neuro_san/service/generic/async_agent_service.py
```diff
@@ -208,7 +213,12 @@ async def streaming_chat(self, request_dict: Dict[str, Any],
...
         # Prepare
         factory = ExternalAgentSessionFactory(use_direct=False)
-        invocation_context = SessionInvocationContext(factory, self.llm_factory, self.toolbox_factory, metadata)
+        invocation_context = SessionInvocationContext(
+            factory,
+            self.async_executors_pool,
+            self.llm_factory,
+            self.toolbox_factory,
+            metadata)
         invocation_context.start()
```
andreidenissov-cog, line 220: We do the same thing as for sync AgentService above.

neuro_san/service/generic/asyncio_executor_pool.py
```diff
@@ -0,0 +1,67 @@
...
+        self.reuse_mode: bool = reuse_mode
+        self.pool = []
+        self.lock: threading.Lock = threading.Lock()
+        self.logger = logging.getLogger(self.__class__.__name__)
+        self.logger.info("AsyncioExecutorPool created: %s reuse: %s max concurrent: %d",
+                         id(self), str(self.reuse_mode), self.max_concurrent)
+
```
andreidenissov-cog: Pretty simple-minded pool of AsyncioExecutors implementation.
Instances move through self.pool sequence in queue mode: fist in - first out.

neuro_san/service/generic/asyncio_executor_pool.py
andreidenissov-cog: Nowhere right now. Was thinking somewhat ahead. Maybe not worth it.

PR #330, opened 2025-07-25
Un 3388 Fixed Gemini LLM test case failure

vince-leaf:
This PR addressed the failure of the Gemini LLM Smoke test case. In summary, three significant changes led to this failure:
1. The fixes for duplication of system messages
2. Update the test case to use "structure_formats"
3. New version of langchain-google-genai-2.1.8 vs. 2.1.6

The current Gemini-2.0-flash model name used for this test is no longer working; it fails 90% of the time. I went with Gemini-2.5-flash and with the existing prompt that we are using for  Anthropic and Ollama test cases.


neuro_san/registries/music_nerd_pro_llm_gemini.hocon
```diff
@@ -15,7 +15,8 @@
 
 {
     "llm_config": {
-        "model_name": "gemini-2.0-flash",
+        "class": "gemini",
+        "model_name": "gemini-2.5-flash",
```
vince-leaf, line 19: I switched to another model name.

neuro_san/registries/music_nerd_pro_llm_gemini.hocon
```diff
@@ -56,13 +57,16 @@ You’re Music Nerd Pro, the go-to brain for all things rock, pop, and everythin
...
-running fees. 
+This service comes at a fee. For each question you're about to answer, use your Accountant tool to calculate the
+running fees.
+
+- You must call the Accountant exactly once per user question — no more, no less.
+- You must not estimate, guess, or invent the cost under any circumstances.
+
+Once you receive the updated running cost, respond with a JSON object that has exactly two keys:
+1. "answer" – your full answer to the user’s question.
+2. "running_cost" – the updated cost returned by the Accountant tool.
```
vince-leaf, line 68: I replaced the prompt used by Ollama and Anthropic LLM test cases.

requirements.txt
```diff
@@ -38,7 +38,7 @@ langchain>=0.3.15,<0.4
 langchain-anthropic>=0.3.11,<0.4
 langchain-aws>=0.2.27,<0.3
 langchain-community>=0.3.19,<0.4
-langchain-google-genai>=2.0.11,<3.0
+langchain-google-genai>=2.1.8,<3.0
```
vince-leaf, line 41: Bumped up the version.

neuro_san/registries/music_nerd_pro_llm_gemini.hocon
d1donlydfink: It is Great and Good that we are trying to use the same prompt across multiple llms

PR #331, opened 2025-07-25
Release <5.0 requirement on protobuf for Python 3.13 compatibility

ofrancon:
This changes lets `pip install` work with python 3.13 without having to compile anything.


requirements.txt
```diff
@@ -7,15 +7,15 @@
 
 # In-house dependencies
 leaf-common>=1.2.23
-leaf-server-common>=0.1.20
+leaf-server-common>=0.1.21
```
ofrancon, line 10: This version of leaf-server-common does NOT depend on protobuf <5.0 anymore

requirements.txt
```diff
@@ -7,15 +7,15 @@
...
 
 # These are needed for generating code from .proto files
 grpcio>=1.62.0
 grpcio-health-checking>=1.62.0
 grpcio-reflection>=1.62.0
 grpcio-tools>=1.62.0
 
-protobuf>=4.25.3,<5.0
+protobuf>=4.25.3
```
ofrancon, line 18: Letting protobuf float

d1donlydfink (approved):
I am in concept good with this, but we need to be sure that removing the protobufs ceiling doesn't mess with the creation of the agent_service.json file for OpenAPI usage and that existing grpc connections are still fine when the new protobuf files are generated.

neuro_san/api/grpc/agent_pb2.py
```diff
@@ -11,13 +11,23 @@
 # END COPYRIGHT
 # -*- coding: utf-8 -*-
 # Generated by the protocol buffer compiler.  DO NOT EDIT!
+# NO CHECKED-IN PROTOBUF GENCODE
 # source: neuro_san/api/grpc/agent.proto
-# Protobuf Python Version: 4.25.1
+# Protobuf Python Version: 6.31.1
```
d1donlydfink: Redid ./neuro_san/api/scripts/do_generate.sh on Python 3.12 with protobuf==6.31.1.
Here are some testing results:
* agent_service.json is generated exactly the same way by the golang tools, as far as I can tell.
* new protobuf 6 server with new protobuf 6 client is fine
* new protobuf 6 server with old protobuf 4 client is fine
* old protobuf 4 server with new protobuf 6 client is fine

What will happen is a ImportError will trip if you have an older protobuf 4 venv and you try to run against these newly generated grpc files...

```
  File "/home//sandbox1/esp/gitrepos/neuro-san/neuro_san/service/main_loop/server_main_loop.py", line 32, in <module>
    from neuro_san.service.grpc.grpc_agent_server import DEFAULT_SERVER_NAME
  File "/home/d/sandbox1/esp/gitrepos/neuro-san/neuro_san/service/grpc/grpc_agent_server.py", line 21, in <module>
    from neuro_san.api.grpc import agent_pb2
  File "/home//sandbox1/esp/gitrepos/neuro-san/neuro_san/api/grpc/agent_pb2.py", line 20, in <module>
    from google.protobuf import runtime_version as _runtime_version
ImportError: cannot import name 'runtime_version' from 'google.protobuf' (/home/venv/ns/lib/python3.12/site-packages/google/protobuf/__init__.py)
```

d1donlydfink (approved):

PR #332, opened 2025-07-25
UN-3352 Add synchronous flag to AsyncCollatingQueue.put() to be very clear as to what's going on

d1donlydfink:
I was looking to re-use AsyncCollatingQueue and I found something I thought to be suspicious:
an asynchronous call put() was putting something in the queue synchronously.
After some digging, I found this to be OK, but its usage needed to be made much more up-front than was currently done.

A couple other minor tweaks.

Tested: http chicken scenario. All agent_thinking comes over as expected.

### DIFF
--- a/neuro_san/deploy/Dockerfile
+++ b/neuro_san/deploy/Dockerfile
@@ -191,7 +191,7 @@ ENV AGENT_REQUEST_LIMIT=-1
 # agent registry files could be added or deleted from registry directory
 # with corresponding changes in manifest file.
 # Update will be done every AGENT_MANIFEST_UPDATE_PERIOD_SECONDS seconds;
-# if value is not specified or <= 0, no dynamic updates will be executed.
+# if value is not specified or <= 0, no such dynamic updates will be executed.
 ENV AGENT_MANIFEST_UPDATE_PERIOD_SECONDS=0
 
 # By default, the HTTP service reports the neuro-san library pip version in its health-check response.
--- a/neuro_san/internals/chat/async_collating_queue.py
+++ b/neuro_san/internals/chat/async_collating_queue.py
@@ -66,21 +66,36 @@ class AsyncCollatingQueue(AsyncIterator, AsyncHopper):
 
         return message
 
-    async def put(self, item: Any):
+    async def put(self, item: Any, synchronous: bool = False):
         """
         Fulfills AsyncHopper interface
 
         :param item: The item to put on the queue.
+        :param synchronous: When False (the default), we use the asynchronous-side
+                of the queue for our put() operation.  This is generally what
+                would be expected inside an async call, which is why it is the default.
+                When True, we use the synchronous side of the queue for put().
+                This ends up being necessary when each end of the queue is serviced
+                in a different asyncio event loop.
         """
-        self.queue.sync_q.put(item)
+        if synchronous:
+            self.queue.sync_q.put(item)
+        else:
+            await self.queue.async_q.put(item)
 
-    async def put_final_item(self):
+    async def put_final_item(self, synchronous: bool = False):
         """
         Puts the final item on the queue indicating that no more data will
         be on the queue and the consumer's iteration can cease when it sees
         this item.
+        :param synchronous: When False (the default), we use the asynchronous-side
+                of the queue for our put() operation.  This is generally what
+                would be expected inside an async call, which is why it is the default.
+                When True, we use the synchronous side of the queue for put().
+                This ends up being necessary when each end of the queue is serviced
+                in a different asyncio event loop.
         """
-        await self.put(self.END_MESSAGE)
+        await self.put(self.END_MESSAGE, synchronous)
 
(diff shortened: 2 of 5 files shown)


neuro_san/deploy/Dockerfile
```diff
@@ -191,7 +191,7 @@ ENV AGENT_REQUEST_LIMIT=-1
 # agent registry files could be added or deleted from registry directory
 # with corresponding changes in manifest file.
 # Update will be done every AGENT_MANIFEST_UPDATE_PERIOD_SECONDS seconds;
-# if value is not specified or <= 0, no dynamic updates will be executed.
+# if value is not specified or <= 0, no such dynamic updates will be executed.
```
d1donlydfink, line 194: Minor verbiage tweak.

neuro_san/internals/chat/async_collating_queue.py
```diff
@@ -66,21 +66,36 @@ async def __anext__(self) -> Dict[str, Any]:
 
         return message
 
-    async def put(self, item: Any):
+    async def put(self, item: Any, synchronous: bool = False):
```
d1donlydfink, line 69: Add this synchronous parameter to allow callers to make the intentions obvious.

neuro_san/internals/chat/data_driven_chat_session.py
```diff
@@ -205,7 +205,12 @@ async def streaming_chat(self, user_input: str,
         # and it doesn't need to wait for any more messages.
         # The consumer await-s for queue.get()
         queue: AsyncCollatingQueue = invocation_context.get_queue()
-        await queue.put_final_item()
+
+        # The synchronous=True is necessary when an async HTTP request is at the get()-ing end of the queue,
+        # as the journal messages come from inside a separate event loop from that request. The lock
+        # taken here ends up being harmless in the synchronous request case (like for gRPC) because
+        # we would only be blocking our own event loop.
+        await queue.put_final_item(synchronous=True)
```
d1donlydfink, line 213: Be very specific as to why we want to do synchronous stuff.
Big-ass comment to boot.

neuro_san/internals/journals/message_journal.py
```diff
@@ -51,4 +51,8 @@ async def write_message(self, message: BaseMessage, origin: List[Dict[str, Any]]
 
         # Queue Producer from this:
         #   https://stackoverflow.com/questions/74130544/asyncio-yielding-results-from-multiple-futures-as-they-arrive
-        await self.hopper.put(message_dict)
+        # The synchronous=True is necessary when an async HTTP request is at the get()-ing end of the queue,
+        # as the journal messages come from inside a separate event loop from that request. The lock
+        # taken here ends up being harmless in the synchronous request case (like for gRPC) because
+        # we would only be blocking our own event loop.
+        await self.hopper.put(message_dict, synchronous=True)
```
d1donlydfink, line 58: Same big-ass comment here too.

neuro_san/internals/network_providers/agent_network_storage.py
```diff
@@ -9,11 +9,12 @@
 #
 # END COPYRIGHT
 
-import logging
-import threading
 from typing import Dict
 from typing import List
 
+import logging
+import threading
```
d1donlydfink, line 16: Beautify imports.

neuro_san/internals/network_providers/agent_network_storage.py
```diff
@@ -56,8 +57,9 @@ def add_agent_network(self, agent_name: str, agent_network: AgentNetwork):
         """
         is_new: bool = False
         with self.lock:
-            is_new = self.agents_table.get(agent_name, None) is None
+            is_new = self.agents_table.get(agent_name) is None
```
d1donlydfink, line 60: None is not necessary, it's the default for get()
Below: Add a little whitespace for readability.

neuro_san/internals/chat/async_collating_queue.py
```diff
@@ -66,21 +66,36 @@ async def __anext__(self) -> Dict[str, Any]:
...
         """
-        self.queue.sync_q.put(item)
+        if synchronous:
+            self.queue.sync_q.put(item)
+        else:
+            await self.queue.async_q.put(item)
 
-    async def put_final_item(self):
+    async def put_final_item(self, synchronous: bool = False):
         """
```
andreidenissov-cog, line 87: So why synchronous: bool is False by default? It looks like we really use it with synchronous == True,
and that's the mode that will work in all our scenarios,
that is: event loop on one side and thread + maybe event loop on the other side.
Then again, queue put operation is fast and not supposed to fail (unless we OOMed) 
so it's OK to sync call it even from running event loop.

d1donlydfink: Because what you would expect from an asynchronous call by default is to call the asynchronous version of put().
Calling the synchronous version of put() is the exception from the expectation, and that's what needs to be called out in the API when it is used like that.

PR #333, opened 2025-07-25
Un-3217 Use AsyncioExecutorPool from leaf-common.

andreidenissov-cog:
This PR uses AsyncExecutorPool from leaf-common for AsyncExecutors we use to run per-request service session.
The main goal here is to compensate for (unknown to us) httpx connections pooling policy, which results in transport connection being closed at some unpredictable moment in time, after our request is already processed and underlying AsyncExecutor is shutdown.
With pooling, we basically keep all our AsyncExecutors alive and running, with their event loops active and available for incoming "close this connection" task. There is also some performance advantage in not creating AsyncExecutor from scratch for each request , with thread creation/starting and all.
We can still specify "backward compatibility" mode for the pool with pass-thru creation and shutdown for each AsyncExecutor.

Tested: usual stress tests for both OpenAI and AzureOpenAI models with no exceptions observed.


neuro_san/client/direct_agent_session_factory.py
```diff
@@ -74,7 +75,8 @@ def create_session(self, agent_name: str, use_direct: bool = False,
         toolbox_factory.load()
 
         factory = ExternalAgentSessionFactory(use_direct=use_direct, network_storage=self.network_storage)
-        invocation_context = SessionInvocationContext(factory, llm_factory, toolbox_factory, metadata)
+        executors_pool: AsyncioExecutorPool = AsyncioExecutorPool()
+        invocation_context = SessionInvocationContext(factory, executors_pool, llm_factory, toolbox_factory, metadata)
         invocation_context.start()
```
andreidenissov-cog, line 80: Directly create AsyncioExecutorPool to use in SessionInvocationContext.

neuro_san/service/generic/agent_service.py
```diff
@@ -214,7 +217,12 @@ def streaming_chat(self, request_dict: Dict[str, Any],
...
         # Prepare
         factory = ExternalAgentSessionFactory(use_direct=False)
-        invocation_context = SessionInvocationContext(factory, self.llm_factory, self.toolbox_factory, metadata)
+        invocation_context = SessionInvocationContext(
+            factory,
+            self.async_executors_pool,
+            self.llm_factory,
+            self.toolbox_factory,
+            metadata)
         invocation_context.start()
```
andreidenissov-cog, line 226: Here and below - get server-wide AsyncioExecutorPool instance to use in our SessionInvocationContext.

requirements.txt
```diff
@@ -6,7 +6,7 @@
 # requirements-build.txt
 
 # In-house dependencies
-leaf-common>=1.2.23
+leaf-common>=1.2.24
 leaf-server-common>=0.1.20
```
andreidenissov-cog, line 11: Use leaf-common with AsyncioExecutorPool.

neuro_san/service/main_loop/server_main_loop.py
```diff
@@ -177,6 +178,8 @@ def main_loop(self):
         metadata_set = metadata_set | set(self.usage_logger_metadata.split())
         metadata_str: str = " ".join(sorted(metadata_set))
 
+        AsyncioExecutorPoolProvider.set_executors_pool(reuse_mode=True)
+
         if self.server_status.grpc_service.is_requested():
```
andreidenissov-cog, line 183: Create executors pool global for the server.

neuro_san/internals/utils/asyncio_executor_pool_provider.py
```diff
@@ -0,0 +1,38 @@
...
+"""
+from leaf_common.asyncio.asyncio_executor_pool import AsyncioExecutorPool
+
+
+class AsyncioExecutorPoolProvider:
+    """
+    Class providing server-wide instance of AsyncioExecutorPool
+    for use by specific agent services.
+    """
+
```
andreidenissov-cog, line 23: Yes, I know - this is a global variable.
But I don't see options that would be less ugly.

d1donlydfink: It's a decent start. Don't have to repair that now. Matters more to get rid of the dreaded error.

FWIW: I have been planting these instances in ServerMainLoop and then doing some plumbing down to where that stuff is needed.  It's possible we might want to consolidate that into its own class so we have to do less plumbing.  We'll get there.   Working title: ServiceContext

d1donlydfink: Question here: Is it reasonable for HTTP server and gRPC server to share this Executor pool?

neuro_san/session/session_invocation_context.py
```diff
@@ -119,7 +127,7 @@ def close(self):
         Release resources owned by this context
         """
         if self.asyncio_executor is not None:
-            self.asyncio_executor.shutdown()
+            self.async_executors_pool.return_executor(self.asyncio_executor)
```
d1donlydfink, line 130: Later on there might be some tightening up w/rt this SessionInvocationObject.
I'm fine with what you have here, but one thing I'd like to keep a longer-term eye on is references on this object that end up providing too much of a hook for into server-land for agent-level mischief.  Not easy!

neuro_san/internals/utils/asyncio_executor_pool_provider.py
andreidenissov-cog: Well, they both use the same AsyncExecutor instances, so I don't see why not technically.
But policy is another thing - we can imagine a dictionary of Executor pools where different server components
use separate pools.

PR #334, opened 2025-07-25
Fix possible bug leading to wrong error message when instantiating LLM

Noravee:
This PR improves error handling during LLM instantiation when factory-based resolution fails.

### Problem:

When all `llm_factories` fail to create an LLM, we fallback to using the `"class"` value in the config to attempt instantiation via a user-defined class. However, `"class"` is always set — even for factory-backed models like `"openai"` or `"bedrock"`.

If we blindly attempt to resolve these default classes again via `create_base_chat_model_from_user_class`, it can raise a secondary error that masks the original factory error, making debugging much harder.

### Fix:

Add a check to skip fallback resolution if `config["class"]` is one of the default classes.

This ensures we preserve and raise the original, meaningful exception instead of a misleading one from reprocessing default class paths.

### DIFF
--- a/neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
+++ b/neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
@@ -36,7 +36,6 @@ from neuro_san.internals.run_context.langchain.util.api_key_error_check import A
 from neuro_san.internals.run_context.langchain.util.argument_validator import ArgumentValidator
 from neuro_san.internals.utils.resolver_util import ResolverUtil
 
-DEFAULT_LLM_CLASSES: Set[str] = {"anthropic", "azure-openai", "bedrock", "gemini", "ollama", "openai", "nvidia"}
 KEYS_TO_REMOVE_FOR_USER_CLASS: Set[str] = {"class", "verbose"}
 
 
@@ -315,10 +314,11 @@ class DefaultLlmFactory(ContextTypeLlmFactory, LangChainLlmFactory):
         # This fallback only applies when the user provides a non-default class path
         # and factory resolution failed.
         class_path: str = config.get("class")
+        default_llm_classes: Set[str] = set(self.llm_infos.get("classes"))
         if (
             llm is None
             and found_exception is not None
-            and class_path not in DEFAULT_LLM_CLASSES
+            and class_path not in default_llm_classes
         ):
             llm = self.create_base_chat_model_from_user_class(class_path, config)
             found_exception = None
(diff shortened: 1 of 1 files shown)


neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
```diff
@@ -36,6 +36,7 @@
 from neuro_san.internals.run_context.langchain.util.argument_validator import ArgumentValidator
 from neuro_san.internals.utils.resolver_util import ResolverUtil
 
+DEFAULT_LLM_CLASSES: Set[str] = {"anthropic", "azure-openai", "bedrock", "gemini", "ollama", "openai", "nvidia"}
```
Noravee: Default classes

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
```diff
@@ -304,9 +305,21 @@ def create_base_chat_model(self, config: Dict[str, Any],
...
+        # a new one from create_base_chat_model_from_user_class.
+        #
+        # This fallback only applies when the user provides a non-default class path
+        # and factory resolution failed.
         class_path: str = config.get("class")
-        if llm is None and found_exception is not None and class_path:
+        if (
+            llm is None
+            and found_exception is not None
+            and class_path not in DEFAULT_LLM_CLASSES
```
Noravee: Check that `class` is not in the default classes.

ofrancon (dismissed):
lgtm

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
d1donlydfink: Why not get these from the keys listed in default_llm_info.hocon "classes" dictionary?
Then when we add more we won't have 2 places to update.

d1donlydfink (changes requested):
Why not get the default classes from the keys listed in default_llm_info.hocon "classes" dictionary?
Then when we add more we won't have 2 places to update.
Single point of truth is waaaay better.

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
```diff
@@ -304,9 +304,22 @@ def create_base_chat_model(self, config: Dict[str, Any],
...
+        # Note: config["class"] is always set — if the user intended to use a default LLMs,
+        # it will point to a known default like "openai" or "bedrock". In those cases,
+        # we avoid re-resolving it here to prevent masking the original error with
+        # a new one from create_base_chat_model_from_user_class.
+        #
+        # This fallback only applies when the user provides a non-default class path
+        # and factory resolution failed.
         class_path: str = config.get("class")
-        if llm is None and found_exception is not None and class_path:
+        default_llm_classes: Set[str] = set(self.llm_infos.get("classes"))
```
Noravee, line 317: Get the default llm classes from the default llm info file.

Thank you @d1donlydfink  for the suggestion.

PR #335, opened 2025-07-25
Modify alias of bedrock models to have "us"

Noravee:
The bedrock models in the default llm info file require access to either us-east-1, us-east-2, or us-west-2 so add "us" to the alias to make in clearer.


ofrancon (approved):
lgtm

PR #336, opened 2025-07-29
UN-3352 ServerContext refactor

d1donlydfink:
* Create a new neuro_san/service/utils directory.
* Move ServerStatus and ServiceStatus into there, breaking a tangle created by having them live in main_loop dir
* Create a new ServerContext that contains global-ish state without singletons which contains:
    * The ServerStatus instance (existing server_main_loop state)
    * The NetworkStorage dictionary (existing server_main_loop state)
    * The AsyncioExecutorPool (new state)
   This ServerContext guy now gets plumbed up and down the service area and eventually is made available to the generic      AgentService and the StorageWatcher.  The main thrust here is to avoid singletons for testability and not have plumbing be the impediment of a proper object ownership tree. (Note: We should probably do something similar for a ServiceContext object, but there's enough going on in this PR already)
* Remove AsyncioExecutorPoolProvider, as its singleton-ness is no longer necessary.

Tested: Chicken scenario on http and grpc


neuro_san/internals/utils/asyncio_executor_pool_provider.py
d1donlydfink, line 1: Singleton-ness of this class is no longer necessary.
Now we carry around a single instance of the AsyncioExecutorPool in a ServerContext object which is owned by the  ServerMainLoop

neuro_san/service/generic/agent_service.py
```diff
@@ -58,7 +58,8 @@ def __init__(self,
                  security_cfg: Dict[str, Any],
                  agent_name: str,
                  agent_network_provider: AgentNetworkProvider,
-                 server_logging: AgentServerLogging):
+                 server_logging: AgentServerLogging,
+                 server_context: ServerContext):
```
d1donlydfink, line 62: I was about to plumb the AsyncioExecutorPool all over the place but then realized that the pain of this plumbing is one of the things that lead to the singleton-ness of the PoolProvider solution in the first place.

So now we pass around a ServerContext object which contains global-sh server state ultimately owned by ServerMainLoop.  This should alleviate the plumbing burden.  Much of what you will see below is (final?) plumbing for this.

Future: It would also make sense to package up the other aspects of this contstructor into a ServiceContext object and plumb that for similar reasons.  Enough going on in here already.

neuro_san/service/generic/async_agent_service.py
```diff
@@ -83,7 +85,7 @@ def __init__(self,
         config: Dict[str, Any] = agent_network.get_config()
         self.llm_factory: ContextTypeLlmFactory = MasterLlmFactory.create_llm_factory(config)
         self.toolbox_factory: ContextTypeToolboxFactory = MasterToolboxFactory.create_toolbox_factory(config)
-        self.async_executors_pool: AsyncioExecutorPool = AsyncioExecutorPoolProvider.get_executors_pool()
+        self.async_executor_pool: AsyncioExecutorPool = server_context.get_executor_pool()
```
d1donlydfink, line 88: Change variable to singular to match the existing class name.  This led to typos.

Seems like maybe we should consolidate the sync/asynchronous versions into an abstract class at some point.

neuro_san/service/grpc/grpc_agent_server.py
```diff
@@ -52,8 +53,7 @@ class GrpcAgentServer(AgentServer, AgentStateListener):
     # pylint: disable=too-many-arguments,too-many-positional-arguments
     def __init__(self, port: int,
                  server_loop_callbacks: ServerLoopCallbacks,
-                 network_storage_dict: Dict[str, AgentNetworkStorage],
-                 server_status: ServerStatus,
+                 server_context: ServerContext,
```
d1donlydfink, line 56: Only plumb the one object that contains both network_storage_dict and server_status.

neuro_san/service/grpc/grpc_agent_server.py
```diff
@@ -164,8 +161,9 @@ def prepare_for_serving(self):
 
         # Add listener to handle adding per-agent gRPC services
         # to our dynamic router:
-        public_storage: AgentNetworkStorage = self.network_storage_dict.get("public")
-        for network_storage in self.network_storage_dict.values():
+        network_storage_dict: Dict[str, AgentNetworkStorage] = self.server_context.get_network_storage_dict()
+        public_storage: AgentNetworkStorage = network_storage_dict.get("public")
+        for network_storage in network_storage_dict.values():
```
d1donlydfink, line 166: Use the objects on the ServerContext now.

neuro_san/service/main_loop/server_main_loop.py
```diff
@@ -66,11 +66,7 @@ def __init__(self):
         self.grpc_server: GrpcAgentServer = None
         self.http_server: HttpServer = None
         self.manifest_files: List[str] = []
-        # Dictionary is string key (describing scope) to AgentNetworkStorage grouping.
-        self.network_storage_dict: Dict[str, AgentNetworkStorage] = {
-            "public": AgentNetworkStorage()
-        }
-        self.server_status: ServerStatus = None
+        self.server_context = ServerContext()
```
d1donlydfink, line 69: network_storage_dict is now created in the ServerContext constructor.

neuro_san/service/main_loop/server_main_loop.py
```diff
@@ -122,14 +118,15 @@ def parse_args(self):
         # See destination below ~ line 139, 154 for explanation.
         args = arg_parser.parse_args()
         self.server_name = args.server_name
-        self.server_status = ServerStatus(self.server_name)
+        server_status = ServerStatus(self.server_name)
+        self.server_context.set_server_status(server_status)
```
d1donlydfink, line 122: ServerStatus object needs to be created later and set on the new ServerContext object because of when the name comes in.

neuro_san/service/utils/server_context.py
```diff
@@ -0,0 +1,60 @@
...
+from leaf_common.asyncio.asyncio_executor_pool import AsyncioExecutorPool
+
+from neuro_san.internals.network_providers.agent_network_storage import AgentNetworkStorage
+from neuro_san.service.utils.server_status import ServerStatus
+
+
+class ServerContext:
+    """
+    Class that contains global-ish state for each instance of a server.
+    """
```
d1donlydfink, line 24: Single object that contains server-wide global-ish state.
We do this instead of use singletons, which increases testability.

neuro_san/service/main_loop/server_main_loop.py
```diff
@@ -28,16 +28,16 @@
...
 from neuro_san.service.grpc.grpc_agent_server import DEFAULT_SERVER_NAME
 from neuro_san.service.grpc.grpc_agent_server import DEFAULT_SERVER_NAME_FOR_LOGS
 from neuro_san.service.grpc.grpc_agent_server import DEFAULT_MAX_CONCURRENT_REQUESTS
 from neuro_san.service.grpc.grpc_agent_server import DEFAULT_REQUEST_LIMIT
 from neuro_san.service.grpc.grpc_agent_server import GrpcAgentServer
 from neuro_san.service.grpc.grpc_agent_service import GrpcAgentService
 from neuro_san.service.http.server.http_server import HttpServer
-from neuro_san.service.main_loop.server_status import ServerStatus
 from neuro_san.service.watcher.main_loop.storage_watcher import StorageWatcher
+from neuro_san.service.utils.server_status import ServerStatus
```
d1donlydfink, line 39: ServerStatus (and ServiceStatus too) moves to service.utils to avoid a tangle

neuro_san/service/main_loop/server_main_loop.py
```diff
@@ -178,58 +175,56 @@ def main_loop(self):
         metadata_set = metadata_set | set(self.usage_logger_metadata.split())
         metadata_str: str = " ".join(sorted(metadata_set))
 
-        AsyncioExecutorPoolProvider.set_executors_pool(reuse_mode=True)
+        server_status: ServerStatus = self.server_context.get_server_status()
```
d1donlydfink, line 178: No longer need the PoolProvider, but we do need to futz with the ServerStatus from here on down, so get that guy from the ServerContext.

neuro_san/service/watcher/main_loop/storage_watcher.py
```diff
@@ -33,32 +32,29 @@ class StorageWatcher(Startable):
     and other changes to AgentNetworkStorage instances.
     """
     def __init__(self,
-                 network_storage_dict: Dict[str, AgentNetworkStorage],
                  manifest_path: str,
                  manifest_update_period_in_seconds: int,
-                 server_status: ServerStatus):
+                 server_context: ServerContext):
```
d1donlydfink, line 37: Plumb the ServerContext to the StorageWatcher as well so the other 2 deleted args can come from there.

andreidenissov-cog (approved):
Looks good.

PR #337, opened 2025-07-29
UN-3376 For langchain tools, make sure the UI knows that the communication between those nodes has been completed

Noravee:
### Summary

This PR introduces a proof of concept for logging (journaling) output from a LangChain tool using a callback-based approach.

### Problem

For coded tools, journaling is handled within the activation class. However, LangChain tools are subclasses of LangChain’s `BaseTool`, and do not invoke our custom implementation (`LangchainOpenaiFunctionTool`) that creates an activation class and handles logging. As a result:

- No journal is created for LangChain tools.

- There is no clear indication when such a tool finishes execution.

### Solution

Leverage `JournalingCallbackHandler`, which defines `on_tool_start()` and `on_tool_end()`—methods that are triggered at the start and end of tool execution.

Previously, callbacks were passed as arguments when instantiating the LLM, which limited their use to LLM-related events (e.g., `on_llm_end()`). By instead using the **invoke config**, callbacks can now respond to a broader set of events, including tool-level events like `on_tool_start()` and `on_tool_end()`.

### Changes

- `ToolboxFactory.create_tool_from_toolbox()`
    
    - Added `agent_name` argument to override or prefix the LangChain tool name. This supports naming for single-tool agents and for agents with toolkits.

    - Added `"langchain_tool"` tag to identify tools that should trigger journaling.

- `JournalingCallbackHandler`
    
    - `__init__`:

        - Renamed `journal` to `calling_agent_journal`.

        - Added new parameters: `base_journal`, `parent_origin`, and `origination`.

    - `on_tool_start`:

        - Checks for the `"langchain_tool"` tag.

        - Creates a `langchain_tool_journal` and logs the tool’s input.

        - Prevents double journaling for coded or branch tools.

    - `on_tool_end`:

        - Logs the tool's output to both the `calling_agent_journal` and the `langchain_tool_journal`.

- `LangChainRunContext`
    - Moved callback assignment from `create_resources()` to `wait_from_run()` to ensure proper callback usage for tools as well as LLMs.

- 'TestToolboxFactory`: Added attribute `name` and `tags` to mock tools.

### Tests

Tested by running the following agents and inspecting their corresponding thinking files to verify tool-level logging:

- `bing_search`

- `music_nerd_pro`

- `music_nerd_pro_multi_agent`

Confirmed that:

- Tool input is logged at the start (`on_tool_start`)

- Tool output is logged at the end (`on_tool_end`)

- Journals are correctly scoped and do not duplicate entries for coded or branch tools

### Future Development

This callback-based approach could be extended to handle journaling for coded tools, branch tools, and other tool types—providing a unified and consistent logging mechanism across tool types.

### DIFF
--- a/neuro_san/internals/run_context/langchain/core/langchain_run_context.py
+++ b/neuro_san/internals/run_context/langchain/core/langchain_run_context.py
@@ -313,7 +313,7 @@ class LangChainRunContext(RunContext):
             if toolbox:
                 toolbox_factory: ContextTypeToolboxFactory = self.invocation_context.get_toolbox_factory()
                 try:
-                    tool_from_toolbox = toolbox_factory.create_tool_from_toolbox(toolbox, agent_spec.get("args"))
+                    tool_from_toolbox = toolbox_factory.create_tool_from_toolbox(toolbox, agent_spec.get("args"), name)
                     # If the tool from toolbox is base tool or list of base tool, return the tool as is
                     # since tool's definition and args schema are predefined in these the class of the tool.
                     if isinstance(tool_from_toolbox, BaseTool) or (
@@ -441,9 +441,12 @@ class LangChainRunContext(RunContext):
             "input": self.recent_human_message
         }
 
-        # Create the list of callbacks to pass to the LLM ChatModel
+        # Create the list of callbacks to pass when invoking
+        parent_origin: List[Dict[str, Any]] = self.get_origin()
+        base_journal: Journal = self.invocation_context.get_journal()
+        origination: Origination = self.invocation_context.get_origination()
         callbacks: List[BaseCallbackHandler] = [
-            JournalingCallbackHandler(self, self.journal)
+            JournalingCallbackHandler(base_journal, self.journal, parent_origin, origination)
         ]
         # Consult the agent spec for level of verbosity as it pertains to callbacks.
         agent_spec: Dict[str, Any] = self.tool_caller.get_agent_tool_spec()
@@ -452,6 +455,8 @@ class LangChainRunContext(RunContext):
             # This particular class adds a *lot* of very detailed messages
             # to the logs.  Add this because some people are interested in it.
             callbacks.append(LoggingCallbackHandler(self.logger))
+
+        # Add callbacks as an invoke config
         invoke_config = {
             "configurable": {
                 "session_id": run.get_id()
--- a/neuro_san/internals/run_context/langchain/core/langchain_run_context.py
+++ b/neuro_san/internals/run_context/langchain/core/langchain_run_context.py
@@ -168,19 +168,8 @@ class LangChainRunContext(RunContext):
         # DEF - Remove the arg if possible
         _ = agent_name
 
-        # Create the list of callbacks to pass to the LLM ChatModel
-        callbacks: List[BaseCallbackHandler] = [
-            JournalingCallbackHandler(self.journal)
-        ]
         full_name: str = Origination.get_full_name_from_origin(self.origin)
-
-        # Consult the agent spec for level of verbosity as it pertains to callbacks.
         agent_spec: Dict[str, Any] = self.tool_caller.get_agent_tool_spec()
-        verbose: Union[bool, str] = agent_spec.get("verbose", False)
-        if isinstance(verbose, str) and verbose.lower() in ("extra", "logging"):
-            # This particular class adds a *lot* of very detailed messages
-            # to the logs.  Add this because some people are interested in it.
-            callbacks.append(LoggingCallbackHandler(self.logger))
 
         # Now that we have a name, we can create an ErrorDetector for the output.
         self.error_detector = ErrorDetector(full_name,
@@ -199,10 +188,10 @@ class LangChainRunContext(RunContext):
 
         prompt_template: ChatPromptTemplate = await self._create_prompt_template(instructions)
 
-        self.agent = self.create_agent_with_fallbacks(prompt_template, callbacks)
+        self.agent = self.create_agent_with_fallbacks(prompt_template)
 
     def create_agent_with_fallbacks(self, prompt_template: ChatPromptTemplate,
-                                    callbacks: List[BaseCallbackHandler]) -> Agent:
+                                    callbacks: List[BaseCallbackHandler] = None) -> Agent:
         """
         Creates an agent with potential fallback llms to use.
         :param prompt_template: The ChatPromptTemplate to use for the agent
@@ -451,10 +440,23 @@ class LangChainRunContext(RunContext):
             "chat_history": previous_chat_history,
             "input": self.recent_human_message
         }
+
+        # Create the list of callbacks to pass to the LLM ChatModel
+        callbacks: List[BaseCallbackHandler] = [
+            JournalingCallbackHandler(self, self.journal)
+        ]
+        # Consult the agent spec for level of verbosity as it pertains to callbacks.
+        agent_spec: Dict[str, Any] = self.tool_caller.get_agent_tool_spec()
+        verbose: Union[bool, str] = agent_spec.get("verbose", False)
+        if isinstance(verbose, str) and verbose.lower() in ("extra", "logging"):
+            # This particular class adds a *lot* of very detailed messages
+            # to the logs.  Add this because some people are interested in it.
+            callbacks.append(LoggingCallbackHandler(self.logger))
         invoke_config = {
             "configurable": {
                 "session_id": run.get_id()
-            }
+            },
+            "callbacks": callbacks
         }
 
         # Chat history is updated in write_message
--- a/neuro_san/internals/run_context/langchain/core/langchain_run_context.py
+++ b/neuro_san/internals/run_context/langchain/core/langchain_run_context.py
@@ -190,12 +190,10 @@ class LangChainRunContext(RunContext):
 
         self.agent = self.create_agent_with_fallbacks(prompt_template)
 
-    def create_agent_with_fallbacks(self, prompt_template: ChatPromptTemplate,
-                                    callbacks: List[BaseCallbackHandler] = None) -> Agent:
+    def create_agent_with_fallbacks(self, prompt_template: ChatPromptTemplate) -> Agent:
         """
         Creates an agent with potential fallback llms to use.
         :param prompt_template: The ChatPromptTemplate to use for the agent
-        :param callbacks: The list of callbacks to use when creating any LLM via the factory
         :return: An Agent (Runnable)
         """
         # Initialize our return value
@@ -215,7 +213,7 @@ class LangChainRunContext(RunContext):
         for index, fallback in enumerate(fallbacks):
 
             # Create a model we might use.
-            one_llm: BaseLanguageModel = llm_factory.create_llm(fallback, callbacks=callbacks)
+            one_llm: BaseLanguageModel = llm_factory.create_llm(fallback)
             one_agent: Agent = self.create_agent(prompt_template, one_llm)
 
             if index == 0:
--- a/neuro_san/internals/run_context/langchain/core/langchain_run_context.py
+++ b/neuro_san/internals/run_context/langchain/core/langchain_run_context.py
@@ -446,7 +446,7 @@ class LangChainRunContext(RunContext):
         base_journal: Journal = self.invocation_context.get_journal()
         origination: Origination = self.invocation_context.get_origination()
         callbacks: List[BaseCallbackHandler] = [
-            JournalingCallbackHandler(base_journal, self.journal, parent_origin, origination)
+            JournalingCallbackHandler(self.journal, base_journal, parent_origin, origination)
         ]
         # Consult the agent spec for level of verbosity as it pertains to callbacks.
         agent_spec: Dict[str, Any] = self.tool_caller.get_agent_tool_spec()
--- a/neuro_san/internals/run_context/langchain/journaling/journaling_callback_handler.py
+++ b/neuro_san/internals/run_context/langchain/journaling/journaling_callback_handler.py
@@ -22,8 +22,11 @@ from langchain_core.documents import Document
 from langchain_core.outputs import LLMResult
 from langchain_core.outputs.chat_generation import ChatGeneration
 
+from neuro_san.internals.journals.journal import Journal
 from neuro_san.internals.journals.originating_journal import OriginatingJournal
+from neuro_san.internals.messages.origination import Origination
 from neuro_san.internals.messages.agent_message import AgentMessage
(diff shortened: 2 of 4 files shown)


neuro_san/internals/run_context/langchain/toolbox/toolbox_factory.py
```diff
@@ -168,8 +171,18 @@ def create_tool_from_toolbox(
...
+                # Prefix the name of the agent to each tool
+                tool.name = f"{agent_name}_{tool.name}"
+                # Add "langchain_tool" tags so journal callback can idenitify it
+                tool.tags = ["langchain_tool"]
+            return toolkit
+
+        # Replace langchain tool's name with agent name
+        instance.name = agent_name
+        # Add "langchain_tool" tags so journal callback can idenitify it
+        instance.tags = ["langchain_tool"]
```
Noravee, line 192: Override langchain tool's name with `agent_anme` and add `langchain_tool` to `tags` as identifier.

neuro_san/internals/run_context/langchain/toolbox/toolbox_factory.py
```diff
@@ -168,8 +171,18 @@ def create_tool_from_toolbox(
 
         # If the instantiated class has "get_tools()", assume it's a toolkit and return a list of tools
         if hasattr(instance, "get_tools") and callable(instance.get_tools):
-            return instance.get_tools()
-
+            toolkit: List[BaseTool] = instance.get_tools()
+            for tool in toolkit:
+                # Prefix the name of the agent to each tool
+                tool.name = f"{agent_name}_{tool.name}"
```
Noravee: For agent with multiple tools (toolkit), prefix the tool names with the agent name.

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
```diff
@@ -451,10 +440,28 @@ async def wait_on_run(self, run: Run, journal: Journal = None) -> Run:
...
+            # to the logs.  Add this because some people are interested in it.
+            callbacks.append(LoggingCallbackHandler(self.logger))
+
+        # Add callbacks as an invoke config
         invoke_config = {
             "configurable": {
                 "session_id": run.get_id()
-            }
+            },
+            "callbacks": callbacks
```
Noravee, line 462: Use callbacks as `invoke_config` in `wait_on_run()` to have access to `on_tool_start()` and `on_tool_run()`

neuro_san/internals/run_context/langchain/journaling/journaling_callback_handler.py
```diff
@@ -76,17 +97,73 @@ async def on_llm_end(self, response: LLMResult,
...
+        origin metadata, and full tool name.
+
+        :param serialized: Serialized representation of the tool, including its name and description.
+        :param input_str: String representation of the tool's input.
+        :param tags: List of tags associated with the tool. Used to determine whether it is a LangChain tool.
+        :param inputs: Structured dictionary of input arguments
+            passed to the tool.
+        """
+
+        if "langchain_tool" in tags:
```
Noravee, line 139: Check if the tool is langchain's tool with `tags`.

neuro_san/internals/run_context/langchain/journaling/journaling_callback_handler.py
```diff
@@ -76,17 +97,73 @@ async def on_llm_end(self, response: LLMResult,
...
+
+            # Combine the original tool inputs with origin metadata
+            combined_args: Dict[str, Any] = inputs.copy()
+            combined_args["origin"] = origin
+            combined_args["origin_str"] = full_name
+
+            # Create a journal entry for this invocation and log the combined inputs
+            self.langchain_tool_journal = OriginatingJournal(self.base_journal, origin)
+            message: BaseMessage = AgentMessage(content=f"Received arguments {combined_args}")
+            await self.langchain_tool_journal.write_message(message)
```
Noravee, line 156: Creating journal for langchain tool and write message.

neuro_san/internals/run_context/langchain/journaling/journaling_callback_handler.py
```diff
@@ -76,17 +97,73 @@ async def on_llm_end(self, response: LLMResult,
...
+        :param tags: List of tags associated with the tool. Used to determine whether it is a LangChain tool.
+        """
+
+        if "langchain_tool" in tags:
+            # Log the tool output to the calling agent's journal
+            await self.calling_agent_journal.write_message(AIMessage(content=output))
+
+            # Also log the tool output to the LangChain tool-specific journal
+            message: BaseMessage = AgentMessage(content=f"Got result: {output}")
+            await self.langchain_tool_journal.write_message(message)
```
Noravee, line 176: Write message in the langchain journal with tool output. This can be used as an indicator that the tool is finished.

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
```diff
@@ -168,19 +168,8 @@ async def create_resources(self, agent_name: str,
...
-        ]
         full_name: str = Origination.get_full_name_from_origin(self.origin)
-
-        # Consult the agent spec for level of verbosity as it pertains to callbacks.
         agent_spec: Dict[str, Any] = self.tool_caller.get_agent_tool_spec()
-        verbose: Union[bool, str] = agent_spec.get("verbose", False)
-        if isinstance(verbose, str) and verbose.lower() in ("extra", "logging"):
-            # This particular class adds a *lot* of very detailed messages
-            # to the logs.  Add this because some people are interested in it.
-            callbacks.append(LoggingCallbackHandler(self.logger))
```
d1donlydfink, line 183: Looks like all this moves down into create_agent_with_fallbacks()

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
```diff
@@ -199,10 +188,10 @@ async def create_resources(self, agent_name: str,
 
         prompt_template: ChatPromptTemplate = await self._create_prompt_template(instructions)
 
-        self.agent = self.create_agent_with_fallbacks(prompt_template, callbacks)
+        self.agent = self.create_agent_with_fallbacks(prompt_template)
 
     def create_agent_with_fallbacks(self, prompt_template: ChatPromptTemplate,
-                                    callbacks: List[BaseCallbackHandler]) -> Agent:
+                                    callbacks: List[BaseCallbackHandler] = None) -> Agent:
```
d1donlydfink: Does anyone still call this method with callbacks not None?
If not, let's get rid of the callbacks arg.

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
```diff
@@ -324,7 +313,7 @@ async def _create_base_tool(self, name: str) -> BaseTool:
             if toolbox:
                 toolbox_factory: ContextTypeToolboxFactory = self.invocation_context.get_toolbox_factory()
                 try:
-                    tool_from_toolbox = toolbox_factory.create_tool_from_toolbox(toolbox, agent_spec.get("args"))
+                    tool_from_toolbox = toolbox_factory.create_tool_from_toolbox(toolbox, agent_spec.get("args"), name)
```
d1donlydfink, line 314: Where does "name" come from here?

neuro_san/internals/run_context/langchain/journaling/journaling_callback_handler.py
```diff
@@ -52,13 +57,29 @@ class JournalingCallbackHandler(AsyncCallbackHandler):
...
+        :param base_journal: The Journal instance that allows message reporting during the course of the AgentSession.
+            This is used to construct the langchain_tool_journal.
+        :param parent_origin: A List of origin dictionaries indicating the origin of the run
+            This is used to construct the langchain_tool_journal.
+        :param origination: The Origination instance carrying state about tool instantation
+            during the course of the AgentSession. This is used to construct the langchain_tool_journal.
         """
-        self.journal: OriginatingJournal = journal
+        self.calling_agent_journal: Journal = calling_agent_journal
+        self.base_journal: Journal = base_journal
```
d1donlydfink, line 89: Why do we need two journals?
Shouldn't all the messages come from a single journal w/ a single origin?

d1donlydfink (changes requested):
Please answer the questions sprinkled throughout the PR.

dsargent (dismissed):
Don't really know enough to be useful here. I will defer to @d1donlydfink's questions.

neuro_san/internals/run_context/langchain/journaling/journaling_callback_handler.py
Noravee: The calling-agent journal logs the execution flow from the perspective of the agent invoking the tool (e.g., `MusicNerdPro`). In contrast, the LangChain tool journal represents the tool's own execution context—similar to how coded tools like `Accountant` have their own journal tied to their run context.

LangChain tools don’t instantiate their own `RunContext`, so they lack a dedicated journal by default. To maintain consistency with how other tools are tracked, we explicitly create a `langchain_tool_journal` when the tool starts. This ensures tool-specific inputs and outputs are captured independently, while still allowing the calling agent to log its own perspective.

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
Noravee: Thank you for your suggestion. Callbacks arg has been removed.

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
Noravee: Yes. `callbacks` used to be an argument for `create_agent_with_fallbacks()` but the argument is now removed, and the `callbacks` is passed as `invoke_config` in `wait_on_run()` instead.

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
Noravee: The argument `name` of `_create_base_tool()` can be traced back to a tool in a list of `tools` for tool-calling agent in the hocon file.

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
```diff
@@ -199,14 +188,12 @@ async def create_resources(self, agent_name: str,
 
         prompt_template: ChatPromptTemplate = await self._create_prompt_template(instructions)
 
-        self.agent = self.create_agent_with_fallbacks(prompt_template, callbacks)
+        self.agent = self.create_agent_with_fallbacks(prompt_template)
 
-    def create_agent_with_fallbacks(self, prompt_template: ChatPromptTemplate,
-                                    callbacks: List[BaseCallbackHandler]) -> Agent:
+    def create_agent_with_fallbacks(self, prompt_template: ChatPromptTemplate) -> Agent:
```
Noravee, line 193: Remove callbacks from the argument list of `create_agent_with_fallbacks`. This eliminates the need for 
a `callbacks` parameter in both `default_llm_factory` and `standard_langchain_llm_factory`. 
However, that cleanup is outside the scope of this PR and will be addressed separately.

neuro_san/internals/run_context/langchain/journaling/journaling_callback_handler.py
swensel: @Noravee Since the UI already checks for `"Got result:"` for coded tool end events on the `text` field, this should work fine.

We also may want to consider adding a way to signify an end event for coded tools or langchain tools on the `structure` key instead, so that we don't have to do string parsing on the `text` field.

cc: @d1donlydfink

neuro_san/internals/run_context/langchain/journaling/journaling_callback_handler.py
d1donlydfink: Ah. That makes sense.
You should add this explanation you just gave me in the PR to the code so our future selves do not have to have this conversation again.
Where we had this conversation is as good a place as any.

neuro_san/internals/run_context/langchain/toolbox/toolbox_factory.py
```diff
@@ -116,13 +116,16 @@ def load(self):
     def create_tool_from_toolbox(
             self,
             tool_name: str,
-            user_args: Dict[str, Any] = None
+            user_args: Dict[str, Any] = None,
+            agent_name: str = None
```
d1donlydfink, line 120: Does it ever make sense to not pass an agent_name here?  (I think maybe no)

If you agree, then perhaps the agent_name should come after the tool_name as required but before the user_args as optional.  A quick grep shows that the only other callers are in tests, and it should be reasonable to update those.

If you don't agree, then it's worth adding a comment to the agent_name docs for the arg below as to what happens when agent_name is the None default.

neuro_san/internals/run_context/langchain/toolbox/toolbox_factory.py
d1donlydfink: Worth noting for above argument that you are not really checking for a None agent_name value,
so that kinda bolsters my argument as to its necessity of not being None.

d1donlydfink (changes requested):
* Please add nice explanation of need for 2 Journals you added in the PR to the code itself
* Make agent_name a required argument of create_tool_from_toolbox(()

neuro_san/internals/run_context/langchain/journaling/journaling_callback_handler.py
Noravee: Comment added

neuro_san/internals/run_context/langchain/toolbox/toolbox_factory.py
Noravee: Thanks for the thoughtful input. I left `agent_name` as optional because, while it isn’t currently needed, I see a potential future use case where tools from the toolbox could be used as external tools, in which case an agent name wouldn’t be necessary.

To clarify, I’ve added the following comment in the code:

> `agent_name` is required when the tool is used as an internal agent. However, tools from the toolbox could potentially be used as external tools, in which case agent_name may not be needed.

Happy to revisit this and make `agent_name` required (and update the tests accordingly) if you'd prefer a stricter default for now.

neuro_san/internals/run_context/langchain/journaling/journaling_callback_handler.py
d1donlydfink: Can do that in a separate PR @swensel for full consistency

swensel: Thanks @d1donlydfink , that would be great.

cc: @Noravee

PR #341, opened 2025-07-30
UN-3377 Add debug check for missing request metadata.

andreidenissov-cog:
This PR adds temporary debug logging for incoming requests with missing metadata.
We are hoping to find out where exactly a problem originates - on the client side
or maybe in the server internals.
Debug mode is controlled by SHOW_ABSENT_METADATA env variable which should be set to enable it.


neuro_san/service/http/handlers/base_request_handler.py
```diff
@@ -63,6 +63,7 @@ def initialize(self,
         self.openapi_service_spec_path: str = openapi_service_spec_path
         self.logger = HttpLogger(forwarded_request_metadata)
         self.network_storage_dict: Dict[str, AgentNetworkStorage] = network_storage_dict
+        self.show_absent: bool = os.environ.get("SHOW_ABSENT_METADATA") is not None
```
andreidenissov-cog, line 67: Set the debug mode from env variable.

d1donlydfink (approved):
Clever way to see where the fault lies.

PR #348, opened 2025-08-01
Add error key when model name is invalid for gemini

Noravee:


neuro_san/internals/run_context/langchain/util/api_key_error_check.py
```diff
@@ -16,7 +16,8 @@
 API_KEY_EXCEPTIONS: Dict[str, List] = {
     "OPENAI_API_KEY": ["OPENAI_API_KEY", "Incorrect API key provided"],
     "ANTHROPIC_API_KEY": ["ANTHROPIC_API_KEY", "anthropic_api_key", "invalid x-api-key", "credit balance"],
-    "GOOGLE_API_KEY": ["Application Default Credentials", "default credentials", "Gemini: 400 API key not valid"],
+    "GOOGLE_API_KEY": ["Application Default Credentials", "default credentials",
+                       "Gemini: 400 API key not valid", "not found for API version v1beta"],
```
Noravee, line 20: "not found for API version v1beta" is an error message when the gemini model name is invalid.

d1donlydfink: An invalid model name is not really the point of the common help text in this class.
When you read the help text below, it's all about dumb mistakes with API keys and firewalls, not invalid model names.

It can be a noble thing to tackle the invalid model name in a consistent manner across LLM providers and common help text.  That doesn't mean it belongs in this class.

d1donlydfink (changes requested):
It can be a noble thing to tackle the invalid model name in a consistent manner across LLM providers and common help text. That doesn't mean it belongs in this class.

PR #351, opened 2025-08-04
UN-3419 No indicator of which tool or output is from in thinking_file

Noravee:
###  Problem

There was no clear indicator in the **thinking file** of tool-calling agents to distinguish whether an output came from a **coded tool** or a **LangChain tool**. This made it difficult to debug or trace the origin of tool-related messages in the journal.

#### Coded Tools

- Tool messages were previously converted to `AIMessage` to avoid issues with Anthropic models.

- Updated `OriginatingJournal.write_message()` to:

    - Still store a valid `AIMessage` in the chat history for compatibility.

    - But retain the original `AgentToolResultMessage` separately for journal output, preserving source fidelity.

#### LangChain Tools

- `AIMessage` was written in place of an `AgentToolResultMessage` in the tool-calling agent journal.

- Fixed by updating `JournalingCallbackHandler.on_tool_end()` to write the correct `AgentToolResultMessage`.

### Tests

Manually verified by running:

- `music_nerd_pro`

- `music_nerd_pro_llm_anthropic`

- `bing_search`

Then inspected thinking file outputs to confirm correct message typing and provenance labeling.


neuro_san/internals/journals/originating_journal.py
```diff
@@ -82,15 +82,17 @@ async def write_message(self, message: BaseMessage, origin: List[Dict[str, Any]]
             #   to include additional metadata like `tool_result_origin` when supported.
             #
             # To avoid problem with any other LLMs, convert "AgentToolResultMessage" to "AIMessage"
-            # when appending it chathistory but allow it to be written in the journal as is to
+            # when appending it chat history but allow it to be written in the journal as is to
             # to maintain the information on tool origin.
             if isinstance(message, AgentToolResultMessage):
-                message = AIMessage(content=message.content)
+                chat_history_message: BaseMessage = AIMessage(content=message.content)
```
Noravee, line 88: Coded tool: do not change `message` to `AIMessage` here. Instead create a new variable to add into chat history.

neuro_san/internals/run_context/langchain/journaling/journaling_callback_handler.py
```diff
@@ -173,7 +174,9 @@ async def on_tool_end(self, output: Any, tags: List[str] = None, **kwargs: Any)
 
         if "langchain_tool" in tags:
             # Log the tool output to the calling agent's journal
-            await self.calling_agent_journal.write_message(AIMessage(content=output))
+            await self.calling_agent_journal.write_message(
+                AgentToolResultMessage(content=output, tool_result_origin=self.origin)
```
Noravee, line 178: Langchain tool: use `AgentToolResultMessage` instead of `AIMessage` for the calling agent journal.

neuro_san/internals/journals/originating_journal.py
d1donlydfink: Theoretically the AgentToolResultMessage should be redundant w/rt chat history.
But that optimization doesn't have to be for now.

neuro_san/internals/run_context/langchain/journaling/journaling_callback_handler.py
```diff
@@ -142,16 +143,16 @@ async def on_tool_start(
             agent_name: str = serialized.get("name")
 
             # Build the origin path
-            origin: List[Dict[str, Any]] = self.origination.add_spec_name_to_origin(self.parent_origin, agent_name)
-            full_name: str = self.origination.get_full_name_from_origin(origin)
+            self.origin: List[Dict[str, Any]] = self.origination.add_spec_name_to_origin(self.parent_origin, agent_name)
+            full_name: str = self.origination.get_full_name_from_origin(self.origin)
```
d1donlydfink, line 147: Interesting this ends up needing to be stateful. Not surprising tho.
That said, is this likely to get called multiple times if an agent has multiple tools?
I guess it doesn't matter either way, it'll get the same result.

neuro_san/internals/run_context/langchain/journaling/journaling_callback_handler.py
d1donlydfink: Yeah this makes sense.

My only outstanding questions are:

* Is the code here:
https://github.com/cognizant-ai-lab/neuro-san/blob/main/neuro_san/internals/run_context/langchain/core/langchain_run_context.py#L620
... ever getting invoked any more?

* If not, what other things happening in that code path are also getting missed? (That's my bigger worry)
* If that is all good: What code could we then safely remove?

Noravee: The code still get invoked for coded tools. The callback only handle langchain tools.

d1donlydfink (approved):

PR #353, opened 2025-08-05
UN-3406 Multiple frontman error when branch also has no parameters

Noravee:
### Change

- Change logic of `AgentNetwork.find_front_man()` so that the first agent that is not coded tool or toolbox is the front man.

### Test

- Ran `music_nerd_pro_multi_agents` and confirmed correct front man selection.

### DIFF
--- a/neuro_san/internals/graph/registry/agent_network.py
+++ b/neuro_san/internals/graph/registry/agent_network.py
@@ -128,10 +128,8 @@ No front man found for the {self.name} agent network.
 
 The front man is the first agent listed under the "tools" section of your agent HOCON file.
 However, the front man must not be:
-
-- A CodedTool (i.e., an agent defined with a "class" field)
-
-- A toolbox agent (i.e., defined with a "toolbox" field)
+* A CodedTool (i.e., an agent defined with a "class" field)
+* A toolbox agent (i.e., defined with a "toolbox" field)
 """)
 
         # The front-man is the first agent that is not coded tool ("class")
--- a/neuro_san/internals/graph/registry/agent_network.py
+++ b/neuro_san/internals/graph/registry/agent_network.py
@@ -104,21 +104,9 @@ Some things to try:
                  This guy will be user facing.  If there are none or > 1,
                  an exception will be raised.
         """
-        front_men: List[str] = []
 
-        # Identify the "front-man" agent.
-        # Primary heuristic: an agent with defined instructions and a function that takes no parameters.
-        # The presence of instructions ensures it was explicitly defined, since users may add parameters
-        # to front-men, making function signature alone unreliable.
-        for name, agent_spec in self.agent_spec_map.items():
-            instructions: str = agent_spec.get("instructions")
-            function: Dict[str, Any] = agent_spec.get("function")
-            if instructions is not None and function is not None and function.get("parameters") is None:
-                front_men.append(name)
-
-        if len(front_men) == 0:
-            # The best way to find a front man is to see which agent was registered first
-            front_men.append(self.first_agent)
+        # Initially all agents are front-man candidates.
+        front_men: List[str] = list(self.agent_spec_map.keys())
 
         # Check for validity of our front-man candidates.
         valid_front_men: List[str] = copy(front_men)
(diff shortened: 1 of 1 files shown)


neuro_san/internals/graph/registry/agent_network.py
```diff
@@ -104,21 +104,9 @@ def find_front_man(self) -> str:
...
-        front_men: List[str] = []
-
-        # Identify the "front-man" agent.
-        # Primary heuristic: an agent with defined instructions and a function that takes no parameters.
-        # The presence of instructions ensures it was explicitly defined, since users may add parameters
-        # to front-men, making function signature alone unreliable.
-        for name, agent_spec in self.agent_spec_map.items():
-            instructions: str = agent_spec.get("instructions")
-            function: Dict[str, Any] = agent_spec.get("function")
-            if instructions is not None and function is not None and function.get("parameters") is None:
```
Noravee, line 116: If the front man is simply defined as the first agent that is not a coded tool or toolbox, this logic may no longer be necessary.
Please let me know your thoughts — or if you'd prefer to keep additional criteria for selecting the front man.

neuro_san/internals/graph/registry/agent_network.py
```diff
@@ -104,21 +104,9 @@ def find_front_man(self) -> str:
...
-                front_men.append(name)
-
-        if len(front_men) == 0:
-            # The best way to find a front man is to see which agent was registered first
-            front_men.append(self.first_agent)
+
+        # Initially all agents are front-man candidates.
+        front_men: List[str] = list(self.agent_spec_map.keys())
```
kxk302: I think this variable name is misleading a bit. We only have one frontman that is the first listed agent. How about we call this variable 'agent_list' or something and simply pick 'agent_list[0]' as front man? Can add validation to ensure the first agent does not have CodedTools or uses toolbox (and any other restriction)

neuro_san/internals/graph/registry/agent_network.py
```diff
@@ -137,19 +125,15 @@ def find_front_man(self) -> str:
...
+However, the front man must not be:
+* A CodedTool (i.e., an agent defined with a "class" field)
+* A toolbox agent (i.e., defined with a "toolbox" field)
 """)
 
-        if len(valid_front_men) > 1:
-            raise ValueError(f"Found > 1 front man for chat for the {self.name} network." +
-                             f" Possibilities: {valid_front_men}")
-
+        # The front-man is the first agent that is not coded tool ("class")
```
kxk302, line 135: I think this comment is outdated. The frontman is the first agent. However, frontman is not supposed to use CodedTools or toolbox tool.

d1donlydfink: I think the constraint in the comments is reasonable.

d1donlydfink (approved):

neuro_san/internals/graph/registry/agent_network.py
Noravee: The variable name is changed to `agent_list`. Then the list is filtered where coded tools and toolbox were removed. Finally, the first one of the `valid_front_men` or `valid_front_men[0]` is the front man.

neuro_san/internals/graph/registry/agent_network.py
Noravee: @kxk302 In NeuroSan, the terms `agent` and `tool` are somewhat interchangeable, since agents are activated through tool-calling, and tool nodes are treated as agents.

For example:

- The `name` field for a tool node is used as both the agent name and the tool name (in the case of coded tools).

- All nodes—whether LLMs or tools—are listed under the `tools` section in the HOCON config.

Given this structure, I believe the comment “The front man is the first agent that is not a coded tool (`class`) or a toolbox tool (`toolbox`)” is still accurate and appropriate.

d1donlydfink: I agree.

FWIW: I'd like to attempt to start clarifying the terminology (if it's possible) so we are very specific about when/where we use the term "tool" vs "agent".

I think "tool" belongs more to the langchain level - something that gets invoked by an "agent".
I think "agent" belongs for the most part at our library level and there has recently been a renaming-fest to get "tool" out of class names where it is not quite the correct usage. 

There are very specific situations where "agent" and "tool" actually do mean the same thing in our code base,
but let's try to keep that footprint small.

kxk302:
We need to discuss the logic for selecting frontman, maybe in a standup parking lot, as I think its a bit different that what we discussed the last time.

neuro_san/internals/graph/registry/agent_network.py
ofrancon: I prefer your definition. We should also update our comments in our example .hocon files, here and in neuro-san-studio. For other PRs.

ofrancon:
For reference, what I mean by "we should update our .hocon files in future PRs", is that most of our files contain this:
```
    "tools": [
        # These tool definitions do not have to be in any particular order
        # How they are linked and call each other is defined within their
        # own specs.  This could be a graph, potentially even with cycles.

        # This first agent definition is regarded as the "Front Man", which
        # does all the talking to the outside world/client.
        # It is identified as such because it is either:
        #   A) The only one with no parameters in his function definition,
        #      and therefore he needs to talk to the outside world to get things rolling.
        #   B) The first agent listed, regardless of function parameters.
        #
        # Some disqualifications from being a front man:
        #   1) Cannot use a CodedTool "class" definition
        #   2) Cannot use a Tool "toolbox" definition
        {
            "name": "MusicNerd",

            # Note that there are no parameters defined for this guy's "function" key.
            # This is the primary way to identify this tool as a front-man,
            # distinguishing it from the rest of the tools.

```
It can probably be replaced by:
```
    "tools": [
        # This list defines the agents in the agent network and their relationship.
        # How they are linked and call each other is defined within their
        # own specs.  This could be a graph, potentially even with cycles.
        # The front man is the first agent in this list that is not a coded tool (class) or a toolbox tool (toolbox)
```

PR #354, opened 2025-08-05
UN-3427 Updated model name on Anthropic test case

vince-leaf:
The new neuro-san-studio user's guide has been updated with a new model on the Anthropic model name example. Since Bedrock is using this newer model, we want this test case to use the same model to be consistent.

### DIFF
--- a/neuro_san/registries/music_nerd_pro_llm_anthropic.hocon
+++ b/neuro_san/registries/music_nerd_pro_llm_anthropic.hocon
@@ -15,7 +15,7 @@
 
 {
     "llm_config": {
-        "model_name": "claude-3-5-haiku",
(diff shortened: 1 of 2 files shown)


neuro_san/registries/music_nerd_pro_llm_anthropic.hocon
```diff
@@ -15,7 +15,7 @@
 
 {
     "llm_config": {
-        "model_name": "claude-3-5-haiku",
+        "model_name": "claude-3-7-sonnet",
```
vince-leaf, line 18: Updated to the new model name, as this is what Bedrock is set up to use.

Noravee (approved):
LGTM.

PR #356, opened 2025-08-07
UN-3407 No assignment for branch with no parameters

Noravee:


neuro_san/internals/graph/activations/argument_assigner.py
```diff
@@ -40,45 +31,49 @@ def assign(self, arguments: Dict[str, Any]) -> List[str]:
         """
         assignments: List[str] = []
 
-        # Start to build the list of assignments, with one sentence for each property
-        # listed (exception for name and description).
-        for one_property, attributes in self.properties.items():
+        # Start to build the list of assignments, with one sentence for each argument
+        for args_name, args_value in arguments.items():
```
Noravee, line 44: Use argument name and value from `arguments` directly.

neuro_san/internals/graph/activations/branch_activation.py
```diff
@@ -83,7 +73,7 @@ def get_command(self) -> str:
         :return: A string describing the objective of the component.
         """
         agent_spec = self.get_agent_tool_spec()
-        return agent_spec.get("command", "Perform your instructions to the best of your ability.")
```
Noravee, line 86: Remove default value for `command`.

neuro_san/internals/graph/activations/branch_activation.py
```diff
@@ -124,10 +114,11 @@ async def build(self) -> str:
...
         command = self.get_command()
-        # If there is assignments, combine it with command to be used as HumanMessage.
-        if assignments:
-            command = assignments + "\n" + command
-        run: Run = await self.run_context.submit_message(command)
+
+        # If there is command, combine it with assignment to be used as HumanMessage.
+        if command:
+            assignments = assignments + "\n" + command
+        run: Run = await self.run_context.submit_message(assignments)
```
Noravee, line 137: Since there may be no `command`, human message is mainly `assignments`.

neuro_san/internals/graph/activations/argument_assigner.py
```diff
@@ -22,15 +22,6 @@ class ArgumentAssigner:
...
     """
 
-    def __init__(self, properties: Dict[str, Any]):
-        """
-        Constructor
-
-        :param properties: The dictionary of function properties to fulfill,
-                as described in the callee agent spec.
-        """
-        self.properties: Dict[str, Any] = properties
```
d1donlydfink: Hmm. Interesting. This is a nice little optimization.
It strikes me than in doing this we are trusting the agent system more to send what is necessary via the args rather than rely on what is in the properties spec.  Maybe that's OK.

neuro_san/internals/graph/activations/argument_assigner.py
```diff
@@ -40,45 +31,49 @@ def assign(self, arguments: Dict[str, Any]) -> List[str]:
...
+            args_value_str: str = self.get_args_value_as_string(args_value)
 
             # No specific attribution text, so we make up a boilerplate
-            # one where it give the property/arg name <is/are> and the value.
+            # one where it give the arg name <is/are> and the value.
 
             # Figure out the attribution verb for singular vs plural
             assignment_verb: str = "is"
-            if attributes.get("type") == "array":
+            if isinstance(args_value, (dict, list)):
```
d1donlydfink: Probably more grammatically correct to stick with just a list and not a dictionary here.
A dict is singular.  Contents of a list is indeed plural.

neuro_san/internals/graph/activations/argument_assigner.py
```diff
@@ -40,45 +31,49 @@ def assign(self, arguments: Dict[str, Any]) -> List[str]:
...
 
-            args_value: Any = arguments.get(one_property)
-            if args_value is None:
-                # If we do not have an argument value for the property,
-                # do not add anything to the attribution list
-                continue
-
-            args_value_str: str = self.get_args_value_as_string(args_value,
-                                                                attributes.get("type"))
+            args_value_str: str = self.get_args_value_as_string(args_value)
```
d1donlydfink: Ah. Interesting.  This is why the properties were there in the first place - to get at the typing in the attributes.
Again, you are very much trusting the underlying agent system giving you the right thing rather than the what the agent spec says it wants.

Maybe there is a compromise here:
* Allow the constructor properties to be None
* If the properties are given , then go with the typing from the attributes from the spec
* If the properties are not given (None the default), then do what you are doing here and trust the args.

This allows the caller to pick who they are going to trust w/rt arguments and very explicitly so.
This will also allow us an easier retreat from this path if it is not working out for us by causing unanticipated trouble.

neuro_san/internals/graph/activations/argument_assigner.py
```diff
@@ -40,45 +31,49 @@ def assign(self, arguments: Dict[str, Any]) -> List[str]:
...
 
             # Figure out the attribution verb for singular vs plural
             assignment_verb: str = "is"
-            if attributes.get("type") == "array":
+            if isinstance(args_value, (dict, list)):
                 assignment_verb = "are"
 
             # Put together the assignment statement
-            assignment: str = f"The {one_property} {assignment_verb} {args_value_str}."
+            assignment: str = f"The {args_name} {assignment_verb} {args_value_str}."
```
d1donlydfink, line 69: Again, trusting the agent system over the spec.  Propose same strategy as above:
* Get from properties if properties were given in constructor, otherwise get from args.

neuro_san/internals/graph/activations/branch_activation.py
```diff
@@ -61,16 +59,8 @@ def get_assignments(self) -> str:
...
-        extractor: FieldExtractor = FieldExtractor()
-        empty: Dict[str, Any] = {}
-
-        agent_spec = self.get_agent_tool_spec()
 
-        # Properties describe the function arguments
-        properties: Dict[str, Any] = extractor.get_field(agent_spec, "function.parameters.properties", empty)
-
-        assigner = ArgumentAssigner(properties)
+        assigner = ArgumentAssigner()
```
d1donlydfink: With strategy above you could still get the properties from the spec and pass them as golden if they exist.
If they do not exist, then you can fall back to relying on the args.  This would mean most of what was removed above would return, except maybe the empty dictionary would go away.

neuro_san/internals/graph/activations/branch_activation.py
d1donlydfink: What happens when the return value is None?
Probably worth a comment.

neuro_san/internals/graph/activations/branch_activation.py
d1donlydfink: But it seems that there might be no assignments as well.  Best to allow for that here. It allows the rest of the stuff to change easier if we do that.

d1donlydfink (changes requested):
See recommendations in various places in PR.
I think this is tricky what you are doing here and I most definitely think you should test before/after cases on much much more than just music_nerd_pro. (This falls into the "know the consequences of your actions as fully as you can" category)

Please test a few more agents like:
* esp_decision_assistant (has rather complex argument passing)
* math_guy_passthrough (uses external agents and no commands)
* intranet_agents (uses AAOSA - which is the linchpin of babak's demos)

neuro_san/internals/graph/activations/branch_activation.py
Noravee: Added the following comment:

The command will be combined with assignments (arguments from an upstream agent) to direct the agent toward a specific task.
Optional: if omitted, the agent will choose an action based on the assignments and the list of available tools.

hshahrzad (approved):
Looks good to me.

neuro_san/internals/graph/activations/branch_activation.py
Noravee: `assignments` will never be `None` since it is generated from `"\n".join()`, but it may be an empty string. If both `assignments` and `command` are empty, the agent will typically respond by asking how it can help.

neuro_san/internals/graph/activations/argument_assigner.py
Noravee: Fixed.

neuro_san/internals/graph/activations/argument_assigner.py
Noravee: The code is modified to use `type` from `properties` first and fallback to argument value.

neuro_san/internals/graph/activations/argument_assigner.py
Noravee: The code is modified to use `type` from `properties` first and fallback to argument value.

neuro_san/internals/graph/activations/argument_assigner.py
Noravee: The code is modified to skip if the argument name from llm does not match with that of properties.

neuro_san/internals/graph/activations/branch_activation.py
Noravee: The code is modified so that the argument value type is taken from properties if possible and to skip if the argument name from llm does not match with that of properties.

Noravee:
@d1donlydfink 
Thank you for your comments and suggestions.

The code has been updated so that the argument value type is taken from properties when available, and arguments from the LLM are skipped if their names do not match any in properties.

The following networks were tested to ensure branch activations with parameters still work as expected:

- `esp_decision_assistant`

- `math_guy_passthrough`

- `intranet_agents`

Additionally, `music_nerd_pro_multi_agents` was tested to verify branch activation with no parameters.

Please let me know if you have any further suggestions or feedback.

neuro_san/internals/graph/activations/argument_assigner.py
```diff
@@ -40,33 +40,33 @@ def assign(self, arguments: Dict[str, Any]) -> List[str]:
...
+            args_value_str: str = self.get_args_value_as_string(args_value, args_value_type)
 
             # No specific attribution text, so we make up a boilerplate
-            # one where it give the property/arg name <is/are> and the value.
+            # one where it give the arg name <is/are> and the value.
 
             # Figure out the attribution verb for singular vs plural
             assignment_verb: str = "is"
-            if attributes.get("type") == "array":
+            if (args_value_type == "array") or isinstance(args_value, list):
```
d1donlydfink: Why the parentheses? Seems like pylint should have called that out.

neuro_san/internals/graph/activations/argument_assigner.py
```diff
@@ -40,33 +40,33 @@ def assign(self, arguments: Dict[str, Any]) -> List[str]:
...
-        # Start to build the list of assignments, with one sentence for each property
-        # listed (exception for name and description).
-        for one_property, attributes in self.properties.items():
+        # Start to build the list of assignments, with one sentence for each argument
+        for args_name, args_value in arguments.items():
 
-            if one_property in ("name", "description"):
-                # Skip, as this does not need to be in the attribution list
+            # Skip if the value of the argument is None or empty
+            if args_value in (None, "", [], {}):
```
d1donlydfink: I can see skipping None, maybe even an empty string,
but empty lists and dictionaries could easily be valid arguments.

What was the motivation for skipping these?
This is not something we were doing before.

neuro_san/internals/graph/activations/argument_assigner.py
```diff
@@ -40,33 +40,33 @@ def assign(self, arguments: Dict[str, Any]) -> List[str]:
...
                 continue
 
-            args_value: Any = arguments.get(one_property)
-            if args_value is None:
-                # If we do not have an argument value for the property,
-                # do not add anything to the attribution list
-                continue
```
d1donlydfink, line 55: Notice here we were only skipping before when the arg_value was None.

neuro_san/internals/graph/activations/branch_activation.py
d1donlydfink: OK.  Worth a comment where assignments is assigned that we expect it to never be None.

d1donlydfink (changes requested):
* Add a comment where assignments is assigned that we expect it should never be None
* Remove unnecessary parens in if statement (dan == pylint)
* Only skip if the value is None, not empty (string, list, dict)
* Please talk about how you tested something with an external agent, like math_guy_passthrough
* Strongly consider adding a unit test

neuro_san/internals/graph/activations/argument_assigner.py
Noravee: Parentheses have been removed.

neuro_san/internals/graph/activations/argument_assigner.py
Noravee: Thank you for pointing that out.
I thought the llm treat None and empty similarly, but maybe not.

neuro_san/internals/graph/activations/argument_assigner.py
Noravee: Thank you for pointing out.

Noravee:
- A comment was added that the assigner will skip any arguments that the value is None.
- Remove unnecessary brackets.
- Only skip if the value is None, not empty (string, list, dict).
- `math_guy_passthrough` was tested by cli with `python -m neuro_san.client.agent_cli --agent math_guy --sly_data '{"x": 28, "y": 17}' --http`
- A unit test for ArgumentAssigner was added.

tests/neuro_san/internals/graph/activations/test_argument_assigner.py
```diff
@@ -0,0 +1,174 @@
...
+#
+# END COPYRIGHT
+
+from unittest import TestCase
+
+from neuro_san.internals.graph.activations.argument_assigner import ArgumentAssigner
+
+
+class TestArgumentAssigner(TestCase):
+    """Unit tests for ArgumentAssigner class"""
```
d1donlydfink, line 19: Nice. Sorry to not have this kind of test in place earlier, it would have come in handy at a time like this.
That is an artifact of how this code came to be in the first place - by refactoring an experiment from Babak.
Theoretically almost all of the examples here would pass given the old code (and maybe a little more properties config).

PR #358, opened 2025-08-08
Add some Tornado knobs to neuro-san server.

andreidenissov-cog:
This PR adds and exposes 3 knobs for Tornado server performance tuning.
Here is GPT-generated description:

======================================
Tornado Performance Tuning Parameters

TCP Connections Backlog
Parameter: backlog in HTTPServer.bind(port, backlog=...)
Function: Sets the maximum number of pending TCP connections waiting to be accepted by the server.
Impact:
Higher values allow the kernel to queue more incoming connections during traffic spikes, reducing the chance of refused connections.
Default: Typically 128 (depends on OS).

Multiple Server Instances (Process Scaling)
Parameter: HTTPServer.start(num_processes)
Function: Forks multiple worker processes, each with its own Tornado event loop, all sharing the same listening socket.
Impact:
Allows utilization of multiple CPU cores.
Improves concurrency without blocking a single IOLoop.

Typical Usage:
start(0) → Automatically spawns one process per CPU core.
start(N) → Manually specifies number of worker processes.

Idle Connections Timeout
Parameter: idle_connection_timeout in HTTPServer(...)
Function: Sets the maximum idle time (in seconds) before Tornado closes a keep-alive connection.
Impact:
Shorter timeouts free up resources faster under high load.
Longer timeouts reduce reconnections for clients that make periodic requests.
Default: 3600 seconds (1 hour)

========================================


d1donlydfink (changes requested):
We do not need sync-upstream.yml here
Be your own first reviewer.

neuro_san/service/main_loop/server_main_loop.py
```diff
@@ -108,6 +113,20 @@ def parse_args(self):
                                 default=int(os.environ.get("AGENT_MANIFEST_UPDATE_PERIOD_SECONDS", "0")),
                                 help="Periodic run-time update period for manifest in seconds."
                                      " Value <= 0 disables updates.")
+        arg_parser.add_argument("--http_connections_backlog", type=int,
+                                default=int(os.environ.get("AGENT_HTTP_CONNECTIONS_BACKLOG",
+                                                           DEFAULT_HTTP_CONNECTIONS_BACKLOG)),
+                                help="Size of backlog for TCP connections to http server.")
+        arg_parser.add_argument("--http_idle_connections_timeout", type=int,
```
d1donlydfink, line 120: These should go in the Dockerfile as documentation.

d1donlydfink (approved):

PR #366, opened 2025-08-14
UN-3441 UserWarning: Importing verbose from langchain root module is no longer supported langchain.globals.set_verbose()

Noravee:
The warning is from
https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/globals.py 
and can be trace back to https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/language_models/base.py#L92.

The fix is to set the value of `verbose` for chat models so it does not have to call the method to figure out the global default value.

Note that this warning only display for concurrent runs and is likely due to concurrent import of `langchain`.

### DIFF
--- a/neuro_san/internals/run_context/langchain/llms/standard_langchain_llm_factory.py
+++ b/neuro_san/internals/run_context/langchain/llms/standard_langchain_llm_factory.py
@@ -98,7 +98,21 @@ class StandardLangChainLlmFactory(LangChainLlmFactory):
                 max_tokens=config.get("max_tokens"),    # This is always for output
                 tiktoken_model_name=config.get("tiktoken_model_name"),
                 stop=config.get("stop"),
-                verbose=config.get("verbose"),
+
+                # If omitted, this defaults to the global verbose value,
+                # accessible via langchain_core.globals.get_verbose():
+                # https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/globals.py#L53
+                #
+                # However, accessing the global verbose value during concurrent initialization
+                # can trigger the following warning:
+                #
+                # UserWarning: Importing verbose from langchain root module is no longer supported.
+                # Please use langchain.globals.set_verbose() / langchain.globals.get_verbose() instead.
+                # old_verbose = langchain.verbose
+                #
+                # To prevent this, we explicitly set verbose=False here (which matches the default
+                # global verbose value) so that the warning is never triggered.
+                verbose=False,
 
                 # Set stream_usage to True in order to get token counting chunks.
                 stream_usage=True
@@ -136,7 +150,21 @@ class StandardLangChainLlmFactory(LangChainLlmFactory):
                 max_tokens=config.get("max_tokens"),    # This is always for output
                 tiktoken_model_name=config.get("tiktoken_model_name"),
                 stop=config.get("stop"),
-                verbose=config.get("verbose"),
+
+                # If omitted, this defaults to the global verbose value,
+                # accessible via langchain_core.globals.get_verbose():
+                # https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/globals.py#L53
+                #
+                # However, accessing the global verbose value during concurrent initialization
+                # can trigger the following warning:
+                #
+                # UserWarning: Importing verbose from langchain root module is no longer supported.
+                # Please use langchain.globals.set_verbose() / langchain.globals.get_verbose() instead.
+                # old_verbose = langchain.verbose
+                #
+                # To prevent this, we explicitly set verbose=False here (which matches the default
+                # global verbose value) so that the warning is never triggered.
+                verbose=False,
 
                 # Azure-specific
                 azure_endpoint=self.get_value_or_env(config, "azure_endpoint",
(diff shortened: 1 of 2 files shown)


neuro_san/internals/run_context/langchain/llms/standard_langchain_llm_factory.py
```diff
@@ -98,6 +98,7 @@ def create_base_chat_model(self, config: Dict[str, Any]) -> BaseLanguageModel:
                 max_tokens=config.get("max_tokens"),    # This is always for output
                 tiktoken_model_name=config.get("tiktoken_model_name"),
                 stop=config.get("stop"),
+                verbose=config.get("verbose"),
```
d1donlydfink: Yes, most often this can be a simple boolean, but as per our documentation here:
https://github.com/cognizant-ai-lab/neuro-san/blob/main/docs/agent_hocon_reference.md#verbose

This can also be a string which might have the value "extra".
I can tell you that the 1C guys use this "extra" string value.

Perhaps best thing to do would be to convert to boolean before assigning verbose here and everywhere else like this:

```
verbose=bool(config.get("verbose")),
```

d1donlydfink (changes requested):
Verbose can sometimes have a string value "extra"
See PR comments to account for that.

neuro_san/internals/run_context/langchain/llms/standard_langchain_llm_factory.py
ofrancon: For strings, any non-empty string evaluates to True. So if `verbose` is set, it will return True. Even if `verbose` is set to `False`.
`BaseChatModel` expects `verbose` to be a boolean, so we have to convert the config's string to a boolean.
Looks like we need a function to do the conversion. Something like:
```
def _is_verbose(config: Dict[str, Any]) -> bool:
    return config.get("verbose", "").strip().lower() in ("true", "1", "yes", "y", "on", "extra")
```
I assume "extra" means we do want to set verbose to true on the underlying model.

ofrancon (changes requested):
Tricky case where we expect strings but `BaseChatModel` expects boolean

neuro_san/internals/run_context/langchain/llms/standard_langchain_llm_factory.py
Noravee: Thank you for reminding me that we already have a key `verbose`  in the `llm_config` , which is for the `AgentExecutor` .
The `verbose`  setting in the PR is for the chat model and actually **useless** when using with `AgentExecutor` .
In general, if the this value is not provided it will default to `False` , but will go get the default value from the method that could trigger the warning in the concurrent scenario.
Can I hard-code `verbose=False` in the `StandardLangchainLlmFactory`  and leave the `default_llm_info`  as is (without adding `verbose`)?
So we don't have to worry about conflict between `verbose` of `AgentExecutor` and chat model?

d1donlydfink:
Talking with Nor about this his alternate strategy of simply setting verbose=False in these cases actually sounds more reasonable at least to get rid of the annoying warning message. We would then not leap into the conflation of the llm_config.verbose being used at AgentExecutor level *and* this level ... right away at least.

neuro_san/internals/run_context/langchain/llms/standard_langchain_llm_factory.py
```diff
@@ -99,6 +99,21 @@ def create_base_chat_model(self, config: Dict[str, Any]) -> BaseLanguageModel:
...
+                # However, accessing the global verbose value during concurrent initialization
+                # can trigger the following warning:
+                #
+                # UserWarning: Importing verbose from langchain root module is no longer supported.
+                # Please use langchain.globals.set_verbose() / langchain.globals.get_verbose() instead.
+                # old_verbose = langchain.verbose
+                #
+                # To prevent this, we explicitly set verbose=False here (which matches the default
+                # global verbose value) so that the warning is never triggered.
+                verbose=False,
```
Noravee, line 115: Set `verbose` to `False` and provide the reason in comments.

neuro_san/internals/run_context/langchain/llms/standard_langchain_llm_factory.py
Noravee: - Remove `verbose` in the class config of `default_llm_info`.

- Set `verbose` to `False` in `StandardLangchainLlmFactory` and provide the following reason:

If omitted, this defaults to the global verbose value,
accessible via langchain_core.globals.get_verbose():https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/globals.py#L53

However, accessing the global verbose value during concurrent initialization
can trigger the following warning:

UserWarning: Importing verbose from langchain root module is no longer supported.
Please use langchain.globals.set_verbose() / langchain.globals.get_verbose() instead.
old_verbose = langchain.verbose

To prevent this, we explicitly set verbose=False here (which matches the default
global verbose value) so that the warning is never triggered.

Noravee:
- Remove `verbose` in the class config of `default_llm_info`.

- Set `verbose` to `False` in `StandardLangchainLlmFactory` and provide the following reason:

If omitted, this defaults to the global verbose value,
accessible via langchain_core.globals.get_verbose():https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/globals.py#L53

However, accessing the global verbose value during concurrent initialization
can trigger the following warning:

UserWarning: Importing verbose from langchain root module is no longer supported.
Please use langchain.globals.set_verbose() / langchain.globals.get_verbose() instead.
old_verbose = langchain.verbose

To prevent this, we explicitly set verbose=False here (which matches the default
global verbose value) so that the warning is never triggered.

ofrancon (approved):
works for me

PR #368, opened 2025-08-14
UN-3432 Fix duplicate request ids.

andreidenissov-cog:
This PR does 2 things:
1. Fixes the logic for generating server-supplied request ids in case a client doesn't provide this information in request metadata;
2. Extends information for periodic logging of server run-time resources used, now it shows what kinds of file descriptors
    are currently opened by the server and describes inbound/outbound socket connections.
    This extra information helps with identifying potential points of resources leakage.

### DIFF
--- a/neuro_san/service/http/handlers/base_request_handler.py
+++ b/neuro_san/service/http/handlers/base_request_handler.py
@@ -164,6 +164,8 @@ class BaseRequestHandler(RequestHandler):
         """
         return
 
+    # Tornado can handle both syns and async versions of "prepare" method
+    # pylint: disable=invalid-overridden-method
     async def prepare(self):
         if not self.application.is_serving():
             self.set_status(503)
--- a/neuro_san/service/http/handlers/base_request_handler.py
+++ b/neuro_san/service/http/handlers/base_request_handler.py
@@ -90,8 +90,10 @@ class BaseRequestHandler(RequestHandler):
                 self.show_absent,
                 self.logger)
         # Put in our own unique id so we have some way to track this request:
-        if "request_id" not in metadata_dict:
-            metadata_dict["request_id"] = self.request_id
+        user_request_id: str = metadata_dict.get("request_id", "None")
+        if user_request_id == "None":
+            metadata_dict["request_id"] = f"request-{self.request_id}"
+        return metadata_dict
 
     @classmethod
     def get_request_metadata(cls, request,
--- a/neuro_san/service/http/handlers/base_request_handler.py
+++ b/neuro_san/service/http/handlers/base_request_handler.py
@@ -24,6 +24,7 @@ import asyncio
 import tornado
 from tornado.web import RequestHandler
 
+from neuro_san.service.utils.async_atomic_counter import AsyncAtomicCounter
 from neuro_san.service.generic.async_agent_service import AsyncAgentService
 from neuro_san.service.generic.async_agent_service_provider import AsyncAgentServiceProvider
 from neuro_san.internals.network_providers.agent_network_storage import AgentNetworkStorage
@@ -37,7 +38,7 @@ class BaseRequestHandler(RequestHandler):
     Provides logic to inject neuro-san service specific data
     into local handler context.
     """
-    request_id: int = 0
+    request_id_counter: AsyncAtomicCounter = AsyncAtomicCounter()
 
     # pylint: disable=attribute-defined-outside-init
     # pylint: disable=too-many-arguments
@@ -64,9 +65,7 @@ class BaseRequestHandler(RequestHandler):
         self.logger = HttpLogger(forwarded_request_metadata)
         self.network_storage_dict: Dict[str, AgentNetworkStorage] = network_storage_dict
         self.show_absent: bool = os.environ.get("SHOW_ABSENT_METADATA") is not None
-
-        # Set default request_id for this request handler in case we will need it:
-        BaseRequestHandler.request_id += 1
+        self.request_id: int = 0
 
         if os.environ.get("AGENT_ALLOW_CORS_HEADERS") is not None:
             self.set_header("Access-Control-Allow-Origin", "*")
@@ -84,11 +83,15 @@ class BaseRequestHandler(RequestHandler):
         from incoming request.
         :return: dictionary of user request metadata; possibly empty
         """
-        return BaseRequestHandler.get_request_metadata(
-            self.request,
-            self.forwarded_request_metadata,
-            self.show_absent,
-            self.logger)
+        metadata_dict: Dict[str, Any] =\
+            BaseRequestHandler.get_request_metadata(
+                self.request,
+                self.forwarded_request_metadata,
+                self.show_absent,
+                self.logger)
+        # Put in our own unique id so we have some way to track this request:
+        if "request_id" not in metadata_dict:
+            metadata_dict["request_id"] = self.request_id
 
     @classmethod
     def get_request_metadata(cls, request,
@@ -112,9 +115,6 @@ class BaseRequestHandler(RequestHandler):
         for item_name in forwarded_request_metadata:
             if item_name in headers.keys():
                 result[item_name] = headers[item_name]
-            elif item_name == "request_id":
-                # Generate unique id so we have some way to track this request:
-                result[item_name] = f"request-{BaseRequestHandler.request_id}"
             else:
                 if show_absent and logger:
                     logger.warning({}, "MISSING METADATA VALUE: %s request %s", item_name, request.uri)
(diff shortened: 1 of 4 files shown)


neuro_san/service/http/handlers/base_request_handler.py
```diff
@@ -37,7 +38,7 @@ class BaseRequestHandler(RequestHandler):
     Provides logic to inject neuro-san service specific data
     into local handler context.
     """
-    request_id: int = 0
+    request_id_counter: AsyncAtomicCounter = AsyncAtomicCounter()
```
andreidenissov-cog, line 42: Async version of atomic counter.

neuro_san/service/http/handlers/base_request_handler.py
```diff
@@ -84,11 +83,17 @@ def get_metadata(self) -> Dict[str, Any]:
...
+            BaseRequestHandler.get_request_metadata(
+                self.request,
+                self.forwarded_request_metadata,
+                self.show_absent,
+                self.logger)
+        # Put in our own unique id so we have some way to track this request:
+        user_request_id: str = metadata_dict.get("request_id", "None")
+        if user_request_id == "None":
+            metadata_dict["request_id"] = f"request-{self.request_id}"
+        return metadata_dict
```
andreidenissov-cog, line 96: Add locally generated request id to metadata if it doesn't have it already.

neuro_san/service/http/handlers/base_request_handler.py
```diff
@@ -112,9 +117,6 @@ def get_request_metadata(cls, request,
         for item_name in forwarded_request_metadata:
             if item_name in headers.keys():
                 result[item_name] = headers[item_name]
-            elif item_name == "request_id":
-                # Generate unique id so we have some way to track this request:
-                result[item_name] = f"request-{BaseRequestHandler.request_id}"
             else:
```
andreidenissov-cog, line 120: We don't do it here anymore.

neuro_san/service/http/handlers/base_request_handler.py
```diff
@@ -162,14 +164,19 @@ def data_received(self, chunk):
         """
         return
 
-    def prepare(self):
+    # Tornado can handle both syns and async versions of "prepare" method
+    # pylint: disable=invalid-overridden-method
+    async def prepare(self):
         if not self.application.is_serving():
```
andreidenissov-cog, line 170: Tornado can handle both sync and async methods thanks to internal magic machinery,
but async is "native" to its event loop.

neuro_san/service/http/server/http_server.py
```diff
@@ -146,11 +147,15 @@ def log_resources_usage(self):
...
-                         fds, soft_limit, hard_limit, conn)
+        fd_dict, soft_limit, hard_limit = ServiceResources.get_fd_usage()
+        sock_classes = ServiceResources.classify_sockets(self.http_port)
+        log_dict: Dict[str, Any] = {
+            "soft_limit": soft_limit,
+            "hard_limit": hard_limit,
+            "file_descroptors": fd_dict,
+            "sockets": sock_classes
+        }
+        self.logger.info({}, "Used: %s", json.dumps(log_dict, indent=4))
```
andreidenissov-cog, line 159: Log extended information about resources usage.

neuro_san/service/utils/async_atomic_counter.py
```diff
@@ -0,0 +1,19 @@
+import asyncio
+
+
```
andreidenissov-cog, line 15: Very simple async atomic counter.

neuro_san/service/utils/service_resources.py
```diff
@@ -24,17 +30,70 @@ class ServiceResources:
     """
```
andreidenissov-cog, line 45: Added to this class: several helper functions to classify and describe currently active file descriptors and socket connections. Mostly compliments of GPT.

neuro_san/service/utils/service_resources.py
```diff
@@ -24,17 +30,70 @@ class ServiceResources:
...
+                if fd in inet_fds:
+                    kind = "socket_inet"
+                elif fd in unix_fds:
+                    kind = "socket_unix"
+                else:
+                    kind = "other"
+            elif stat.S_ISFIFO(mode):
+                kind = "fifo_pipe"
+            else:
+                kind = "other"
```
andreidenissov-cog, line 101: There are more file descriptor kinds, but we are interested in regular files,
sockets - which also count towards file descriptors resources,
and FIFO pipes - not sure what they do, but probably related to organizing internal server connectivity.

neuro_san/service/utils/service_resources.py
```diff
@@ -47,3 +106,114 @@ def active_tcp_on_port(cls, port: int) -> int:
...
+            return os.readlink(path)
+        except Exception:  # pylint: disable=broad-exception-caught
+            return None
+
+    @classmethod
+    def classify_unix_sockets(cls):
+        """
+        Classify Unix sockets open for current process
+        """
+        p = psutil.Process()
```
andreidenissov-cog: Not using it for now.

neuro_san/service/utils/async_atomic_counter.py
```diff
@@ -0,0 +1,19 @@
+import asyncio
```
d1donlydfink, line 13: Add a copyright notice here please.

FWIW: This is a reasonable thing to eventually push down into leaf-server-common next to its synchronous counterpart there.  Ideally they'd have the same interface.

neuro_san/service/utils/async_atomic_counter.py
```diff
@@ -0,0 +1,19 @@
...
+
+class AsyncAtomicCounter:
+    """
+    Class implements atomic incrementing counter for async execution environment.
+    """
+    def __init__(self, start: int = 1):
+        self._value = start
+        self._lock = asyncio.Lock()
+
+    async def next(self):
```
d1donlydfink: Add a return type hint and a :return: comment in the method docs please

neuro_san/service/utils/service_resources.py
```diff
@@ -24,17 +30,70 @@ class ServiceResources:
...
 
     @classmethod
-    def get_fd_usage(cls) -> Tuple[int, int, int]:
+    def iter_fds(cls):
         """
-        :return: tuple of 3 integer values:
-            number of file descriptors in use;
+        Iterator for all open file descriptors in current process:
+        """
+        fd_dir = "/proc/self/fd" if sys.platform.startswith("linux") else "/dev/fd"
```
d1donlydfink: FWIW: We do have people running on windows now but no real resources to test it.
Will this at least not crash on windows?

neuro_san/service/http/server/http_server.py
```diff
@@ -146,11 +147,15 @@ def log_resources_usage(self):
         file descriptors and open inet connections on server port.
         """
         # Get used file descriptors:
-        fds, soft_limit, hard_limit = ServiceResources.get_fd_usage()
-        # Get active TCP connections to our http port:
-        conn = ServiceResources.active_tcp_on_port(self.http_port)
-        self.logger.info({}, "Used: file descriptors %d (%d, %d) connections: %d",
-                         fds, soft_limit, hard_limit, conn)
+        fd_dict, soft_limit, hard_limit = ServiceResources.get_fd_usage()
+        sock_classes = ServiceResources.classify_sockets(self.http_port)
```
d1donlydfink, line 151: Does this take awhile to do?
Also: If I recall this happens periodically, but the thing I worry about more is while this logging is going on are there locks within psutil that are going to have everything come to a screeching halt?

neuro_san/service/utils/service_resources.py
andreidenissov-cog: I believe not, but the only real way to be sure is to run this on Windows.
At least it does not crash when I force "you run on Windows" mode.

neuro_san/service/http/server/http_server.py
andreidenissov-cog: That's actually a good point if somebody will fancy running these stats every second (we don't allow to run it with <1 second intervals). So I moved the logic out of event loop to be run in a thread alongside.

neuro_san/service/utils/async_atomic_counter.py
```diff
@@ -0,0 +1,55 @@
...
+# END COPYRIGHT
+
+import asyncio
+
+
+class AsyncAtomicCounter:
+    """
+    Class implements atomic incrementing counter for async execution environment.
+    """
+    def __init__(self, start: int = 0):
```
andreidenissov-cog, line 20: Made API closer to AtomicCounter from leaf-server-common.

neuro_san/service/utils/service_resources.py
andreidenissov-cog: I have added code to deal with Windows platform - again, GPT is here to help.

d1donlydfink (approved):

PR #370, opened 2025-08-15
Factor out arg prep in ServerMainLoop

d1donlydfink:
In another branch when adding new args, pylint started complaining about parse_args() being to big. 
So this breaks it up.

### DIFF
--- a/neuro_san/service/main_loop/server_main_loop.py
+++ b/neuro_san/service/main_loop/server_main_loop.py
@@ -71,9 +71,9 @@ class ServerMainLoop(ServerLoopCallbacks):
         self.http_server_config = HttpServerConfig()
         self.watcher_config: Dict[str, Any] = {}
 
-    def parse_args(self):
+    def prepare_args(self) -> ArgumentParser:
         """
-        Parse command-line arguments into member variables
+        :return: An ArgumentParser set up to parse command-line arguments
         """
         # Set up the CLI parser
         arg_parser = ArgumentParser()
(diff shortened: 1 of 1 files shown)


neuro_san/service/main_loop/server_main_loop.py
```diff
@@ -71,9 +71,9 @@ def __init__(self):
         self.http_server_config = HttpServerConfig()
         self.watcher_config: Dict[str, Any] = {}
 
-    def parse_args(self):
+    def prepare_args(self) -> ArgumentParser:
         """
-        Parse command-line arguments into member variables
+        :return: An ArgumentParser set up to parse command-line arguments
         """
```
d1donlydfink, line 77: Break a prepare_args() method out of ServerMainLoop which simply prepares the ArgumentParser and returns it for later processing.

neuro_san/service/main_loop/server_main_loop.py
```diff
@@ -133,6 +133,13 @@ def parse_args(self):
                                                            DEFAULT_HTTP_SERVER_MONITOR_INTERVAL_SECONDS)),
                                 help="Http server resources monitoring/logging interval in seconds "
                                      "0 means no logging")
+        return arg_parser
```
d1donlydfink, line 136: Return the ArgumentParser

neuro_san/service/main_loop/server_main_loop.py
```diff
@@ -133,6 +133,13 @@ def parse_args(self):
...
+
+    def parse_args(self):
+        """
+        Parse command-line arguments into member variables
+        """
+        arg_parser: ArgumentParser = self.prepare_args()
```
d1donlydfink, line 142: Call the new method to get the arg parser.
This nicely breaks out responsibilities so that parse_args() doesn't get too big for pylint to complain about how many lines it is.  It is on the threshold of that now.

andreidenissov-cog (approved):
LGTM.

PR #371, opened 2025-08-15
close() objects on sly_data that can be closed at the end of a session

d1donlydfink:
This will allow folks to share longer-lived objects amongst coded tools more properly.
Sharing could be amongst different coded tools, or different invocations of the same coded tool within the same agent network invocation.  This should also contribute to the resistance of anyone needing to use a global variable.

### DIFF
--- a/neuro_san/coded_tools/math_guy/async_closeable.py
+++ b/neuro_san/coded_tools/math_guy/async_closeable.py
@@ -1,0 +1,27 @@
+
+# Copyright (C) 2023-2025 Cognizant Digital Business, Evolutionary AI.
+# All Rights Reserved.
+# Issued under the Academic Public License.
+#
+# You can be released from the terms, and requirements of the Academic Public
+# License by purchasing a commercial license.
+# Purchase of a commercial license is mandatory for any use of the
+# neuro-san SDK Software in commercial settings.
+#
+# END COPYRIGHT
+
+from logging import getLogger
+from logging import Logger
+
+
+class AsyncCloseable:
+    """
+    A simple object used to test the asynchronous close()-ing of objects on sly_data.
+    """
+
+    async def close(self):
+        """
+        Close the object
+        """
+        logger: Logger = getLogger(self.__class__.__name__)
+        logger.info("async close() called")
--- a/neuro_san/coded_tools/math_guy/calculator.py
+++ b/neuro_san/coded_tools/math_guy/calculator.py
@@ -14,6 +14,8 @@ from typing import Any
 from typing import Dict
 
 from neuro_san.interfaces.coded_tool import CodedTool
+from neuro_san.coded_tools.math_guy.async_closeable import AsyncCloseable
+from neuro_san.coded_tools.math_guy.sync_closeable import SyncCloseable
 
 
 class Calculator(CodedTool):
@@ -77,5 +79,11 @@ class Calculator(CodedTool):
                 return "Can't divide by 0"
 
         sly_data["equals"] = retval
+
+        # Add close()-able objects to test closing of sly_data
+        # These are only to enhance testing coverage and are non-essential
+        # to the function of this coded tool.
+        sly_data["sync_closeable"] = SyncCloseable()
+        sly_data["async_closeable"] = AsyncCloseable()
 
         return "Check sly_data['equals'] for the result"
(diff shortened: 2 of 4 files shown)


neuro_san/internals/chat/data_driven_chat_session.py
```diff
@@ -212,6 +215,9 @@ async def streaming_chat(self, user_input: str,
         # we would only be blocking our own event loop.
         await queue.put_final_item(synchronous=True)
 
+        # Close any objects on sly data that can be closed.
+        await self.close_sly_data()
```
d1donlydfink, line 219: Now that we know we are all done, close out objects in the sly data that can be closed.

neuro_san/internals/chat/data_driven_chat_session.py
```diff
@@ -269,3 +275,45 @@ def create_outgoing_message_processor(self) -> MessageProcessor:
...
+        """
+        if self.sly_data is None or len(self.sly_data) == 0:
+            # Nothing to close
+            return
+
+        # Try closing anything that can be close()-d in the sly_data.
+        for key, value in self.sly_data.items():
+            # Note: It's possible we might want to try closing keys as well
+            description: str = f"sly_data[{key}]"
+            await self.close_one(value, description)
```
d1donlydfink, line 291: New method to loop through sly_data.

neuro_san/internals/chat/data_driven_chat_session.py
```diff
@@ -269,3 +275,45 @@ def create_outgoing_message_processor(self) -> MessageProcessor:
...
+
+    async def close_one(self, value: Any, description: str):
+        """
+        Close a single object if it can be closed.
+        :param value: The object to close
+        :param description: Where the object to close came from
+        """
```
d1donlydfink, line 298: Method to close out a single sly_data entity.
Most of the time we won't need this because most of the time sly_data is "Just Data".
But when people put objects on sly_data that have sessions and can be shared between CodedTools, then that's when we need this.

neuro_san/coded_tools/math_guy/async_closeable.py
```diff
@@ -0,0 +1,27 @@
...
+# END COPYRIGHT
+
+from logging import getLogger
+from logging import Logger
+
+
+class AsyncCloseable:
+    """
+    A simple object used to test the asynchronous close()-ing of objects on sly_data.
+    """
```
d1donlydfink, line 20: Used for testing asynchronous close()-ing of sly_data

neuro_san/coded_tools/math_guy/calculator.py
```diff
@@ -78,4 +80,10 @@ async def async_invoke(self, args: Dict[str, Any], sly_data: Dict[str, Any]) ->
 
         sly_data["equals"] = retval
 
+        # Add close()-able objects to test closing of sly_data
+        # These are only to enhance testing coverage and are non-essential
+        # to the function of this coded tool.
+        sly_data["sync_closeable"] = SyncCloseable()
+        sly_data["async_closeable"] = AsyncCloseable()
```
d1donlydfink, line 87: Add two kinds of close()-able objects to sly_data for testing.

neuro_san/coded_tools/math_guy/sync_closeable.py
```diff
@@ -0,0 +1,27 @@
...
+# END COPYRIGHT
+
+from logging import getLogger
+from logging import Logger
+
+
+class SyncCloseable:
+    """
+    A simple object used to test the synchronous close()-ing of objects on sly_data.
+    """
```
d1donlydfink, line 20: Used for testing synchronous close()-ing of sly_data

Noravee (approved):
LGTM.

neuro_san/coded_tools/math_guy/async_closeable.py
ofrancon: It really belongs to the `tests` package then. Otherwise it's going to be systematically pip installed by everyone, for no valid reason.

ofrancon (approved):
That should really be tested by a unit test instead of bastardizing math_guy...

neuro_san/coded_tools/math_guy/async_closeable.py
d1donlydfink: There is a fine line here that I don't yet know what to do about.
Some of the networks here are bare-bones examples and some are tests.

Maybe the only networks/coded_tools in neuro-san should be the test infrastructure ones
and the rest moved to the test/directory. That would require:
* Some discussion as to which examples need to be promoted to neuro-san-studio (maybe none)
* Some infrastructure so that networks/coded tool paths could come from > 1 place, more like a PYTHONPATH.

That's out of scope here.
In the mean time, these 2 files are very small and quite harmless.

PR #373, opened 2025-08-19
Remove yaml for codefresh, now using github actions

donn-leaf:
Just removing this Codefresh yaml now that the test is accomplished in github actions.


ofrancon (approved):
lgtm

PR #375, opened 2025-08-20
UN-3443: Add "reasoning", "reasoning_effort", and "verbosity" parameters to "openai" class in default llm info.

Noravee:
Add `reasoning`, `reasoning_effort`, and `verbosity` for class `openai`.


ofrancon (approved):
lgtm

PR #379, opened 2025-08-26
UN-3377 Initialize logging setup for AsyncioExecutor.

andreidenissov-cog:
This PR attempts to address missing logging metadata issue by running logging setup
for each AsyncioExecutor instance as blocking call - making sure it is done
before we hand over this AsyncioExecutor for request processing.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,7 +6,7 @@
 # requirements-build.txt
 
 # In-house dependencies
-leaf-common>=1.2.26
+leaf-common>=1.2.27
 leaf-server-common>=0.1.21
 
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -6,7 +6,7 @@
 # requirements-build.txt
 
 # In-house dependencies
-leaf-common>=1.2.26
+leaf-common>=1.2.27
 leaf-server-common>=0.1.21
```
andreidenissov-cog, line 10: Switch to latest leaf-common with AsyncioExecutor support for initialization step.

donn-leaf (approved):
LGTM

PR #382, opened 2025-09-03
UN-3483 Fix large async HTTP JSON line ingestion ("Chunk too big" ValueError)

d1donlydfink:
* Allow asynchronously ingested streams to have large single-line JSON structures, avoiding the "Chunk too big" ValueError in aiohttp library as seen by 1C guys
* Worth noting that this error will not show itself in synchronous requests code (code inspected)
* Also fix typo in docs

Tested: math_guy_passthrough. External agents use the async_http_service_agent_session path.

### DIFF
--- a/neuro_san/session/async_http_service_agent_session.py
+++ b/neuro_san/session/async_http_service_agent_session.py
@@ -127,7 +127,7 @@ class AsyncHttpServiceAgentSession(AbstractHttpServiceAgentSession, AsyncAgentSe
 
                             # Grab a single line
                             line: bytes = accumulator[:index]
-                            unicode_line = line.decode('utf-8')
+                            unicode_line = line.decode("utf-8")
                             if unicode_line.strip():    # Skip empty lines
 
                                 # We have a line with something in it.
@@ -143,7 +143,7 @@ class AsyncHttpServiceAgentSession(AbstractHttpServiceAgentSession, AsyncAgentSe
 
                     # If there is anything left in the accumulator, yield it
                     if len(accumulator) > 0:
-                        result_dict = json.loads(accumulator.decode('utf-8'))
+                        result_dict = json.loads(accumulator.decode("utf-8"))
                         yield result_dict
 
         except (asyncio.TimeoutError, ClientOSError) as exc:


docs/agent_hocon_reference.md
```diff
@@ -35,7 +35,7 @@ Sub-keys to those dictionaries will be described in the next-level down heading
     - [error_fragments](#error_fragments)
     - [tools](#tools)
     - [metadata](#metadata)
-        - [decription](#decription)
+        - [description](#description)
```
d1donlydfink, line 38: Fix typo in agent_hocon_reference.md

neuro_san/session/async_http_service_agent_session.py
```diff
@@ -107,12 +108,34 @@ async def streaming_chat(self, request_dict: Dict[str, Any]) -> Generator[Dict[s
...
+                            unicode_line = line.decode('utf-8')
+                            if unicode_line.strip():    # Skip empty lines
+
+                                # We have a line with something in it.
+                                # Decode and yield as a dictionary
+                                result_dict = json.loads(unicode_line)
+                                yield result_dict
+
+                            # Remove the previous line from the accumulator
+                            accumulator = accumulator[index + len(separator):]
```
d1donlydfink, line 139: Look for lines but do our own chunk accumulation for potentially large json structures coming from the server.

d1donlydfink: Worth noting that we are sacrificing some memory compaction/re-use efficiency here for being able to handle the arbitrary size.

andreidenissov-cog: GPT sez: even with response.content.iter_any() there is no guarantee the other side will not dump a huge chunk in one go that will overrun httpx internal buffer.
Suggestion is to use response.aiter_bytes(N) for some reasonable N bytes (like 16-32K).
Also - you assume that all contents ends with newline always?
What if not?

andreidenissov-cog (changes requested):
Some possible issues with that fix?

neuro_san/session/async_http_service_agent_session.py
d1donlydfink: > GPT sez: even with response.content.iter_any() there is no guarantee the other side will not dump a huge chunk in one go that will overrun httpx internal buffer.

1) There is no aiter_bytes() anywhere in the aiohttp library.  GPT got you.  Closest thing is iter_chunked(n).

2) What iter_any() gives us that iter_chunks() or iter_chunked(n) does not is potential for a bit less latency in parsing a response when waiting for a certain number of bytes to roll by.  Remember our clients are expecting exactly 1 line per JSON response in streaming_chat().  iter_any() allows the server to flush a buffer on a line boundary and have the client take it up immediately.

3) As for huge chunks that overrun ... we'll see.  I think 1C guys have an example for us and tornado. This is the first step of change to see how extreme we have to be.  The original line-based iteration used StreamReader.readany() and that path specifically had the "Chunk too big" error path. The assumption there was that any one line would fit within a chunk, which is no longer the case for us, hence the new logic. Specifically in the iter_any() approach, there is no exception path due to buffer size at all. There should still room for give and take for buffer sizes at lower levels of the network IO stack.

> you assume that all contents ends with newline always?

Yes, that is already the contract - 1 line (separator = b"\n") per JSON structure streamed.
We wouldn't have the comms we have today without that as part of the streaming_chat() contract.

andreidenissov-cog: 1. Hallucinations make this fun;
2. iter_chunked(N) doesn't mean "wait till you get exactly N bytes". It is an upper limit on what you can get in one transaction, a safety rail. I think using something like 64K is prudent;
3. Newline-at-the-end contract - yeah, maybe. Being paranoid, I would check that when we are done with input stream, accumulator is indeed empty. ValueError() or something;
4. What happens if your chunk contains several full lines, not just one or zero?

d1donlydfink: I can try some of that.

d1donlydfink: OK. Added 2, 3, 4.  1 seemed unnecessary. :)

neuro_san/session/async_http_service_agent_session.py
```diff
@@ -107,12 +109,42 @@ async def streaming_chat(self, request_dict: Dict[str, Any]) -> Generator[Dict[s
...
+
+                            # Allow for case of multiple lines in one chunk
+                            index = accumulator.find(separator)
```
d1donlydfink, line 142: Allow for > 1 line in chunk

neuro_san/session/async_http_service_agent_session.py
```diff
@@ -107,12 +109,42 @@ async def streaming_chat(self, request_dict: Dict[str, Any]) -> Generator[Dict[s
...
+
+                    # If there is anything left in the accumulator, yield it
+                    if len(accumulator) > 0:
+                        result_dict = json.loads(accumulator.decode('utf-8'))
+                        yield result_dict
```
d1donlydfink, line 147: Allow for leftovers without last newline.

andreidenissov-cog (approved):
Looks good.

PR #383, opened 2025-09-03
UN-3385 lazy llm-provider imports

d1donlydfink:


requirements.txt
```diff
@@ -6,7 +6,7 @@
 # requirements-build.txt
 
 # In-house dependencies
-leaf-common>=1.2.27
+leaf-common>=1.2.28
```
d1donlydfink, line 9: Use the latest leaf-common with new Resolver and ResolverUtil functionality.

requirements.txt
```diff
@@ -32,20 +32,22 @@ ruamel.yaml>=0.18.6
...
+# Keep all these guys limited for now, as it can be the wild wild west out there
+# We don't want to be on a dependencies treadmill with timing not of our choosing.
 langchain>=0.3.15,<0.4
-langchain-anthropic>=0.3.11,<0.4
-langchain-aws>=0.2.27,<0.3
 langchain-community>=0.3.19,<0.4
-langchain-google-genai>=2.1.8,<3.0
-langchain-openai>=0.3.29,<0.4
-langchain-nvidia-ai-endpoints>=0.3.8,<0.4
-langchain-ollama>=0.3.4,<0.4
```
d1donlydfink, line 44: Move all of the langchain_* LLM vendor libs to the requirements-build.txt
This allows us to not import as much of the world as a basis.

requirements.txt
```diff
@@ -32,20 +32,22 @@ ruamel.yaml>=0.18.6
...
 bs4>=0.0.2,<0.1
 pydantic>=2.9.2,<3.0
 
+# OpenAI is the only LLM provider we require out of the box so that users
+# have a decent experience running the demos out of the box.
+# Other LLM provider's langchain-<blah> can be optionally installed so that
+# those guys work, but we do not want to make everyone install everything
+# and all the ensuing transient dependencies for the world.
+# For a list of other LLMs we support out of the box, see requirements-build.txt
+langchain-openai>=0.3.29,<0.4
```
d1donlydfink, line 49: OpenAI is now the only langchain-* library we force imports for, for purposes of a smooth out-of-the-box experience.  Given this, it's possible for installations to pip _uninstall_ langchain-openai and openai packages and install their own langchain-* of choice all this will still be good.

neuro_san/internals/chat/data_driven_chat_session.py
```diff
@@ -40,6 +42,11 @@
 from neuro_san.message_processing.answer_message_processor import AnswerMessageProcessor
 from neuro_san.message_processing.structure_message_processor import StructureMessageProcessor
 
+# Lazily import specific errors from llm providers
+PATIENCE_ERRORS: Tuple[Type[Any], ...] = ResolverUtil.create_type_tuple([
+                                            "openai.BadRequestError",
+                                         ])
```
d1donlydfink, line 48: Typical pattern of using the ResolverUtil to lazily import a list of classes and return a tuple of these classes for use in except clauses (see below) or isinstance() cases.   Only classes that are installed in the venv are included in the end list. 

 If *none* of the classes listed are installed, then what is returned here is the empty tuple () and any implied conditionals within except and isinstance() clauses simply evaluate to False. Kinda nice actually.

For this case in particular, this means that openai libs do not necessarily have to be installed in order for neuro-san to work with (say) only Anthropic.

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
```diff
@@ -72,6 +71,13 @@
 
 MINUTES: float = 60.0
 
+# Lazily import specific errors from llm providers
+API_ERROR_TYPES: Tuple[Type[Any], ...] = ResolverUtil.create_type_tuple([
+                                            "openai.APIError",
+                                            "anthropic.APIError",
+                                            "langchain_google_genai.chat_models.ChatGoogleGenerativeAIError",
+                                         ])
```
d1donlydfink, line 79: Another lazy install case from multiple LLM vendors.
Again, if any of these are not installed, they do not get included in the resulting Tuple for use in the except clause below.

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
```diff
@@ -33,10 +31,16 @@
...
-from neuro_san.internals.utils.resolver_util import ResolverUtil
 
 KEYS_TO_REMOVE_FOR_USER_CLASS: Set[str] = {"class", "verbose"}
 
+# Lazily import specific errors from llm providers
+API_KEY_ERRORS: Tuple[Type[Any], ...] = ResolverUtil.create_type_tuple([
+                                            "google.auth.exceptions.DefaultCredentialsError",
+                                            "openai.OpenAIError",
+                                            "pydantic_core.ValidationError",
+                                        ])
```
d1donlydfink, line 42: More of the same.
In this case pydantic_core is something we always want installed, but that ends up always coming.
In order for the code below to more or less appear as it did before we include the mandatory ValidationError as part of the not-so-mandatory list. 🤷

neuro_san/internals/run_context/langchain/llms/standard_langchain_llm_factory.py
```diff
@@ -13,14 +13,9 @@
...
-from langchain_anthropic.chat_models import ChatAnthropic
-from langchain_aws import ChatBedrock
 from langchain_core.language_models.base import BaseLanguageModel
-from langchain_google_genai.chat_models import ChatGoogleGenerativeAI
-from langchain_nvidia_ai_endpoints import ChatNVIDIA
-from langchain_ollama import ChatOllama
-from langchain_openai.chat_models.azure import AzureChatOpenAI
-from langchain_openai.chat_models.base import ChatOpenAI
+
+from leaf_common.config.resolver import Resolver
```
d1donlydfink, line 18: Note that we are exchanging all the langchain_* vendor imports for the Resolver.

neuro_san/internals/run_context/langchain/llms/standard_langchain_llm_factory.py
```diff
@@ -72,7 +67,20 @@ def create_base_chat_model(self, config: Dict[str, Any]) -> BaseLanguageModel:
...
         if chat_class == "openai":
+
+            # OpenAI is the one chat class that we do not require any extra installs.
+            # This is what we want to work out of the box.
+            # Nevertheless, have it go through the same lazy-loading resolver rigamarole as the others.
+
+            # pylint: disable=invalid-name
+            ChatOpenAI = resolver.resolve_class_in_module("ChatOpenAI",
+                                                          module_name="langchain_openai.chat_models.base",
+                                                          install_if_missing="langchain-openai")
```
d1donlydfink, line 83: For any chat class we use the resolver to find the correct type so all the LLM vendors do not have to be installed at the same time, thus reducing our dependency footprint.  
This makes us less of a pandasai clown-car importing the world.

The install_if_missing parameter here is the package to pip install should the class itself not be found.
Note that we are only using the install_if_missing when we resolve the BaseLanguageModel/ChatModel and if the correct langchain_* vendor LLM package is not installed, the resultant ValueError here will have a message telling people to "pip install <install_if_missing>" to get the functionality going.

Same goes for all the core-supported LLMs below in this file.

neuro_san/internals/run_context/langchain/token_counting/langchain_token_counter.py
```diff
@@ -47,6 +48,12 @@
 # See: https://docs.python.org/3/library/contextvars.html
 ORIGIN_INFO: ContextVar[str] = ContextVar('origin_info', default=None)
 
+# Keep a single lazy resolution of OpenAI chat model types.
+OPENAI_CHAT_TYPES: Tuple[Type[Any], ...] = ResolverUtil.create_type_tuple([
+                                                "langchain_openai.chat_models.base.ChatOpenAI",
+                                                "langchain_openai.chat_models.azure.AzureChatOpenAI",
+                                           ])
```
d1donlydfink, line 55: Another example of lazily loading a tuple of OpenAI-only types.

neuro_san/internals/utils/resolver_util.py
```diff
@@ -1,98 +0,0 @@
-
```
d1donlydfink, line 1: We no longer need this file in neuro-san as we are using the one coming from leaf-common now.

requirements-build.txt
```diff
@@ -1,6 +1,13 @@
 # We separate out requirements that are specific to the build, but not
 # necessary for operation to minimize the size of containers
 
+# For LLM Access
+langchain-anthropic>=0.3.11,<0.4
+langchain-aws>=0.2.27,<0.3
+langchain-google-genai>=2.1.8,<3.0
+langchain-nvidia-ai-endpoints>=0.3.8,<0.4
+langchain-ollama>=0.3.4,<0.4
```
d1donlydfink, line 9: Worth noting that we now include all the langchain_* LLM vendor libs in the requirements-build.txt so that our unit tests can still probe each of these. But since these are not part of the regular requirements.txt, they are not forced to be installed with every neuro-san distribution.

deepsaia (approved):
LGTM.

requirements.txt
ofrancon: Nice

PR #387, opened 2025-09-10
Add ChatMockLlm and hocon file for running it

Noravee:


neuro_san/test/chat_mock_llm/chat_mock_llm.py
```diff
@@ -0,0 +1,194 @@
...
+                "input_tokens": input_tokens,
+                "output_tokens": input_tokens,
+                "total_tokens": 2*input_tokens,
+            },
+        )
+
+        generation = ChatGeneration(message=message)
+        return ChatResult(generations=[generation])
+
+    def _stream(
```
Noravee: This is the method required to run the agent executor.

neuro_san/test/chat_mock_llm/chat_mock_llm.py
```diff
@@ -0,0 +1,194 @@
...
+            )
+        )
+        if run_manager:
+            # This is optional in newer versions of LangChain
+            # The on_llm_new_token will be called automatically
+            run_manager.on_llm_new_token(content, chunk=chunk)
+
+        yield chunk
+
+    def _num_tokens_from_string(self, string: str, encoding_name: str = "o200k_base") -> int:
```
Noravee: Using the same model for token calucation as gpt-4o.

neuro_san/registries/echo.hocon
```diff
@@ -0,0 +1,53 @@
...
+        # Some disqualifications from being a front man:
+        #   1) Cannot use a CodedTool "class" definition
+        #   2) Cannot use a Tool "toolbox" definition
+        #
+        # Besides the first agent being the front man, these tool definitions
+        # do not have to be in any particular order. How they are linked and
+        # call each other is defined within their own specs.
+        # This could be a graph, potentially even with cycles.
+        {
+            "name": "echoer",
```
Noravee: HOCON file with only the frontman to use with the mock llm.

neuro_san/registries/echo.hocon
```diff
@@ -0,0 +1,53 @@
...
+# https://github.com/cognizant-ai-lab/neuro-san/blob/main/docs/agent_hocon_reference.md
+
+{
+    # Optional metadata describing this agent network
+    "metadata": {
+        "description": "Testing neuro-san infrastructure by using a mock llm.",
+        "tags": ["test"],
+    },
+    "llm_config": {
+        "class": "neuro_san.test.chat_mock_llm.chat_mock_llm.ChatMockLlm",
```
d1donlydfink: This ends up being a decent example of "Bring your own LLM" functionality. Nice.

neuro_san/test/chat_mock_llm/chat_mock_llm.py
```diff
@@ -0,0 +1,194 @@
+
```
d1donlydfink: Nit: Let's call the directory this is in:

neuro_san/test/llms

This way if there are other test or mock LLMs to be created, they are all in the same place.
If they start to get complicated we can create subdirectories off of that.

d1donlydfink: On the other hand, it is conceivable that this class could eventually go into something like langchain_community. Most certainly don't have to go there yet.

neuro_san/test/chat_mock_llm/chat_mock_llm.py
d1donlydfink: Would it be better to use an asynchronous version of this like _astream() given our environment?

neuro_san/test/chat_mock_llm/chat_mock_llm.py
```diff
@@ -0,0 +1,194 @@
...
+    # This is required field and it is possible to have multiple test models.
+    model_name: str = Field(default=None, alias="model")
+    # Maybe useful for testing
+    max_retries: Optional[int] = None
+
+    # Accept both argument name and alias
+    model_config = ConfigDict(populate_by_name=True)
+
+    def _generate(
```
d1donlydfink: Would it be better to use an asynchronous version of this like _agenerate() given our async environment?

d1donlydfink (changes requested):
* Please rename directory to be neuro_san/test/llms
* Given our asynchronous environment, should we be (also/instead?) be implementing _agenerate() and _astream()?

neuro_san/registries/echo.hocon
```diff
@@ -0,0 +1,53 @@
...
+# neuro-san SDK Software in commercial settings.
+#
+# END COPYRIGHT
+
+# The schema specifications for this file are documented here:
+# https://github.com/cognizant-ai-lab/neuro-san/blob/main/docs/agent_hocon_reference.md
+
+{
+    # Optional metadata describing this agent network
+    "metadata": {
```
vince-leaf: Maybe rename the echo.hocon to ChatMockLLM_echo.hocon

neuro_san/test/chat_mock_llm/chat_mock_llm.py
andreidenissov-cog: Silly question: how will it be "hosted"? I mean this model.
Will it be just a class instance inside running neuro-san service?

neuro_san/registries/echo.hocon
Noravee: Done.

neuro_san/test/chat_mock_llm/chat_mock_llm.py
Noravee: Done.

neuro_san/test/chat_mock_llm/chat_mock_llm.py
Noravee: The agent executor actually uses `_astream()` when calling `ainvoke()` method.
 
The `_astream()` implementation uses `run_in_executor()` to launch the sync `_stream()` in a separate thread if `_stream()` is implemented. Thus, it is not required to override the async method, especially when the code is not natively async.

d1donlydfink (approved):

PR #388, opened 2025-09-10
UN-3499: user input contains metadata

Noravee:
This is because the user message was put into `HumanMessage` in `langchain_run_context.submit_message()`. However, the prompt template is
```python
[
("system", instructions))
("placeholder", "{chat_history}"),
("human", "{input}"), # This is equivalent to HumanMessage(content={input})
("placeholder", "{agent_scratchpad}"),
]
```

so the input becomes HumanMessage(content=HumanMessage(content=<real_input>, metadata...)).

### DIFF
--- a/neuro_san/registries/chat_mock_llm_echo.hocon
+++ b/neuro_san/registries/chat_mock_llm_echo.hocon
@@ -1,0 +1,54 @@
+
+# Copyright (C) 2023-2025 Cognizant Digital Business, Evolutionary AI.
+# All Rights Reserved.
+# Issued under the Academic Public License.
+#
+# You can be released from the terms, and requirements of the Academic Public
+# License by purchasing a commercial license.
+# Purchase of a commercial license is mandatory for any use of the
+# neuro-san SDK Software in commercial settings.
+#
+# END COPYRIGHT
+
+# The schema specifications for this file are documented here:
+# https://github.com/cognizant-ai-lab/neuro-san/blob/main/docs/agent_hocon_reference.md
+
+{
+    # Optional metadata describing this agent network
+    "metadata": {
+        "description": "Testing neuro-san infrastructure by using a mock llm.",
+        "tags": ["test"],
+    },
+    "llm_config": {
+        "class": "neuro_san.test.llms.chat_mock_llm.ChatMockLlm",
+        # As of Oct 10, 2025 the model name can be anything since the model only echoes.
+        "model_name": "echo",
+    },
+    "tools": [
+        # This first agent definition is regarded as the "Front Man", which
+        # does all the talking to the outside world/client.
+        #
+        # Some disqualifications from being a front man:
+        #   1) Cannot use a CodedTool "class" definition
+        #   2) Cannot use a Tool "toolbox" definition
+        #
+        # Besides the first agent being the front man, these tool definitions
+        # do not have to be in any particular order. How they are linked and
+        # call each other is defined within their own specs.
+        # This could be a graph, potentially even with cycles.
+        {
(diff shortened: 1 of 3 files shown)


neuro_san/internals/run_context/langchain/core/langchain_run_context.py
```diff
@@ -393,7 +393,7 @@ async def submit_message(self, user_message: str) -> Run:
         # Contruct a human message out of the text of the user message
         # Don't add this to the chat history yet.
         try:
-            self.recent_human_message = HumanMessage(user_message)
+            self.recent_human_message = user_message
```
Noravee: Do not put the user message in the human message.

d1donlydfink:
I guess I don't fully understand the problem that this resolves.
Can you maybe give a more concrete example of how this was a problem?

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
d1donlydfink: So, the typing of LangChainRunContext.recent_human_message in the contstructor is actually HumanMessage.
I can see that this is used as an input to a dictionary field for "input" that is sent to the token_counter in wait_on_run(), but in another place in wait_on_run(), recent_human_message is passed to the journal.write_message() call, which definitely wants a BaseMessage and not a string.

So maybe you are looking to simplify something in the token counter? If so, maybe such a switch belongs at that level?

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
```diff
@@ -440,7 +440,7 @@ async def wait_on_run(self, run: Run, journal: Journal = None) -> Run:
 
         inputs = {
             "chat_history": previous_chat_history,
-            "input": self.recent_human_message
+            "input": self.recent_human_message.content
```
Noravee, line 443: Just put the content of human message which is string into input so there is no need to change the data type of `recent_human_message`.

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
Noravee: Thank you for your feedback.

Please let me explain more in details.

As the prompt template is written as

```python
[
    ("system", instructions)),
    ("placeholder", "{chat_history}"),
    ("human", "{input}"),
    ("placeholder", "{agent_scratchpad}"),
]
```

When we call `ainvoke({"input": <str>})`, the LLM receives:

- `HumanMessage(content=<str>, response_metadata={}, usage_metadata={})` as the message

- `<str>` as the content

But in the LangChain run context, when we call `ainvoke({"input": HumanMessage(content=<str>)})`, the LLM instead receives:

- `HumanMessage(content=HumanMessage(content=<str>, ...))` as the message

- `content=<str>, response_metadata={}, usage_metadata={}` as the content

While the functional impact is minimal (the LLM still sees the user’s string, just with some unnecessary wrapping), I think this makes things less clear.

I agree with your point that changing `LangChainRunContext.recent_human_message` directly could have broader side effects.

My proposal:

Keep `LangChainRunContext.recent_human_message` as a HumanMessage (unchanged)

But, when building the dictionary input, only use the `content` attribute of that `HumanMessage` for `"input"`.

This way, we maintain clarity without introducing unintended changes elsewhere.

Please let me know what you think.

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
d1donlydfink: This makes more sense.

PR #390, opened 2025-09-11
UN-2876 Activation Simplifications towards better intra-agent JSON

d1donlydfink:
Checkpoint on the way towards saner intra-agent JSON.
This one becomes a little risky as it contains some subtle logical changes deep in the bowels of neuro-san:

* Change the return value type of Activation.build() to be List[BaseMessage] instead of a string

   This removes one layer of the wonky "role" structures that are reported between agents.
   This in turn allows for less string conversion to happen inside the activations as they are merely reporting their raw data.
   This can likely eventually be reduced to a single BaseMessage, but baby steps for stability.

* Specifically in BranchActivation we are now returning directly what came back from integrate_callable_response().

   This means we can get rid of the evil IntraAgentMessageUtils, as this was the only place it was being called.

* Only ever use the last message from anything returned by build()

   This still goes through some string conversion in how tool results are reported (to be addressed in next layer PR), but my investigations tell me this is OK to do.

Tested:
* esp_decision_assistant with the chicken scenario
* intranet_agent with the marriage scenario (since it uses early AAOSA)
* math_guy_passthrough

All seem fine.
If anyone sees issues after this goes in, especially with AAOSA networks, I'll be interested to know.


neuro_san/internals/graph/activations/branch_activation.py
```diff
@@ -143,8 +141,7 @@ async def build(self) -> str:
 
         messages = await self.integrate_callable_response(run, messages)
 
-        response: str = IntraAgentMessageUtils.generate_response(messages)
-        return response
+        return messages
```
d1donlydfink, line 144: Just return the list of messages from integrate_callable_response which are all already BaseMessges.
With this as the interface change, now we can get rid of the evil IntraAgentMessageUtils.

neuro_san/internals/graph/activations/branch_activation.py
```diff
@@ -176,10 +173,9 @@ async def use_tool(self, tool_name: str, tool_args: Dict[str, Any], sly_data: Di
                                                                                        tool_name,
                                                                                        sly_data,
                                                                                        tool_args)
-        message: str = await callable_activation.build()
+        message_list: List[BaseMessage] = await callable_activation.build()
 
         # We got a list of messages back as a string. Take the last.
-        message_list: List[Dict[str, Any]] = json.loads(message)
```
d1donlydfink, line 182: Skip a conversion from BaseMessage -> txt in BranchActivation.  Just use what came back from the build() call above.

neuro_san/internals/graph/activations/calling_activation.py
```diff
@@ -221,7 +223,16 @@ async def make_one_tool_function_call(self, component_tool_call: ToolCall) -> Di
...
-        output: str = await callable_component.build()
+        message_list: List[BaseMessage] = await callable_component.build()
+
+        # We got a list of messages back as a string. Take the last.
+        last_message: BaseMessage = message_list[-1]
+        message_list = [{
+            "role": "assistant",
+            "content": last_message.content
+        }]
+        output: str = json.dumps(message_list)
```
d1donlydfink, line 234: This is yet another place where message lists get turned into JSON which can be removed eventually.
There is more work to remove this here and in LangChainRunContext.parse_tool_output(),
but I really want to go slow to be sure I am not breaking anything.  Hence this checkpoint PR.

neuro_san/internals/graph/activations/external_activation.py
```diff
@@ -186,14 +189,10 @@ async def build(self) -> str:
         # Eventually we will care about a fuller chat history.
 
         # Prepare the output
-        message: Dict[str, Any] = {
-            "role": "assistant",
-            "content": answer
-        }
-        message_list.append(message)
+        ai_message = AIMessage(content=answer)
+        message_list.append(ai_message)
```
d1donlydfink, line 193: Use an AIMessage for the answer to return from build instead of the wonky intermediate "role" structure business.

neuro_san/internals/graph/activations/intra_agent_message_utils.py
```diff
@@ -1,128 +0,0 @@
-
```
d1donlydfink, line 1: Ding, dong! The witch is dead!

neuro_san/internals/graph/activations/abstract_class_activation.py
```diff
@@ -100,13 +101,13 @@ class name (e.g., "my_module.MyToolClass") of the CodedTool that should
         raise NotImplementedError
 
     # pylint: disable=too-many-locals
-    async def build(self) -> str:
+    async def build(self) -> List[BaseMessage]:
```
d1donlydfink, line 104: Make a stronger claim as to what build() returns now - a List of BaseMessages.
While there is more work to do for a future PR, this begins to allow less JSON <-> text conversion.

Likely: This can probably be reduced to a single BaseMessage instead of a list when the time is right.

neuro_san/internals/graph/activations/abstract_class_activation.py
```diff
@@ -206,14 +207,10 @@ async def build(self) -> str:
 
         # Change the result into a message
         retval_str: str = f"{retval}"
-        message: Dict[str, Any] = {
-            "role": "assistant",
-            "content": retval_str
-        }
+        message = AIMessage(content=retval_str)
```
d1donlydfink, line 210: Use an AIMessage as the BaseMessage to return instead of the wonky intermediate "role" structure list
with string conversion of that.

neuro_san/internals/graph/activations/calling_activation.py
```diff
@@ -221,7 +223,16 @@ async def make_one_tool_function_call(self, component_tool_call: ToolCall) -> Di
             self.factory.create_agent_activation(self.run_context, our_agent_spec, use_tool_name,
                                                  self.sly_data, tool_arguments)
 
-        output: str = await callable_component.build()
+        message_list: List[BaseMessage] = await callable_component.build()
+
+        # We got a list of messages back as a string. Take the last.
+        last_message: BaseMessage = message_list[-1]
```
d1donlydfink, line 229: Only use the last message of what was returned by build().
This is a bit risky, but if my investigations are telling me what I think they are, this is the only message used at the consumer end anyway.

We are still making a string here (see below), but now at least it can be smaller.
Future PR will remove that entirely.

andreidenissov-cog (approved):
AFAICT, looks OK.

Noravee (approved):
LGTM.

PR #391, opened 2025-09-12
Fix some pytest warnings for smoother sailing

d1donlydfink:
There were a couple of warnings that have crept into the running of pytest over time.
Fixing one of them about asyncio might actually lead to more stable runs / less intermittent failures.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -17,4 +17,4 @@ filterwarnings =
     # Ignore warnings about protobuf 4's PyType_Spec metaclass usage
     ignore:Type google\._upb\._message.* uses PyType_Spec with a metaclass that has custom tp_new:DeprecationWarning
 
-ayncio_default_fixture_loop_scope = function
+asyncio_default_fixture_loop_scope = function


tests/neuro_san/coded_tools/music_nerd_pro/test_accountant.py
```diff
@@ -9,14 +9,14 @@
...
-from unittest import TestCase
+from unittest import IsolatedAsyncioTestCase
 
 import pytest
 
 from neuro_san.coded_tools.accountant import Accountant
 
 
-class TestAccountant(TestCase):
+class TestAccountant(IsolatedAsyncioTestCase):
```
d1donlydfink, line 19: A separate warning said:

"DeprecationWarning: It is deprecated to return a value that is not None from a test case"

Looking at the docs for pytest-asyncio mentioned that we should be deriving async tests from this base class, and that indeed fixes the warning.

pytest.ini
```diff
@@ -16,3 +16,5 @@ markers =
 filterwarnings =
     # Ignore warnings about protobuf 4's PyType_Spec metaclass usage
     ignore:Type google\._upb\._message.* uses PyType_Spec with a metaclass that has custom tp_new:DeprecationWarning
+
+asyncio_default_fixture_loop_scope = function
```
d1donlydfink, line 20: One warning in the build output said we should be specifying this.
Google says this "function" a reasonable setting. Seems to clear the warning anyway.

It's possible this might also fix some intermittent failures as well.

PR #392, opened 2025-09-15
Make math_guy unit test more reliable

d1donlydfink:
Existing prompting on math_guy has started to fail intermittently in unit tests.
Most of the time the LLM was worried that it wasn't receiving any input and asking for it, when the input was actually coming from the sly_data.  This expectation of further input is what was making the unit tests fail because they only expected a single interaction.

This change makes the prompting less vague and asks the LLM to trust that the calculator tool will get its inputs by other means.  It's less anxious now and so far I haven't seen a failure.

### DIFF
--- a/neuro_san/registries/math_guy.hocon
+++ b/neuro_san/registries/math_guy.hocon
@@ -84,15 +84,19 @@ on the secret numbers.
             },
 
             "instructions": """
-Always call the calculator tool to populate the return sly_data correctly.
-Return verbatim whatever it returns.
+You are an interface for a calculator tool.
+Always call the real_calculator tool to perform the arithmetic calculation,
+but know that the operand values will always be hidden from you. Do not ask
+for them. Instead trust that the real_calculator tool will get the operands
+by some other means, and just pass the name of the operator to the real_calculator tool.
+Return verbatim whatever real_calculator returns.
 """,
             "allow": {
                 "to_upstream": {
                     "sly_data": ["equals"]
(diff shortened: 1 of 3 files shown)


neuro_san/registries/math_guy.hocon
```diff
@@ -84,15 +84,19 @@ on the secret numbers.
...
 
             "instructions": """
-Always call the calculator tool to populate the return sly_data correctly.
-Return verbatim whatever it returns.
+You are an interface for a calculator tool.
+Always call the real_calculator tool to perform the arithmetic calculation,
+but know that the operand values will always be hidden from you. Do not ask
+for them. Instead trust that the real_calculator tool will get the operands
+by some other means, and just pass the name of the operator to the real_calculator tool.
+Return verbatim whatever real_calculator returns.
```
d1donlydfink, line 92: Be specific about the fact that the front-man will not know the operands.
Also: Change the name of the tool to real_calculator so the prompt text will not be ambiguous in its usage of the term "calculator".

tests/fixtures/math_guy/basic_sly_data.hocon
```diff
@@ -22,7 +22,7 @@
     "interactions": [
         {
             # This is what we send as input to streaming_chat()
-            "text": "times",
+            "text": "multiply",
```
d1donlydfink, line 25: Be more specific in the unit tests' input.

PR #399, opened 2025-09-17
UN-3477 Run-time resources for models.

andreidenissov-cog:


neuro_san/internals/interfaces/context_type_llm_factory.py
```diff
@@ -48,12 +50,12 @@ def load(self):
...
-    def create_llm(self, config: Dict[str, Any]) -> Any:
+    def create_llm(self, config: Dict[str, Any]) -> LangChainLlmResources:
         """
-        Create an llm instance BaseLanguageModel from the fully-specified llm config.
+        Create a llm instance BaseLanguageModel from the fully-specified llm config.
         :param config: The fully specified llm config from which the LLM instance
                     should be created.
-        :return: An llm instance native to the context type.
+        :return: An LangChainLlmResources instance containing a model native to the context type.
                 Can raise a ValueError if the config's class or model_name value is
```
andreidenissov-cog, line 57: Change of interface - create_llm() now returns LangChainLlmResources = model + run-time resources.

neuro_san/internals/run_context/langchain/llms/langchain_llm_resources.py
```diff
@@ -0,0 +1,50 @@
...
+from langchain_core.language_models.base import BaseLanguageModel
+
+
+class LangChainLlmResources:
+    """
+    Class for representing a LangChain model
+    together with run-time resources necessary for model usage by the service.
+    """
+
+    def __init__(self, model: BaseLanguageModel, http_client: httpx.AsyncClient = None):
```
andreidenissov-cog: For now, we just pair model itself as BaseLanguageModel and httpx client we create ourselves
which will be used by the model as LLM host client.
Lifecycle of http client will be managed by neuro-san service directly through release_resources() method.

neuro_san/session/async_direct_agent_session.py
```diff
@@ -173,18 +174,21 @@ async def streaming_chat(self, request_dict: Dict[str, Any]) -> Generator[Dict[s
...
+                    # We expect the message to be a dictionary form of chat.ChatMessage
+                    if message_processor is not None:
+                        message_type: ChatMessageType = message.get("type")
+                        # Can modify message
+                        await message_processor.async_process_message(message, message_type)
+                    response_dict["response"] = message
+                    yield response_dict
+        finally:
+            with contextlib.suppress(Exception):
+                await chat_session.delete_resources()
```
andreidenissov-cog, line 199: Change of logic:
1. We tell underlying chat_session to delete its resources since we are done with this request;
2. We do this in "finally" block so this releasing of resources happens in any case, including getting our async generator (which is implicitly constructed by lines 179-188 and returned by this method) interrupted by caller-side "aclose" method.
3. And we suppress all exceptions while deleting resources to keep things quieter.

neuro_san/session/direct_agent_session.py
```diff
@@ -187,17 +188,20 @@ def streaming_chat(self, request_dict: Dict[str, Any]) -> Generator[Dict[str, An
...
+                    # We expect the message to be a dictionary form of chat.ChatMessage
+                    if message_processor is not None:
+                        message_type: ChatMessageType = message.get("type")
+                        # Can modify message
+                        message_processor.process_message(message, message_type)
+                    response_dict["response"] = message
+                    yield response_dict
+        finally:
+            with contextlib.suppress(Exception):
+                asyncio.run(chat_session.delete_resources())
```
andreidenissov-cog, line 215: Duplication of logic from async_direct_agent_session.py for the case of sync iterators.

tests/neuro_san/internals/run_context/langchain/llms/test_llm_factory.py
```diff
@@ -51,13 +52,15 @@ class TestLlmFactory(LangChainLlmFactory):
     """
 
     def create_base_chat_model(self, config: Dict[str, Any],
-                               callbacks: List[BaseCallbackHandler] = None) -> BaseLanguageModel:
+                               callbacks: List[BaseCallbackHandler] = None) -> LangChainLlmResources:
         """
```
andreidenissov-cog: Comply to API change.

tests/neuro_san/internals/run_context/langchain/llms/test_llm_factory.py
```diff
@@ -106,4 +109,4 @@ def create_base_chat_model(self, config: Dict[str, Any],
         else:
             raise ValueError(f"Class {chat_class} for model_name {model_name} is unrecognized.")
 
-        return llm
+        return LangChainLlmResources(llm, http_client=None)
```
andreidenissov-cog: But here we don't really care about managing run-time http client to LLM host.

neuro_san/internals/run_context/langchain/llms/langchain_llm_factory.py
```diff
@@ -45,12 +45,14 @@ class LangChainLlmFactory:
                                     the model description in this class.
     """
 
-    def create_base_chat_model(self, config: Dict[str, Any]) -> BaseLanguageModel:
+    def create_base_chat_model(self, config: Dict[str, Any]) -> LangChainLlmResources:
```
andreidenissov-cog: API change.

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
```diff
@@ -266,25 +269,27 @@ def get_chat_class_args(self, chat_class_name: str, use_model_name: str = None)
...
                     _create_full_llm_config() above.
-        :return: A BaseLanguageModel (can be Chat or LLM)
+        :return: A LangChainLlmResources instance containing
+                a BaseLanguageModel (can be Chat or LLM) and all related resources
+                necessary for managing the model run-time lifecycle.
                 Can raise a ValueError if the config's class or model_name value is
                 unknown to this method.
         """
-        llm: BaseLanguageModel = None
+        llm_resources: LangChainLlmResources = None
```
andreidenissov-cog, line 298: Switch to new API for LLM models construction.

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
```diff
@@ -315,17 +320,18 @@ def create_base_chat_model(self, config: Dict[str, Any]) -> BaseLanguageModel:
...
         if (
-            llm is None
+            llm_resources is None
             and found_exception is not None
             and class_path not in default_llm_classes
         ):
-            llm = self.create_base_chat_model_from_user_class(class_path, config)
+            llm: BaseLanguageModel = self.create_base_chat_model_from_user_class(class_path, config)
+            llm_resources = LangChainLlmResources(llm)
             found_exception = None
```
andreidenissov-cog, line 361: And here is the potential problem, because just instantiating a model instance from a class will not give us access to run-time model resources we'll need.
Everything should still work, but no guarantees for proper resources management by a model itself == possible resources leaks.
We'll need to address this thing later.

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
```diff
@@ -108,7 +109,7 @@ def __init__(self, llm_config: Dict[str, Any],
         """
         self.chat_history: List[BaseMessage] = []
         self.journal: OriginatingJournal = None
-        self.llm: BaseLanguageModel = None
+        self.llm_resources: LangChainLlmResources = None
         self.agent: Agent = None
```
andreidenissov-cog, line 115: Use LangChainLlmResources now instead of just BaseLanguageModel.

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
```diff
@@ -122,6 +123,8 @@ def __init__(self, llm_config: Dict[str, Any],
         self.invocation_context: InvocationContext = invocation_context
         self.chat_context: Dict[str, Any] = chat_context
         self.origin: List[Dict[str, Any]] = []
+        # Have we already created resources for this RunContext:
+        self.resources_created: bool = False
         # Default logger
```
andreidenissov-cog, line 130: The new flag to guard against creating resources twice for the same RunContext - which is happening now in the service.

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
```diff
@@ -217,14 +226,14 @@ def create_agent_with_fallbacks(self, prompt_template: ChatPromptTemplate) -> Ag
         for index, fallback in enumerate(fallbacks):
 
             # Create a model we might use.
-            one_llm: BaseLanguageModel = llm_factory.create_llm(fallback)
-            one_agent: Agent = self.create_agent(prompt_template, one_llm)
+            one_llm: LangChainLlmResources = llm_factory.create_llm(fallback)
+            one_agent: Agent = self.create_agent(prompt_template, one_llm.get_model())
```
andreidenissov-cog, line 233: create_llm() now returns an instance of LangChainLlmResources.

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
```diff
@@ -653,11 +663,16 @@ async def delete_resources(self, parent_run_context: RunContext = None):
         :param parent_run_context: A parent RunContext perhaps the same instance,
                         but perhaps not.  Default is None
         """
+
+        # Release model related resources:
+        if self.llm_resources:
+            await self.llm_resources.release_resources()
+
```
andreidenissov-cog, line 826: This is where we actually release model-related resources when we clean up our RunContext.

neuro_san/internals/run_context/langchain/llms/standard_langchain_llm_factory.py
```diff
@@ -81,29 +88,38 @@ def create_base_chat_model(self, config: Dict[str, Any]) -> BaseLanguageModel:
             ChatOpenAI = resolver.resolve_class_in_module("ChatOpenAI",
                                                           module_name="langchain_openai.chat_models.base",
                                                           install_if_missing="langchain-openai")
+
```
andreidenissov-cog: So this is how we need to do things if we want to control model's http client connecting to LLM host:
we create http client ourselves, construct AsyncOpenAI client using this http client in its constructor,
and only then construct ChatOpenAI client using AsyncOpenAI client as a "base client".
Due to peculiarities of OpenAI SDK, we cannot just "inject" desired http client into say ChatOpenAI instance,
we have to build things from the bottom up.

neuro_san/internals/run_context/langchain/llms/standard_langchain_llm_factory.py
```diff
@@ -81,29 +88,38 @@ def create_base_chat_model(self, config: Dict[str, Any]) -> BaseLanguageModel:
...
+            async_openai_client: AsyncOpenAI = AsyncOpenAI(
+                api_key=self.get_value_or_env(config, "openai_api_key",
+                                              "OPENAI_API_KEY"),
+                base_url=self.get_value_or_env(config, "openai_api_base", "OPENAI_API_BASE"),
+                organization=self.get_value_or_env(config, "openai_organization", "OPENAI_ORG_ID"),
+                timeout=request_timeout,
+                max_retries=config.get("max_retries"),
+                http_client=http_client
+            )
+
```
andreidenissov-cog: Note that all constructor parameters of ChatOpenAI are now distributed between http_client,
AsyncOpenAI and ChatOpenAI. According of course to GPT recommendations.

neuro_san/internals/run_context/langchain/llms/standard_langchain_llm_factory.py
```diff
@@ -406,4 +422,4 @@ def create_base_chat_model(self, config: Dict[str, Any]) -> BaseLanguageModel:
         else:
             raise ValueError(f"Class {chat_class} for model_name {model_name} is unrecognized.")
 
-        return llm
+        return LangChainLlmResources(llm, http_client=http_client)
```
andreidenissov-cog: And finally, construct instance of LangChainLlmResources to return.
Note that http_client could be None here. In this case actual http connections release may not happen
when we are done with request.

neuro_san/internals/run_context/langchain/llms/standard_langchain_llm_factory.py
andreidenissov-cog: Another gotcha of OpenAI SDK: given an instance of ChatOpenAI, we _cannot_ get its http client,
that's why we have to combine them in LangChainLlmResources class.

neuro_san/internals/interfaces/context_type_llm_factory.py
```diff
@@ -48,12 +50,12 @@ def load(self):
         """
         raise NotImplementedError
 
-    def create_llm(self, config: Dict[str, Any]) -> Any:
+    def create_llm(self, config: Dict[str, Any]) -> LangChainLlmResources:
```
d1donlydfink: Hmm. Complicated here.
This should probably remain Any as a return type. Here is the history:

Once upon a time, not too long ago we maintained two different implementations for neuro-san's runtime API infrastructure.  The aboriginal version used OpenAI Assistants. The one everyone uses now is based on langchain.  Maybe someday we might have something based on native langgraph (for instance).

This notion of "ContextType" is the acknowledgement of that implementation split, and as such we have ContextTypeLlmFactory and ContextTypeToolFactory in a couple of different high-level places throughout the code. The big idea is that any implementation has its own kind of RunContext and all the other code should be able to deal with that abstractly.

So, what you have here is an API change that turns something very abstract and API-agnostic into something very much tied down to one of these APIs

Short of the long: The should remain Any, and hopefully now you can begin to see why.

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
d1donlydfink: Interesting!  Any info as to the paths that enable this?

neuro_san/internals/run_context/langchain/llms/langchain_llm_factory.py
d1donlydfink: OK, here is the most troubling rub:

You are changing an externally facing API that we encourage people to derive their own stuff from.
See the docs here: 
https://github.com/cognizant-ai-lab/neuro-san/blob/main/docs/llm_info_hocon_reference.md#factories

Given the necessity and how you are solving the problem below in StandardLangChainLlmFactory, we'll need to think very carefully about how to set this up before this goes in.

---

 FWIW: This all said, Here now that we are in langchain territory, it would be OK to more strongly type to the langchain-oriented type. Just not at the uber-abstract ContextTypeLlmFactory above.

neuro_san/internals/run_context/langchain/llms/langchain_llm_resources.py
d1donlydfink: Very interesting indeed!

neuro_san/session/async_direct_agent_session.py
d1donlydfink: Guts of this comment should be in the code, not in ephemeral PR.

Very interesting indeed.
This makes me wonder if this would end up being a different way of addressing the problem of lingering httpx connections other than the AsyncioExecutor pooling solution you had come up with previously?

What's nice about this though is that the cleanup is delegated to ContextType-agnostic delete_resources(). 
That is a very good thing.

neuro_san/session/direct_agent_session.py
d1donlydfink: Similar comment should go in the code here.

tests/neuro_san/internals/run_context/langchain/llms/test_llm_factory.py
d1donlydfink: See what I mean about API breakage for extensions ?

Ideally we shouldn't have to require extensions to make this kind of change unless they deeply cared about being able to abort requests, which perhaps should be optional.   Which perhaps means ideally in the end-state of this PR, this file should not have to change.

neuro_san/internals/run_context/langchain/llms/standard_langchain_llm_factory.py
d1donlydfink: Before going forward with this PR, it would be a good idea to see if this concept also translates to Anthropic ChatModels as well.    There is a small possibility that changes in langchain 1.0 might help here, but that is more a prayer stemming from current ignorance.

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
```diff
@@ -217,14 +226,14 @@ def create_agent_with_fallbacks(self, prompt_template: ChatPromptTemplate) -> Ag
...
-            one_agent: Agent = self.create_agent(prompt_template, one_llm)
+            one_llm: LangChainLlmResources = llm_factory.create_llm(fallback)
+            one_agent: Agent = self.create_agent(prompt_template, one_llm.get_model())
 
             if index == 0:
                 # The first agent is the one we want to be our main guy.
                 agent = one_agent
                 # For now. Could be problems with different providers w/ token counting.
-                self.llm = one_llm
+                self.llm_resources = one_llm
```
d1donlydfink: This is likely not going to be quite right with fallbacks.

The concept of fallbacks is a langchain thing and here we are adopting it whole-cloth.
The idea is that you have a list of models and you try to use the first one, but if that one fails,
then try to use the next one.  One or two in the list could be from OpenAI, another in the list could be from Anthropic.  In most demo cases there is only one of these to worry about, however as people move towards production they will be more concerned about uptime and gravitate toward these fallbacks.

I think instead what you want to do here for correctness is store a list of llm_resources so each of those can be closed out in delete_resources() properly given how you are taking control over the client.

This presents its own puzzle. Let's say we have a primary of OpenAI and a fallback of Anthropic.
It feels like in this scheme we would have to at least open connections to both, though we are not likely to actually use the connection to the Anthropic fallback in this case.

Thinking about this, the fallback notion could contribute to the reasons as to why you can't simply just get the client from the langchain entity itself - it could likely be lazily instantiated.

neuro_san/internals/run_context/langchain/llms/standard_langchain_llm_factory.py
d1donlydfink: Another thought:  Wasn't langchain pooling these clients?
Seems like with this scheme we are opening a new one for each LLM.
This could easily slow things down.
Do we now need to pool at our level?
Do we need to lazily instantiate (somehow) the connection as well?

neuro_san/internals/run_context/langchain/llms/langchain_llm_resources.py
```diff
@@ -0,0 +1,50 @@
...
+    def get_http_client(self) -> httpx.AsyncClient:
+        """
+        Get the http client used by the model
+        """
+        return self.http_client
+
+    async def release_resources(self):
+        """
+        Release the run-time resources used by the model
+        """
```
d1donlydfink, line 48: In other places we use delete_resources(), let's not create another convention here.

neuro_san/internals/run_context/langchain/llms/langchain_llm_resources.py
```diff
@@ -0,0 +1,50 @@
...
+        Get the language model
+        """
+        return self.model
+
+    def get_http_client(self) -> httpx.AsyncClient:
+        """
+        Get the http client used by the model
+        """
+        return self.http_client
```
d1donlydfink: It's possible that we might have other kinds of clients with other kinds of llms.
There might be yet another layer of abstraction to provide here rather than going directly adhering to httpx.
If this is going to be part of a public API, we need to think about that kind of step before we take it.

d1donlydfink (changes requested):
Further fantastic sleuthing and very interesting direction on solving very tricky problem.
I can now see why you were in deep for so long.

Before this goes in we need to think even more deeply about the approach w/rt API.

Big Concerns, not necessarily with answers yet:

* API breakage at externally facing LangChainLlmFactory interface
  Ideally TestLlmFactory should not change. ✅ 
* Do we need to do our own pooling/lazy connection of httpx.AsyncClients?
* Should there be a connection client abstraction as well so that llm libs that use something
  other than httpx can do the equivalent of aclose()?  ✅ 

Must:
* Change return type of ContextTypeLlmFactory to Any ✅ 
* Change release_resources() to delete_resources() on LangChainLlmResources ✅ 
* delete_resources() of all LlmResources in RunContext for fallbacks

Please:
* See if the LlmResources concept translates to Anthropic as well.
  If it does, it's a good sign your new construct abstracts going forward.  ✅ 
* Rename one_llm variable in LangChainRunContext.create_agent_with_fallbacks() to one_llm_resources ✅ 
* Move salient PR comments to code in the Sessions.

Will need to file tickets to be sure LlmClient concept propagates beyond just OpenAI.

We can discuss the Big Concerns out-of-band of this PR I think, otherwise the comments will overwhelm the changes.

neuro_san/internals/run_context/langchain/llms/langchain_llm_resources.py
```diff
@@ -0,0 +1,50 @@
...
+
+    async def release_resources(self):
+        """
+        Release the run-time resources used by the model
+        """
+        if self.http_client:
+            with contextlib.suppress(Exception):
+                await self.http_client.aclose()
```
d1donlydfink: Local LLMs like from Ollama models might not have a connecting client to close, so it's good you check for None

neuro_san/internals/run_context/langchain/llms/standard_langchain_llm_factory.py
```diff
@@ -13,11 +12,15 @@
 from typing import Any
 from typing import Dict
 
+import httpx
+from openai import AsyncOpenAI
```
d1donlydfink: This should be lazily resolved within the appropriate case.  We are trying to make it so that we are not required to import the world

neuro_san/session/async_direct_agent_session.py
andreidenissov-cog: OK. Comments in code.

neuro_san/session/direct_agent_session.py
andreidenissov-cog: Comments in code.

tests/neuro_san/internals/run_context/langchain/llms/test_llm_factory.py
d1donlydfink: Resolved with optional new method on LlmFactory

neuro_san/internals/run_context/langchain/llms/langchain_llm_factory.py
d1donlydfink: Resolved

neuro_san/internals/run_context/langchain/llms/standard_langchain_llm_factory.py
d1donlydfink: This is more structured now with LlmPolicy

neuro_san/internals/run_context/langchain/llms/standard_langchain_llm_factory.py
d1donlydfink: Someday you'll read and answer all of my questions <sigh>

PR #400, opened 2025-09-17
UN-3493 use asyncio.to_thread() to run CodedTool.invoke() synchronous code in separate thread

d1donlydfink:
Use asyncio.to_thread() when calling any CodedTool's synchronous invoke() method.
The idea is to allow Easy development of CodedTools without giving people rope to hang themselves with w/rt blocking the asyncio event loop.

Also update the CodedTool interface docs to reflect the change.

Special shout out to @Noravee for finding this nicety.

### DIFF
--- a/neuro_san/coded_tools/math_guy/calculator.py
+++ b/neuro_san/coded_tools/math_guy/calculator.py
@@ -32,9 +32,9 @@ class Calculator(CodedTool):
         """
         Called when the coded tool is invoked asynchronously by the agent hierarchy.
         Strongly consider overriding this method instead of the "easier" synchronous
-        version above when the possibility of making any kind of call that could block
+        invoke() version above when the possibility of making any kind of call that could block
         (like sleep() or a socket read/write out to a web service) is within the
-        scope of your CodedTool.
+        scope of your CodedTool and can be done asynchronously.
 
         :param args: An argument dictionary whose keys are the parameters
                 to the coded tool and whose values are the values passed for them


neuro_san/internals/graph/activations/abstract_class_activation.py
```diff
@@ -234,24 +234,11 @@ async def attempt_invoke(self, coded_tool: CodedTool, arguments: Dict[str, Any],
...
-
-            # Try to run in the executor.
-            invocation_context = self.run_context.get_invocation_context()
-            executor: AsyncioExecutor = invocation_context.get_asyncio_executor()
-            loop: AbstractEventLoop = executor.get_event_loop()
-            retval = await loop.run_in_executor(None, coded_tool.invoke, arguments, sly_data)
+        except NotImplementedError:
+            # That didn't work, so try running the synchronous method
+            # in a separate thread so as not to distrupt the main event loop.
+            retval = await to_thread(coded_tool.invoke, arguments, sly_data)
```
d1donlydfink, line 241: Instead of complaining about the user's code, send any specifically synchronous invoke() execution to a separate thread so the main event loop thread will not be blocked.

This allows noobs to practice in paradigms they are familiar with.  async_invoke() is still a preferred method to override, but here we give people less rope to hang themselves with.

neuro_san/coded_tools/math_guy/calculator.py
```diff
@@ -32,9 +32,9 @@ async def async_invoke(self, args: Dict[str, Any], sly_data: Dict[str, Any]) ->
         """
         Called when the coded tool is invoked asynchronously by the agent hierarchy.
         Strongly consider overriding this method instead of the "easier" synchronous
-        version above when the possibility of making any kind of call that could block
+        invoke() version above when the possibility of making any kind of call that could block
         (like sleep() or a socket read/write out to a web service) is within the
-        scope of your CodedTool.
+        scope of your CodedTool and can be done asynchronously.
```
d1donlydfink, line 37: Update the docs of Calculator per the CodedTool changes below.

neuro_san/interfaces/coded_tool.py
```diff
@@ -19,24 +19,27 @@ class CodedTool:
     Interface contract for a coded tool to squelch the anti-pattern of
     Static Cling.
 
-    Upon activation by the agent hierarchy, a CodedTool will have its
-    invoke() call called by the system.
+    Upon activation by the agent hierarchy, a CodedTool will have either its
+    async_invoke() (preferred) or synchronous invoke() method called by the system.
```
d1donlydfink, line 23: Update the docs for the CodedTool to communicate the new world order.

deepsaia (approved):
LGTM.

neuro_san/internals/graph/activations/abstract_class_activation.py
andreidenissov-cog: Nice higher-level wrapper API method, which should be functionally the same. If we are in the running env of our request AsyncioExecutor. But that should be the case here, so we are sending the workload to the same event loop.

ofrancon: That does NOT look good to me. It's effectively converting the `invoke` function into `async_invoke`. Which is NOT what we want.
Suppose I have 2 functions A and B that manipulate the same file. Which is typically something people do to persist a state or memory for instance. I'm now running into concurrency issues I didn't have before. 
Before: each function is executed 1 after the other, in order. Each function can open a file, modify it, close it.
After: both function are executed in parallel and now... anything can happen. A can overwrite B, B can overwrite A, exceptions can be thrown because the file is closed, etc.

I think we should revert this change. Created #402 for that.

andreidenissov-cog: That's a very good thought in general, but: if we are talking about potential concurrency issues, we need at least 2 threads to have some concurrency issues. Right? So one thread is started by asyncio.to_thread() OK. What is another one? If it is our per-request thread we run in our AsyncioExecutor, than it actually only runs our (per-request) event loop.

d1donlydfink: @andreidenissov-cog : The next thread would be called by the next synchronous call to any CodedTool.invoke().

@ofrancon 's point is a very subtle but good one. I am going to revert this new logic in favor of documentation for the to_thread() scenario.

ofrancon: I should have said 2 `CodedTools` A and B that manipulate the same file, not `functions`.
I they both run synchronously in the event loop, no problem, they'll run one after the other. If they each run in a separate thread, they could interfere with each other.

PR #413, opened 2025-09-26
UN-3526 More flexible structure for agents registry.

andreidenissov-cog:
This PR adds a mechanism for more flexible mapping between agents (actually, agent names by which agents are identified in the service) and the location of their .hocon definition files.
Trivial default mapping (class AgentFileTreeMapper) simply maps agent name in the form abc/xyz/agent to a file location registry-root/abc/xyz/agent.hocon;
Another one (AgentStandaloneMapper) is used for testing and command line clients.
Many others are also possible, as long as they guarantee one-2-one map between agent names and storage locations.

Tested by setting up tree file structure under registry root and running several requests  referencing non-root agents.
Also tested that registry updates watchdog still functions.


neuro_san/internals/interfaces/agent_name_mapper.py
```diff
@@ -0,0 +1,32 @@
...
+# All Rights Reserved.
+# Issued under the Academic Public License.
+#
+# You can be released from the terms, and requirements of the Academic Public
+# License by purchasing a commercial license.
+# Purchase of a commercial license is mandatory for any use of the
+# neuro-san SDK Software in commercial settings.
+#
+# END COPYRIGHT
+
```
andreidenissov-cog, line 12: Interface for abstract agent mapping policy.

neuro_san/internals/graph/persistence/agent_filetree_mapper.py
```diff
@@ -0,0 +1,35 @@
...
+# You can be released from the terms, and requirements of the Academic Public
+# License by purchasing a commercial license.
+# Purchase of a commercial license is mandatory for any use of the
+# neuro-san SDK Software in commercial settings.
+#
+# END COPYRIGHT
+from pathlib import Path
+
+from neuro_san.interfaces.agent_name_mapper import AgentNameMapper
+
```
andreidenissov-cog, line 15: Simple default implementation of AgentNameMapper policy,
where agent name in the form abc/xyz/agent maps to its definition file location:
registry-root/abc/xyz/agent.hocon.

neuro_san/internals/graph/persistence/registry_manifest_restorer.py
```diff
@@ -47,6 +48,10 @@ def __init__(self, manifest_files: Union[str, List[str]] = None):
             * A list of local names for multiple manifest files to host
             * None (the default) which gets a single manifest file from a known source.
         """
+        self.agent_mapper = agent_mapper
+        if not self.agent_mapper:
+            self.agent_mapper = AgentFileTreeMapper()
+
```
andreidenissov-cog, line 56: If no mapping policy is specified, use default AgentFileTreeMapper.

neuro_san/internals/graph/persistence/registry_manifest_restorer.py
```diff
@@ -38,7 +39,7 @@ class RegistryManifestRestorer(Restorer):
     for agent networks/registries.
     """
 
-    def __init__(self, manifest_files: Union[str, List[str]] = None):
+    def __init__(self, manifest_files: Union[str, List[str]] = None, agent_mapper: AgentNameMapper = None):
         """
```
andreidenissov-cog, line 43: Add optional AgentNameMapper policy.
It will determine how we locate agent definition file given its name;
and how we find in-service agent name form its filepath.

neuro_san/internals/graph/persistence/agent_network_restorer.py
```diff
@@ -13,7 +13,6 @@
 from typing import Any
 from typing import Dict
 
-from os import path
 from pathlib import Path
```
andreidenissov-cog, line 16: No os dependency no more.

neuro_san/service/watcher/registries/event_registry_observer.py
```diff
@@ -38,7 +38,7 @@ def start(self):
         """
         Start running observer
         """
-        self.observer.schedule(self.event_handler, path=self.registry_path, recursive=False)
+        self.observer.schedule(self.event_handler, path=self.registry_path, recursive=True)
         self.observer.start()
```
andreidenissov-cog, line 42: We need to watch the whole file tree now.

neuro_san/service/watcher/registries/polling_registry_observer.py
```diff
@@ -39,7 +39,7 @@ def start(self):
         """
         Start running observer
         """
-        self.observer.schedule(self.event_handler, path=self.registry_path, recursive=False)
+        self.observer.schedule(self.event_handler, path=self.registry_path, recursive=True)
         self.observer.start()
```
andreidenissov-cog, line 43: We need to watch the whole file tree now.

neuro_san/service/http/server/http_server.py
```diff
@@ -192,9 +192,9 @@ def make_app(self, requests_limit: int, logger: EventLoopLogger):
 
         # Register templated request paths for agent API methods:
         # regexp format used here is that of Python Re standard library.
-        handlers.append((r"/api/v1/([^/]+)/function", FunctionHandler, request_initialize_data))
-        handlers.append((r"/api/v1/([^/]+)/connectivity", ConnectivityHandler, request_initialize_data))
-        handlers.append((r"/api/v1/([^/]+)/streaming_chat", StreamingChatHandler, request_initialize_data))
+        handlers.append((r"/api/v1/(.+)/function", FunctionHandler, request_initialize_data))
+        handlers.append((r"/api/v1/(.+)/connectivity", ConnectivityHandler, request_initialize_data))
+        handlers.append((r"/api/v1/(.+)/streaming_chat", StreamingChatHandler, request_initialize_data))
```
andreidenissov-cog, line 198: Change regexp pattern so that we capture agent name in general form,
including things like abc/xyz/my_agent.

neuro_san/internals/graph/persistence/agent_network_restorer.py
```diff
@@ -64,7 +76,7 @@ def restore(self, file_reference: str = None):
 
         use_file: str = file_reference
         if self.registry_dir is not None:
-            use_file = path.join(self.registry_dir, file_reference)
+            use_file = str(Path(self.registry_dir) / file_reference)
```
andreidenissov-cog, line 80: Use Path instead of os.path

neuro_san/internals/graph/persistence/registry_manifest_restorer.py
```diff
@@ -118,23 +125,27 @@ def restore_from_files(self, file_references: Sequence[str]) -> Dict[str, AgentN
...
 
+                # Key here is an agent name in a form that we chose,
+                # and we'll need to use an agent mapper to get to this agent definition file.
                 # Keys sometimes come with quotes.
                 use_key: str = key.replace(r'"', "")
-
-                if not bool(value):
-                    continue
+                use_key = use_key.strip()
+                agent_filepath: str = self.agent_mapper.agent_name_to_filepath(use_key)
```
andreidenissov-cog, line 137: Get agent name - as specified in registry manifest.

neuro_san/internals/graph/persistence/registry_manifest_restorer.py
```diff
@@ -118,23 +125,27 @@ def restore_from_files(self, file_references: Sequence[str]) -> Dict[str, AgentN
...
                 try:
-                    agent_network: AgentNetwork = registry_restorer.restore(file_reference=use_key)
+                    agent_network: AgentNetwork = registry_restorer.restore(file_reference=agent_filepath)
                 except FileNotFoundError as exc:
                     self.logger.error("Failed to restore registry item %s - %s", use_key, str(exc))
                     agent_network = None
                 if agent_network is not None:
-                    network_name: str = Path(use_key).stem
+                    network_name: str = self.agent_mapper.filepath_to_agent_network_name(agent_filepath)
                     agent_networks[network_name] = agent_network
```
andreidenissov-cog, line 149: From agent definition file location, get the agent network name that will identify it to the service.

neuro_san/internals/interfaces/agent_name_mapper.py
d1donlydfink: Interesting. Makes me wonder if we could use this for separating out mcp vs neuro-san vs a2a agent namespaces as well. (Thoughts?)

neuro_san/internals/graph/persistence/agent_filetree_mapper.py
```diff
@@ -0,0 +1,35 @@
...
+        Converts an agent name from registry manifest to file path to this agent definition file.
+        """
+        return agent_name
+
+    def filepath_to_agent_network_name(self, filepath: str) -> str:
+        """
+        Converts a file path to agent definition file (relative to registry root directory)
+        to agent network name identifying it to the service.
+        """
+        return str(Path(filepath).with_suffix(""))
```
d1donlydfink, line 36: Why is with_suffix() needed?
Maybe add a code comment

neuro_san/internals/graph/persistence/agent_network_restorer.py
```diff
@@ -40,15 +42,25 @@ class AgentNetworkRestorer(Restorer):
...
+                if registry_dir is None, AgentStandaloneMapper instance will be used;
+                otherwise, we use AgentFileTreeMapper.
         """
         self.registry_dir: str = registry_dir
+        self.agent_mapper = agent_mapper
+        if not self.agent_mapper:
+            if self.registry_dir is not None:
+                self.agent_mapper = AgentFileTreeMapper()
+            else:
+                self.agent_mapper = AgentStandaloneMapper()
```
d1donlydfink, line 63: Make this a factory?

neuro_san/internals/graph/persistence/agent_standalone_mapper.py
```diff
@@ -0,0 +1,35 @@
...
+        Agent name is its filepath.
+        """
+        return agent_name
+
+    def filepath_to_agent_network_name(self, filepath: str) -> str:
+        """
+        Converts a file path to agent standalone definition file
+        to agent network name identifying it to the service.
+        """
+        return str(Path(filepath).stem)
```
d1donlydfink, line 37: stem is needed to remove .hocon suffix?
Worth a comment.

neuro_san/service/http/server/http_server.py
d1donlydfink: How can we do this for gRPC?

neuro_san/internals/graph/persistence/agent_network_restorer.py
d1donlydfink: When do we want Standalone vs FileTree?
Also, what is ultimately doing the choosing here?

d1donlydfink (changes requested):
In general good, so approve in spirit.

Some ?s:
* What does the choosing between Stanalone vs FileTree (not obvious)
* How to do grpc (can come in later PR)?
* Maybe add some code comments in a few places. (See PR)
* Maybe add a blurb in the summary as to how you tested.
* Could AgentNameMapper conceivably be used for mcp vs neuro-san vs a2a divisions?

neuro_san/internals/interfaces/agent_name_mapper.py
andreidenissov-cog: Hmm. Do we need separate namespaces? I thought the idea is that we could provide the same functionality (set of agent networks) presented by different protocols (MCP, A2A, whatever).
That said though - in general yes. When we enable some protocol, we can specify any agent -> agent definition file mapping as necessary. Maybe with different registry roots? Or not.

neuro_san/service/http/server/http_server.py
andreidenissov-cog: Interesting question. I don't exactly know right now.
I'll get to it when we have any real client using gRPC to work with neuro-san.

neuro_san/internals/graph/persistence/agent_network_restorer.py
andreidenissov-cog: Well, Standalone is a kinda special case I see used in some tests - when we don't bother with specifying a registry root, and simple saying - here is an absolute path to a single .hocon file, go run my test for this agent.
FileTree is a proposed new default - when agents could be organized in a real file tree instead of flat collection under registry root.
Choice is done here by  "self.registry_dir" parameter - whether or not we specify it (We don't specify it in our neuro-san tests only, from what I see).

neuro_san/internals/interfaces/agent_name_mapper.py
d1donlydfink: I was thinking since mcp can specify via:

http://myserver.com/mcp/my_agent

... this could end up being its own namespace.

To eventually place nice in that top-level world, we might also want to allow

http://myserver.com/neuro_san/my_agent

Or a2a, or your_mother_in_law, or whatever.

neuro_san/service/http/server/http_server.py
d1donlydfink: Just make sure it's possible.
We might have to document somewhere that some agent names are not allowed like:
streaming_chat, connectivity, function, mcp, neuro_san.

... and maybe even error out on these. (Can be for later PR)

neuro_san/internals/graph/persistence/agent_network_restorer.py
d1donlydfink: Callers with a registry_dir would be in:
./neuro_san/internals/graph/persistence/registry_manifest_restorer.py

... who in turn gets called by:
./neuro_san/service/watcher/registries/registry_storage_updater.py
./neuro_san/service/main_loop/server_main_loop.py
./neuro_san/client/direct_agent_storage_util.py


Callers without would be in:
./neuro_san/client/direct_agent_session_factory.py  (reasonable for library call as opposed to server)
./tests/neuro_san/internals/chat/test_connectivity_reporter.py (test - always reasonable).

---

1C guys would use this a lot with external agents. We might want to be sure that the use case of a 
"deep" (to coin a term for what is going on here) external agent is handled correctly both in http and direct modes from agent_cli.

You could try this by moving math_guy.hocon to be a deep network and then have math_guy_passthrough reference the deep math_guy.

neuro_san/internals/graph/registry/activation_factory.py
```diff
@@ -130,6 +133,7 @@ def create_agent_activation(self, parent_run_context: RunContext,
         :param sly_data: A mapping whose keys might be referenceable by agents, but whose
                  values should not appear in agent chat text. Can be an empty dictionary.
         :param arguments: A dictionary of arguments for the newly constructed agent
+        :param factory: A factory that will be used to create the agent tool
         :return: The CallableActivation agent referred to by the name.
```
andreidenissov-cog, line 137: Put in missing parameter description.

neuro_san/internals/run_context/utils/external_agent_parsing.py
```diff
@@ -60,7 +60,7 @@ def parse_external_agent(agent_url: str) -> Dict[str, str]:
             split: List[str] = parse_result.netloc.split(":")
             host = split[0]
             if len(split) > 1:
-                port = split[0]
+                port = split[1]
```
andreidenissov-cog, line 64: Fixed a bad typo.

neuro_san/internals/interfaces/agent_name_mapper.py
d1donlydfink: Looking at this again, this isn't exactly a client-facing interface.
This is better off in neuro_san/internals/interfaces

neuro_san/internals/run_context/utils/external_agent_parsing.py
d1donlydfink: Thanks!

neuro_san/internals/graph/registry/activation_factory.py
```diff
@@ -105,6 +105,9 @@ def _determine_agent_tool_path(self) -> str:
         if not agent_tool_path.endswith(self.agent_network.get_network_name()):
             agent_tool_path = f"{agent_tool_path}.{self.agent_network.get_network_name()}"
 
+        # Now, agent network name itself can contain "os.sep" symbols
+        # in case of hierarchical agents structure. Replace those too.
+        agent_tool_path = agent_tool_path.replace(os.sep, ".")
```
d1donlydfink, line 110: So here we are replacing os.sep with ".", but below we are replacing "/" with "__".
Why should this be different?

andreidenissov-cog: Here we are constructing Python module path, which requires "." as separator.
We replace "/" with "__" to make our tool/agent name Langchain/OpenAI compliant, and this should be the only place we do this. Inside neuro-san, any string is good for identifying an agent network.

d1donlydfink: Ah missed that. Thanks!

PR #417, opened 2025-10-01
UN-3325 Reservations for temporary AgentNetworks

d1donlydfink:
CodedTools can now request Reservations for temporary agent networks.
(Patent pending)

A Reservation lets the CodedTool know before a network is deployed:

what its id will be (agent name) so it can be referenced by other agent networks created in at the same time for linkage. This is at least a UUID4, but you can also specify a prefix to go with it so a human can remember what it's for.
when this temporary network is going to expire
To register a Reservation for an agent network, a CodedTool calls a Reservationist instance that comes to it via the "reservationist" key in the args dictionary. The Reservationist:

Actually gives the CodedTool any Reservation instance it needs via reserve()
Allows for groups of agent network Reservations to be deploy()-ed together with optional async Event notification as to when they become available
To make this happen in a multi-threaded async server environment, the ServerMainLoop holds a single instance of a Queue of Queues. Any time a new request sees that it might create new temp agent network deployments, a new "inner" queue for each request is dropped down the chute of the "outer" Queue of Queues.

When a CodedTool calls its Reservationist.deploy(), it drops a structure down its inner queue for each request containing all the information needed for a single agent network deployment group.

On the consuming end of all this is a TempNetworkStorageUpdater that wakes up perdiocially as part of the Watcher. This guy looks at the queue-of-queues to see if there are any new inner queues it needs to look at. If so it adds those to a pool and begins to read each of these inner queues for new agent network deployment information.

For each group of agent networks to deploy, the TempNetworkStorageUpdater updates a special "temp" agent network storage whose contents are not broadcast as part of the Concierge service. After each deployment group is stored, any Event corresponding to the deployment is tripped so the CodedTool who initiated all this can know that it's safe to use any agent references (if it wants to)

This all works for the single server case.
There is another simpler implementation that works for the direct/library case that doesn't involve all the queueing.

As an example, a new copy_cat network is provided. Give this guy the name of an existing agent network and it will make a temporary copy and return you the temporary reservation reference via sly_data. If you give it optional calling instructions, it will immediately call the new temporary agent with the text you supply it and return the results.

The copy_cat network uses a CodedTool called the Copyist to do all this. The Copyist uses a new ReservationUtils class to do all the reserve() and deploy() business for a single agent network in reusable, exemplary fashion as well as a new use_reservation() method on BranchActivation (which it also inherits from) to do the calling of the new temp network.

Also included is a multi-pod implementation called S3ReservationsStorage that stores temporary networks across S3 network storage for consistent access across multiple replicated pods. I was able to test it by having 2 http servers running (on different ports) and pointing to the same s3 storage bucket.


hshahrzad (approved):
Looks good to me.

Noravee (approved):
lgtm.

PR #420, opened 2025-10-02
UN-3477 Client Policy for LLMs

d1donlydfink:
Previously we had a LangChainLlmClient object that held references for use in closing.
Upon further looking into how to do this kind of thing for providers other than OpenAI,
we find that most other ChatModels do not actually let you pass in your own async client.
In at least one case (Bedrock), the client itself can't even be an async client.

So with this PR, we are moving to a ClientPolicy model which serves two different styles in the same interface.

The first style is the OpenAI style.  The create_client() method returns the correct object
to pass into the ChatModel's constructor as a client.  Inside create_client(), the ClientPolicy
has access to the full llm config so that client objects can also be configured, and it has the freedom
to store objects that are relevant to shutting down web clients for llm access correctly.

The second style is one where the ChatModel does not allow an external client instance coming in
its constructor or being set anywhere on the object.  In this case we pass in the LLM ChatModel
itself as the basis reference.

Both styles need to implement a delete_resources() method, which correctly cleans up.
Most often for style 2, this involves a reach-in to private member variables to do the right thing
for clean up (at least until we find a better way of doing it.)

Even with all this change, I still think there might be yet another iteration on this notion of a ClientPolicy interface.
See ClientPolicy to comment on near-future direction.


neuro_san/internals/run_context/langchain/llms/anthropic_client_policy.py
```diff
@@ -0,0 +1,48 @@
...
+from neuro_san.internals.run_context.langchain.llms.client_policy import ClientPolicy
+
+
+class AnthropicClientPolicy(ClientPolicy):
+    """
+    Implementation of the ClientPolicy for Anthtropic chat models.
+
+    Anthropic chat models do not allow for passing in an externally managed
+    async web client.
+    """
```
d1donlydfink, line 26: Anthtopic implementation of the ClientPolicy interface.
Still need to ve sure I test the non-OpenAI ones.

neuro_san/internals/run_context/langchain/llms/anthropic_client_policy.py
```diff
@@ -0,0 +1,48 @@
...
+        if self.llm is None:
+            return
+
+        # Do the necessary reach-ins to successfully shut down the web client
+
+        # This is really an anthropic.AsyncClient, but we don't really want to do the Resolver here.
+        # Note we don't want to do this in the constructor, as AnthropicChat lazily
+        # creates these as needed via a cached_property that needs to be done in its own time
+        # via Anthropic infrastructure.  By the time we get here, it's already been created.
+        anthropic_async_client: Any = self.llm._async_client     # pylint:disable=protected-access
```
d1donlydfink, line 41: Do the reach-ins in order to shut down the client as correctly as we can inside the delete_resources() method of the ClientPolicy interface.

This is the general pattern to follow when the LLM class itself does not allow you to pass in a client class at all. - all we can do.

neuro_san/internals/run_context/langchain/llms/azure_client_policy.py
```diff
@@ -0,0 +1,98 @@
...
+        do not support that (even though they should!).
+
+        Implementations should retain any references to state that needs to be cleaned up
+        in the delete_resources() method.
+
+        :param config: The fully specified llm config
+        :return: The web client that accesses the LLM.
+                By default this is None, as many BaseLanguageModels
+                do not allow a web client to be passed in as an arg.
+        """
```
d1donlydfink, line 44: Azure implementation follows the make-a-client-first pattern.
Worth noting that this implementation relies on the OpenAI implementation to do the proper delete_resources() policy

neuro_san/internals/run_context/langchain/llms/bedrock_client_policy.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+class BedrockClientPolicy(ClientPolicy):
+    """
+    ClientPolicy implementation for Bedrock.
+
+    Bedrock does not allow for passing in async web clients.
+    As a matter of fact, all of its clients are synchronous,
+    which is not the best for an async service.
+    """
```
d1donlydfink, line 23: Bedrock is another one of those reach-in during delete_resources, as it does not allow a client to be passed in.  During this exercise I also learned that bedrock doesn't even allow an async client, so not-so-great performance is nearly guaranteed within the async server model.

neuro_san/internals/run_context/langchain/llms/client_policy.py
```diff
@@ -0,0 +1,93 @@
...
+
+    1) When BaseLanguageModels can have web clients passed into their constructor,
+       implementations should use the create_client() method to retain any references necessary
+       to help them clean up nicely in the delete_resources() method.
+
+    2) When BaseLanguageModels cannot have web clients passed into their constructor,
+       implementations should pass the already created llm into their implementation's
+       constructor. Later delete_resources() implementations will need to do a reach-in
+       to the llm instance to clean up any references related to the web client.
+    """
```
d1donlydfink, line 36: This is the basic interface for LLM ClientPolicy.

This is the 2nd iteration on this kind of interface, moving from the data-only class w/ external policy of the LangChainLlmClient we had previously to a class that is more policy based.  (Verb-centered as opposed to Noun-centered).

Having slept on this, I think there is perhaps another 3rd iteration to be had which adds a notion of 2 more
methods looking something like this:

```

def get_llm_class_name(self) -> str:
     """
     :return: The string class name for the LLM represented so as to register this policy class
                   with the neuro-san LlmFactory system
     """
     # Would return something like "openai" or "anthropic" or whatever the llm class name for the llm_info.hocon files would be
     raise NotImplementedError

def create_chat_model(self, config: Dict[str, Any], client: Any = None) -> BaseLanguageModel
```

This last create_chat_model() would allow for LLM creation and deletion policy to exist within the same class, and with standardized external structure could allow for standardized calling and registration as long as the class was listed in the user's llm_info.hocon file.  The idea would be we could also register our own classes this way in default_llm_info.hocon as well for an example.   This would mean that LlmFactory would no longer be the interface to override (though we could keep it for backwards compatibility) but this new LlmPolicy interface would be.

Thoughts about this direction anyone?

neuro_san/internals/run_context/langchain/llms/client_policy.py
```diff
@@ -0,0 +1,93 @@
...
+        do not support that (even though they should!).
+
+        Implementations should retain any references to state that needs to be cleaned up
+        in the delete_resources() method.
+
+        :param config: The fully specified llm config
+        :return: The web client that accesses the LLM.
+                By default this is None, as many BaseLanguageModels
+                do not allow a web client to be passed in as an arg.
+        """
```
d1donlydfink, line 61: In the meantime, this interface allows for a create_client() method, which saves a place for OpenAI/Azure models to create a client to access LLM services separately from the BaseLanguageModel itself, a la @andreidenissov-cog 's original methodology.

neuro_san/internals/run_context/langchain/llms/client_policy.py
```diff
@@ -0,0 +1,93 @@
...
+        return None
+
+    async def delete_resources(self):
+        """
+        Release the run-time resources used by the instance.
+
+        Unfortunately for many BaseLanguageModels, this tends to involve
+        a reach-in to its private internals in order to shutting down
+        any web client references in there.
+        """
```
d1donlydfink, line 72: Specify exactly how the LLM instance's resources should be handled when we are shutting it down.
There is a case to be made for having this be part of its own ResourceHandler interface.
Will go there in subsequent PR.

neuro_san/internals/interfaces/environment_configuration.py
```diff
@@ -9,44 +9,38 @@
...
 from typing import Any
 from typing import Dict
 
 import os
 
-from neuro_san.internals.run_context.langchain.llms.langchain_llm_client import LangChainLlmClient
-
 
-class LangChainLlmClientFactory:
+class EnvironmentConfiguration:
```
d1donlydfink, line 18: Not the greatest diff here.

This is actually a new interface to tack on to things that need to get config values from a dictionary or an env var.  This originally comes from LangChainLlmFactory, but I added the notion of a "none_object", which returns None for the value when the object passed in for that is something. You'll see this used for the OpenAI cases when passing in a client.

neuro_san/internals/interfaces/environment_configuration.py
```diff
@@ -9,44 +9,38 @@
...
-
-                Can raise a ValueError if the config's class or model_name value is
-                unknown to this method.
-        """
-        raise NotImplementedError
-
-    def get_value_or_env(self, config: Dict[str, Any], key: str, env_key: str) -> Any:
+    @staticmethod
+    def get_value_or_env(config: Dict[str, Any], key: str, env_key: str,
+                         none_obj: Any = None) -> Any:
```
d1donlydfink, line 26: Even though this is a static method, classes that multiply inherit can still use self.<foo> to access this method.  This is more concise for one-line value reading than "EnvironmentConfiguration.get_value_or_env(blah blah)".

At some point we could add a DictionaryExtractor in here to allow for deep keys.

neuro_san/internals/interfaces/environment_configuration.py
```diff
@@ -9,44 +9,38 @@
...
         :param config: The config dictionary to search
         :param key: The key for the config to look for
         :param env_key: The os.environ key whose value should be gotten if either
                         the key does not exist or the value for the key is None
+        :param none_obj:  An optional object instance to test.
+                          If present this method will return None, implying
+                          that some other external object/mechanism is supplying the values.
         """
+        if none_obj is not None:
+            return None
```
d1donlydfink, line 37: Do the none_obj policy now.

neuro_san/internals/interfaces/environment_configuration.py
```diff
@@ -9,44 +9,38 @@
...
+
         value = None
         if config is not None:
             value = config.get(key)
 
-        if value is None:
+        if value is None and env_key is not None:
```
d1donlydfink, line 43: Only refer to the environ if the env_key is something.

neuro_san/internals/run_context/langchain/llms/azure_client_policy.py
```diff
@@ -0,0 +1,98 @@
...
+        # From lanchain_openai.chat_models.azure.py
+        default_headers: Dict[str, str] = {}
+        default_headers = config.get("default_headers", default_headers)
+        default_headers.update({
+            "User-Agent": "langchain-partner-python-azure-openai",
+        })
+
+        self.async_openai_client = AsyncAzureOpenAI(
+            azure_endpoint=self.get_value_or_env(config, "azure_endpoint",
+                                                 "AZURE_OPENAI_ENDPOINT"),
```
d1donlydfink, line 74: Because ClientPolicy derives from EnvironmentConfiguration, we can easily use the self.get_value_or_env() method when configuring the clients too, now.

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
```diff
@@ -86,9 +82,6 @@ def __init__(self, config: Dict[str, Any] = None):
         self.llm_factories: List[LangChainLlmFactory] = [
             StandardLangChainLlmFactory()
         ]
-        self.llm_client_factories: List[LangChainLlmClientFactory] = [
-            StandardLangChainLlmClientFactory()
-        ]
```
d1donlydfink, line 91: Abandon the separate factory for clients creation (Nouns), now that we are doing these policy clients (Verbs).

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
```diff
@@ -178,8 +171,7 @@ def create_llm(
                 unknown to this method.
         """
         full_config: Dict[str, Any] = self.create_full_llm_config(config)
-        llm_client: LangChainLlmClient = self.create_llm_client(full_config)
-        llm_resources: LangChainLlmResources = self.create_llm_resources_with_client(full_config, llm_client)
+        llm_resources: LangChainLlmResources = self.create_llm_resources(full_config)
```
d1donlydfink, line 174: Abandon the LangChainLlmClient and factory idea.
That said, we still need a new LlmFactory method that properly returns us LlmResources.

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
```diff
@@ -361,7 +301,7 @@ def create_llm_resources_with_client(self, config: Dict[str, Any],
         found_exception: Exception = None
         for llm_factory in self.llm_factories:
             try:
-                llm_resources = llm_factory.create_llm_resources_with_client(config, llm_client)
+                llm_resources = llm_factory.create_llm_resources(config)
```
d1donlydfink, line 304: Prefer the create_llm_resources() method before trying the older create_chat_model() method.

neuro_san/internals/run_context/langchain/llms/langchain_llm_factory.py
```diff
@@ -65,54 +65,17 @@ def create_base_chat_model(self, config: Dict[str, Any]) -> BaseLanguageModel:
...
+                a BaseLanguageModel (can be Chat or LLM) and a ClientPolicy
+                object that contains all related resources
                 necessary for managing the model run-time lifecycle.
                 Can raise a ValueError if the config's class or model_name value is
                 unknown to this method.
         """
         raise NotImplementedError
-
-    def get_value_or_env(self, config: Dict[str, Any], key: str, env_key: str,
-                         llm_client: Any = None) -> Any:
```
d1donlydfink, line 89: This method moves to EnvironmentConfiguration interface.

neuro_san/internals/run_context/langchain/llms/langchain_llm_factory.py
```diff
@@ -65,54 +65,17 @@ def create_base_chat_model(self, config: Dict[str, Any]) -> BaseLanguageModel:
         """
         raise NotImplementedError
 
-    def create_llm_resources_with_client(self, config: Dict[str, Any],
-                                         llm_client: LangChainLlmClient = None) -> LangChainLlmResources:
+    def create_llm_resources(self, config: Dict[str, Any]) -> LangChainLlmResources:
```
d1donlydfink, line 68: Change the interface before going live with the first merge released.
There still might be more changes here. See comments in ClientPolicy above.

neuro_san/internals/run_context/langchain/llms/langchain_llm_resources.py
```diff
@@ -12,39 +12,39 @@
...
 
 class LangChainLlmResources:
     """
     Class for representing a LangChain model
-    together with run-time resources necessary for model usage by the service.
+    together with run-time policy necessary for model usage by the service.
     """
 
-    def __init__(self, model: BaseLanguageModel, llm_client: LangChainLlmClient = None):
+    def __init__(self, model: BaseLanguageModel, client_policy: ClientPolicy = None):
```
d1donlydfink, line 24: Store LLM + ClientPolicy in this data class.
This allows older LlmFactorys to simply populate an LLM with no ClientPolicy and any resource cleanup/leakage will happen as before.

neuro_san/internals/run_context/langchain/llms/langchain_llm_resources.py
```diff
@@ -12,39 +12,39 @@
...
+        return self.client_policy
 
     async def delete_resources(self):
         """
         Release the run-time resources used by the model
         """
-        if self.llm_client:
-            await self.llm_client.delete_resources()
+        if self.client_policy is not None:
+            await self.client_policy.delete_resources()
```
d1donlydfink, line 50: If there's no ClientPolicy, then there's nothing to delete_resources() on.

neuro_san/internals/run_context/langchain/llms/openai_client_policy.py
```diff
@@ -0,0 +1,111 @@
...
+from httpx import AsyncClient
+
+from langchain_core.language_models.base import BaseLanguageModel
+
+from leaf_common.config.resolver import Resolver
+
+from neuro_san.internals.run_context.langchain.llms.client_policy import ClientPolicy
+
+
+class OpenAIClientPolicy(ClientPolicy):
```
d1donlydfink, line 25: Base ClientPolicy for OpenAI.
@andreidenissov-cog should recognize his efforts in this class instead of in the LlmFactory.

neuro_san/internals/run_context/langchain/llms/standard_langchain_llm_factory.py
```diff
@@ -66,17 +70,11 @@ def create_base_chat_model(self, config: Dict[str, Any]) -> BaseLanguageModel:
         raise NotImplementedError
 
     # pylint: disable=too-many-branches
-    def create_llm_resources_with_client(self, config: Dict[str, Any],
-                                         llm_client: LangChainLlmClient = None) -> LangChainLlmResources:
+    def create_llm_resources(self, config: Dict[str, Any]) -> LangChainLlmResources:
```
d1donlydfink, line 73: Rename for the new interface.

neuro_san/internals/run_context/langchain/llms/standard_langchain_llm_factory.py
```diff
@@ -110,13 +110,9 @@ def create_llm_resources_with_client(self, config: Dict[str, Any],
...
-            # See if there is an async_client to be had from the llm_client passed in
-            async_client: Any = None
-            if llm_client is not None:
-                async_openai_client = llm_client.get_client()
-                if async_openai_client is not None:
-                    # Necessary reach-in.
-                    async_client = async_openai_client.chat.completions
+            # Create the policy object that allows us to manage the model run-time lifecycle
+            client_policy = OpenAIClientPolicy()
+            async_client: Any = client_policy.create_client(config)
```
d1donlydfink, line 115: Consult the OpenAIClientPolicy object in order to create the client _before_ we create the LLM itself.

neuro_san/internals/run_context/langchain/llms/standard_langchain_llm_factory.py
```diff
@@ -178,27 +174,38 @@ def create_llm_resources_with_client(self, config: Dict[str, Any],
...
             llm = AzureChatOpenAI(
+                async_client=async_client,
                 model_name=model_name,
                 temperature=config.get("temperature"),
+
+                # This next group of params should always be None when we have async_client
                 openai_api_key=openai_api_key,
                 openai_api_base=self.get_value_or_env(config, "openai_api_base",
-                                                      "OPENAI_API_BASE"),
+                                                      "OPENAI_API_BASE", async_client),
```
d1donlydfink, line 201: Here is a case where we use that "none_object" in the get_value_or_env() method.

neuro_san/internals/run_context/langchain/llms/standard_langchain_llm_factory.py
```diff
@@ -178,27 +174,38 @@ def create_llm_resources_with_client(self, config: Dict[str, Any],
...
                 openai_organization=self.get_value_or_env(config, "openai_organization",
-                                                          "OPENAI_ORG_ID"),
-                openai_proxy=self.get_value_or_env(config, "openai_organization",
-                                                   "OPENAI_PROXY"),
```
d1donlydfink, line 199: There was a conflation of OPENAI_ORG_ID and OPENAI_PROXY before. Now these are more properly separated.

neuro_san/internals/run_context/langchain/llms/standard_langchain_llm_factory.py
```diff
@@ -264,9 +277,14 @@ def create_llm_resources_with_client(self, config: Dict[str, Any],
...
                 anthropic_api_key=self.get_value_or_env(config, "anthropic_api_key",
                                                         "ANTHROPIC_API_KEY"),
+                default_headers=config.get("default_headers"),
+                betas=config.get("betas"),
                 streaming=True,  # streaming is always on. Without it token counting will not work.
                 # Set stream_usage to True in order to get token counting chunks.
                 stream_usage=True,
+                thinking=config.get("thinking"),
+                mcp_servers=config.get("mcp_servers"),
+                context_management=config.get("context_management"),
```
d1donlydfink, line 287: Add some params I discovered while kicking around with Anthropic.

neuro_san/internals/run_context/langchain/llms/bedrock_client_policy.py
d1donlydfink: Might be worth poking langchain-aws maintainers to allow for an asynchronous client.

neuro_san/internals/run_context/langchain/llms/standard_langchain_llm_factory.py
```diff
@@ -227,30 +234,36 @@ def create_llm_resources_with_client(self, config: Dict[str, Any],
...
         elif chat_class == "anthropic":
 
             # Use lazy loading to prevent installing the world
             # pylint: disable=invalid-name
             ChatAnthropic = resolver.resolve_class_in_module("ChatAnthropic",
                                                              module_name="langchain_anthropic.chat_models",
                                                              install_if_missing="langchain-anthropic")
+
+            # ChatAnthropic currently only supports _async_client() as a cached_property,
+            # not as a constructor arg.
```
d1donlydfink, line 265: Comment of note.

neuro_san/internals/run_context/langchain/llms/standard_langchain_llm_factory.py
```diff
@@ -283,6 +301,10 @@ def create_llm_resources_with_client(self, config: Dict[str, Any],
                 # global verbose value) so that the warning is never triggered.
                 verbose=False,
             )
+
+            # Create the client_policy after the fact, with reach-in
+            client_policy = AnthropicClientPolicy(llm)
+
         elif chat_class == "ollama":
```
d1donlydfink, line 308: Not doing ollama or gemini just yet.

neuro_san/internals/run_context/langchain/llms/standard_langchain_llm_factory.py
```diff
@@ -443,11 +469,15 @@ def create_llm_resources_with_client(self, config: Dict[str, Any],
...
         elif chat_class is None:
             raise ValueError(f"Class name {chat_class} for model_name {model_name} is unspecified.")
         else:
             raise ValueError(f"Class {chat_class} for model_name {model_name} is unrecognized.")
 
-        # Return the LlmResources with the llm_client that was passed in.
+        # Return the LlmResources with the client_policy that was created.
         # That might be None, and that's OK.
-        return LangChainLlmResources(llm, llm_client=llm_client)
+        return LangChainLlmResources(llm, client_policy=client_policy)
```
d1donlydfink, line 483: Stash the ClientPolicy created. If this is still None, then that's till fine.

tests/neuro_san/internals/run_context/langchain/llms/test_llm_factory.py
```diff
@@ -67,26 +68,22 @@ def create_base_chat_model(self, config: Dict[str, Any]) -> BaseLanguageModel:
         """
         raise NotImplementedError
 
-    def create_llm_resources_with_client(self, config: Dict[str, Any],
-                                         llm_client: LangChainLlmClient = None) -> LangChainLlmResources:
+    def create_llm_resources(self, config: Dict[str, Any]) -> LangChainLlmResources:
```
d1donlydfink, line 71: Change the TestLlmFactory to be exemplary.

andreidenissov-cog (approved):
Looks good to me.

PR #422, opened 2025-10-07
Update to aiohttp>=3.13.0

d1donlydfink:
@sidcog says this should help us clear a BlackDuck scan

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -25,7 +25,7 @@ boto3>=1.34.51
 botocore>=1.34.51
 idna>=3.6
 urllib3>=1.26.18
-aiohttp>=3.10.5,<4.0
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -25,7 +25,7 @@ boto3>=1.34.51
 botocore>=1.34.51
 idna>=3.6
 urllib3>=1.26.18
-aiohttp>=3.10.5,<4.0
+aiohttp>=3.13.0,<4.0
```
d1donlydfink, line 28: Increase base version to something that BlackDuck likes better.

sidcog: Sweet, I will get a new scan tonight.

PR #423, opened 2025-10-07
Simple persistence for MCP clients info.

andreidenissov-cog:
This PR adds simple persistence policy for MCP clients info, including auth-related headers,
which is used by LangChainMcpAdapter class when establishing session with external MCP servers.
Mostly modeled after RegistryManifestRestorer logic.
Tested against simple local MCP server requiring static bearer token.


neuro_san/internals/run_context/langchain/mcp/langchain_mcp_adapter.py
```diff
@@ -37,17 +45,24 @@ async def get_mcp_tools(
...
+            LangChainMcpAdapter._mcp_clients_info = McpClientsInfoRestorer().restore()
+            if LangChainMcpAdapter._mcp_clients_info is None:
+                # Prevent further attempts to read the file if first attempt failed.
+                LangChainMcpAdapter._mcp_clients_info = {}
+
+        mcp_tool_dict: Dict[str, Any] = {
+            "url": server_url,
+            "transport": "streamable_http",
+        }
+        # Try to look up authentication details from the URL
```
Noravee, line 70: Should we also look for tool filtering in the MCP info file? For example, maybe we can have a key `"tools"` where its value is a list of allowed tools. 

If user provide `allowed_tools` in the network hocon file, we use that one. Otherwise, the one in the MCP info file is used.

Please let me know what you think about this.

andreidenissov-cog: I've updated the logic - now we look also at tools list provided by MCP client info.
Feel free to tune this part if I missed something.

neuro_san/internals/run_context/langchain/mcp/mcp_clients_info_restorer.py
```diff
@@ -0,0 +1,92 @@
+
+# Copyright (C) 2023-2025 Cognizant Digital Business, Evolutionary AI.
+# All Rights Reserved.
+# Issued under the Academic Public License.
+#
+# You can be released from the terms, and requirements of the Academic Public
+# License by purchasing a commercial license.
+# Purchase of a commercial license is mandatory for any use of the
+# neuro-san SDK Software in commercial settings.
```
d1donlydfink, line 9: You should move this file to the langchain/mcp directory.
It has nothing to do with graph/persistence and introduces an import tangle

neuro_san/internals/run_context/langchain/mcp/langchain_mcp_adapter.py
```diff
@@ -23,6 +28,9 @@ class LangChainMcpAdapter:
     LangChain-compatible tools. This class provides static methods for interacting with MCP servers.
     """
 
+    # Cached MCP clients info to avoid repeated file reads
+    _mcp_clients_info: Dict[str, Any] = None
```
d1donlydfink, line 34: class variables are ultimately very dangerous in a multi-request world.

neuro_san/internals/run_context/langchain/mcp/mcp_clients_info_restorer.py
```diff
@@ -0,0 +1,92 @@
...
+from typing import Any, Dict
+
+import json
+from pyparsing.exceptions import ParseException
+from pyparsing.exceptions import ParseSyntaxException
+
+from leaf_common.persistence.easy.easy_hocon_persistence import EasyHoconPersistence
+from leaf_common.persistence.interface.restorer import Restorer
+
+from neuro_san import DEPLOY_DIR
```
d1donlydfink: FWIW: DEPLOY_DIR within the neuro_san repo is neuro_san/deploy
This is where all the Dockerfile-related stuff lives. Are you sure this is where you want to look for the mcp_info.hocon file?  Seems a little odd.  Maybe the same directory that the toolbox hocon is looked for would be better?

neuro_san/internals/run_context/langchain/mcp/langchain_mcp_adapter.py
Noravee: I think this is good enough for now. Thanks.

Noravee (approved):
I think this is good enough for me to use with the agent network designer. Thank you.

neuro_san/internals/run_context/langchain/mcp/mcp_clients_info_restorer.py
andreidenissov-cog: OK done.

neuro_san/internals/run_context/langchain/mcp/langchain_mcp_adapter.py
andreidenissov-cog: They are, unless properly protected.
Anyway, this PR was thrown together as a quick throw-away demo-purpose contraption,
but if we go for more stable thing, let's add protection.

neuro_san/internals/run_context/langchain/mcp/mcp_clients_info_restorer.py
andreidenissov-cog: Might as well. I don't really have any preference here. We could consider external MCPs connection info to be a deployment artifact. Or not.

neuro_san/internals/run_context/langchain/mcp/mcp_clients_info_restorer.py
```diff
@@ -0,0 +1,95 @@
...
+                Default is None, implying the file reference is up to the
+                implementation.
+        :return: a dictionary with MCP clients information
+        """
+        file_path: str = file_reference
+        if not file_path:
+            file_path = os.environ.get("MCP_CLIENTS_INFO_FILE")
+            if not file_path:
+                # No env var, so fallback to what is coded in this repo.
+                path_from_top: str = "internals/run_context/langchain/toolbox/mcp_clients_info.hocon"
```
Noravee: The path should be `internals/run_context/langchain/mcp/mcp_clients_info.hocon`. However, I think we may not need this at all. If there is no mcp client info file, get the tool from the given url in the network hocon without authentication or tool filtering.

PR #459, opened 2025-10-09
Fix MCP tool filtering

Noravee:
### Issue
- `LangChainMcpAdapter.get_mcp_tools` fails to filter the tools.
- This is because we are looking for field "tools" before the URL. 

### Solution
- Look for "tools" after the URL.


neuro_san/internals/run_context/langchain/mcp/langchain_mcp_adapter.py
```diff
@@ -90,7 +90,7 @@ async def get_mcp_tools(
         client_allowed_tools: List[str] = allowed_tools
         if client_allowed_tools is None:
             # Check if MCP client info has a "tools" field to use as allowed tools.
-            client_allowed_tools = LangChainMcpAdapter._mcp_clients_info.get("tools")
+            client_allowed_tools = LangChainMcpAdapter._mcp_clients_info.get(server_url, {}).get("tools")
```
Noravee, line 93: `tools` is inside `server_url`.

Issue #418, opened 2025-10-01
Validate agent names


ofrancon:
Using special characters in agent names makes neuro-san throw exceptions that are difficult to analyze.
Calling an agent `Agent.1`or `Agent 2` for instance will make neuro-san crash when the agents are invoked.
It would be better to check the agent names beforehand and gracefully handle the issue by reporting a user friendly error at startup or when the agent network is loaded, like:
> Agent network: some_agent_network.hocon contains invalid agent names: Agent.1, Agent 2. Special characters `.` and ` ` are not supported.

Fail fast, fail gracefully. That way I'd know what's wrong and how to fix it.

To reproduce: a network like this:
```
llm_config {
    class = "openai"
    model_name = "gpt-4.1"
}

tools = [
    # This is the agent network's entry point.
    {
        name = "Frontman"
        function {
            description = """Some description"""
        }
        instructions =
"""
Some instructions
"""
        tools = ["Agent.1",
            "Agent 2"]
...
```
Throws this exception when we start chatting with it:
```
File "/Users/***/***/venv/lib/python3.13/site-packages/openai/_base_client.py", line 1594, in request
raise self.make_status_error_from_response(err.response) from None
openai.BadRequestError: Error code: 400 - {'error': {'message': "Invalid 'tools[0].function.name': string does not match pattern. Expected a string that matches the pattern '^[a-zA-Z0-9-]+$'.", 'type': 'invalid_request_error', 'param': 'tools[0].function.name', 'code': 'invalid_value'}}
)
```

PR #460, opened 2025-10-09
418 Validate agent hocon. Skip files that fail validation.

d1donlydfink:
Take @Noravee 's and @hshahrzad 's original validator from neuro-san-studio's agent_network_designer and bring it into the main neuro-san fold.  In doing so, I am breaking up the monolithy with a single AgentNetworkValidator interface as the basis.

* Create an AgentNetworkValidator interface as the basis for all the other implementations. This guy has a single abstract validate() method which takes an agent network description to validate as its only arg. It returns a list of validation error strings.
* Take the original monolithic implementation from neuro-san-studio and break it up into the following validators:
    * KeywordNetworkValidator
    * StructureNetworkValidator
    * ToolboxNetworkValidator
    * UrlNetworkValidator
    ... where each is its own pass over an agent network spec, generating its own errors, with its own scope of concerns.
* There are a couple of validation policy additions that are slightly different from @Noravee 's original implementation:
    * The deepwiki network caused some issues because it has an mcp structure as a tool.  There is a notion of "safe tools" now which sidesteps checking on those. Perhaps more work is needed to achieve further robustness. I just don't know at this point.
    * There is a flag for checking for cycles.  I can see that the Designer might not want to have cycles going on, but when reading hand-crafted files, that is indeed valid.
    * It's also OK for agent networks to have a single node in the network. Things like gist.hocon were failing from this oversight.
    * We are no longer reading files in every time for some of the validation data. The style is now to pass this information in once to a validator's constructor.  This info includes:
        * Valid external agent references (UrlNetworkValidator)
        * Valid mcp server references (UrlNetworkValidator)
        * Toolbox contents (ToolboxNetworkValidator)
* Create a new ToolNameNetworkValidator which addresses @ofrancon 's original concern from #418 
* Create unit tests for all of these (some omissions are noted in the PR)
* Create a notion of a CompositeNetworkValidator, whose validate() method can loop over those validate() methods from a list of Validators
* Create a notion of a ManifestNetworkValidator, which is a Composite with a specific list of Validators we will use when reading in agent network files from the manifest. Designer can have its own.
* When reading in agent network files from the manifest, do a validation check with the ManifestNetworkValidator. Skip any networks that have errors and spit out why we are skipping over it to the error logger.

---
I was able to test starting up the neuro-san repo server with no problem after all these additions.
It's possible that neuro-san-studio's networks and/or our close development partners might uncover some fallout.
"I defined it this way and it worked before!"

A preliminary look at neuro-san-studio gives these errors which will have to be fixed separately:
```
["Agent Network HTML Creator must match the regex '^[a-zA-Z0-9_-]+$'"]
manifest registry agent_network_html_creator.hocon has validation errors. Skipping. Errors: [
    "Agent Network HTML Creator must match the regex '^[a-zA-Z0-9_-]+$'"
]
["PDF RAG Assistant must match the regex '^[a-zA-Z0-9_-]+$'"]
manifest registry pdf_rag.hocon has validation errors. Skipping. Errors: [
    "PDF RAG Assistant must match the regex '^[a-zA-Z0-9_-]+$'"
]
["Airline 360 Assistant must match the regex '^[a-zA-Z0-9_-]+$'"]
manifest registry airline_policy.hocon has validation errors. Skipping. Errors: [
    "Airline 360 Assistant must match the regex '^[a-zA-Z0-9_-]+$'"
]
```
These will be fixed with https://github.com/cognizant-ai-lab/neuro-san-studio/pull/337

We'll see about our partners.

---
Future work:
* Get @Noravee to write a unit test for ToolboxNetworkValidator
* Hook up validation to the Reservations infrastructure.
* Integrate this stuff into agent_network_designer

### DIFF
--- a/neuro_san/internals/run_context/langchain/core/langchain_run_context.py
+++ b/neuro_san/internals/run_context/langchain/core/langchain_run_context.py
@@ -367,16 +367,17 @@ class LangChainRunContext(RunContext):
         :param mcp_info: MCP server URL (string) or a configuration dictionary
         :return: A list of MCP tools as base tools
         """
-
+        # By default, assume no allowed tools. This may get updated below or in the LangChainMcpAdadter.
+        allowed_tools: List[str] = None
         if isinstance(mcp_info, str):
             server_url: str = mcp_info
-            allowed_tools: List[str] = []
         else:
             server_url = mcp_info.get("url")
             allowed_tools = mcp_info.get("tools")
 
         try:
-            mcp_tools: List[BaseTool] = await LangChainMcpAdapter.get_mcp_tools(server_url, allowed_tools)
+            mcp_adapter = LangChainMcpAdapter()
+            mcp_tools: List[BaseTool] = await mcp_adapter.get_mcp_tools(server_url, allowed_tools)
 
         # MCP errors are nested exceptions.
         except ExceptionGroup as nested_exception:
@@ -388,6 +389,8 @@ class LangChainRunContext(RunContext):
             self.logger.info(message)
             return None
 
+        # The allowed tools list might have been updated by the MCP adapter
+        allowed_tools: List[str] = mcp_adapter.client_allowed_tools
         tool_names: List[str] = [tool.name for tool in mcp_tools]
         invalid_names: Set[str] = set(allowed_tools) - set(tool_names)
         # Check if there are invalid tool names in the list.
--- a/neuro_san/internals/run_context/langchain/mcp/langchain_mcp_adapter.py
+++ b/neuro_san/internals/run_context/langchain/mcp/langchain_mcp_adapter.py
@@ -29,9 +29,14 @@ class LangChainMcpAdapter:
     LangChain-compatible tools. This class provides static methods for interacting with MCP servers.
     """
 
-    # Cached MCP clients info to avoid repeated file reads
     _mcp_info_lock: threading.Lock = threading.Lock()
     _mcp_clients_info: Dict[str, Any] = None
+
+    def __init__(self):
+        """
+        Constructor
+        """
+        self.client_allowed_tools: List[str] = []
 
     @staticmethod
     def _load_mcp_clients_info():
@@ -46,10 +51,10 @@ class LangChainMcpAdapter:
                     # Prevent further attempts to load info.
                     LangChainMcpAdapter._mcp_clients_info = {}
 
-    @staticmethod
     async def get_mcp_tools(
-        server_url: str,
-        allowed_tools: Optional[List[str]] = None,
+            self,
+            server_url: str,
+            allowed_tools: Optional[List[str]] = None,
     ) -> List[BaseTool]:
         """
         Fetches tools from the given MCP server and returns them as a list of LangChain-compatible tools.
@@ -60,8 +65,8 @@ class LangChainMcpAdapter:
 
         :return: A list of LangChain BaseTool instances retrieved from the MCP server.
         """
-        if LangChainMcpAdapter._mcp_clients_info is None:
-            LangChainMcpAdapter._load_mcp_clients_info()
+        if self._mcp_clients_info is None:
+            self._load_mcp_clients_info()
 
         mcp_tool_dict: Dict[str, Any] = {
             "url": server_url,
@@ -69,7 +74,7 @@ class LangChainMcpAdapter:
         }
         # Try to look up authentication details from the URL
         headers_dict: Dict[str, Any] =\
-            LangChainMcpAdapter._mcp_clients_info.get(server_url, {}).get("headers")
+            self._mcp_clients_info.get(server_url, {}).get("headers")
         if headers_dict:
             mcp_tool_dict["headers"] = copy.copy(headers_dict)
 
@@ -90,9 +95,12 @@ class LangChainMcpAdapter:
         client_allowed_tools: List[str] = allowed_tools
         if client_allowed_tools is None:
             # Check if MCP client info has a "tools" field to use as allowed tools.
-            client_allowed_tools = LangChainMcpAdapter._mcp_clients_info.get(server_url, {}).get("tools")
+            client_allowed_tools = self._mcp_clients_info.get(server_url, {}).get("tools", [])
+        # If client allowed tools is an empty list, do not filter the tools.
         if client_allowed_tools:
             mcp_tools = [tool for tool in mcp_tools if tool.name in client_allowed_tools]
+
+        self.client_allowed_tools = client_allowed_tools
 
         for tool in mcp_tools:
             # Add "langchain_tool" tags so journal callback can idenitify it.


neuro_san/internals/graph/persistence/registry_manifest_restorer.py
```diff
@@ -124,6 +125,12 @@ def restore_from_files(self, file_references: Sequence[str]) -> Dict[str, AgentN
 """
                 raise FileNotFoundError(message)
 
+            # Find the list of agent network names
+            external_network_names: List[str] = self.find_external_network_names(one_manifest)
+
+            # DEF - need mcp servers as well at some point
+            validator = ManifestNetworkValidator(external_network_names)
```
d1donlydfink, line 132: Validate based on local names for external networks.
Note that this step is currently not validating mcp server stuff, as that is not necessarily fully supported just yet.

neuro_san/internals/graph/persistence/registry_manifest_restorer.py
```diff
@@ -144,7 +151,16 @@ def restore_from_files(self, file_references: Sequence[str]) -> Dict[str, AgentN
...
                     agent_network = None
+
                 if agent_network is not None:
+
+                    validation_errors: List[str] = validator.validate(agent_network.get_config())
+                    if len(validation_errors) > 0:
+                        self.logger.error("manifest registry %s has validation errors. Skipping. Errors: %s",
+                                          agent_filepath,
+                                          json.dumps(validation_errors, indent=4, sort_keys=True))
+                        continue
```
d1donlydfink, line 162: Skip a network if it has validation errors.
Spit out those validation errors via the logger.

neuro_san/internals/graph/persistence/registry_manifest_restorer.py
```diff
@@ -172,3 +188,28 @@ def get_manifest_files(self) -> List[str]:
...
         """
         return self.manifest_files
+
+    def find_external_network_names(self, manifest_entries: Dict[str, Any]) -> List[str]:
+        """
+        Find the list of valid external agent network names
+
+        :param manifest_entries: The manifest entries
+        :return: A list of valid external network references.
+        """
```
d1donlydfink, line 198: New method to pre-determine valid external network names on this server for validation purposes.

neuro_san/internals/interfaces/agent_network_validator.py
```diff
@@ -0,0 +1,75 @@
...
+# END COPYRIGHT
+from typing import Any
+from typing import Dict
+from typing import List
+
+
+class AgentNetworkValidator:
+    """
+    An interface for validating agent network content.
+    """
```
d1donlydfink, line 20: Basic interface for validating an agent network.
This can take something that was just read in from a hocon file, or the name -> agent_spec format that the network designer uses.

neuro_san/internals/interfaces/agent_network_validator.py
```diff
@@ -0,0 +1,75 @@
...
+        :return: A list of error messages
+        """
+        raise NotImplementedError
+
+    @staticmethod
+    def get_name_to_spec(agent_network: Dict[str, Any]) -> Dict[str, Any]:
+        """
+        :param agent_network: The top-level agent network or tools dictionary
+        :return: The agent name -> single agent spec dictionary of the agent network
+        """
```
d1donlydfink, line 36: Some convenience methods that a couple of implementations tend to use over and over again.

neuro_san/internals/validation/composite_network_validator.py
```diff
@@ -0,0 +1,48 @@
...
+from typing import Dict
+from typing import List
+
+from neuro_san.internals.interfaces.agent_network_validator import AgentNetworkValidator
+
+
+class CompositeNetworkValidator(AgentNetworkValidator):
+    """
+    Implementation of the AgentNetworkValidator interface that uses multiple validators
+    """
```
d1donlydfink, line 22: Semi-abstract Validator that allows multiple validation steps to be run, given a list of Validator instances.
We might want to use different kinds of validators for different situations.

neuro_san/internals/validation/keyword_network_validator.py
```diff
@@ -0,0 +1,63 @@
...
+        """
+        self.logger: Logger = getLogger(self.__class__.__name__)
+
+    def validate(self, agent_network: Dict[str, Any]) -> List[str]:
+        """
+        Validate the agent network.
+
+        :param agent_network: The agent network or name -> spec dictionary to validate
+        :return: List of errors indicating agents and missing keywords
+        """
```
d1donlydfink, line 40: This implementation comes from what was in Nor's original monolithic validator in neuro-san-studio.

neuro_san/internals/validation/keyword_network_validator.py
```diff
@@ -0,0 +1,63 @@
...
+
+        self.logger.info("Validating agent network keywords...")
+
+        if not agent_network:
+            errors.append("Agent network is empty.")
+            return errors
+
+        # We can validate either from a top-level agent network,
+        # or from the list of tools from the agent spec.
+        name_to_spec: Dict[str, Any] = self.get_name_to_spec(agent_network)
```
d1donlydfink, line 51: Most of these implementations have this step to get the incoming network description in a mapping of agent name -> agent spec dictionary.  @Noravee 's original implementations worked on these, and most validations care about what's inside the tools.  But we do also want to easily do the same validation on a freshly-read hocon file, so that's what this step is about.

neuro_san/internals/validation/keyword_network_validator.py
```diff
@@ -0,0 +1,63 @@
...
+
+        # Currently, only required "instructions" for non-function agents.
+        for agent_name, agent in name_to_spec.items():
+            if agent.get("instructions") == "":
+                error_msg = f"{agent_name} 'instructions' cannot be empty."
+                errors.append(error_msg)
```
d1donlydfink, line 57: Could easily do more here. This was @Noravee 's original implementation.

neuro_san/internals/validation/manifest_network_validator.py
```diff
@@ -0,0 +1,42 @@
...
+        :param external_network_names: A list of external network names
+        :param mcp_servers: A list of MCP servers, as read in from a mcp_info.hocon file
+        """
+        validators: List[AgentNetworkValidator] = [
+            KeywordNetworkValidator(),
+            StructureNetworkValidator(cyclical_agents_ok=True),
+            # No ToolBoxNetworkValidator yet.
+            ToolNameNetworkValidator(),
+            UrlNetworkValidator(external_network_names, mcp_servers),
+        ]
```
d1donlydfink, line 41: A list of validators we want to use when reading an agent network in from the manifest file.
Note that we are not yet doing the ToolBox stuff yet.  I don't know enough about how to properly test it and set it up in the spot where this gets invoked.   Will leave that to @Noravee to do for a future date.

neuro_san/internals/validation/structure_network_validator.py
```diff
@@ -0,0 +1,303 @@
...
+
+class StructureNetworkValidator(AgentNetworkValidator):
+    """
+    AgentNetworkValidator that looks for topological issues in an agent network.
+    """
+
+    # State tracking for graph visitation
+    UNVISITED: int = 0
+    CURRENTLY_BEING_PROCESSED: int = 1
+    FULLY_PROCESSED: int = 2
```
d1donlydfink, line 32: @Noravee 's original implementation had some magic numbers. Turn those into constants.

neuro_san/internals/validation/structure_network_validator.py
```diff
@@ -0,0 +1,303 @@
...
+
+    def __init__(self, cyclical_agents_ok: bool = False):
+        """
+        Constructor
+        """
+        self.logger: Logger = getLogger(self.__class__.__name__)
+        self.cyclical_agents_ok: bool = cyclical_agents_ok
```
d1donlydfink, line 39: Original implementation for agent network designer asserted that cycles in the graph were bad.
That's not necessarily the case, so make that detection optional.

neuro_san/internals/validation/structure_network_validator.py
```diff
@@ -0,0 +1,303 @@
...
+        all_down_chains: Set[str] = set()
+        has_down_chains: Set[str] = set()
+
+        for agent_name, agent_config in name_to_spec.items():
+            down_chains: List[str] = agent_config.get("tools", [])
+            if down_chains:
+
+                has_down_chains.add(agent_name)
+
+                safe_down_chains: List[str] = self.remove_dictionary_tools(down_chains)
```
d1donlydfink, line 113: Here and in a few more places I am calling this remove_dictionary_tools() so that a network like the deepwiki one, which has an mcp-spec dictionary as a tool, can pass validation.

neuro_san/internals/validation/structure_network_validator.py
```diff
@@ -0,0 +1,303 @@
...
+                all_down_chains.update(safe_down_chains)
+
+        # Potential top agents are agents that have down-chains but are not down-chains of others
+        top_agents: Set[str] = has_down_chains - all_down_chains
+
+        # Special case: If there's only one agent in the network, it's always a top agent
+        if len(top_agents) == 0 and len(name_to_spec) == 1:
+            # It's OK to have a single top agent with no down-chains
+            one_top: str = list(name_to_spec.keys())[0]
+            top_agents.add(one_top)
```
d1donlydfink, line 123: This is another addition of mine from the original.
A few agent networks only have a single tool in them to operate as a prompted llm that can be called generically. For example, the gist network does this.

 That's OK too.

neuro_san/internals/interfaces/agent_network_validator.py
```diff
@@ -0,0 +1,75 @@
...
+from typing import Dict
+from typing import List
+
+
+class AgentNetworkValidator:
+    """
+    An interface for validating agent network content.
+    """
+
+    def validate(self, agent_network: Dict[str, Any]) -> List[str]:
```
d1donlydfink, line 22: Note that we are passing in an agent network to validate.
@Noravee 's original implementation had it so that the network needed to be passed into the constructor.
That approach discourages re-use of configured instances, as is convenient for the manifest checking situation.   Plus its just much better style to have an interface defined as `def verb(noun)`, reading: "Do this operation on this object"

neuro_san/internals/validation/tool_name_network_validator.py
```diff
@@ -0,0 +1,70 @@
...
+
+import re
+
+from neuro_san.internals.interfaces.agent_network_validator import AgentNetworkValidator
+
+
+class ToolNameNetworkValidator(AgentNetworkValidator):
+    """
+    AgentNetworkValidator that looks for correct tool names in an agent network
+    """
```
d1donlydfink, line 28: Add a tool name validation spec that @ofrancon was originally looking for in #418 
This was not part of @Noravee 's original implementation. Extra bonus.

neuro_san/internals/validation/toolbox_network_validator.py
```diff
@@ -0,0 +1,63 @@
...
+from logging import getLogger
+from logging import Logger
+
+from neuro_san.internals.interfaces.agent_network_validator import AgentNetworkValidator
+
+
+class ToolboxNetworkValidator(AgentNetworkValidator):
+    """
+    Agent network validator for toolbox references.
+    """
```
d1donlydfink, line 26: This is the one I know the least about, so I was not really able to write a proper test for it.
Will leave that to @Noravee .  Note that this one is not yet included in the ManifestNetworkValidator because of my lack of experience.

neuro_san/internals/validation/toolbox_network_validator.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def __init__(self, tools: Dict[str, Any]):
+        """
+        Constructor
+
+        :param tools: A dictionary of tools, as read in from a toolbox_info.hocon file
+        """
```
d1donlydfink, line 33: The one change here from @Noravee 's original implementation is that we are not reading in a toolbox file every time this is called.  The idea is that we construct one of these once and pass in what we need from the toolbox file

neuro_san/internals/validation/url_network_validator.py
```diff
@@ -0,0 +1,89 @@
...
+
+from neuro_san.internals.interfaces.agent_network_validator import AgentNetworkValidator
+
+
+class UrlNetworkValidator(AgentNetworkValidator):
+    """
+    AgentNetworkValidator that looks for correct URLs in an agent network
+    """
+
+    def __init__(self, external_agents: List[str] = None, mcp_servers: List[str] = None):
```
d1donlydfink, line 28: Similar to toolbox validator above, this validator passes in valid external agent references and mcp servers that are each determined externally so we don't read a file every time we run the validator.

neuro_san/internals/validation/url_network_validator.py
```diff
@@ -0,0 +1,89 @@
...
+        if self.mcp_servers:
+            urls.extend(self.mcp_servers)
+
+        self.logger.info("Validating URLs for MCP tools and subnetwork...")
+
+        for agent_name, agent in name_to_spec.items():
+            if agent.get("tools"):
+                tools: List[str] = agent.get("tools")
+                if tools:
+                    safe_tools: List[str] = self.remove_dictionary_tools(tools)
```
d1donlydfink, line 69: Another place where we have to look at "safe" tool references that are not dictionaries.
This allows the deepwiki to pass snuff.

neuro_san/service/watcher/temp_networks/reservation_dictionary_converter.py
```diff
@@ -1,4 +1,5 @@
 
+# Copyright (C) 2023-2025 Cognizant Digital Business, Evolutionary AI.
```
d1donlydfink, line 2: FWIW: I did a pass on the copyright generator and it caught stuff like this

tests/neuro_san/internals/validation/test_keyword_network_validator.py
```diff
@@ -0,0 +1,83 @@
...
+from neuro_san import REGISTRIES_DIR
+from neuro_san.internals.graph.persistence.agent_network_restorer import AgentNetworkRestorer
+from neuro_san.internals.graph.registry.agent_network import AgentNetwork
+from neuro_san.internals.validation.keyword_network_validator import KeywordNetworkValidator
+
+
+class TestKeywordNetworkValidator(TestCase):
+    """
+    Unit tests for KeywordNetworkValidator class.
+    """
```
d1donlydfink, line 27: Write a bunch of unit tests.

Most notable omission is a tes for the ToolboxNetworkValidator.
I'll leave that to @Noravee who understands it better.

tests/neuro_san/internals/validation/test_url_network_validator.py
```diff
@@ -0,0 +1,83 @@
...
+
+        failure_message: str = None
+        if len(errors) > 0:
+            failure_message = errors[0]
+        self.assertEqual(0, len(errors), failure_message)
+
+    def test_no_external_network(self):
+        """
+        Tests a network where at least one of the nodes does not have a listed external network
+        """
```
d1donlydfink, line 70: Worth noting that while I am testing for valid external agents, I am not yet testing for valid mcp business. That's kind of in flux, and will let that settle before going there.

neuro_san/internals/interfaces/agent_network_validator.py
ofrancon, line 22: I suppose this class is not supposed to be instantiated on its own. You could have it extend ABC and mark this method with `@abstractmethod`

neuro_san/internals/validation/structure_network_validator.py
ofrancon: You could even use an `enum`

neuro_san/internals/validation/tool_name_network_validator.py
ofrancon: Thank you

tests/neuro_san/internals/validation/test_keyword_network_validator.py
ofrancon: Yeah, unit tests!

neuro_san/internals/validation/tool_name_network_validator.py
```diff
@@ -0,0 +1,70 @@
...
+        # or from the list of tools from the agent spec.
+        name_to_spec: Dict[str, Any] = self.get_name_to_spec(agent_network)
+
+        # Be sure all agent names are valid per the regex above.
+        for agent_name, agent in name_to_spec.items():
+            spec_name: str = agent.get("name")
+            if not re.match(self.TOOL_NAME_PATTERN, agent_name) or \
+                    not re.match(self.TOOL_NAME_PATTERN, spec_name):
+                error_msg = f"{agent_name} must match the regex '{self.TOOL_NAME_PATTERN}'"
+                errors.append(error_msg)
```
andreidenissov-cog, line 64: What about structured tool/agent names we now allow?
Like abc/def/my_agent.

d1donlydfink: We would need a test network and a unit test for that, wouldn't we?  ;)

d1donlydfink: I will start on an Afterthoughts PR today, and I will do what I can to go there.
Enough is going on in here already.

PR #462, opened 2025-10-10
UN-3526 Make working with agents paths OS-agnostic.

andreidenissov-cog:
This PR fixes agent names => local file paths mapping logic so it becomes independent of underlying OS conventions,
specifically file path separators being "/" or "\" or whatever.

Tested (MacOS only though) with math_guy_passthrough using:
/deep/dive/math_guy as external tool;
and also calling deep/dive/math_guy directly.


neuro_san/client/thinking_file_message_processor.py
```diff
@@ -133,6 +133,9 @@ def _write_to_file(self, origin_filename: str, origin_str: str,
                 return
             filename = Path(self.thinking_dir, origin_filename)
 
+        # Ensure the directories exist:
+        filename.parent.mkdir(parents=True, exist_ok=True)
+
```
andreidenissov-cog: In case of structured agent name abc/xyz/my_agent, make sure our directory path for thinking file exists.

neuro_san/internals/graph/persistence/agent_filetree_mapper.py
```diff
@@ -25,12 +25,15 @@ def agent_name_to_filepath(self, agent_name: str) -> str:
...
 
     def filepath_to_agent_network_name(self, filepath: str) -> str:
         """
         Converts a file path to agent definition file (relative to registry root directory)
         to agent network name identifying it to the service.
         """
-        # Remove file name extension, leave everything else in place.
-        return str(Path(filepath).with_suffix(""))
+        # Remove file name extension, convert it to "/" separator.
+        return str(Path(Path(filepath).as_posix()).with_suffix(""))
```
andreidenissov-cog, line 39: Our structured agent name format always uses "/", even on say Windows.
Manifest file format is OS-agnostic.

neuro_san/internals/graph/persistence/agent_filetree_mapper.py
```diff
@@ -25,12 +25,15 @@ def agent_name_to_filepath(self, agent_name: str) -> str:
         """
         Converts an agent name from registry manifest to file path to this agent definition file.
         """
-        return agent_name
+        # agent_name from registry manifest uses '/' as separator,
+        # we need to convert it to the local OS file path format.
+
+        return str(Path(PurePosixPath(agent_name)))
```
andreidenissov-cog, line 32: File path is generated in local OS format.

neuro_san/internals/graph/persistence/agent_standalone_mapper.py
```diff
@@ -25,7 +25,7 @@ def agent_name_to_filepath(self, agent_name: str) -> str:
         """
         Agent name is its filepath.
         """
-        return agent_name
+        return str(Path(PurePosixPath(agent_name)))
```
andreidenissov-cog, line 29: The same thing here - generate local OS format.

neuro_san/internals/graph/registry/activation_factory.py
```diff
@@ -105,9 +105,9 @@ def _determine_agent_tool_path(self) -> str:
         if not agent_tool_path.endswith(self.agent_network.get_network_name()):
             agent_tool_path = f"{agent_tool_path}.{self.agent_network.get_network_name()}"
 
-        # Now, agent network name itself can contain "os.sep" symbols
+        # Now, agent network name itself can contain "/" symbols (regardless of underlying OS)
         # in case of hierarchical agents structure. Replace those too.
-        agent_tool_path = agent_tool_path.replace(os.sep, ".")
+        agent_tool_path = agent_tool_path.replace("/", ".")
         return agent_tool_path
```
andreidenissov-cog, line 112: That is the tricky part which I believe is the reason for 1C team problem:
agent name itself should never contain "\", only "/".

neuro_san/client/thinking_file_message_processor.py
d1donlydfink: We shouldn't be creating directories for these messages. We should be using the safe agent name which replaces "/" with "__".  Otherwise people will never find them.

neuro_san/internals/graph/persistence/agent_filetree_mapper.py
d1donlydfink: Ah. This could help them.

neuro_san/internals/graph/persistence/agent_filetree_mapper.py
d1donlydfink: Tool names themselves need to conform to this regex:

```
   # This comes from the langchain error message that happens when a tool name is not valid
    TOOL_NAME_PATTERN: str = r"^[a-zA-Z0-9_-]+$"
```

Note that "/" or "\" is not among those characters.
This is why we need that safe agent name thing in certain places. This might be one of them.

neuro_san/internals/graph/registry/activation_factory.py
d1donlydfink: See comment about about regex for agent names.
This might be OK because it's all about coded tool paths.

neuro_san/internals/run_context/utils/external_agent_parsing.py
```diff
@@ -135,6 +135,7 @@ def get_safe_agent_name(agent_url: str) -> str:
             # FWIW: langchain internal tool references must satisfy the regex: "^[a-zA-Z0-9_-]+$"
             # It's possible that more complex external references might have the agent_name
             # needing further mangling.  Cross that bridge when we have a real example.
+            # As a part of valid URL, agent_name can only have "/" in it.
             safe_name = "__" + agent_location.get("agent_name").replace("/", "__")
```
d1donlydfink, line 139: And you have discovered where the safe name thing can be called.

neuro_san/client/thinking_file_message_processor.py
d1donlydfink: Theoretically this shouldn't be necessary *if* the safe names are already happening at this level.
I have seen them for external references, so I think this is unnecessary.

neuro_san/internals/run_context/utils/external_agent_parsing.py
andreidenissov-cog: Yes, and it happened in previous PR already merged for some time.

neuro_san/internals/graph/persistence/agent_filetree_mapper.py
andreidenissov-cog: No it is not.
This is about mapping our agent names - which we agreed have general format of abc/xyz/my_agent
and local file path of agent definition file in registry file tree.
This TOOL_NAME_PATTERN is about LangChain limitations on their tool names.

neuro_san/client/thinking_file_message_processor.py
andreidenissov-cog: Without creating these missing directories, client will crash when calling structured agent abc/xyz.
So it's either this or creating abc__xyz thinking file.

d1donlydfink: Yes. abc__xyz thinking file.  Or make a flag for dirs vs not dirs in the client.
Personally I would not want to see a directory hierarchy.

neuro_san/coded_tools/deep/math_guy/__init__.py
andreidenissov-cog, line 1: All 4 files under neuro_san/coded_tools/deep/math_guy 
are put here for deep/math_guy agent network to work properly - 
path under coded_tools must replicate agent network path.

neuro_san/client/thinking_file_message_processor.py
```diff
@@ -131,6 +131,8 @@ def _write_to_file(self, origin_filename: str, origin_str: str,
         if self.thinking_dir:
             if origin_filename is None or len(origin_filename) == 0:
                 return
+            # In case origin_filename is a structured agent name, convert it with "__"
+            origin_filename = origin_filename.replace("/", "__")
             filename = Path(self.thinking_dir, origin_filename)
```
andreidenissov-cog, line 136: So instead of building a file tree, we maintain a flat structure with thinking files renaming.

PR #473, opened 2025-10-11
Refactor Validators to allow for future kinds of validation

d1donlydfink:
The thought occurred to me that while we do indeed want to validate AgentNetworks, there are other kinds of dictionary representations we get from hocon files that we will want to validate as well.  LLM info is likely the next thing.
So this is a relatively simple refactor to allow room for that:

* Rename AgentNetworkValidator to DictionaryValidator and rename the arg passed to validate() to be "nominee"
* Rename CompositeNetworkValidator to be CompositeDictionaryValidator and move it to internals.validation.common, as it is generic enough to be useful for anything else we want to validate
* Move the existing network validators into internals.validation.network.  This will allow other kinds of validators to be grouped according to what they validate.

Just about all of this is renames and import moves - so easy.

Wanted to do this sooner so that Agent Network Designer can start depending on this in relative peace.

### DIFF
--- a/neuro_san/internals/interfaces/dictionary_validator.py
+++ b/neuro_san/internals/interfaces/dictionary_validator.py
@@ -19,11 +19,11 @@ class DictionaryValidator:
     An interface for validating dictionaries of various meanings.
     """
 
-    def validate(self, nominee: Dict[str, Any]) -> List[str]:
+    def validate(self, candidate: Dict[str, Any]) -> List[str]:
         """
         Validate the given dictionary
 
-        :param nominee: The dictionary to validate
+        :param candidate: The dictionary to validate
         :return: A list of error messages
         """
         raise NotImplementedError
--- a/neuro_san/internals/validation/common/composite_dictionary_validator.py
+++ b/neuro_san/internals/validation/common/composite_dictionary_validator.py
@@ -29,16 +29,16 @@ class CompositeDictionaryValidator(DictionaryValidator):
         """
         self.validators: List[DictionaryValidator] = validators
 
-    def validate(self, nominee: Dict[str, Any]) -> List[str]:
+    def validate(self, candidate: Dict[str, Any]) -> List[str]:
         """
         Validate the agent network.
 
-        :param nominee: The dictionary to validate
+        :param candidate: The dictionary to validate
         :return: A list of error messages
         """
         errors: List[str] = []
 
-        if not nominee:
+        if not candidate:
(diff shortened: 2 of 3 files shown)


neuro_san/internals/graph/persistence/registry_manifest_restorer.py
```diff
@@ -32,7 +32,7 @@
 from neuro_san.internals.graph.persistence.agent_filetree_mapper import AgentFileTreeMapper
 from neuro_san.internals.graph.persistence.agent_network_restorer import AgentNetworkRestorer
 from neuro_san.internals.graph.registry.agent_network import AgentNetwork
-from neuro_san.internals.validation.manifest_network_validator import ManifestNetworkValidator
+from neuro_san.internals.validation.network.manifest_network_validator import ManifestNetworkValidator
```
d1donlydfink, line 35: Typical import move for existing Validator implementations.

neuro_san/internals/interfaces/dictionary_validator.py
```diff
@@ -14,16 +14,16 @@
 from typing import List
 
 
-class AgentNetworkValidator:
+class DictionaryValidator:
```
d1donlydfink, line 17: Rename for higher level re-use.

neuro_san/internals/validation/common/composite_dictionary_validator.py
```diff
@@ -13,36 +13,40 @@
 from typing import Dict
 from typing import List
 
-from neuro_san.internals.interfaces.agent_network_validator import AgentNetworkValidator
+from neuro_san.internals.interfaces.dictionary_validator import DictionaryValidator
 
 
-class CompositeNetworkValidator(AgentNetworkValidator):
+class CompositeDictionaryValidator(DictionaryValidator):
```
d1donlydfink, line 19: Rename for higher level reuse.

neuro_san/internals/validation/common/composite_dictionary_validator.py
```diff
@@ -13,36 +13,40 @@
...
         errors: List[str] = []
 
-        if not agent_network:
-            errors.append("Agent network is empty.")
+        if not nominee:
+            errors.append("Nothing to validate.")
+            return errors
+
+        if self.validators is None or len(self.validators) == 0:
+            errors.append("No validation policy.")
```
d1donlydfink, line 46: Add a check in case there is no policy to validate against.

PR #478, opened 2025-10-14
Use pypi Trusted Publishing as per best practice

donn-leaf:
Based on best practice as outlined in this blog post. Hat tip to @sidcog for pointing this out to me.

https://blog.pypi.org/posts/2025-09-16-github-actions-token-exfiltration/

This PR gets us away from a long-lived token and instead uses pypi's Trusted Publishing mechanism. The change is three parts. 
1. Configure project at pypi.org to use Trusted Publishing
2. Give the workflow the necessary permissions to use Trusted Publishing
3. Remove the long-lived token from the workflow

 Once this gets merged and we test the full release process, we should convert our other pypi-publishing repos in the same way.

### DIFF
--- a/docs/manifest_hocon_reference.md
+++ b/docs/manifest_hocon_reference.md
@@ -26,15 +26,38 @@ are file references ending with the .hocon extension. Each of these *keys* point
 
 ### Values for File Name Keys
 
-The value for any filename key is currently a boolean value.
+The value for any filename key is either a boolean value or a dictionary.
+
+#### Boolean value
 
 When the value is true, the agent described by the file key is served by the neuro-san server infrastructure
 and listed in the Concierge Service which lists all the available agents on the server.
 
 When the value is false, the agent described by the file key is neither served nor listed by the Concierge Service.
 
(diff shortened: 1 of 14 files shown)


.github/workflows/publish.yml
```diff
@@ -14,6 +14,8 @@ jobs:
     if: github.repository == 'cognizant-ai-lab/neuro-san'
     runs-on: ubuntu-latest
     environment: pypi  # Match this to your configured GitHub Environment.
+    permissions:
+      id-token: write #Mandatory for PyPi Trusted Publishing (configured at pypi.org)
```
donn-leaf, line 18: The comment speaks for itself.

.github/workflows/publish.yml
```diff
@@ -34,8 +36,6 @@ jobs:
 
       - name: Publish to PyPI
         uses: pypa/gh-action-pypi-publish@release/v1
-        with:
-          password: ${{ secrets.PYPI_API_TOKEN }}
```
donn-leaf, line 38: Et voilà, we no longer need a long-lived secret.

PR #485, opened 2025-10-20
Initial version bump to langchain==1.0.x

d1donlydfink:
Do what's necessary to upgrade to langchain==1.0.x:

* Change requirements.txt and requirements-build.txt to have new version caps.
* **NVIDIA is the only LLM provider which is not on top of the new dependencies for langchain==1.0.x.**  Commenting it out for now, and our lazy dependency loading for LLM providers will help most cases. This could change in a matter of days (I'd hope). See https://pypi.org/project/langchain-nvidia-ai-endpoints/#history for latest
* A number of imports have to change to conform to langchain==1.0.x. FWIW: **With this change nothing comes from langchain any more**. Only langchain_core/classic/community.  This should propagate up to the dependent repos eventually as well (studio, nsflow, etc).
* Some of our imports need to come from a new "langchain-classic" package.  Comments in source say agent stuff in particular will need to come from langgraph. This is already a transient dependency, so might not be so bad. Not for this PR though.
* Move the readme_renderer dependencies out of the GitHub action definition and into requirements-build.txt
* readme_renderer was having some problems installing from a new release as of this morning, so cap the version there to get past those problems.

### DIFF
--- a/neuro_san/internals/run_context/langchain/toolbox/toolbox_factory.py
+++ b/neuro_san/internals/run_context/langchain/toolbox/toolbox_factory.py
@@ -21,7 +21,7 @@ from typing import Union
 
 import os
 
-from langchain.tools import BaseTool
+from langchain_core.tools.base import BaseTool
 from langchain_community.agent_toolkits.base import BaseToolkit
 from pydantic import BaseModel
 
--- a/tests/neuro_san/internals/run_context/langchain/toolbox/test_toolbox_factory.py
+++ b/tests/neuro_san/internals/run_context/langchain/toolbox/test_toolbox_factory.py
@@ -12,9 +12,10 @@
 
 from unittest.mock import patch, MagicMock
 
-from langchain.tools import BaseTool
-from langchain_community.agent_toolkits.base import BaseToolkit
 import pytest
+
+from langchain_community.agent_toolkits.base import BaseToolkit
+from langchain_core.tools.base import BaseTool
 
 from neuro_san.internals.run_context.langchain.toolbox.toolbox_factory import ToolboxFactory
 


.github/workflows/tests.yml
```diff
@@ -56,7 +56,6 @@ jobs:
 
       - name: Check README renders correctly on PyPI
         run: |
-          pip install readme_renderer readme_renderer[md]
```
d1donlydfink, line 59: Don't install build dependencies in the action itself.
List them in requirements-build.txt

neuro_san/internals/run_context/langchain/core/langchain_run.py
```diff
@@ -12,7 +12,7 @@
 from typing import Any
 from typing import List
 
-from langchain.schema import BaseMessage
+from langchain_core.messages.base import BaseMessage
```
d1donlydfink, line 15: Some imports need to be fixed for adoption of langchain==1.0.x

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
```diff
@@ -27,12 +27,12 @@
...
 
-from langchain.agents import Agent
-from langchain.agents import AgentExecutor
-from langchain.agents.tool_calling_agent.base import create_tool_calling_agent
-from langchain.base_language import BaseLanguageModel
-from langchain.callbacks.tracers.logging import LoggingCallbackHandler
+from langchain_classic.agents import Agent
+from langchain_classic.agents import AgentExecutor
+from langchain_classic.agents.tool_calling_agent.base import create_tool_calling_agent
+from langchain_classic.callbacks.tracers.logging import LoggingCallbackHandler
```
d1donlydfink, line 33: Interesting to see some necessary imports move to what's now called "langchain_classic".
Comments in langchain_core point to using langgraph as the basis library for agents.

For this first blush we will use langchain_classic.

requirements-build.txt
```diff
@@ -2,11 +2,12 @@
...
-langchain-aws>=0.2.27,<0.3
-langchain-google-genai>=2.1.8,<3.0
-langchain-nvidia-ai-endpoints>=0.3.8,<0.4
-langchain-ollama>=0.3.4,<0.4
+langchain-anthropic>=0.3.11,<1.1
+langchain-aws>=0.2.27,<1.1
+langchain-google-genai>=2.1.8,<3.1
+# Note: As of 10/20/2025 langchain-nvidia is not ready for langchain==1.0.x
+#langchain-nvidia-ai-endpoints>=0.3.8,<0.4
+langchain-ollama>=0.3.4,<1.1
```
d1donlydfink, line 10: Put a new cap for versions at <1.1 so we only get langchain==1.0.x

Also worth noting that nvidia is not necessarily on top of things.
nvidia llms are already lazily loaded, so we are generally fine.

@vince-leaf : If we decide to push this as a new version of neuro-san to release early, we might need to press pause on nvidia-specific tests until they catch up.  It might only take a few days though.

requirements-build.txt
```diff
@@ -32,3 +33,9 @@ build
 setuptools>=61.0
 wheel
 setuptools-scm==8.1.0
+
+# Tools that check that README renders correctly on PyPi
+readme_renderer
+readme_renderer[md]
```
d1donlydfink, line 39: Move the readme_renderer dependencies from the tests.yml into requirements-build.txt.

requirements-build.txt
```diff
@@ -32,3 +33,9 @@ build
...
+# Note: Workaround for install problem with cmarkgfm==2025.11.20
+cmarkgfm<=2024.11.20
```
d1donlydfink, line 41: There is an install problem with the latest version of cmarkgfm that was just released early this morning.

requirements.txt
```diff
@@ -34,8 +34,8 @@ hvac>=1.1.0
 # For LLM access
 # Keep all these guys limited for now, as it can be the wild wild west out there
 # We don't want to be on a dependencies treadmill with timing not of our choosing.
-langchain>=0.3.15,<0.4
-langchain-community>=0.3.19,<0.4
+langchain>=0.3.15,<1.1
+langchain-community>=0.3.19,<1.1
```
d1donlydfink, line 38: Bump base requirements up to allow for langchain==1.0.x

.github/workflows/tests.yml
```diff
@@ -26,8 +26,7 @@ jobs:
       - name: Install dependencies
         run: |
           apt-get update && apt-get install -y shellcheck
-          pip install -r requirements-build.txt
-          pip install -r requirements.txt
+          pip install -r requirements.txt -r requirements-build.txt
```
d1donlydfink, line 29: Install both requirements files in the same command line.
This captures that requirements interactions between the 2 a little better.

neuro_san/internals/run_context/langchain/core/langchain_run.py
Noravee: We can also use `langchain.messages` since it is re-exported from `langchain-core`.

neuro_san/internals/run_context/langchain/toolbox/toolbox_factory.py
```diff
@@ -21,7 +21,7 @@
 
 import os
 
-from langchain.tools import BaseTool
+from langchain_core.tools.base import BaseTool
```
Noravee, line 24: We can also use `langchain.messages` since it is re-exported from `langchain-core`.

requirements-build.txt
donn-leaf: We have this readme_renderer install in all of our public python repos.

donn-leaf (approved):
LGTM WRT the github actions stuff.

neuro_san/internals/run_context/langchain/toolbox/toolbox_factory.py
d1donlydfink: I can see a near-future PR where we no longer actually depend on langchain any more.
This is why I head in the more separable langchain_core direction.

andreidenissov-cog (approved):
Looks good.

requirements-build.txt
vince-leaf: From my side, there is currently no NVIDIA test case.

requirements.txt
ofrancon: Will `lanchain_classic` be available for people using `langchain<1.0`?
We have a direct dependency to it now, because of imports like:
```
from langchain_classic.agents.output_parsers.tools import ToolsAgentOutputParser
```
So I don't think our code is compatible with older versions of lanchain anymore, is it?

PR #492, opened 2025-10-21
ASD-GH-436 MCP server for neuro-san

andreidenissov-cog:
This PR is a first MVP-level implementation of MCP service for neuro-san.
Tested by running an agent using MCP tool through Langchain MCP adapter, MCP server for this tool being neuro-san instance itself. We used both separate neuro-san instance and also the same one.
Things to note:

1. MCP service implements (kinda) latest 2026-06-18 protocol, in partial subset;
2. MCP service always runs - when neuro-san runs http service, MCP endpoint is also available for the same set of agent networks;
3. Since MCP payload is a JSON-RPC structure, we have actually a single Tornado-level handler, dispatching various MCP requests to different internal functions;
4. Underlying sessions we use are async;
5. True streaming http requests are not yet implemented, "tool/call" request returns a single result from chat request stream: message(s) of type AGENT_FRAMEWORK;
6. All incoming MCP requests are validated against published protocol schema, which is pretty convenient - if validation passed, you know exactly what you are dealing with at least format-wise.


neuro_san/service/http/server/http_server.py
```diff
@@ -68,6 +70,7 @@ class HttpServer(AgentAuthorizer, AgentStateListener):
     def __init__(self,
                  server_context: ServerContext,
                  server_config: HttpServerConfig,
+                 mcp_server_context: MCPServerContext,
```
d1donlydfink: Would this be better off as a part of ServerContext?

neuro_san/service/mcp/context/mcp_server_context.py
```diff
@@ -0,0 +1,63 @@
...
+See cccccbulijhliuvfhlirlbhncljehjjhttvunhdnhfeg
+class comment for details
+"""
+import json
+
+from neuro_san.internals.interfaces.dictionary_validator import DictionaryValidator
+from neuro_san.service.mcp.validation.mcp_request_validator import MCPRequestValidator
+from neuro_san.service.mcp.session.mcp_session_manager import MCPSessionManager
+
+MCP_VERSION: str = "2025-06-18"
```
d1donlydfink: How about a comment describing what this corresponds to with a link?

neuro_san/service/mcp/context/mcp_server_context.py
```diff
@@ -0,0 +1,63 @@
...
+# Issued under the Academic Public License.
+#
+# You can be released from the terms, and requirements of the Academic Public
+# License by purchasing a commercial license.
+# Purchase of a commercial license is mandatory for any use of the
+# neuro-san SDK Software in commercial settings.
+#
+# END COPYRIGHT
+"""
+See cccccbulijhliuvfhlirlbhncljehjjhttvunhdnhfeg
```
d1donlydfink: See what?

neuro_san/service/mcp/handlers/mcp_root_handler.py
```diff
@@ -0,0 +1,296 @@
...
+            error_msg: Dict[str, Any] =\
+                MCPErrorsUtil.get_protocol_error(request_id, MCPError.InvalidSession, extra_error)
+            self.set_status(HTTPStatus.UNAUTHORIZED)
+            self.write(error_msg)
+            self.logger.error(self.get_metadata(), f"error: {extra_error}")
+            self.do_finish()
+            return
+
+        try:
+            if method == "tools/list":
```
d1donlydfink, line 153: Would be really nice to have links to protocol documentation here for each if-block and in your processors as to which piece of the protocol you are implementing.

neuro_san/service/mcp/session/mcp_client_session.py
```diff
@@ -0,0 +1,45 @@
...
+# END COPYRIGHT
+"""
+See class comment for details
+"""
+
+
+class MCPClientSession:
+    """
+    Class representing a client session with the MCP service.
+    """
```
d1donlydfink, line 20: Maybe add some docs about the expected lifetime of these McpClientSessions.
My fear is the expectation that these will live beyond a single request.
If so, this is not going scale across pods, and we'll need to be very very clear about scaling expectation in our docs/md files.

neuro_san/service/http/server/http_server.py
```diff
@@ -204,6 +208,10 @@ def make_app(self, requests_limit: int, logger: EventLoopLogger):
         handlers.append((r"/api/v1/(.+)/connectivity", ConnectivityHandler, request_initialize_data))
         handlers.append((r"/api/v1/(.+)/streaming_chat", StreamingChatHandler, request_initialize_data))
 
+        # Register MCP "root" handler:
+        mcp_request_initialize_data: Dict[str, Any] = self.build_mcp_request_data()
+        handlers.append((r"/mcp", MCPRootHandler, mcp_request_initialize_data))
```
d1donlydfink: It might be best to put this new capability behind an environment variable switch.
It's conceivable that not every server would want mcp-ness activated.
You could use this as a franken-switch as well with a default value of False for now if you are considering this still under construction, and then later on True for easy bring-up.

neuro_san/service/mcp/processors/mcp_tools_processor.py
```diff
@@ -0,0 +1,151 @@
...
+        List available MCP tools.
+        :param request_id: MCP request id;
+        :param metadata: http-level request metadata;
+        :return: json dictionary with tools list in MCP format
+        """
+        public_storage: AgentNetworkStorage = self.network_storage_dict.get("public")
+        data: Dict[str, Any] = {}
+        session: ConciergeSession = DirectConciergeSession(public_storage, metadata=metadata)
+        result_dict: Dict[str, Any] = session.list(data)
+        tools_description: List[Dict[str, Any]] = []
```
d1donlydfink, line 56: It would make sense to expose an mcp switch in the default dictionary that is returned for each entry in  ManifestDictConfigFilter for greater flexibility of mixing/matching than just what the Concierge service returns.

It might be the case that:
* not every public network wants to be mcp exposed
* maybe some protected networks want to be mcp exposed.

That said it might make sense to have mcp exposure on by default for ease of use.

neuro_san/service/http/server/http_server.py
andreidenissov-cog: OK put it as a component of ServerContext.

neuro_san/service/http/handlers/base_request_handler.py
```diff
@@ -43,11 +43,7 @@ class BaseRequestHandler(RequestHandler):
     # pylint: disable=attribute-defined-outside-init
     # pylint: disable=too-many-arguments
     # pylint: disable=too-many-positional-arguments
-    def initialize(self,
-                   agent_policy: AgentAuthorizer,
-                   forwarded_request_metadata: List[str],
-                   openapi_service_spec_path: str,
-                   network_storage_dict: Dict[str, AgentNetworkStorage]):
+    def initialize(self, **kwargs):
         """
```
andreidenissov-cog, line 47: Change how we pass parameters to handler "initialize" method - explicit parameters dictionary
is much more flexible in that we can split parameters passed to a base class vs. parameters used by a sub-class.

neuro_san/service/mcp/processors/mcp_prompts_processor.py
```diff
@@ -0,0 +1,44 @@
...
+        :return: json dictionary with prompts list in MCP format
+        """
+        _ = metadata
+        return {
+            "jsonrpc": "2.0",
+            "id": RequestsUtil.safe_request_id(request_id),
+            "result": {
+                "prompts": []
+            }
+        }
```
andreidenissov-cog, line 45: We don't really have anything for MCP resources or prompts, so we just return empty lists.

neuro_san/service/utils/server_status.py
```diff
@@ -42,6 +43,7 @@ def is_server_ready(self) -> bool:
         return \
             (not self.grpc_service.is_requested() or self.grpc_service.is_ready()) and \
             (not self.http_service.is_requested() or self.http_service.is_ready()) and \
+            (not self.mcp_service.is_requested() or self.mcp_service.is_ready()) and \
             (not self.updater.is_requested() or self.updater.is_ready())
```
andreidenissov-cog, line 47: Extend server status with "mcp" service status.

requirements.txt
```diff
@@ -57,6 +57,9 @@ psutil>=7.0.0
 # HTTP server part
 tornado>=6.4.2
 
+# For JSON schema validation
+jsonschema>=4.19.0
+
```
andreidenissov-cog, line 62: Add a dependency for validating a json structure against provided schema.

neuro_san/service/http/server/http_server.py
andreidenissov-cog: OK default is False now for enabling MCP.
As for enabling/disabling individual agent networks in MCP service,
that's for another PR, because logic will not be so trivial.

neuro_san/service/mcp/handlers/mcp_root_handler.py
andreidenissov-cog: Several main links are added - for major classes of MCP requests.

neuro_san/service/main_loop/server_main_loop.py
```diff
@@ -136,6 +136,9 @@ def prepare_args(self) -> ArgumentParser:
                                                            DEFAULT_HTTP_SERVER_MONITOR_INTERVAL_SECONDS)),
                                 help="Http server resources monitoring/logging interval in seconds "
                                      "0 means no logging")
+        arg_parser.add_argument("--mcp_enable", type=str,
+                                default=os.environ.get("AGENT_MCP_ENABLE", "false"),
+                                help="'true' if MCP protocol service should be enabled")
```
andreidenissov-cog, line 141: Default is False - we will not enable MCP service unless we specify that explicitly.

d1donlydfink: Very good choice for now. We might want to reverse this as this feature becomes more mature.
If you think this feature is ready enough with this PR, you should also put this in the Dockerfile with docs.
If you want to wait, please at least file a ticket for it.

neuro_san/service/mcp/handlers/mcp_root_handler.py
```diff
@@ -0,0 +1,297 @@
...
+                return
+        except json.JSONDecodeError as exc:
+            error_msg: Dict[str, Any] =\
+                McpErrorsUtil.get_protocol_error(
+                    request_id,
+                    McpError.ParseError,
+                    "Invalid JSON format")
+            self.set_status(HTTPStatus.BAD_REQUEST)
+            self.write(error_msg)
+            self.logger.error(self.get_metadata(), "error: Invalid JSON format: %s", str(exc))
```
d1donlydfink, line 105: This seems like a reasonable request from sec-bot.
I think the expectation is to not give the client any information about the exception cuz those contribute to
hacking attacks. Seems like the ask here is to not do the self.write() with information from the exception itself. Here are a few other places below.

neuro_san/service/mcp/handlers/mcp_root_handler.py
```diff
@@ -0,0 +1,297 @@
...
+        # Set all allowed headers:
+        self.set_header("Access-Control-Allow-Headers", headers)
+
+    async def post(self):
+        """
+        Implementation of top-level POST request handler for MCP call.
+        """
+        # pylint: disable=too-many-locals
+        # pylint: disable=too-many-branches
+        # pylint: disable=too-many-statements
```
d1donlydfink, line 76: These pylint complaints are fair.
Should probably break this method up so it becomes an outline method.
Doesn't have to happen now, but soon.

neuro_san/service/mcp/handlers/mcp_root_handler.py
```diff
@@ -0,0 +1,297 @@
...
+            self.do_finish()
+            return
+        session_active: bool = False
+        if session_id is not None:
+            session_manager: McpSessionManager = self.mcp_context.get_session_manager()
+            session_active = session_manager.is_session_active(session_id)
+        if not session_active:
+            extra_error: str = "invalid or inactive session id"
+            error_msg: Dict[str, Any] =\
+                McpErrorsUtil.get_protocol_error(request_id, McpError.InvalidSession, extra_error)
```
d1donlydfink, line 145: OK so I get that MCP wants us to return a session id. That's fine, we can play that game.
Here it wants us to check on that session_id to be sure it's valid.

It's this session_id business that is the complicating/unscalable part.
Seems like nowhere else is this session_id actually used except for this bit of gatekeeper logic.
What if we made this session_id check at this point somewhat optional for our implementation?

See proposal in McpSessionManager below.

neuro_san/service/mcp/handlers/mcp_root_handler.py
```diff
@@ -0,0 +1,297 @@
...
+            self.set_status(HTTPStatus.UNAUTHORIZED)
+            self.write(error_msg)
+            self.logger.error(self.get_metadata(), f"error: {extra_error}")
+            self.do_finish()
+            return
+
+        try:
```
d1donlydfink, line 152: FWIW: If statements in this try-block would be ripe for a factory approach to clear some room for the too-many-* pylint issues in this post() method.

neuro_san/service/mcp/processors/mcp_tools_processor.py
```diff
@@ -0,0 +1,155 @@
...
+        self.network_storage_dict: AgentNetworkStorage = network_storage_dict
+        self.agent_policy: AgentAuthorizer = agent_policy
+
+    async def list_tools(self, request_id, metadata: Dict[str, Any]) -> Dict[str, Any]:
+        """
+        List available MCP tools.
+        :param request_id: MCP request id;
+        :param metadata: http-level request metadata;
+        :return: json dictionary with tools list in MCP format
+        """
```
d1donlydfink, line 51: If you took the aforementioned factory approach, ListTools would likely be in a separate class from CallTool

neuro_san/service/mcp/processors/mcp_tools_processor.py
d1donlydfink: Reading this again, this is still a good idea. Doesn't have to be for this PR tho.

neuro_san/service/mcp/processors/mcp_tools_processor.py
```diff
@@ -0,0 +1,155 @@
...
+                },
+                "required": ["input"]
+            }
+        }
+
+    async def _extract_tool_response_part(self, response_dict: Dict[str, Any]) -> str:
+        response_part_dict: Dict[str, Any] = response_dict.get("response", {})
+        response_type: str = response_part_dict.get("type", "")
+        if response_type == "AGENT_FRAMEWORK":
+            return response_part_dict.get("text", None)
```
d1donlydfink, line 139: This is OK for getting things going, but please file a ticket to be sure the structure information gets in here as well.  See BasicMessageProcessor.get_compiled_answer()

Also separate future work would involve allowing specific sly_data fields through. Nothing comes through by default, so you have that covered already.  Will need to think on best place in hocon to do that.

neuro_san/service/mcp/session/mcp_session_manager.py
```diff
@@ -0,0 +1,102 @@
...
+from neuro_san.service.mcp.session.mcp_client_session import McpClientSession
+
+MCP_SESSION_ID: str = "Mcp-Session-Id"
+MCP_PROTOCOL_VERSION: str = "MCP-Protocol-Version"
+
+
+class McpSessionManager:
+    """
+    Class creating and managing client sessions with the MCP service.
+    """
```
d1donlydfink, line 31: Problem:
SessionManager session id checks prevent horizontal scalability across pods as-is.
Luckily, the way we actually use session_ids in McpRootHandler, they are really just some non-functional gatekeeping for our implementation.

Proposal:
* McpSessionManager becomes an interface
* PermissiveMcpSessionManager.is_session_active() (at least) always returns true
* SingleServerLockdownMcpSessionManager.is_session_active() does lookup like now
* MultiServerLockdownMcpSessionManager.is_session_active() does some kind of db lookup for cross-pod and needs more configuration.
* have lazy binding of which class to use via env var AGENT_MCP_SESSION_MANAGER (say) with SingleServer the default.

Probably fine/necessary to issue session_id as BS to keep clients happy in all cases.

We could defer the implementation of MultiServerLockdownMcpSessionManager to the community since there are probably too many implementations for us to want to fathom at this point (not unlike syncing temporary agent network storage).

Default could be SingleServerLockdown. We could use Permissive in our initial deployments until we come up with a MultiServer solution at our liesure.

Thoughts?

neuro_san/service/mcp/util/mcp_errors_util.py
```diff
@@ -0,0 +1,85 @@
...
+#
+# END COPYRIGHT
+"""
+See class comment for details
+"""
+from typing import Any
+from typing import Dict
+
+from neuro_san.service.mcp.mcp_errors import McpError
+from neuro_san.service.mcp.util.requests_util import RequestsUtil
```
d1donlydfink, line 19: FWIW: Seems like mcp_errors should be pulled into this util package.
Makes inter-package dependencies clearer.

d1donlydfink (changes requested):
* Still need to address sec-bot complaints before this goes in.
    We will get dinged from corp-scans and might even have to issue a security alert
    if this goes in as-is.

* Consider:
    *  breaking up RootHandler.post() using a factory pattern
    *  Pulling mcp_errors into utils package
    * McpSessionManager proposal for pod-scalability (see PR comment there)

Please file the following tickets for future work:
* env var turning mcp on/off + Dockerfile example and docs in there
* Response needs more than just text from AGENT_FRAMEWORK message
    * structure (if present always report)  See BasicMessageProcessor.get_compiled_answer()
    * Enabling selective sly_data fields in mcp result. Default is None.
        Need to figure out best place to but this kind of knob in hocon specs 
* Enabling tools in manifest dictionaries for mcp exposure (public or protected)
* Work on McpSessionManager options in proposal

PR #498, opened 2025-10-22
Remove agent_llm_info_file and agent_toolbox_info_file keys

Noravee:
According to [user guide](https://github.com/cognizant-ai-lab/neuro-san-studio/blob/main/docs/user_guide.md#adding-tools-in-toolbox)
`agent_llm_info_file` and `agent_toolbox_info_file` will remain supported until neuro-san==0.6.0. Since the current version is 0.6.1, they should be removed accordingly.


Noravee:
The references have been removed from agent_hocon_reference.md.

`agent_llm_info_file` and `agent_toolbox_info_file` are keys for the agent network hocon file. I don't believe we should remove the ones in the docker file as those are env vars.

Issue #496, opened 2025-10-22
Infrequent exception when trying to access /external network


d1donlydfink:
Traceback (most recent call last): File '/usr/local/lib/python3.12/site-packages/aiohttp/client_proto.py', line 144, in connection_lost uncompleted = self._parser.feed_eof()
^^^^^^^^^^^^^^^^^^^^^^^ File 'aiohttp/_http_parser.pyx', line 508, in aiohttp._http_parser.HttpParser.feed_eof aiohttp.http_exceptions.TransferEncodingError: 400, message: Not enough data to satisfy transfer length header. 
The above exception was the direct cause of the following exception:
 Traceback (most recent call last): 
 File '/usr/local/lib/python3.12/site-packages/neuro_san/session/async_http_service_agent_session.py', line 119, in streaming_chat async for data in response.content.iter_chunked(max_chunk_size): 
 File '/usr/local/lib/python3.12/site-packages/aiohttp/streams.py', line 52, in __anext__ rv = await self.read_func()
 ^^^^^^^^^^^^^^^^^^^^^^ 
 File '/usr/local/lib/python3.12/site-packages/aiohttp/streams.py', line 428, in read await self._wait('read') 
 File '/usr/local/lib/python3.12/site-packages/aiohttp/streams.py', line 347, in _wait await waiter aiohttp.client_exceptions.ClientPayloadError: Response payload is not completed: 
 The above exception was the direct cause of the following exception:
 Traceback (most recent call last): File '/usr/local/lib/python3.12/site-packages/neuro_san/internals/run_context/langchain/core/langchain_openai_function_tool.py', line 213, in _arun run = await self.tool_caller.make_tool_function_calls(initial_run)
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 
 File '/usr/local/lib/python3.12/site-packages/neuro_san/internals/graph/activations/calling_activation.py', line 184, in make_tool_function_calls tool_output: Dict[str, Any] = await self.make_one_tool_function_call(component_tool_call)
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 
 File '/usr/local/lib/python3.12/site-packages/neuro_san/internals/graph/activations/calling_activation.py', line 224, in make_one_tool_function_call output: str = await callable_component.build()
[...]

PR #499, opened 2025-10-22
Add a retry if we hit a ClientPayloadError in an external tool

d1donlydfink:
Addresses #496

### DIFF
--- a/docs/agent_hocon_reference.md
+++ b/docs/agent_hocon_reference.md
@@ -144,10 +144,6 @@ This results in a final interpretation where the `function` value is:
 Value replacement only happens once, but you can have replacement_strings references within
 your string values within your replacement_values and things will work out as you might expect.
 
-### agent_llm_info_file
-
-Deprecated since version 0.5.46: Use `llm_info_file` instead. It will not be removed until `neuro-san==0.6.0`.
-
 ### llm_info_file
 
 The `llm_info_file` key allows you to specify a custom HOCON file that extends the default list of available LLMs used
@@ -155,10 +151,6 @@ by agents in a neuro-san network. This is especially useful if you're using mode
 the default configuration (e.g., newly released models or organization-specific endpoints).
 
 For more information on selecting and customizing models, see the [model_name](#model_name) and [class](#class) section below.
-
-### agent_toolbox_info_file
-
-Deprecated since version 0.5.46: Use `toolbox_info_file` instead. It will not be removed until `neuro-san==0.6.0`.
 
 ### toolbox_info_file
 
--- a/neuro_san/internals/graph/persistence/agent_filetree_mapper.py
+++ b/neuro_san/internals/graph/persistence/agent_filetree_mapper.py
@@ -20,6 +20,14 @@ class AgentFileTreeMapper(AgentNameMapper):
     between agent name as specified in a manifest file
     and a file path (relative to registry root directory) to this agent definition file.
     """
+    def __init__(self, path_method=Path):
+        """
+        Constructor
+
+        :param path_method: Optional Path method to use for path manipulations.
+            Default is pathlib.Path, but can be overridden for testing purposes.
+        """
+        self.path_method = path_method
 
     def agent_name_to_filepath(self, agent_name: str) -> str:
         """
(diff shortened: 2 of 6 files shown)


neuro_san/internals/graph/activations/external_activation.py
```diff
@@ -151,20 +153,33 @@ async def build(self) -> BaseMessage:
...
-                                "Cannot rely on results from it as a tool."
-            full_name: str = Origination.get_full_name_from_origin(self.run_context.get_origin())
-            logger: Logger = getLogger(full_name)
-            logger.info(messages_str)
-            ai_message = AIMessage(content=messages_str)
-            return ai_message
+        chat_responses: AsyncGenerator[Dict[str, Any], None] = None
+        error_str: str = None
+        retries_remaining: int = 2
+        while chat_responses is None and retries_remaining > 0:
```
d1donlydfink, line 162: Better viewed without whitespace.

Make a retry loop to try again if we ever get the infrequent ClientPayloadError

neuro_san/internals/graph/activations/external_activation.py
```diff
@@ -151,20 +153,33 @@ async def build(self) -> BaseMessage:
...
+                # Note that we are not await-ing the response here because what is returned is a generator.
+                # Proper await-ing for generator results is done in the "async for"-loop below.
+                chat_responses = self.session.streaming_chat(chat_request)
+
+            except ClientPayloadError:
+                # This error happens infrequently. Worth a retry to get past it.
+                retries_remaining -= 1
+                if retries_remaining == 0:
+                    error_str: str = f"Agent/tool {self.agent_url} was unreachable due to ClientPayloadError. " + \
+                                     "Cannot rely on results from it as a tool."
```
d1donlydfink, line 173: Make sure the error message is unique from its friend downstairs.

neuro_san/session/async_http_service_agent_session.py
```diff
@@ -146,7 +147,7 @@ async def streaming_chat(self, request_dict: Dict[str, Any]) -> Generator[Dict[s
                         result_dict = json.loads(accumulator.decode("utf-8"))
                         yield result_dict
 
-        except (asyncio.TimeoutError, ClientOSError) as exc:
+        except (asyncio.TimeoutError, ClientOSError, ClientPayloadError) as exc:
```
d1donlydfink, line 150: Let the infrequent ClientPayloadError through so it can be handled by calling code.

neuro_san/internals/graph/activations/external_activation.py
```diff
@@ -151,20 +153,36 @@ async def build(self) -> BaseMessage:
...
+                else:
+                    logger.warning("Agent/tool %s was unreachable due to ClientPayloadError. Retrying.", self.agent_url)
```
d1donlydfink, line 175: Log when we are retrying.

neuro_san/internals/graph/activations/external_activation.py
```diff
@@ -151,20 +153,36 @@ async def build(self) -> BaseMessage:
...
+
+            except ValueError:
+                # Could not reach the server for the external agent, so tell about it
+                error_str: str = f"Agent/tool {self.agent_url} was unreachable. " + \
+                                 "Cannot rely on results from it as a tool."
+
+            if error_str is not None:
+                logger.info(error_str)
+                ai_message = AIMessage(content=error_str)
+                return ai_message
```
d1donlydfink, line 185: Common error/early return code.

neuro_san/internals/graph/activations/external_activation.py
andreidenissov-cog: Are you sure that's the place where ClientPayloadError will be raised?
Not in the async loop itself somewhere?

d1donlydfink: TBH, no I'm not sure, but this is where the backtrace was telling us to look.
Can regroup on proposed solution if there is still a problem.

d1donlydfink: This issue was resolved after in-person discussion and explanation about how the retry actually works.

Issue #507, opened 2025-10-24
Unit tests do not generate thinking files correctly


Noravee:
There should be thinking files at `tmp/agent_test` when running tests, but it is not generated.

PR #511, opened 2025-10-24
#507 Unit tests do not generate thinking files correctly

Noravee:
#507 

### Problem

When running tests, thinking files should be generated under `tmp/agent_test`, but they are currently not being created.

### Root Cause

During initialization, the `StreamingInputProcessor` sets the `thinking_file` to `None`, which prevents the `ThinkingFileMessageProcessor` from being added to the processing pipeline.

See the relevant code:
https://github.com/cognizant-ai-lab/neuro-san/blob/main/neuro_san/client/streaming_input_processor.py#L45.

### Test Verification

- Ran unit and integration tests and confirmed that thinking files are now correctly generated.

#### Additional Notes

There is an existing integration test for MCP tools (used as LangChain tools) that should have detected this issue #502. However, the bug occurs during text parsing within the thinking file, and since the thinking file processor was previously not included, the test passed silently.

### DIFF
--- a/neuro_san/deploy/Dockerfile
+++ b/neuro_san/deploy/Dockerfile
@@ -248,11 +248,11 @@ ENV AGENT_EXTERNAL_SERVER_URL=""
 # When not set, this defaults to a null implementation.
 # A simple default implementation is provided by the class
 # "neuro_san.service.watcher.temp_networks.s3_reservations_storage.S3ReservationsStorage"
-ENV AGENT_EXTERNAL_RESERVATIONS_STORAGE
+ENV AGENT_EXTERNAL_RESERVATIONS_STORAGE=""
 
 # When the AGENT_EXTERNAL_RESERVATIONS_STORAGE (immediately above) is set to the
 # S3ReservationsStorage class mentioned there, this env var must be set to the S3 Bucket
 # to use for cross-pod reservations storage.
-ENV AGENT_RESERVATIONS_S3_BUCKET
+ENV AGENT_RESERVATIONS_S3_BUCKET=""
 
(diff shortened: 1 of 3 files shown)


neuro_san/test/driver/data_driven_agent_test_driver.py
```diff
@@ -222,8 +224,18 @@ def interact(self, agent: str, session: AgentSession, interaction: Dict[str, Any
         # Prepare the processor
         now = datetime.now()
         datestr: str = now.strftime("%Y-%m-%d-%H:%M:%S")
+        thinking_file: str = f"/tmp/agent_test/{datestr}_agent.txt"
         thinking_dir: str = f"/tmp/agent_test/{datestr}_agent"
```
Noravee, line 228: Set thinking file to `/tmp/agent_test/{datestr}_agent.txt`

neuro_san/test/driver/data_driven_agent_test_driver.py
```diff
@@ -222,8 +224,18 @@ def interact(self, agent: str, session: AgentSession, interaction: Dict[str, Any
...
-        input_processor = StreamingInputProcessor("", None, session, thinking_dir)
+
+        # Remove any contents that might be there already.
+        # Writing over existing dir will just confuse output.
+        # Although it is unlikely that two tests run at the same time...
+        if os.path.exists(thinking_dir):
+            shutil.rmtree(thinking_dir)
+        # Create the directory anew
+        os.makedirs(thinking_dir)
```
Noravee, line 236: Create the directory.

Issue #509, opened 2025-10-24
Verifier scoffs at reference to localhost


d1donlydfink:
Per Siva:

When using the tool URL http://localhost/OnecPla, we encountered the following error:
       ["Agent 'learning' references non-existent agent(s) in tools: ' http://localhost/OnecPla'"]

PR #519, opened 2025-10-27
NS-509 Fix validator scoffing at http urls

d1donlydfink:
Addresses #509 which brings up the issue that the UrlNetworkValidator will scoff at any http:// references


neuro_san/internals/validation/network/url_network_validator.py
```diff
@@ -76,6 +76,8 @@ def check_safe_urls(self, agent_name: str, safe_tools: List[str], urls: List[str
         for tool in safe_tools:
             if self.is_url_or_path(tool) and \
                     tool not in urls and \
+                    not tool.startswith("http://") and \
+                    not tool.startswith("https://") and \
```
d1donlydfink, line 81: Allow any external agents that are over the network. We do not have enough information to know in this simple internal network validation step if the external network has the tool or not.

neuro_san/registries/deep/math_guy_passthrough.hocon
```diff
@@ -58,7 +58,7 @@ Return verbatim whatever it returns.
             # Call the math_guy as an "external tool". Tools that are prefixed
             # with a slash "/" are expected to reside as agents on the same server.
             # This allows us to test agent references in a directory hierachy.
-            "tools": ["/deep/math_guy"],
+            "tools": ["http://localhost/deep/math_guy"],
```
d1donlydfink, line 61: Make deep/math_guy_passthrough the test case

Issue #457, opened 2025-10-08
AgentToolResult message should not be in the chat history


d1donlydfink:
Dupe of: UN-3532

Issue #454, opened 2025-10-08
Smoke test Ollama test case failed tag [0.5.59]


d1donlydfink:
This failure smoke test case has been disabled. Please re-enable it once the issue has been addressed.

Note: No other LLM has this issue except Ollama. Also, I have confirmed this issue did NOT reproduce on tag 0.5.58. The problem started on tag 0.5.59.
[Releases · cognizant-ai-lab/neuro-san](https://github.com/cognizant-ai-lab/neuro-san/releases) 

I could reproduce this failure 100% locally on my laptop with agent_cli using connection http & direct.

PR #318, opened 2025-07-16
UN-3202 Support for Amazon Bedrock models and UN-3342 Add bedrock Amazon Bedrock models error due to system message in the chat history

Noravee:
### Summary
This PR adds support for `bedrock` models and fixes a bug involving **duplicate system prompts** in the message list sent to the LLM.

The issue occurs because both our implementation and LangChain independently insert a system message into the chat history. This becomes problematic for models—like those on `bedrock`—that **require the system message to appear only once and strictly in the first position.**

Before this fix, the system prompt was added manually **only in the first round** (when `chat_history` is `None` or empty), so we didn’t notice the issue with other chat models (e.g., OpenAI) because they tolerate system messages in any position. However, `bedrock` models are stricter and throw an error when multiple system messages are detected or when the system message is not the first in the sequence.

Additionally, we were treating `assignment` messages for branch nodes as system messages, which also violated this constraint and triggered similar errors. We’ve now updated `assignment` to be sent as a **human message** instead.

### Changes
- Fix double system prompts bug
    - `LangChainRunContext._create_prompt_template()`: Change the logic so that the template is always
    ```python
    [
        ("system", <system_prompt>),
        ("placeholder", "{chat_history}"),
        ("human", "{input}"),
        ("placeholder", "{agent_scratchpad}")
    ]
    ```
    - `OriginatingJournal.write_message()`
    Ensures that system messages are **not duplicated** by checking message type before writing to `chat_history`.
    > Note: This does **not** change any journal-writing behavior—only filters for system messages.

- Change `assignment` from system message to human message
    -  `LangChainRunContext._create_prompt_template()`
    Removed handling of `assignment` as system message
    - `BranchActivation.build()`
    Ensures `assignment` is included in the `command` and passed as a human message instead.

- Add `bedrock` class and models in default llm info file
    - `default_llm_info.hocon`
    Added `bedrock` class and the following models
        - `us.anthropic.claude-opus-4-20250514-v1:0`
        - `us.anthropic.claude-sonnet-4-20250514-v1:0`
        - `us.anthropic.claude-3-7-sonnet-20250219-v1:0`
    - `StandardLangChainLlmFactory.create_base_chat_model()`
    Added support for `ChatBedrock`
    - `agent_hocon_reference.md`
    Added `bedrock` in the table of the `model_name` section 
    - `requirements.txt`
    Added `langchain-aws>=0.2.27,<0.3`

### Tests
- Ran music nerd and intranet with `bedrock` and `openai` models
[...]

PR #520, opened 2025-10-28
#454 Smoke test Ollama test case failed tag [0.5.59]

Noravee:
## Update - Oct 30, 2025

- The focus of this PR has shifted to address **#454**.
(The discussion around tool output and chat history in **#457** will continue separately, particularly with the move from `AgentExecutor` to `create_agent`.)
- No changes are required in `OriginatingJournal`.

### Bug Summary (Issue #454)

This bug is particularly tricky because it is **prompt- and model-specific:**

- The **LLM gets stuck in a tool-call loop**, repeatedly invoking the same tool without ever producing a final response, until it reaches the maximum iteration limit.
- The issue occurs only with `qwen3` models.
- The bug is triggered when the first prompt is: “Who made Yellow Submarine?”
- However:
        - If the first prompt is “Where were the Beatles from?”, the model responds correctly, and you can ask “Who made Yellow Submarine?” afterward with no issue.
        - If the conversation starts with something simple like “hi” and then you ask “Who made Yellow Submarine?”, it also responds correctly.

### Key Finding

- Including the `tool_result_origin` field in the `AgentToolResult` unexpectedly prevents the tool-call loop.
- If the tool output is placed only in the content field (without the origin), the bug still occurs.

### Issue

In general, chat history should consist of alternated human and AI messages. Currently we add the `AgentToolResult` message to chat history by turning it to `AIMessage` to the chat history becomes something like [human, ai, ai, human , ...]
instead fo [human, ai, human, ai, ...]

### Changes

- Don't put tool message in the chat history in `OriginatingJournal`
- Create `get_tool_message()` method for `LangChainRun` and use it in `LangChainOpenaiFunctionTool._arun()`
- Put tool_message in when instantiate run in `LangChainRunContext.submit_tool_outputs()`

**Note:** For some reasons, this fixs #454 as well.

### Tests

- `music_nerd_pro`, `music_nerd_pro_llm_anthropic`, `music_nerd_pro_llm_gemini`, `music_nerd_pro_llm_ollama`, and `music_nerd_pro_multi_agents`
-  `math_guy`, `math_guy_passthrough`
- `esp_decision_assistant`

### DIFF
--- a/neuro_san/internals/utils/metadata_util.py
+++ b/neuro_san/internals/utils/metadata_util.py
@@ -43,6 +43,10 @@ class MetadataUtil:
         """
         minimized: Dict[str, Any] = {}
 
+        if request_metadata is None:
+            # No request data, nothing to fill
+            return minimized
+
         # If there are no keys, there is nothing to fill.
         if not keys:
             return minimized


neuro_san/internals/journals/originating_journal.py
```diff
@@ -68,31 +68,12 @@ async def write_message(self, message: BaseMessage, origin: List[Dict[str, Any]]
...
-            else:
-                chat_history_message = message
 
             # Langchain automatically adds the system prompt to the beginning of the chat history.
             # Ensure that the system message does not get added into the chat history.
-            if not isinstance(chat_history_message, SystemMessage):
-                self.chat_history.append(chat_history_message)
+            # Also avoid adding AgentToolResultMessages to chat history since chat history is for user/AI exchanges.
+            if not isinstance(message, SystemMessage) and not isinstance(message, AgentToolResultMessage):
+                self.chat_history.append(message)
```
Noravee: Avoid adding `SystemMessage` and `AgentToolResultMessage`.

neuro_san/internals/run_context/langchain/core/langchain_run.py
```diff
@@ -34,11 +36,14 @@ def __init__(self, run_id_base: str, chat_history: List[BaseMessage],
...
         :param args: The arguments used for the tool during the run
                     represented by this instance.
+        :param tool_message: The most recent tool message produced during the run
+                             represented by this instance.
         """
         self.id: str = self._create_run_id(run_id_base, chat_history)
         self.chat_history: List[BaseMessage] = chat_history
         self.tool_name: str = tool_name
         self.args: Any = args
+        self.tool_message: BaseMessage = tool_message
```
Noravee, line 51: Add `tool_message` argument in `LangChainRun` constructor.

neuro_san/internals/run_context/langchain/core/langchain_run.py
```diff
@@ -95,10 +100,17 @@ def model_dump_json(self) -> str:
         return "{ " + f"'id': '{self.id}'" + " }"
 
     def get_chat_history(self) -> List[BaseMessage]:
+        """
+        Get the most recent chat history.
+        :return: The list of messages comprising the chat history
+        """
+        return self.chat_history
+
+    def get_tool_message(self) -> BaseMessage:
```
Noravee, line 114: Create `get_tool_message()` to return the latest tool output.

neuro_san/internals/run_context/langchain/core/langchain_openai_function_tool.py
```diff
@@ -225,8 +226,6 @@ async def _arun(
         if run is None:
             return None
 
-        # Assume the last message in the Run's chat_history holds the answer.
-        # This eventually gets integrated into the CallingTool/tool_caller
-        # in its submit_tool_outputs() call for its RunContext.
-        the_answer = run.get_chat_history()[-1]
+        # Assume the answer is latest tool message in the Run.
+        the_answer: BaseMessage = run.get_tool_message()
```
Noravee, line 235: Use `get_tool_message()` to return tool output to LLM.

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
```diff
@@ -371,15 +371,18 @@ async def submit_tool_outputs(self, run: Run, tool_outputs: List[Dict[str, Any]]
...
                 if tool_message is not None:
-                    # Chat history is updated in write_message()
+                    # Write the tool message to the journal.
+                    # Chat history should only contain user and AI messages.
+                    # It will not be updated in the write_message() call below.
                     await self.journal.write_message(tool_message)
 
         # Create a run to return
-        run = LangChainRun(self.run_id_base, self.chat_history)
+        run = LangChainRun(self.run_id_base, self.chat_history, tool_message=tool_message)
```
Noravee, line 390: Add tool message to `LangChainRun` constructor in `LangChainRunContext.submit_tool_outputs()`.

neuro_san/internals/journals/originating_journal.py
andreidenissov-cog: Could be shorter Pythonic:
not isinstance(message, (SystemMessage, AgentToolResultMessage)):
just saying.

d1donlydfink: You are completely missing the spirit of the original if statement here by 2 items here.
There's no need to change the output of the other models because of Ollama at this level:

1. AgentToolResultMessages were translated into AI messages. That has always been reasonable for all other models, as AI messages, as a langchain-level message has always been valid in chat history.
2. You are skipping the SystemMessages. Now agent_thinking output will never know what the SystemMessage was, complicating debugging.

You need to look at the agent_thinking output for a few of the test networks like math_guy_passthrough and esp_decision_assistant and music_nerd_pro.

If the agent_thinking output is not exactly the same as it was before, this change is a fail.

Noravee: The `SystemMessage` has been skipped since https://github.com/cognizant-ai-lab/neuro-san/pull/318. I have also tested and checked the thinking files for following networks:
- music_nerd_pro, music_nerd_pro_llm_anthropic, music_nerd_pro_llm_gemini, music_nerd_pro_llm_ollama, and music_nerd_pro_multi_agents
- math_guy, math_guy_passthrough
- esp_decision_assistant

But there could be a better solution. I will keep looking.

d1donlydfink: OK I was missing that this is only effecting what is accumulating in chat history, and you are correct the system message is added elsewhere. I was conflating the chat history with the entire journal.

That still leaves the AIMessage though.  There are a couple purposes to this chat_history I believe.
1. This chat_history is what the chat_context that goes back to the user is based on. Removing the AI Messages can affect subsequent conversations. music_nerd_pro's 2nd response is likely to fail a test.
2. . The chat_history is used when A caller agent has multiple conversations with its callee agent(s).     This would imply that AAOSA conversations might not be as robust.

(1) would be moot if you could prove that the chat_context is not change (and not just for Ollama) when sent to the user
(2) could be moot if you could show that AAOSA conversation are still robust as well as (1) for an AAOSA network.

---

It still doesn’t make sense to me that we should be touching OriginatingJournal for an LLM-specific fix unless you can show that there are duplicate messages in the chat_history for all LLMs happening without that fix.

Worth noting that there is an OriginatingJournal.write_message_if_next_not_dupe() method that helps in avoiding duplicate messages.  Maybe that is part of the solution.

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
d1donlydfink: This part seems reasonable.

ofrancon (approved):

vince-leaf:
I ran the ollama test case on the smoketest on this local repo, and it passed.

PR #526, opened 2025-10-30
436 - fix generation of safe MCP server response.

andreidenissov-cog:
Turns out it matters:
when generating an "HTTP safe" service response,
do not convert request id from int to a string - MCP client get confused and freezes.


d1donlydfink (approved):
If you say so :)

PR #541, opened 2025-10-31
Apache2.0

sidcog:
This branch has an updated LICENSE.tx file and build_scripts/source_available_copyright.txt file. It has the four new files required by security. The pyproject.yaml and requirements.txt files have been updated. All the python scripts have been updated with new copyright headers.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,7 +13,7 @@ description = "NeuroAI data-driven System for multi-Agent Networks - client, lib
 keywords = ["LLM", "langchain", "agent", "multi-agent"]
 
 # This license name comes from doc recommendation here
-# https://www.apache.org/licenses/LICENSE-2.0
+# https://packaging.python.org/en/latest/guides/licensing-examples-and-user-scenarios/
 license = "Apache-2.0"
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -13,8 +13,8 @@ description = "NeuroAI data-driven System for multi-Agent Networks - client, lib
 keywords = ["LLM", "langchain", "agent", "multi-agent"]
 
 # This license name comes from doc recommendation here
-# https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license
-license = "LicenseRef-CognizantAcademicSource"
+# https://www.apache.org/licenses/LICENSE-2.0
```
d1donlydfink: The comment needs to pertain to the semantics of pyproject.toml, not the license itself.
That link is already in the license file
This should be just like it was in the pyproject.toml of the other 2 repos you’ve already done

sidcog: Fixing with  https://packaging.python.org/en/latest/guides/licensing-examples-and-user-scenarios/

d1donlydfink (approved):

Issue #540, opened 2025-10-31
Nightly integration test failed on Github Action


vince-leaf:
Two test cases failed. This failure started on Oct 25.
https://github.com/cognizant-ai-lab/neuro-san/actions/workflows/integration.yml

`------------------------------ Captured log call -------------------------------
ERROR    asyncio:base_events.py:1833 Task was destroyed but it is pending!
task: <Task pending name='Task-1698' coro=<AsyncToSyncGenerator.my_anext() running at /usr/local/lib/python3.12/site-packages/leaf_common/asyncio/async_to_sync_generator.py:89> wait_for=<Future pending cb=[Task.task_wakeup()]> cb=[_chain_future.<locals>._call_set_state() at /usr/local/lib/python3.12/asyncio/futures.py:396]>
=============================== warnings summary ===============================
tests/neuro_san/internals/run_context/langchain/llms/test_llm_factory.py:21
  /__w/neuro-san/neuro-san/tests/neuro_san/internals/run_context/langchain/llms/test_llm_factory.py:21: PytestCollectionWarning: cannot collect test class 'TestLlmFactory' because it has a __init__ constructor (from: tests/neuro_san/internals/run_context/langchain/llms/test_llm_factory.py)
    class TestLlmFactory(StandardLangChainLlmFactory):

<frozen abc>:106
  <frozen abc>:106: LangGraphDeprecatedSinceV10: AgentStatePydantic has been moved to `langchain.agents`. Please update your import to `from langchain.agents import AgentStatePydantic`. Deprecated in LangGraph V1.0 to be removed in V2.0.

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
================================= pytest-timer =================================
[fail] 50.00% tests/neuro_san/zzz_hocons/test_integration_test_hocons.py::TestIntegrationTestHocons::test_hocon_1_mcp_deepwiki_dict_cail_neuro_san: 90.0009s
[fail] 50.00% tests/neuro_san/zzz_hocons/test_integration_test_hocons.py::TestIntegrationTestHocons::test_hocon_0_esp_decision_assistant_fallbacks_and_commondef_values: 90.0005s`

PR #544, opened 2025-11-04
NS-540 Fix None value in minimize_metadata()

d1donlydfink:
Addresses #540 for me locally where integration tests were crashing.

### DIFF
--- a/neuro_san/internals/utils/metadata_util.py
+++ b/neuro_san/internals/utils/metadata_util.py
@@ -43,6 +43,10 @@ class MetadataUtil:
         """
         minimized: Dict[str, Any] = {}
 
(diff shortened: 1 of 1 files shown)


neuro_san/internals/utils/metadata_util.py
```diff
@@ -43,6 +43,10 @@ def minimize_metadata(request_metadata: Dict[str, Any],
         """
         minimized: Dict[str, Any] = {}
 
+        if request_metadata is None:
+            # No request data, nothing to fill
+            return minimized
```
d1donlydfink, line 48: Fix a corner case for None metadata that was causing a crash.

PR #553, opened 2025-11-05
Remove retired LLM models

Noravee:
The following models have been removed:
- `o1-preview-2024-09-12`
- `claude-3-5-sonnet-20241022`
- `gemini-1.5-flash-8b`
- `gemini-1.5-flash`
- `gemini-1.5-pro`

### Tests
All the above models have been tested to make sure that they are no longer available.

### References
- https://platform.openai.com/docs/deprecations
- https://docs.claude.com/en/docs/about-claude/model-deprecations
- https://ai.google.dev/gemini-api/docs/deprecations

### DIFF
--- a/neuro_san/internals/run_context/langchain/llms/default_llm_info.hocon
+++ b/neuro_san/internals/run_context/langchain/llms/default_llm_info.hocon
@@ -54,21 +54,6 @@
         "max_output_tokens": 100000,
         "knowledge_cutoff": "09/30/2023",
     },
-    "o1-preview": {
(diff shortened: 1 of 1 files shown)


ofrancon (approved):
lgtm

d1donlydfink (approved):
This is all fine the way it is.

It makes me wonder if we should leave a "use_model_name" pointer to the blessed version of the model as a breadcrumb.   Don't have to do that now on the bleeding edge. This is fine.

PR #556, opened 2025-11-06
450 - Ensure closing of external tool message generator.

andreidenissov-cog:
Whenever we are getting a messages stream out of async generator,
it's necessary to ensure that generator will be closed on every possible code path.

### DIFF
--- a/neuro_san/internals/graph/activations/external_activation.py
+++ b/neuro_san/internals/graph/activations/external_activation.py
@@ -20,6 +20,7 @@ from typing import Dict
 from typing import List
 from typing import Union
 
+import contextlib
 import json
 
 from logging import getLogger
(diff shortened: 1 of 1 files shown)


neuro_san/internals/graph/activations/external_activation.py
```diff
@@ -192,10 +193,19 @@ async def build(self) -> BaseMessage:
...
+                await self.processor.async_process_message(response)
+        finally:
+            # We are done with response stream, make sure to close it properly.
+            # We don't handle any possible exceptions here
+            # but response stream must be closed in any case.
+            if chat_responses is not None:
+                with contextlib.suppress(Exception):
+                    # It is possible we will call .aclose() twice
+                    # on our chat_responses - it is allowed and has no effect.
+                    await chat_responses.aclose()
```
d1donlydfink, line 208: This might be that 3rd case from the previous PR?

andreidenissov-cog: Yes, possible.

Issue #543, opened 2025-11-03
Return error messages of the coded tools to agent


Noravee:
Currently, when a tool fails, the agent simply reports that it cannot access the tool, without providing any explanation. This could be because when the run fails, we return None instead of the error message.

PR #561, opened 2025-11-10
Return error messages of the coded tools to agent and journal

Noravee:
#543

## Issues

When a coded tool fails, the agent currently reports only that it cannot access the tool, without explaining why.
This happens because, upon failure, the tool returns None instead of including the actual error message.

## Changes

- Added a boolean field `tool_error` to the coded tool message structure in the journal.
- Captured coded tool exceptions and included the error details as part of the tool output.
- Updated `langchain_openai_function_tool` to return the actual tool error instead of `None`.

## Note
- This update applies only to coded tools — not LangChain or MCP tools.
- Even when the tool error is included in the output, the LLM rarely provides details about what went wrong or why the tool failed.
- **This is a draft PR to discuss the following points:**
    - **Should tool errors be recorded in the chat history and journal?**
    - **Should the `tool_error` field be part of the coded tool response structure?**

### DIFF
--- a/neuro_san/internals/graph/activations/abstract_class_activation.py
+++ b/neuro_san/internals/graph/activations/abstract_class_activation.py
@@ -273,6 +273,8 @@ Some hints:
                 retval = await coded_tool.async_invoke(self.arguments, self.sly_data)
 
             except NotImplementedError:
+                # That didn't work, so try running the synchronous method as an async task
+                # within the confines of the proper executor.
                 # Warn that there is a better alternative
                 message = f"""
 Running CodedTool class {coded_tool.__class__.__name__}.invoke() synchronously in an asynchronous environment.
@@ -288,6 +290,8 @@ This can lead to performance problems when running within a server. Consider por
                 retval = await loop.run_in_executor(None, coded_tool.invoke, arguments, sly_data)
         # pylint: disable=broad-exception-caught
         except Exception as exception:
+            # There was an error invoking the CodedTool.
+            # Log it and return an error string.
             tool_error = True
             retval = f"Error: {str(exception)}"
             self.logger.error("Error invoking CodedTool %s: %s", coded_tool.__class__.__name__, str(exception))


neuro_san/internals/graph/activations/abstract_class_activation.py
```diff
@@ -266,30 +267,39 @@ async def attempt_invoke(self, coded_tool: CodedTool, arguments: Dict[str, Any],
...
+            # There was an error invoking the CodedTool.
+            # Log it and return an error string.
+            tool_error = True
+            retval = f"Error: {str(exception)}"
+            self.logger.error("Error invoking CodedTool %s: %s", coded_tool.__class__.__name__, str(exception))
+            self.logger.error(traceback.format_exc())
 
         retval_dict: Dict[str, Any] = {
             "tool_end": True,
+            "tool_error": tool_error,
```
Noravee, line 302: Add `tool_error` field in the structure of coded tool response in journal so that client knows if there is an error.

neuro_san/internals/graph/activations/abstract_class_activation.py
```diff
@@ -266,30 +267,39 @@ async def attempt_invoke(self, coded_tool: CodedTool, arguments: Dict[str, Any],
...
+                invocation_context = self.run_context.get_invocation_context()
+                executor: AsyncioExecutor = invocation_context.get_asyncio_executor()
+                loop: AbstractEventLoop = executor.get_event_loop()
+                retval = await loop.run_in_executor(None, coded_tool.invoke, arguments, sly_data)
+        # pylint: disable=broad-exception-caught
+        except Exception as exception:
+            # There was an error invoking the CodedTool.
+            # Log it and return an error string.
+            tool_error = True
+            retval = f"Error: {str(exception)}"
```
Noravee, line 296: Use the exception as the return value when there is a tool error.

neuro_san/internals/run_context/langchain/core/langchain_openai_function_tool.py
```diff
@@ -220,16 +220,14 @@ async def _arun(
...
             # This actually allows LLMs to recognize that something is wrong
             # and verbally report on that.
             logger = logging.getLogger(self.__class__.__name__)
             logger.error("Tool._arun() got Exception: %s", str(exception))
             logger.error(traceback.format_exc())
             run = None
-
-        if run is None:
-            return None
+            return str(exception)
```
Noravee, line 230: Return exception instead of None when there is an error when creating run.

Note that this should not be triggered when invoking coded tool since the exceptions are caught in `abstract_class_activation` and returned as tool output instead.

neuro_san/internals/graph/activations/abstract_class_activation.py
ofrancon: Getting this exception back is very useful: if you run into some import error because you forgot to pip install requirements for a specific tool, now you'll know.

PR #564, opened 2025-11-13
Fix MCP handler construction.

andreidenissov-cog:
This PR fixes MCP handler construction. 
Bug has crept in somehow.

### DIFF
--- a/neuro_san/service/http/server/http_server.py
+++ b/neuro_san/service/http/server/http_server.py
@@ -219,8 +219,7 @@ class HttpServer(AgentAuthorizer, AgentStateListener):
         # Register MCP "root" handler for all MCP requests
         # if MCP server is enabled:
         if self.server_context.get_mcp_server_context().is_enabled():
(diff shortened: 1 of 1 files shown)


neuro_san/service/http/server/http_server.py
```diff
@@ -219,8 +219,7 @@ def make_app(self, requests_limit: int, logger: EventLoopLogger):
         # Register MCP "root" handler for all MCP requests
         # if MCP server is enabled:
         if self.server_context.get_mcp_server_context().is_enabled():
-            mcp_request_initialize_data: Dict[str, Any] = self.build_mcp_request_data()
-            handlers.append((r"/mcp", McpRootHandler, mcp_request_initialize_data))
+            handlers.append((r"/mcp", McpRootHandler, request_initialize_data))
```
andreidenissov-cog, line 223: Use "standard" request initialization data - it has everything we need for MCP handler.

PR #568, opened 2025-11-14
Use sly_data for authentication

Noravee:
# Update 11/17/2025

- Authentication credentials can be passed from the client side via `sly_data` using field `http_headers`
- The required fields depend on the authentication scheme expected by each MCP server.
- Users may specify different authorization credentials for different MCP URLs.
For example, sly_data can contain:
```json
{
     "http_headers": {
         "<MCP_URL_1>": {      
             "Authorization": "Bearer <token_value>"
         },
         "<MCP_URL_2>": {
             "client_id": "<client_id_value>",
             "client_secret": "<client_secret_value>"
         },
     }
 }
```
- Add unit test for `LangChainMcpAdapter`.

# Draft PR: Enable Client-Side Authentication for MCP Tools

**Note: This is a draft for discussion on passing authentication credentials (e.g., bearer tokens) from the client side, potentially via `sly_data`. Feedback welcome before converting to a formal PR.**

## Problem
MCP tools currently lack access to client-side authentication credentials. There's no mechanism to pass bearer tokens or similar authentication data from the client, as MCP tools don't have access to `sly_data`.

## Proposed Solution
This PR enables authentication by making `sly_data` available earlier in the request lifecycle and exposing it to MCP tools:

- **Timing adjustment**: Move `sly_data` value initialization from immediately before `submit_message()` in `DataDrivenChatSession.chat()` to before `create_resource()` in `DataDrivenChatSession.set_up()`
- **MCP integration**: Pass `sly_data` from `tool_caller` to MCP tools
- **Example configuration**: Add `mcp_github.hocon` network configuration as a working reference implementation

## Testing
Validated using the CLI with both `math_guy_passthrough` and `mcp_github` configurations. Verified correct behavior by inspecting AGENT_FRAMEWORK trace inputs/outputs in LangSmith.

### DIFF
--- a/neuro_san/registries/mcp_github.hocon
+++ b/neuro_san/registries/mcp_github.hocon
@@ -91,7 +91,7 @@
                     # Users may specify different authorization credentials for different MCP URLs.
                     # For example, sly_data can contain:
                     # {
-                    #     "headers": {
+                    #     "http_headers": {
                     #         "<MCP_URL_1>": {      
                     #             "Authorization": "Bearer <token_value>"
                     #         },


neuro_san/internals/chat/data_driven_chat_session.py
```diff
@@ -105,6 +108,14 @@ async def set_up(self, invocation_context: InvocationContext,
         # Reset what we might have created before.
         await self.delete_resources()
 
+        # Update sly data, if any.
+        # Note that since this instance is the owner of the sly_data,
+        # any update here should get transmitted to all the other graph components
+        # because it is expected they share the reference and only interact with it
+        # in a read-only fashion.
+        if sly_data is not None:
+            self.sly_data.update(sly_data)
```
Noravee, line 117: Move update `sly_data` to before `self.front_man.create_any_resources()` since the MCP tools are instantiate in this process.

neuro_san/internals/run_context/langchain/core/base_tool_factory.py
```diff
@@ -153,6 +153,9 @@ async def create_mcp_tool(self, mcp_info: Union[str, Dict[str, Any]]) -> List[Ba
         """
         # By default, assume no allowed tools. This may get updated below or in the LangChainMcpAdadter.
         allowed_tools: List[str] = None
+        # Get headers from sly_data if available
+        headers: Dict[str, Any] = self.tool_caller.sly_data.get("headers")
```
Noravee: Look for field `headers` in `sly_data`.

neuro_san/internals/run_context/langchain/mcp/langchain_mcp_adapter.py
```diff
@@ -77,9 +79,8 @@ async def get_mcp_tools(
             "url": server_url,
             "transport": "streamable_http",
         }
-        # Try to look up authentication details from the URL
-        headers_dict: Dict[str, Any] =\
-            self._mcp_clients_info.get(server_url, {}).get("headers")
+        # Try to look up authentication details first from the sly data then from the URL
+        headers_dict: Dict[str, Any] = headers or self._mcp_clients_info.get(server_url, {}).get("headers")
```
Noravee, line 86: `headers` from `sly_data` has the priority over from mcp info file.

neuro_san/registries/mcp_github.hocon
```diff
@@ -0,0 +1,76 @@
...
+#
+# END COPYRIGHT
+
+# The schema specifications for this file are documented here:
+# https://github.com/cognizant-ai-lab/neuro-san/blob/main/docs/agent_hocon_reference.md
+
+{
+    # Optional metadata describing this agent network
+    "metadata": {
+        "description": """Simple agent network consisting of a frontman, a lower level agent,
```
Noravee, line 24: `mcp_github` has 3 levels; frontman, branch, and MCP. This is to make sure that the `sly_data` can go through the layer to MCP tools.

neuro_san/registries/mcp_github.hocon
```diff
@@ -0,0 +1,76 @@
...
+            }
+            "tools": [
+                {
+                    # Only support streamable http protocal
+                    # For other modes of transport, use coded tool implementation
+                    "url": "https://api.githubcopilot.com/mcp",
+
+                    # Optional: restricts which tools are available.
+                    # If omitted, all tools are enabled.
+                    "tools": ["issue_read"]
```
Noravee, line 85: This tool read issue from GitHub.

neuro_san/registries/mcp_github.hocon
```diff
@@ -0,0 +1,76 @@
...
+            "function": {
+                "description": "Assist user with answer github tools."
+            },
+            "tools": ["issue_getter"]
+        },
+        {
+            "name": "issue_getter",
+            "instructions": "Use your tool to respond to the inquiry.",
+            "function": {
+            "description": "Assist user with answer from deepwiki."
```
d1donlydfink: Indentation is off here for the description key

neuro_san/registries/mcp_github.hocon
```diff
@@ -0,0 +1,76 @@
...
+                    # This specific tool requires issue number, repo name, and owner name to be provided
+                    # Furthermore, it requires authentication via an Authorization header.
+
+                    # Authentication can be provided via field "headers" in sly_data.
+                    # For example, sly_data can contain:
+                    # {
+                    #     "headers": {
+                    #         "Authorization": "Bearer <token_value>"
+                    #      }
+                    # }
```
d1donlydfink, line 103: Per the agent network hocon docs here:
https://github.com/cognizant-ai-lab/neuro-san/blob/main/docs/agent_hocon_reference.md#sly_data_schema

You can specify what the sly_data schema should look like in the front-man's function().
That would be a much better thing to do than simply documenting it in code, as clients can then read that and potentially offer fields to fill in for the request.

d1donlydfink: FWIW: From an end-user point of view, they will never really know how to fill in the Authorization/Bearer token stuff.  Usually some other authorization service supplies that.  

What you have is good enough for now, but a better example for the future (perhaps a studio level) is one where the sly_data contains a GitHub username (at least) and a direction to a website that fills in the auth information for them in sly_data.  There's hand-waving there because it requires some client interaction with another service, but it'd be a better experience (but still a good demo showing auth interaction).

neuro_san/registries/mcp_github.hocon
Noravee: The indentation has been fixed.

neuro_san/registries/mcp_github.hocon
Noravee: Thank you for highlighting the sly_data_schema option — I’ve implemented it accordingly.

I also appreciate the suggestion around improving the auth experience. I agree that a flow where the client handles the token for the user would create a much smoother and more realistic example, especially for future studio-level demos.

neuro_san/registries/mcp_github.hocon
```diff
@@ -0,0 +1,102 @@
...
+
+        # This first guy is the "Front Man".  He is identified as such because
+        # he is the only one with no parameters in his function definition,
+        # and therefore he needs to talk to the outside world to get things rolling.
+        {
+            "name": "github_manager",
+            "instructions": "Use your tool to respond to the inquiry on a specific github repo.",
+            "function": {
+                "description": "Assist user with answer github tools.",
+                "sly_data_schema": {
```
Noravee, line 44: Add `sly_data_schema`.

Noravee:
✔️ Fixed the HOCON indentation.

✔️ Added sly_data_schema to the mcp_github.hocon frontman function.

- Regarding the test: this new network is specifically meant to demonstrate an authorization workflow. For a non-authorization MCP example, we already have a test here:
https://github.com/cognizant-ai-lab/neuro-san/blob/main/tests/fixtures/mcp_deepwiki_dict/cail_neuro_san.hocon

neuro_san/internals/run_context/langchain/core/base_tool_factory.py
ofrancon: That means `headers` is now a reserved keyword in `sly_data`. 
That also means it will be expose in the UIs that show sly_data, like `nsflow`. Are we ok with that?

neuro_san/internals/chat/data_driven_chat_session.py
```diff
@@ -105,6 +108,14 @@ async def set_up(self, invocation_context: InvocationContext,
         # Reset what we might have created before.
         await self.delete_resources()
 
+        # Update sly data, if any.
+        # Note that since this instance is the owner of the sly_data,
+        # any update here should get transmitted to all the other graph components
+        # because it is expected they share the reference and only interact with it
+        # in a read-only fashion.
```
ofrancon, line 115: I think the other components can update `sly_data`, can't they?  I understand this piece of code just moved up - just wondering.

neuro_san/internals/run_context/langchain/core/base_tool_factory.py
ofrancon: I mean we might need to expose it in `sly_data`, so that a user can paste a token for instance. Or a client interface. That's the goal of sly_data after all. 
Maybe `headers` is a bit too generic? Should we use `http_headers`?

ofrancon (commented):
Looks reasonable to me. I'd just change the sly_data key to `http_headers` instead of simply `headers`.

neuro_san/internals/chat/data_driven_chat_session.py
d1donlydfink: Yes. This is simply moving when the single sly_data instance for the entire request gets initialized earlier.

neuro_san/internals/run_context/langchain/core/base_tool_factory.py
d1donlydfink: I like "http_headers" better than "headers" as well because it's more specific.
That said, there would be nothing preventing a single agent network from needing > 1 set of headers for contacting > 1 mcp tool, and then this naming goes up in smoke.

I think the "http_headers" name change will be good enough for now, but we'll have to think about how to expose some kind of sly_data -> http_header translation to the outside world. That doesn't have to be now as long as this is kept on the down-low.

d1donlydfink (changes requested):
I think @ofrancon 's suggestion of changing the sly_data key to "http_headers" (or even "mcp_http_headers" if you wanted to get that specific) would be good for this PR.

We'll need to think more about how to expose a translation from sly_data -> headers for any one MCP tool, but don't need to do that before this goes in. This is good progress for the demo and we shouldn't let that get in the way.

neuro_san/internals/run_context/langchain/core/base_tool_factory.py
```diff
@@ -153,15 +153,21 @@ async def create_mcp_tool(self, mcp_info: Union[str, Dict[str, Any]]) -> List[Ba
         """
         # By default, assume no allowed tools. This may get updated below or in the LangChainMcpAdadter.
         allowed_tools: List[str] = None
+        # Get HTTP headers from sly_data if available
+        http_headers: Dict[str, Any] = self.tool_caller.sly_data.get("http_headers", {})
```
Noravee, line 157: Look for key `http_headers` in `sly_data`.

neuro_san/internals/run_context/langchain/core/base_tool_factory.py
```diff
@@ -153,15 +153,21 @@ async def create_mcp_tool(self, mcp_info: Union[str, Dict[str, Any]]) -> List[Ba
...
+
         if isinstance(mcp_info, str):
             server_url: str = mcp_info
         else:
             server_url = mcp_info.get("url")
             allowed_tools = mcp_info.get("tools")
 
+        # Get specific headers for the MCP server if available
+        headers: Dict[str, Any] = http_headers.get(server_url)
```
Noravee, line 166: Headers should be for specific URL.

neuro_san/internals/run_context/langchain/mcp/langchain_mcp_adapter.py
```diff
@@ -77,11 +82,14 @@ async def get_mcp_tools(
...
-            self._mcp_clients_info.get(server_url, {}).get("headers")
+        # Try to look up authentication details first from the sly data then from the MCP clients info.
+        headers_dict: Dict[str, Any] = headers or self._mcp_clients_info.get(server_url, {}).get("headers")
         if headers_dict:
-            mcp_tool_dict["headers"] = copy.copy(headers_dict)
+            if isinstance(headers_dict, dict):
+                # Use a copy to avoid modifying the original headers dictionary.
+                mcp_tool_dict["headers"] = copy.copy(headers_dict)
+            else:
+                self.logger.error("MCP client headers for server %s must be a dictionary.",  server_url)
```
Noravee, line 92: Check if the headers is a dictionary otherwise log error.

Note that there are also logging and journaling in `BaseToolFactory.create_mcp_tool()`.

tests/neuro_san/internals/run_context/langchain/mcp/test_langchain_mcp_adapter.py
```diff
@@ -0,0 +1,214 @@
...
+from unittest.mock import MagicMock
+from unittest.mock import patch
+import pytest
+
+from langchain_core.tools import StructuredTool
+
+from neuro_san.internals.run_context.langchain.mcp.langchain_mcp_adapter import LangChainMcpAdapter
+
+
+class TestLangChainMcpAdapter:
```
Noravee, line 28: Add unit test for `LangChainMcpAdapter`.

Noravee:
- Authentication credentials can be passed from the client side via `sly_data` using field `http_headers`
- The required fields depend on the authentication scheme expected by each MCP server.
- Users may specify different authorization credentials for different MCP URLs.
For example, sly_data can contain:
```json
{
     "http_headers": {
         "<MCP_URL_1>": {      
             "Authorization": "Bearer <token_value>"
         },
         "<MCP_URL_2>": {
             "client_id": "<client_id_value>",
             "client_secret": "<client_secret_value>"
         },
     }
 }
```
- Add unit test for `LangChainMcpAdapter`.

Issue #570, opened 2025-11-15
`AbstractClassActivation` should keep looking up for a class


ofrancon:
Agent networks have been re-organized by directory in Neuro SAN Studio:
See https://github.com/cognizant-ai-lab/neuro-san-studio/issues/406

`coffee_finder_advanced` has moved to the `basic` subfolder for instance.
However now `neuro-san` doesn't find the `kwik_agents` coded tools anymore:

```
[2025-11-14 17:30:51 PST] ERROR    NeuroSan - {"message": "Tool._arun() got Exception: Could not find class "ListTopics" in module "kwik_agents.list_topics" under AGENT_TOOL_PATH "coded_tools.basic" for the agent called
                                   "known_users" in the agent network "coffee_finder_advanced". Check these things: 1.  Is there a typo in your AGENT_TOOL_PATH? 2.  Expected to find a specific CodedTool for the given
                                   agent network in: <AGENT_TOOL_PATH>/<agent_network>/<coded_tool_name>.py Global CodedTools (shared across networks) should be located at: <AGENT_TOOL_PATH>/<coded_tool_name>.py a)  Does
                                   your AGENT_TOOL_PATH point to the correct directory? b)  Does your CodedTool actually live in a module appropriately named for your agent network? c)  Does the module in the "class"
                                   designation for the agent known_users match what is in the filesystem? d)  Does the specified class name match what is actually implemented in the file? e)  If an agent network contains
                                   both specific and global CodedTools, the global module must not have the same name as the agent network. 3. Is AGENT_TOOL_PATH findable from what is set for your PYTHONPATH? ",
                                   "user_id": "754337", "Timestamp": "2025-11-14T17:30:50.980649", "source": "HttpServer", "message_type": "Error", "request_id": "request-7"}
[2025-11-14 17:30:51 PST] ERROR    NeuroSan - {"message": "Traceback (most recent call last): File
                                   "/Users/754337/workspace/neuro-san-studio/venv/lib/python3.13/site-packages/neuro_san/internals/graph/activations/abstract_class_activation.py", line 166, in build python_class =
                                   resolver.resolve_class_in_module(class_name, module_name) File "/Users/754337/workspace/neuro-san-studio/venv/lib/python3.13/site-packages/leaf_common/config/resolver.py", line 91, in
[...]

PR #572, opened 2025-11-16
Modify AbstractClassActivation to find the correct agent_tool_path

Noravee:
#570 

### Issue

Neuro-SAN currently searches for coded tools only within the agent network module and one level above it.
Now that agent network names can be hierarchical, coded tools located in the root module may not be discovered, making them effectively unreachable.

### Fix
Updated the lookup logic to search for coded tools first in the agent network module and then in the root module, rather than only one level up.

### Tests
- `music_nerd_pro`
- `math_guy`

### DIFF
--- a/neuro_san/client/simple_one_shot.py
+++ b/neuro_san/client/simple_one_shot.py
@@ -46,6 +46,7 @@ class SimpleOneShot:
         self.connection_type: str = connection_type
         self.host: str = host
         self.port: int = port
+        self.processor: BasicMessageProcessor = None
 
     def get_answer_for(self, text: str) -> str:
         """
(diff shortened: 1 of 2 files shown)


neuro_san/internals/graph/activations/abstract_class_activation.py
```diff
@@ -158,32 +158,34 @@ async def build(self) -> BaseMessage:
...
         except (ValueError, AttributeError):
-            # Drop the last segment to go one level up in the module path
-            parent_path: str = this_agent_tool_path.rsplit(".", 1)[0]
-            packages = [parent_path]
+            # Since the agent network can have hierarchical structure with "/", we have to reconstruct
+            # the AGENT_TOOL_PATH accordingly to find the correct root module.
+            agent_network_name: str = self.factory.agent_network.get_network_name()
+            agent_network_name_parts: List[str] = agent_network_name.split("/")
+            this_agent_tool_path_parts: List[str] = this_agent_tool_path.split(".")
+            agent_tool_path: str = ".".join(this_agent_tool_path_parts[:-len(agent_network_name_parts)])
```
Noravee, line 175: If coded tool cannot be found in the agent network module, look at the root module instead of one level up.

PR #573, opened 2025-11-17
Easier one-shot to get more than text

d1donlydfink:
Was writing an app that needed to use the SimpleOneShot and realized I could only get the text that was associated with that.  Make it so that the BasicMessageProcessor is accessible for all the different parts of the answer.

Plus, add some typing in StreamingInputProcessor logic for sanity.

### DIFF
--- a/neuro_san/__init__.py
+++ b/neuro_san/__init__.py
@@ -1,12 +1,18 @@
 
-# Copyright (C) 2023-2025 Cognizant Digital Business, Evolutionary AI.
-# All Rights Reserved.
-# Issued under the Academic Public License.
+
+# Copyright © 2023-2025 Cognizant Technology Solutions Corp, www.cognizant.com.
 #
-# You can be released from the terms, and requirements of the Academic Public
-# License by purchasing a commercial license.
-# Purchase of a commercial license is mandatory for any use of the
-# neuro-san SDK Software in commercial settings.
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
 #
 # END COPYRIGHT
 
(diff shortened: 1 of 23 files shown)


neuro_san/client/simple_one_shot.py
```diff
@@ -46,6 +46,7 @@ def __init__(self, agent: str, connection_type: str = "direct",
         self.connection_type: str = connection_type
         self.host: str = host
         self.port: int = port
+        self.processor: BasicMessageProcessor = None
```
d1donlydfink, line 49: Keep the BasicMessageProcessor around as part of the instance so we can ask it more detailed questions later.

neuro_san/client/simple_one_shot.py
```diff
@@ -58,15 +59,21 @@ def get_answer_for(self, text: str) -> str:
...
 
         # Call streaming_chat()
         empty: Dict[str, Any] = {}
         for chat_response in session.streaming_chat(request):
             message: Dict[str, Any] = chat_response.get("response", empty)
-            processor.process_message(message, chat_response.get("type"))
+            self.processor.process_message(message, chat_response.get("type"))
 
-        raw_answer: str = processor.get_compiled_answer()
+        raw_answer: str = self.processor.get_compiled_answer()
```
d1donlydfink, line 71: Use the member variable.

neuro_san/client/simple_one_shot.py
```diff
@@ -58,15 +59,21 @@ def get_answer_for(self, text: str) -> str:
...
 
         return raw_answer
+
+    def get_processor(self) -> BasicMessageProcessor:
+        """
+        :return: The message processor holding the details of the answer from the agent
+        """
+        return self.processor
```
d1donlydfink, line 79: Allow access to the message processor to get more details from the final answer (like structure, sly_data, chat_context, etc)

neuro_san/client/streaming_input_processor.py
```diff
@@ -68,10 +68,10 @@ def process_once(self, state: Dict[str, Any]) -> Dict[str, Any]:
...
         empty: Dict[str, Any] = {}
         user_input: str = state.get("user_input")
-        last_chat_response = state.get("last_chat_response")
-        num_input = state.get("num_input")
-        chat_context = state.get("chat_context", empty)
-        chat_filter = state.get("chat_filter", empty)
+        last_chat_response: str = state.get("last_chat_response")
+        num_input: int = state.get("num_input", 0)
+        chat_context: Dict[str, Any] = state.get("chat_context", empty)
+        chat_filter: Dict[str, Any] = state.get("chat_filter", empty)
```
d1donlydfink, line 74: Add typing. Allow num_input to not have to be defined in the dictionary as an integer (as it is used in addition later on in the code)

PR #575, opened 2025-11-17
574: Fixes for multi-instance http server service threads.

andreidenissov-cog:
This PR fixes a found problem with multi-instances http neuro-san server.
Summary: if we have some service thread running in neuro-san system, this thread will not survive process forking when we spin up several tornado instances using say AGENT_HTTP_SERVER_INSTANCES configuration parameter.
Solution is to start all those service threads after we do the forking.


neuro_san/service/http/server/http_server.py
```diff
@@ -115,10 +116,12 @@ def __init__(self,
...
             network_storage.add_listener(self)
 
-    def start(self):
+    def start(self, startables: List[Startable]):
         """
         Method to be called by a thread running tornado HTTP server
         to actually start serving requests.
+        :param startables: List of Startable instances to start once server
+            has forked its multiple running instances.
         """
```
andreidenissov-cog, line 125: Pass the list of things we need to start to http server.

neuro_san/service/http/server/http_server.py
```diff
@@ -130,6 +133,14 @@ def start(self):
...
         )
 
+        if self.server_config.http_server_monitor_interval_seconds > 0:
+            # Add resources usage logger to list of things we need to start
+            # after http server is spun up:
+            resources_logger: Startable =\
+                ResourcesUsageLogger(
+                    self.server_config.http_server_monitor_interval_seconds, self.http_port, self.logger)
+            startables.append(resources_logger)
+
```
andreidenissov-cog, line 143: If requested, add ResourcesUsageLogger to the "startables" list.

neuro_san/service/http/server/http_server.py
```diff
@@ -152,41 +163,14 @@ def start(self):
...
 
-        if self.server_config.http_server_monitor_interval_seconds > 0:
-            # Start periodic logging of server resources used:
-            tornado.ioloop.PeriodicCallback(
-                self.run_resources_usage,
-                self.server_config.http_server_monitor_interval_seconds * 1000).start()
+        if startables:
+            for startable in startables:
+                with contextlib.suppress(Exception):
+                    startable.start()
```
andreidenissov-cog, line 170: And here we do start them, but after we do:
server.start(self.server_config.http_server_instances)
where we will potentially fork out several tornado instances.

neuro_san/service/http/server/http_server.py
```diff
@@ -152,41 +163,14 @@ def start(self):
...
-        """
-        Execute collecting and logging of server run-time resources
-        in on-blocking mode w.r.t. server event loop.
-        This is done because enumerating of some system resources
-        could be relatively slow.
-        """
-        loop = tornado.ioloop.IOLoop.current()
-        return await loop.run_in_executor(None, self.log_resources_usage)
-
     def make_app(self, requests_limit: int, logger: EventLoopLogger):
```
andreidenissov-cog, line 174: Moved out to separate class now.

neuro_san/service/http/server/resources_usage_logger.py
```diff
@@ -0,0 +1,82 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# END COPYRIGHT
+"""
+See class comment for details
+"""
+
```
andreidenissov-cog, line 20: Class is factored out of http_server.py.
No change in functionality.

neuro_san/service/main_loop/server_main_loop.py
```diff
@@ -239,14 +240,17 @@ def main_loop(self):
...
+        components_to_start: List[Startable] = []
         if server_status.updater.is_requested():
             current_dir: str = os.path.dirname(os.path.abspath(__file__))
             setup_logging(server_status.updater.get_service_name(),
                           current_dir,
                           'AGENT_SERVICE_LOG_JSON',
                           'AGENT_SERVICE_LOG_LEVEL')
             watcher = StorageWatcher(self.watcher_config, self.server_context)
-            watcher.start()
+            components_to_start.append(watcher)
```
andreidenissov-cog, line 254: The root of the issue:
watcher thread that was started too soon. It was not running in spun-up neuro-san processes.

neuro_san/service/watcher/interfaces/storage_updater.py
```diff
@@ -15,7 +15,7 @@
 #
 # END COPYRIGHT
 
-from neuro_san.service.watcher.interfaces.startable import Startable
+from neuro_san.service.interfaces.startable import Startable
```
andreidenissov-cog, line 19: Here and below: import change.
Startable is a generally useful interface, so move it up in code base.

d1donlydfink: We might eventually want to push it down to leaf-common or leaf-server-common

PR #580, opened 2025-11-18
NS-571 Add local port setting for localhost situation

### LINKED ISSUE #571: External agents are not found when neuro SAN server is started on an http port that is not 8080
To reproduce, use the `math_guy_passthrough` agent network that delegates to the external agent `/math_guy`.

Steps:

1. Start the Neuro SAN server on http port `8042`, or any other port that is not the default `8080`:
```
 python -m neuro_san.service.main_loop.server_main_loop --http_port 8042
```
2. Start a client to query it, on that port:
```
curl --request POST --url localhost:8042/api/v1/math_guy_passthrough/streaming_chat --data '{
    "user_message": {
        "text": "multiply"
    },
    "sly_data": {
        "x": 7,
        "y": 6
    }
}'
```

**ISSUE**: The server reports the following exception:
```
{"message": "Agent/tool /math_guy was unreachable. Not including it as a tool.
Problem accessing external agent /math_guy.

The server (which could be your own localhost) is currently not serving up
an agent network by that name. Try these hints:
1. Check to see that you do not have a typo in your reference to the external agent
   in the calling hocon file.
2. If you have control over the server, check to see that the agent network your are trying
   to reach has an entry in the manifest.hocon file whose value is set to true.
3. Consider restarting the server, as perhaps a server does not continually look
   for changes to hocon files or manifest files during normal operation.
4. If you are new to calling external agent networks, know that:
    a. Simply referencing an agent within your own hocon file with a / prefix
       does not mean the server is serving that agent up separately.
    b. Every agent that is externally referenceable needs its own hocon file
       which must also have an entry in the manifest.hocon file for the server.
    c. There is one and only one "front man" agent note in each network described
       by a hocon file that receives input on behalf of the network.
    d. In order to be called by external agents, that front man must have a full
       "function" definition, which includes a description, and at least one parameter
       defined.  These are how calling agents know how to interact with the agent network.
", "user_id": "None", "Timestamp": "2025-11-14T18:22:59.342381", "source": "HttpServer", "message_type": "Other", "request_id": "request-1"}
```
And the cURL command returns:
```
{"response": {"type": "AGENT_FRAMEWORK", "text": "I need specific numbers to multiply. Could you please provide them?", "chat_context": {"chat_histories": [{"origin": [{"tool": "passthrough", "instantiation_index": 1}], "messages": [{"type": "HUMAN", "origin": [{"tool": "passthrough", "instantiation_index": 1}], "text": "multiply"}, {"type": "AI", "origin": [{"tool": "passthrough", "instantiation_index": 1}], "text": "I need specific numbers to multiply. Could you please provide them?"}]}]}}}
```
instead of the expected
```
{"response": {"type": "AGENT_FRAMEWORK", "text": "\"Check sly_data['equals'] for the result\"", "chat_context": {"chat_histories": [{"origin": [{"tool": "passthrough", "instantiation_index": 1}], "messages": [{"type": "HUMAN", "origin": [{"tool": "passthrough", "instantiation_index": 1}], "text": "multiply"}, {"type": "AI", "origin": [{"tool": "passthrough", "instantiation_index": 1}], "text": "\"Check sly_data['equals'] for the result\""}, {"type": "AI", "origin": [{"tool": "passthrough", "instantiation_index": 1}], "text": "\"Check sly_data['equals'] for the result\""}]}]}, "sly_data": {"equals": 42.0}}}
```

d1donlydfink:
Fix local external agent connections when local server is running on something other than the default AGENT_HTTP_PORT.

Manually tested math_guy_passthrough and it's now happy.  Fixes #571 

@Noravee : Unclear if there is a separate mcp server component that would need to be addressed for the same situation.

### DIFF
--- a/neuro_san/internals/interfaces/invocation_context.py
+++ b/neuro_san/internals/interfaces/invocation_context.py
@@ -111,3 +111,9 @@ class InvocationContext:
         :return: The Reservationist instance for the session
         """
         raise NotImplementedError
+
+    def get_port(self) -> int:
+        """
+        :return: The port on which the server was started
+        """
+        raise NotImplementedError
--- a/neuro_san/internals/run_context/utils/external_agent_parsing.py
+++ b/neuro_san/internals/run_context/utils/external_agent_parsing.py
@@ -19,7 +19,6 @@ from typing import Dict
 from typing import List
 from typing import Union
 
-from os import environ
 from urllib.parse import ParseResult
 from urllib.parse import urlparse
 
@@ -31,9 +30,11 @@ class ExternalAgentParsing:
     """
 
     @staticmethod
-    def parse_external_agent(agent_url: str) -> Dict[str, str]:
+    def parse_external_agent(agent_url: str, server_port: int = None) -> Dict[str, str]:
         """
         :param agent_url: The URL describing where to find the desired agent.
+        :param server_port: The port that the server is listening on
+                            Does not have to be set for all operations.
         :return: A Dictionary with the following keys:
                 "host" - the hostname where the agent lives
                 "port" - the port on the host which serves up the agent (if any)
@@ -76,7 +77,7 @@ class ExternalAgentParsing:
             # If we are localhost and no port was specified in the url,
             # use the port that was configured for the server.
             # If this is not set, it will default to None, which is fine.
-            port = environ.get("AGENT_HTTP_PORT")
+            port = server_port
 
         # Get the agent name from the URL by looking at the path
         # Remove any leading slashes from the path for the agent name.
--- a/neuro_san/service/generic/agent_service.py
+++ b/neuro_san/service/generic/agent_service.py
@@ -105,6 +105,7 @@ class AgentService:
         self.llm_factory: ContextTypeLlmFactory = MasterLlmFactory.create_llm_factory(config)
         self.toolbox_factory: ContextTypeToolboxFactory = MasterToolboxFactory.create_toolbox_factory(config)
         self.async_executor_pool: AsyncioExecutorPool = server_context.get_executor_pool()
+        self.port: int = server_context.get_server_port()
 
         self.request_timeout_seconds: float = agent_network.get_request_timeout_seconds()
 
@@ -255,7 +256,8 @@ class AgentService:
             self.llm_factory,
             self.toolbox_factory,
             metadata,
-            reservationist)
+            reservationist,
+            self.port)
         invocation_context.start()
 
         # Set up logging inside async thread
--- a/neuro_san/service/main_loop/server_main_loop.py
+++ b/neuro_san/service/main_loop/server_main_loop.py
@@ -167,6 +167,7 @@ class ServerMainLoop:
         self.http_port = args.http_port
         if self.http_port == 0:
             server_status.http_service.set_requested(False)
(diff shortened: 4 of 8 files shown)


neuro_san/internals/run_context/utils/external_agent_parsing.py
```diff
@@ -71,6 +72,12 @@ def parse_external_agent(agent_url: str) -> Dict[str, str]:
         if host is None or len(host) == 0:
             host = "localhost"
 
+        if host == "localhost" and port is None:
+            # If we are localhost and no port was specified in the url,
+            # use the port that was configured for the server.
+            # If this is not set, it will default to None, which is fine.
+            port = environ.get("AGENT_HTTP_PORT")
```
d1donlydfink: Simple fix for localhost case.

andreidenissov-cog: What if server http port has been specified in cmd line? This is allowed.

d1donlydfink: That is a very good point.
I will look into the proper plumbing for that.

neuro_san/internals/interfaces/invocation_context.py
```diff
@@ -111,3 +111,9 @@ def get_reservationist(self) -> Reservationist:
         :return: The Reservationist instance for the session
         """
         raise NotImplementedError
+
+    def get_port(self) -> int:
+        """
+        :return: The port on which the server was started
+        """
```
d1donlydfink, line 118: Now we keep the server port on the invocation context

neuro_san/internals/run_context/utils/external_agent_parsing.py
```diff
@@ -71,6 +73,12 @@ def parse_external_agent(agent_url: str) -> Dict[str, str]:
         if host is None or len(host) == 0:
             host = "localhost"
 
+        if host == "localhost" and port is None:
+            # If we are localhost and no port was specified in the url,
+            # use the port that was configured for the server.
+            # If this is not set, it will default to None, which is fine.
+            port = server_port
```
d1donlydfink, line 80: Use that port when it comes down to parsing the external agent reference

neuro_san/service/generic/agent_service.py
```diff
@@ -255,7 +256,8 @@ def streaming_chat(self, request_dict: Dict[str, Any],
             self.llm_factory,
             self.toolbox_factory,
             metadata,
-            reservationist)
+            reservationist,
+            self.port)
```
d1donlydfink, line 260: Pass the port from the ServerContext to the InvocationContext

neuro_san/service/main_loop/server_main_loop.py
```diff
@@ -167,6 +167,7 @@ def parse_args(self):
         self.http_port = args.http_port
         if self.http_port == 0:
             server_status.http_service.set_requested(False)
+        self.server_context.set_server_port(self.http_port)
```
d1donlydfink, line 170: Set the port on the ServerContext

neuro_san/internals/run_context/utils/external_agent_parsing.py
d1donlydfink: Done. See new changes.

PR #581, opened 2025-11-19
NS-533 Simple no-sessions MCP policy.

andreidenissov-cog:
This PR factors out MCP session-handling policy in an interface and also implements a simple policy with essentially no sessions maintained on the server side.
More careful reading through MCP protocol spec seems to indicate that we only MUST maintain client-server sessions if our MCP service provides streamable http responses. 
Which we do not currently. Out tool call results are single json payloads, 
which opens a way to going basically stateless, and thus potentially scalable.
More analysis will be needed, but now we have 2 sessions handling policies to choose from.

### DIFF
--- a/neuro_san/service/utils/mcp_server_context.py
+++ b/neuro_san/service/utils/mcp_server_context.py
@@ -23,7 +23,6 @@ from neuro_san import TOP_LEVEL_DIR
 from neuro_san.internals.interfaces.dictionary_validator import DictionaryValidator
 from neuro_san.service.mcp.validation.mcp_request_validator import McpRequestValidator
 from neuro_san.service.mcp.interfaces.client_session_policy import ClientSessionPolicy
-from neuro_san.service.mcp.session.mcp_session_manager import McpSessionManager
 from neuro_san.service.mcp.session.mcp_no_sessions_policy import McpNoSessionsPolicy
 
 # MCP protocol version supported by this service


neuro_san/service/mcp/interfaces/client_session_policy.py
```diff
@@ -0,0 +1,71 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# END COPYRIGHT
+"""
+See class comment for details
+"""
+
```
andreidenissov-cog, line 20: Interface for sessions handling policy.

neuro_san/service/mcp/session/mcp_client_session.py
```diff
@@ -18,8 +18,10 @@
 See class comment for details
 """
 
+from neuro_san.service.mcp.interfaces.client_session import ClientSession
 
-class McpClientSession:
+
+class McpClientSession(ClientSession):
     """
```
andreidenissov-cog, line 25: Now implementation of ClientSession interface.

neuro_san/service/mcp/session/mcp_session_manager.py
```diff
@@ -86,6 +89,8 @@ def is_session_active(self, session_id: str) -> bool:
         :return: True if session exists and is active;
                  False otherwise
         """
+        if session_id is None:
+            return False
         with self.lock:
```
andreidenissov-cog, line 94: We should be able to handle session_id being None also, 
this check is not done in a caller now.

neuro_san/service/utils/mcp_server_context.py
```diff
@@ -39,7 +40,7 @@ class McpServerContext:
     def __init__(self):
         self.protocol_schema_filepath = None
         self.protocol_schema = None
-        self.session_manager = None
+        self.session_policy = None
         self.request_validator = None
```
andreidenissov-cog, line 44: Rename to match an interface type.

neuro_san/service/utils/mcp_server_context.py
```diff
@@ -61,7 +62,7 @@ def set_enabled(self, enabled: bool) -> None:
                 raise RuntimeError(f"Cannot load MCP protocol schema from "
                                    f"'{self.protocol_schema_filepath}': {str(exc)}") from exc
             # Create new session manager:
-            self.session_manager = McpSessionManager()
+            self.session_policy = McpNoSessionsPolicy()
         self.enabled = enabled
```
andreidenissov-cog, line 66: Set default session policy to "no real sessions" mode. For now.

neuro_san/service/mcp/handlers/mcp_root_handler.py
```diff
@@ -140,10 +142,8 @@ async def post(self):
             self.logger.error(self.get_metadata(), f"error: {extra_error}")
             self.do_finish()
             return
-        session_active: bool = False
-        if session_id is not None:
-            session_manager: McpSessionManager = self.mcp_context.get_session_manager()
-            session_active = session_manager.is_session_active(session_id)
+        session_policy: ClientSessionPolicy = self.mcp_context.get_session_policy()
+        session_active = session_policy.is_session_active(session_id)
         if not session_active:
```
andreidenissov-cog, line 147: session_id == None is now a valid case, it is up to current sessions policy to decide what to do.

neuro_san/service/mcp/handlers/mcp_root_handler.py
```diff
@@ -233,7 +233,8 @@ async def handle_handshake(
                 handshake_processor: McpInitializeProcessor = McpInitializeProcessor(self.mcp_context, self.logger)
                 result_dict, session_id =\
                     await handshake_processor.initialize_handshake(request_id, metadata, request_data["params"])
-                self.set_header(MCP_SESSION_ID, session_id)
+                if session_id is not None:
+                    self.set_header(MCP_SESSION_ID, session_id)
                 self.set_status(HTTPStatus.OK)
```
andreidenissov-cog, line 238: Here and below - only put session_id into response headers if it is not None.
If header is absent, this is essentially a signal to a client that we are not handling client-server sessions.

neuro_san/service/mcp/handlers/mcp_root_handler.py
```diff
@@ -281,8 +283,8 @@ async def delete(self):
 
         request_status: int = HTTPStatus.NO_CONTENT
         if session_id is not None:
-            session_manager: McpSessionManager = self.mcp_context.get_session_manager()
-            deleted: bool = session_manager.delete_session(session_id)
+            session_policy: ClientSessionPolicy = self.mcp_context.get_session_policy()
+            deleted: bool = session_policy.delete_session(session_id)
             if deleted:
```
andreidenissov-cog, line 288: Renaming to "policy" to match interface name.

neuro_san/service/mcp/interfaces/client_session.py
```diff
@@ -0,0 +1,37 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# END COPYRIGHT
+"""
+See class comment for details
+"""
+
+
```
andreidenissov-cog, line 21: Interface for client session instance.

neuro_san/service/mcp/processors/mcp_initialize_processor.py
```diff
@@ -55,9 +55,11 @@ async def initialize_handshake(
...
         _ = params
         # Create new client session:
-        session: McpClientSession = self.mcp_context.get_session_manager().create_session()
-        session_id: str = session.get_id()
-        self.logger.info(metadata, "Created new MCP client session with id: %s", session_id)
+        session: ClientSession = self.mcp_context.get_session_policy().create_session()
+        session_id: str = None
+        if session:
+            session_id = session.get_id()
+            self.logger.info(metadata, "Created new MCP client session with id: %s", session_id)
```
andreidenissov-cog, line 63: Created session could be returned as None,
indicating "no sessions" policy.

neuro_san/service/mcp/session/mcp_no_sessions_policy.py
```diff
@@ -0,0 +1,53 @@
...
+
+from neuro_san.service.mcp.interfaces.client_session_policy import ClientSessionPolicy
+from neuro_san.service.mcp.interfaces.client_session import ClientSession
+
+
+class McpNoSessionsPolicy(ClientSessionPolicy):
+    """
+    Policy class for scenario when client sessions are not supported by the MCP service.
+    """
+
```
andreidenissov-cog, line 29: Trivial implementation for policy not really handling client sessions.

neuro_san/service/mcp/processors/mcp_initialize_processor.py
```diff
@@ -95,7 +97,9 @@ async def activate_session(
         :return: True if successful;
                  False if session with given id does not exist
         """
-        success: bool = self.mcp_context.get_session_manager().activate_session(session_id)
+        success: bool = self.mcp_context.get_session_policy().activate_session(session_id)
+        if not session_id:
+            session_id = "N/A"
```
d1donlydfink, line 102: Is this "N/A" a blessed constant from the MCP spec?
If so, consider graduating it to a constant and providing a pointer to the documentation that says this has special status.

neuro_san/service/utils/mcp_server_context.py
d1donlydfink: I can see where eventually there is a factory here. 
Nice.

PR #588, opened 2025-11-20
Extend MCP tool call parameters.

andreidenissov-cog:
This PR syncs up MCP tool call API with neuro-san service definition which is published as OpenAPI specification document.
To do this, we extract MCP tool call schema specific to our service from neuro-san OpenAPI spec programmatically on service start. Then each incoming tool call request is validated against this schema (after each incoming MCP request is validated against general MCP protocol schema). 
Note that sly_data is passed to MCP tool call if present, but it's never sent back as a tool call result.

### DIFF
--- a/docs/mcp_service.md
+++ b/docs/mcp_service.md
@@ -28,30 +28,175 @@ easy scalability of neuro-san/MCP deployment.
 In the scope of MCP protocol, each public neuro-san agent network is represented by an MCP tool
 (see [MCP tools](https://modelcontextprotocol.io/specification/2025-06-18/server/tools))
 with the name of a tool being the same as the network name.
-Chat request to an agent network becomes a tool call, with the following json schema:
+Chat request to an agent network becomes a tool call, with the following json schema,
+replicated from neuro-san OpenAPI specification:
 
     ```json
     {
             "name": agent_name,
             "description": tool_description,
             "inputSchema": {
-                "type": "object",
-                "properties": {
-                    "input": {
-                        "type": "string",
-                        "description": "text input for chat request"
+              "components": {
+                  "schemas": {
+                    "ChatContext": {
+                      "type": "object",
+                      "properties": {
+                        "chat_histories": {
+                          "type": "array",
+                          "items": {
+                            "$ref": "#/components/schemas/ChatHistory"
+                          },
+                          "description": "A potentially full list of chat histories that pertain to the node. "
+                        }
+                      },
+                      "description": "Message for holding the state of play for any chat session"
                     },
-                    "chat_context": {
-                        "type": "string",
-                        "description": "optional chat context"
+                    "ChatFilter": {
+                      "type": "object",
+                      "properties": {
+                        "chat_filter_type": {
+                          "enum": [
+                            "UNKNOWN",
+                            "MINIMAL",
+                            "MAXIMAL"
+                          ],
+                          "type": "string",
+                          "description": "For now allow for an enum to describe how we want chat messages streamed. ",
+                          "format": "enum"
+                        }
+                      },
+                      "description": "Allows for controlling the messages that get streamed via StreamingChat."
                     },
-                    "sly_data": {
-                        "type": "object",
-                        "description": "optional arbitrary JSON dictionary",
-                        "additionalProperties": True
+                    "ChatHistory": {
+                      "type": "object",
+                      "properties": {
+                        "origin": {
+                          "type": "array",
+                          "items": {
+                            "$ref": "#/components/schemas/Origin"
+                          }
+                        },
+                        "messages": {
+                          "type": "array",
+                          "items": {
+                            "$ref": "#/components/schemas/ChatMessage"
+                          }
+                        }
+                      },
+                      "description": "A structure for storing chat history for a given node in the graph described by the origin."
+                    },
+                    "ChatMessage": {
+                      "type": "object",
+                      "properties": {
+                        "type": {
+                          "enum": [
+                            "UNKNOWN",
+                            "SYSTEM",
+                            "HUMAN",
+                            "AI",
+                            "AGENT",
+                            "AGENT_FRAMEWORK",
+                            "AGENT_TOOL_RESULT",
+                            "AGENT_PROGRESS"
+                          ],
+                          "type": "string",
+                          "description": "The type of chat message",
+                          "format": "enum"
+                        },
+                        "text": {
+                          "type": "string",
+                          "description": "String contents of any chat message"
+                        },
+                        "mime_data": {
+                          "type": "array",
+                          "items": {
+                            "$ref": "#/components/schemas/MimeData"
+                          },
+                          "description": "Optional bytes for any non-text media referenced by this message. "
+                        },
+                        "origin": {
+                          "type": "array",
+                          "items": {
+                            "$ref": "#/components/schemas/Origin"
+                          },
+                          "description": "Optional list of Origin structures (see above) describing the origin of the chat message."
+                        },
+                        "structure": {
+                          "type": "object",
+                          "description": "Optional structure for a message whose contents are parsed JSON."
+                        },
+                        "chat_context": {
+                          "$ref": "#/components/schemas/ChatContext"
+                        },
+                        "tool_result_origin": {
+                          "type": "array",
+                          "items": {
+                            "$ref": "#/components/schemas/Origin"
+                          },
+                          "description": "Optional list of Origin structures (see above) describing the origin of a tool result."
+                        },
+                        "sly_data": {
+                          "type": "object",
+                          "description": "This is an entirely optional map whose keys refer to data that is better left out of the LLM chat stream."
+                        }
+                      },
+                      "description": "Structure describing a single chat message."
+                    },
+                    "ChatRequest": {
+                      "type": "object",
+                      "properties": {
+                        "sly_data": {
+                          "type": "object",
+                          "description": "This is an entirely optional map whose keys refer to data that is better left out of the LLM chat stream."
+                        },
+                        "user_message": {
+                          "$ref": "#/components/schemas/ChatMessage"
+                        },
+                        "chat_context": {
+                          "$ref": "#/components/schemas/ChatContext"
+                        },
+                        "chat_filter": {
+                          "$ref": "#/components/schemas/ChatFilter"
+                        }
+                      },
+                      "description": "Request structure for Chat gRPC method",
+                      "required": [
+                        "user_message"
+                      ],
+                      "additionalProperties": false
+                    },
+                    "MimeData": {
+                      "type": "object",
+                      "properties": {
+                        "mime_type": {
+                          "type": "string",
+                          "description": "MIME type of the image data"
+                        },
+                        "mime_bytes": {
+                          "type": "string",
+                          "description": "Raw bytes of the image",
+                          "format": "byte"
+                        }
+                      },
+                      "description": "A Message identifying image data"
+                    },
+                    "Origin": {
+                      "type": "object",
+                      "properties": {
+                        "tool": {
+                          "type": "string",
+                          "description": "String name of the originating tool, as per the agent spec."
+                        },
+                        "instantiation_index": {
+                          "type": "integer",
+                          "description": "Some tools can be called more than once by one or more different paths. Allow for an instantiation index to distinguish these in the chat stream. Index counting starts at 0.",
+                          "format": "int32"
+                        }
+                      }
                     }
-                },
-                "required": ["input"]
+                  }
+              },                
+              "$ref": "#/components/schemas/ChatRequest"
             }
     }
     ```
--- a/neuro_san/service/mcp/handlers/mcp_root_handler.py
+++ b/neuro_san/service/mcp/handlers/mcp_root_handler.py
@@ -190,8 +190,8 @@ class McpRootHandler(BaseRequestHandler):
                     self.logger.error(self.get_metadata(), f"Error: Invalid tool call request: {extra_error}")
                     return
 
-                prompt: str = call_args.get("user_message", {})
-                chat_context: str = call_args.get("chat_context", None)
+                prompt: Dict[str, Any] = call_args.get("user_message", {})
+                chat_context: Dict[str, Any] = call_args.get("chat_context", None)
                 chat_filter: Dict[str, Any] = call_args.get("chat_filter", None)
                 sly_data: Dict[str, Any] = call_args.get("sly_data", None)
                 result_dict: Dict[str, Any] =\
--- a/neuro_san/service/mcp/validation/tool_request_validator.py
+++ b/neuro_san/service/mcp/validation/tool_request_validator.py
@@ -47,8 +47,6 @@ class ToolRequestValidator(DictionaryValidator):
             self.tool_request_method,
             self.required_property)
 
-        print(f"Tool request validation schema initialized: {self.request_schema}")
-
     def validate(self, candidate: Dict[str, Any]) -> List[str]:
         """
         Validate the dictionary data of an incoming tool call request against MCP protocol schema.


neuro_san/service/mcp/handlers/mcp_root_handler.py
```diff
@@ -172,8 +172,12 @@ async def post(self):
...
                 tool_name: str = call_params.get("name")
-                prompt: str = call_params.get("arguments", {}).get("input", "")
-                result_dict: Dict[str, Any] = await tools_processor.call_tool(request_id, metadata, tool_name, prompt)
+                call_args: Dict[str, Any] = call_params.get("arguments", {})
+                prompt: str = call_args.get("input", "")
+                chat_context: str = call_args.get("chat_context", None)
+                sly_data: Dict[str, Any] = call_args.get("sly_data", None)
+                result_dict: Dict[str, Any] =\
+                    await tools_processor.call_tool(request_id, metadata, tool_name, prompt, chat_context, sly_data)
                 self.set_status(HTTPStatus.OK)
```
andreidenissov-cog, line 205: In addition to "input", which is user prompt, extract also "chat context" and "sly data" parameters from MCP tool call request.

neuro_san/service/mcp/processors/mcp_tools_processor.py
```diff
@@ -66,8 +66,7 @@ async def list_tools(self, request_id, metadata: Dict[str, Any]) -> Dict[str, An
         result_dict: Dict[str, Any] = session.list(data)
         tools_description: List[Dict[str, Any]] = []
         for agent_dict in result_dict.get("agents", []):
-            agent_name: str = agent_dict["agent_name"]
-            tool_dict: Dict[str, Any] = await self._get_tool_description(agent_name, metadata)
+            tool_dict: Dict[str, Any] = await self._get_tool_description(agent_dict)
             tools_description.append(tool_dict)
```
andreidenissov-cog: Extract MCP tool description directly from data provided by ConciergeService,
do not mess with agent network "function" API call.

neuro_san/service/mcp/processors/mcp_tools_processor.py
```diff
@@ -99,18 +103,21 @@ async def call_tool(self, request_id, metadata: Dict[str, Any], tool_name: str,
...
         response_text: str = ""
         response_structure: Dict[str, Any] = None
+        response_sly_data: Dict[str, Any] = None
         try:
             async with asyncio.timeout(tool_timeout_seconds):
                 result_generator = service.streaming_chat(input_request, metadata)
                 async for result_dict in result_generator:
-                    partial_response, structure_data = await self._extract_tool_response_part(result_dict)
+                    partial_response, sly_data, structure_data = await self._extract_tool_response_part(result_dict)
                     if partial_response is not None:
```
andreidenissov-cog, line 127: From the result of neuro-san chat request, extract optional sly data part in case it was generated.

neuro_san/service/mcp/processors/mcp_tools_processor.py
```diff
@@ -136,18 +143,21 @@ async def call_tool(self, request_id, metadata: Dict[str, Any], tool_name: str,
                     await result_generator.aclose()
 
         # Return tool call result:
-        call_result: Dict[str, Any] = await self.build_tool_call_result(request_id, response_text, response_structure)
+        call_result: Dict[str, Any] =\
+            await self.build_tool_call_result(request_id, response_text, response_sly_data, response_structure)
         return call_result
```
andreidenissov-cog, line 158: sly_data doesn't make it into tool call result.

neuro_san/service/mcp/processors/mcp_tools_processor.py
```diff
@@ -166,54 +176,78 @@ async def build_tool_call_result(
...
         if result_structure is not None:
-            call_result["result"]["structuredContent"] = result_structure
+            call_result["result"]["structuredContent"] = {}
+            call_result["result"]["structuredContent"]["result"] = result_structure
             # For backward compatibility, also add text version of structure:
             structure_str: str = f"```json\n{json.dumps(result_structure, indent=2)}\n```"
             result_text = result_text + structure_str
+        if result_sly_data is not None:
+            call_result["result"]["structuredContent"]["sly_data"] = result_sly_data
         call_result["result"]["content"][0]["text"] = RequestsUtil.safe_message(result_text)
```
andreidenissov-cog, line 192: Per MCP spec recommendation, all structured data - sly data and maybe chat response structured part -
should go into "structuredContent" part of tool call result.

neuro_san/service/mcp/processors/mcp_tools_processor.py
```diff
@@ -166,54 +176,78 @@ async def build_tool_call_result(
...
+                    },
+                    "chat_context": {
+                        "type": "string",
+                        "description": "optional chat context"
+                    },
+                    "sly_data": {
+                        "type": "object",
+                        "description": "optional arbitrary JSON dictionary",
+                        "additionalProperties": True
                     }
```
andreidenissov-cog: Extend published MCP tools schema with additional parameters.

neuro_san/service/mcp/processors/mcp_tools_processor.py
```diff
@@ -166,54 +176,78 @@ async def build_tool_call_result(
...
+        :param agent_info_dict: dictionary with agent information;
+        :return: MCP tool description dictionary
+        """
         return {
-            "name": agent_name,
-            "description": tool_description,
+            "name": agent_info_dict.get("agent_name", ""),
+            "description": agent_info_dict.get("description", ""),
+            "tags": agent_info_dict.get("tags", ""),
             "inputSchema": {
```
andreidenissov-cog: Take tool description directly from what ConciergeService will give us.

neuro_san/service/mcp/processors/mcp_tools_processor.py
```diff
@@ -222,3 +256,8 @@ def _get_chat_input_request(self, input_text: str) -> Dict[str, Any]:
                 "chat_filter_type": "MAXIMAL"
             }
         }
+        if chat_context is not None:
+            request_dict["chat_context"] = chat_context
+        if sly_data is not None:
+            request_dict["sly_data"] = sly_data
+        return request_dict
```
andreidenissov-cog, line 247: Construct what we will send to internal neuro-san chat request using chat_context and sly_data if they are in tool call parameters.

neuro_san/service/mcp/processors/mcp_tools_processor.py
d1donlydfink: I have deep reservations about sly_data going into structuredContent.  I believe this is effectively turning the sly_data back over to the chat stream of the caller (neuro-san or not), to be processed by an LLM, and this is a no-no. This would mean that your secrets (keys, passwords, tokens, etc) that happen to live in the sly_data are going to be sent to your friendly neighborhood LLM provider. This is not what we want by default.

If this is somehow not the case, please provide a reference to the MCP docs/spec that tells otherwise.

Perhaps longer term there might be an "allow" block that lets certain sly_data fields go back to mcp clients,
but that seems farther away.

neuro_san/service/mcp/processors/mcp_tools_processor.py
d1donlydfink: No sly_data back to caller's chat stream

neuro_san/service/mcp/handlers/mcp_root_handler.py
d1donlydfink: I don't think we should be sending any sly_data back and forth over MCP.
Not without thorough discussion and knowledge of the ramifications.

neuro_san/service/mcp/processors/mcp_tools_processor.py
d1donlydfink: This is not correct.

The "description" from the Concierge service is from optional metadata and is for ingestion by humans via UI tooltips. It is expected to be brief. (Perhaps docs need a distinction here)

The description that is provided by Function is the correct information to provide here.

What is desired is a caller facing description of what the tool can do from the front-man. For humans, this description from Function is the chat prompt. For bots, this is the specific upward projection of what the tool can do for the caller, in its own semantics, which can actually be technical and/or subtle given the arrangement of agents.  The description in Function is also always required.

The other thing that bots require is a description and semantics of any structured function parameters that must be passed. How does that information get communicated to the caller in this scheme?

A specific simple example here is math_guy.hocon.
The metadata description (for humans) reads: "Simple agent network using a CodedTool and sly_data"
Which is very different from what is returned by Function() which is:
```
            "function": {
                
                # The description acts as an initial prompt.
                "description": """ 
I am a test agent network that can do math with values in sly_data.
Just give me the name of an arithmetic operator and I will do the work
on the secret numbers.
""",            
                # This parameters section isn't strictly needed if the agent network
                # is not expected to be called as an external agent.
                # This one is used with math_guy_passthrough, so we need to add this.
                "parameters": {
                    "type": "object",
                    "properties": { 
                        "operator": {
                            "type": "string",
                            "description": "The single-word name of the math operator to perform"
                        }
                    },
                    "required": ["operator"]
                },
                
                # You can also optionally specify the schema needed for any sly_data input.
                # This is to allow generic clients to prompt for private-data inputs
                # that do not belong in the chat stream.
                "sly_data_schema": {
                    "type": "object",
                    "properties": {
                        "x": {
                            "type": "float",
                            "description": "The first operand for the arithmetic operation"
                        },
                        "y": {
                            "type": "float",
                            "description": "The second operand for the arithmetic operation"
                        }
                    },
                    "required": ["x", "y"]
                }
            },
```

What is returned by the metadata is what is needed by a browsing curious human to see if they might want to click on this agent network.   What is returned by Function() is more detailed and essential to the operation of any calling bot (like math_guy_passthrough).  math_guy_passthrough's agent could never figure out what to do with the agent with the chintzy description in the metadata.

neuro_san/service/mcp/processors/mcp_tools_processor.py
d1donlydfink: Wrong place to get description.  Needs to come from Function().
See larger comment in other PR response.

neuro_san/service/mcp/handlers/mcp_root_handler.py
andreidenissov-cog: We are sending sly_data over http to our agents.
I agree that getting sly_data back from them could be a concern since we don't control MCP clients.
But cutting sly_data off completely will degrade neuro-san functionality when used through MCP.

docs/mcp_service.md
```diff
@@ -40,6 +40,15 @@ Chat request to an agent network becomes a tool call, with the following json sc
...
+                    },
+                    "chat_context": {
+                        "type": "string",
+                        "description": "optional chat context"
+                    },
+                    "sly_data": {
+                        "type": "object",
+                        "description": "optional arbitrary JSON dictionary",
+                        "additionalProperties": True
                     }
```
andreidenissov-cog, line 196: Get in sync with native neuro-san API.

neuro_san/service/mcp/processors/mcp_tools_processor.py
andreidenissov-cog: Updated: no sly_data in tool call result.

neuro_san/service/mcp/handlers/mcp_root_handler.py
d1donlydfink: Reasonable approach. sly_data in if provided, but not out.

docs/mcp_service.md
```diff
@@ -40,6 +40,15 @@ Chat request to an agent network becomes a tool call, with the following json sc
                     "input": {
                         "type": "string",
                         "description": "text input for chat request"
+                    },
+                    "chat_context": {
+                        "type": "string",
+                        "description": "optional chat context"
```
d1donlydfink: chat_context is not a string, it is a list of structures.  See neuro_san/api/grpc/chat.proto
This is only docs, so can be repaired later.

neuro_san/service/mcp/handlers/mcp_root_handler.py
```diff
@@ -172,8 +172,12 @@ async def post(self):
                         self.agent_policy)
                 call_params: Dict[str, Any] = data.get("params", {})
                 tool_name: str = call_params.get("name")
-                prompt: str = call_params.get("arguments", {}).get("input", "")
-                result_dict: Dict[str, Any] = await tools_processor.call_tool(request_id, metadata, tool_name, prompt)
+                call_args: Dict[str, Any] = call_params.get("arguments", {})
+                prompt: str = call_args.get("input", "")
+                chat_context: str = call_args.get("chat_context", None)
```
d1donlydfink: chat_context is not a string. See neuro_san/api/grpc/chat.proto.
Again, let's not invent a new schema just for MCP.

This is code, so not reparable later.

neuro_san/service/mcp/handlers/mcp_root_handler.py
```diff
@@ -172,8 +172,12 @@ async def post(self):
                         self.agent_policy)
                 call_params: Dict[str, Any] = data.get("params", {})
                 tool_name: str = call_params.get("name")
-                prompt: str = call_params.get("arguments", {}).get("input", "")
-                result_dict: Dict[str, Any] = await tools_processor.call_tool(request_id, metadata, tool_name, prompt)
+                call_args: Dict[str, Any] = call_params.get("arguments", {})
+                prompt: str = call_args.get("input", "")
```
d1donlydfink: It would be a lot better if we were not inventing a new schema just for MCP. 😮‍💨 
The schema for a single ChatRequest is in neuro_san/api/grpc/agent.proto
Why can't we use that as the basis for our input schema?

neuro_san/service/mcp/processors/mcp_tools_processor.py
d1donlydfink: More of the same schema errors.
Why do we need a new schema for MCP?
Why can't this follow what is needed for ChatRequest in neuro_san/api/grpc/agent.proto?

neuro_san/service/mcp/processors/mcp_tools_processor.py
```diff
@@ -193,31 +200,50 @@ async def _get_tool_description(self, agent_name: str, metadata: Dict[str, Any])
...
+                                sly_data: Dict[str, Any]) -> Dict[str, Any]:
         """
         Construct Python dictionary expected by "streaming_chat" service API call.
         :param input_text: input user prompt;
+        :param chat_context: chat context string, could be None;
+        :param sly_data: arbitrary JSON dictionary containing sly_data, could be None;
         :return: "streaming_chat" service API input dictionary
         """
-        return {
+        request_dict: Dict[str, Any] = {
```
d1donlydfink, line 238: If we had the same input schema for MCP, then this translation would be a lot simpler.

neuro_san/service/http/handlers/base_request_handler.py
```diff
@@ -65,7 +65,7 @@ def initialize(self, **kwargs):
         # type: List[str]
         self.forwarded_request_metadata: List[str] = kwargs.pop("forwarded_request_metadata", [])
         # type: str
-        self.openapi_service_spec_path: str = kwargs.pop("openapi_service_spec_path", None)
+        self.openapi_service_spec: Dict[str, Any] = kwargs.pop("openapi_service_spec", None)
         # type: ServerContext
```
andreidenissov-cog, line 69: Pass around OpenAPI dictionary itself instead of just a filepath:
we will need it in several places, so avoid reloading it from file again and again.

neuro_san/service/http/handlers/openapi_publish_handler.py
```diff
@@ -37,14 +35,8 @@ def get(self):
...
-            self.write(result_dict)
-        except Exception as exc:  # pylint: disable=broad-exception-caught
-            self.process_exception(exc)
-        finally:
-            self.do_finish()
-            self.application.finish_client_request(metadata, "/api/v1/docs")
+        # Return json data to the HTTP client
+        self.set_header("Content-Type", "application/json")
+        self.write(self.openapi_service_spec)
+        self.do_finish()
```
andreidenissov-cog, line 41: self.openapi_service_spec is already loaded - just send it out.

neuro_san/service/http/server/http_server.py
```diff
@@ -264,9 +265,16 @@ def build_request_data(self) -> Dict[str, Any]:
         Build request data for Http handlers.
         :return: a dictionary with request data to be passed to an http handler.
         """
+        open_api_dict: Dict[str, Any] = None
+        try:
+            with open(self.openapi_service_spec_path, "r", encoding='utf-8') as f_out:
+                open_api_dict = json.load(f_out)
+        except Exception as exc:  # pylint: disable=broad-exception-caught
+            raise ValueError(f"Failed to load '{self.openapi_service_spec_path}'") from exc
+
```
andreidenissov-cog, line 274: Load neuro-san API spec here and pass it to our http server to use.

neuro_san/service/utils/mcp_server_context.py
```diff
@@ -42,6 +42,7 @@ def __init__(self):
         self.protocol_schema = None
         self.session_policy = None
         self.request_validator = None
+        self.tool_call_validator = None
         self.enabled: bool = False
```
andreidenissov-cog, line 45: Include a json schema based validator for tool call processing.

neuro_san/service/mcp/processors/mcp_tools_processor.py
```diff
@@ -77,17 +77,27 @@ async def list_tools(self, request_id, metadata: Dict[str, Any]) -> Dict[str, An
...
         }
 
-    async def call_tool(self, request_id, metadata: Dict[str, Any], tool_name: str, prompt: str) -> Dict[str, Any]:
+    async def call_tool(self, request_id, metadata: Dict[str, Any],
+                        tool_name: str,
+                        prompt: Dict[str, Any],
+                        chat_context: Dict[str, Any],
+                        chat_filter: Dict[str, Any],
+                        sly_data: Dict[str, Any]) -> Dict[str, Any]:
         """
```
andreidenissov-cog, line 89: Extend tool call arguments with all elements from service API.
Note that all of them are general dictionaries, not "just" text strings.

neuro_san/service/mcp/handlers/mcp_root_handler.py
d1donlydfink: It's good that the MCP API exposed now matches our own native API with user_message.
Still some gripes:
* The docs in mcp_service.md do not reflect this
* user_message and chat_context here are strings and not dictionaries.

neuro_san/service/mcp/handlers/mcp_root_handler.py
```diff
@@ -169,11 +174,34 @@ async def post(self):
...
+                    extra_error: str = "; ".join(validation_errors)
+                    error_msg: Dict[str, Any] = \
+                        McpErrorsUtil.get_protocol_error(request_id, McpError.InvalidRequest, extra_error)
+                    self.set_status(HTTPStatus.BAD_REQUEST)
+                    self.write(error_msg)
+                    self.logger.error(self.get_metadata(), f"Error: Invalid tool call request: {extra_error}")
+                    return
+
+                prompt: Dict[str, Any] = call_args.get("user_message", {})
+                chat_context: Dict[str, Any] = call_args.get("chat_context", None)
```
d1donlydfink, line 194: Worth noting that here we have user_message and chat_context as dictionaries.
Why do we need to use string types for these in the other locations?

neuro_san/service/mcp/processors/mcp_tools_processor.py
d1donlydfink: Perhaps worth an explicit code comment at some point

neuro_san/service/mcp/validation/tool_request_validator.py
```diff
@@ -0,0 +1,163 @@
...
+        """
+        Return the set of referenced schema names found in json schema via $ref.
+
+        Recognizes local refs of the form:
+          - "#/components/schemas/Name"
+        """
+        _ref_defs_re = re.compile(r"^#/components/schemas/([^/]+)$")
+        out: Set[str] = set()
+
+        def visit(node: Any):
```
d1donlydfink, line 82: We pretty much never need internal function definitions.
Why is this necessary?
(Can fix later, enough going on in this PR already)

d1donlydfink (approved):

PR #591, opened 2025-11-21
NS-529: Enable MCP service per agent network.

andreidenissov-cog:
This PR adds additional "mcp" designation for declaring agent networks in a manifest file.
Also there is now new "mcp only" configuration parameter for neuro-san server,
if we only want to expose MCP protocol endpoint, but not native neuro-san API.

### DIFF
--- a/neuro_san/internals/graph/persistence/manifest_dict_config_filter.py
+++ b/neuro_san/internals/graph/persistence/manifest_dict_config_filter.py
@@ -30,6 +30,8 @@ class ManifestDictConfigFilter(ConfigFilter):
     any Easy boolean values to a specific dictionary.
     """
 
+    MCP_DEFAULT_MODE: bool = True
+
     def __init__(self, manifest_file: str):
         """
         Constructor
@@ -63,7 +65,7 @@ class ManifestDictConfigFilter(ConfigFilter):
             expanded_value: Dict[str, Any] = {
                 "serve": True,
                 "public": True,
-                "mcp": False
+                "mcp": self.MCP_DEFAULT_MODE
             }
 
             # Traditional, easy entry in a manifest file.
@@ -72,7 +74,7 @@ class ManifestDictConfigFilter(ConfigFilter):
                     expanded_value = {
                         "serve": False,
                         "public": False,
-                        "mcp": False
+                        "mcp": self.MCP_DEFAULT_MODE
                     }
             elif isinstance(value, Dict):
                 expanded_value = value
--- a/neuro_san/service/main_loop/server_main_loop.py
+++ b/neuro_san/service/main_loop/server_main_loop.py
@@ -143,7 +143,7 @@ class ServerMainLoop:
                                 help="Http server resources monitoring/logging interval in seconds "
                                      "0 means no logging")
         arg_parser.add_argument("--mcp_enable", type=str,
-                                default=os.environ.get("AGENT_MCP_ENABLE", "false"),
+                                default=os.environ.get("AGENT_MCP_ENABLE", "true"),
                                 help="'true' if MCP protocol service should be enabled")
         arg_parser.add_argument("--mcp_only", type=str,
                                 default=os.environ.get("AGENT_MCP_ONLY", "false"),
--- a/neuro_san/deploy/Dockerfile
+++ b/neuro_san/deploy/Dockerfile
@@ -270,4 +270,13 @@ ENV AGENT_EXTERNAL_RESERVATIONS_STORAGE=""
 # to use for cross-pod reservations storage.
 ENV AGENT_RESERVATIONS_S3_BUCKET=""
 
+ENV AGENT_MCP_ENABLE="true"
+# When set, this parameter enables MCP service protocol for running neuro-san server.
+# Service endpoint is http://host:port/mcp
+# (neuro-san own http API is also enabled in this case)
+
+ENV AGENT_MCP_ONLY="false"
+# When set, this parameter will only enable MCP service protocol for running neuro-san server,
+# while disabling neuro-san own http API.
+ 
 ENTRYPOINT "${APP_ENTRYPOINT}"


neuro_san/internals/graph/persistence/manifest_dict_config_filter.py
```diff
@@ -63,6 +63,7 @@ def filter_config(self, basis_config: Dict[str, Any]) -> Dict[str, Dict[str, Any
             expanded_value: Dict[str, Any] = {
                 "serve": True,
                 "public": True,
+                "mcp": False
             }
```
andreidenissov-cog, line 69: Extend default agent network dictionary with "mcp" key.
Default is False - any MCP exposed network must be specified explicitly.

neuro_san/internals/graph/persistence/manifest_dict_config_filter.py
```diff
@@ -80,6 +82,12 @@ def filter_config(self, basis_config: Dict[str, Any]) -> Dict[str, Dict[str, Any
                                     key, self.manifest_file)
                 continue
 
+            # MCP designated entries are considered public by default.
+            if "mcp" not in expanded_value:
+                expanded_value["mcp"] = False
+            if expanded_value["mcp"]:
+                expanded_value["public"] = True
+
```
andreidenissov-cog, line 92: If network is exposed as MCP service, it should be public.

neuro_san/internals/graph/persistence/registry_manifest_restorer.py
```diff
@@ -148,6 +148,10 @@ def restore_one_manifest(self, manifest_file: str) -> Dict[str, Dict[str, AgentN
 
             network_name: str = self.agent_mapper.filepath_to_agent_network_name(agent_filepath)
 
+            # Check if this agent network has been declared as MCP tool:
+            if manifest_dict.get("mcp", False):
+                agent_network.set_as_mcp_tool()
+
```
andreidenissov-cog, line 154: Set "I am exposed as MCP" property directly for agent network instance.

neuro_san/internals/graph/registry/agent_network.py
```diff
@@ -41,6 +41,10 @@ def __init__(self, config: Dict[str, Any], name: str):
         self.name = name
         self.agent_spec_map: Dict[str, Dict[str, Any]] = {}
 
+        # True if this agent network is set to be served as an MCP tool;
+        # False otherwise.
+        self.is_mcp_network: bool = False
+
```
andreidenissov-cog, line 47: Add a property to AgentNetwork - True if we want it to be exposed through MCP service.

neuro_san/service/generic/agent_service.py
```diff
@@ -125,6 +125,12 @@ def get_request_timeout_seconds(self) -> float:
         """
         return self.request_timeout_seconds
 
+    def is_mcp_tool(self) -> bool:
+        """
+        :return: True if the agent served by this service is an MCP tool
+        """
+        return self.agent_network_provider.get_agent_network().is_mcp_tool()
+
```
andreidenissov-cog, line 133: Here and below: route "MCP exposed" property to AgentService instance.

neuro_san/service/mcp/processors/mcp_tools_processor.py
```diff
@@ -61,14 +61,14 @@ async def list_tools(self, request_id, metadata: Dict[str, Any]) -> Dict[str, An
...
-            agent_name: str = agent_dict["agent_name"]
-            tool_dict: Dict[str, Any] = await self._get_tool_description(agent_name, metadata)
-            tools_description.append(tool_dict)
+        for agent_name in public_storage.get_agent_names():
+            provider: AgentNetworkProvider = public_storage.get_agent_network_provider(agent_name)
+            if provider is not None:
+                agent_network: AgentNetwork = provider.get_agent_network()
+                if agent_network.is_mcp_tool():
+                    tool_dict: Dict[str, Any] = await self._get_tool_description(agent_name, metadata)
+                    tools_description.append(tool_dict)
```
andreidenissov-cog, line 71: Instead of going through ConciergeService, which lists all public networks,
iterate over public "MCP tool" network instances.

neuro_san/internals/graph/persistence/manifest_dict_config_filter.py
d1donlydfink: Having mcp off by default is actually much harder to explain for the simplest case of 
"my_network.hocon": true

It's much easier to explain that the entire dictionary collapses down to a single true value when every value in the dictionary is true.

Another way to look at this:
What would happen if this mcp default remained false and the --mcp-only flag/env var were turned on for the server with a typical single-boolean value manifest?   I'd guess a server that serves nothing because of these extra unnecessary steps.

This means that MCP should be on by default, which I think is actually OK, even desirable.
In the end, all these networks have endpoints that are available anyway.

neuro_san/internals/graph/persistence/manifest_dict_config_filter.py
d1donlydfink: I agree with that.
Maybe even send up a warning when those 2 are at odds.

neuro_san/service/http/server/http_server.py
```diff
@@ -187,18 +190,21 @@ def make_app(self, requests_limit: int, logger: EventLoopLogger):
...
+
+        if enable_http_handlers:
+            handlers.append(("/api/v1/list", ConciergeHandler, request_initialize_data))
+            handlers.append(("/api/v1/docs", OpenApiPublishHandler, request_initialize_data))
+
+            # Register templated request paths for agent API methods:
+            # regexp format used here is that of Python Re standard library.
+            handlers.append((r"/api/v1/(.+)/function", FunctionHandler, request_initialize_data))
+            handlers.append((r"/api/v1/(.+)/connectivity", ConnectivityHandler, request_initialize_data))
+            handlers.append((r"/api/v1/(.+)/streaming_chat", StreamingChatHandler, request_initialize_data))
```
d1donlydfink, line 207: At some point soon we might want a forward-compatibility path w/rt neuro-san endpoints
`api/v1/neuro_san/<network>/{function,connectivity,streaming_chat}`
Or something like that.

neuro_san/service/main_loop/server_main_loop.py
```diff
@@ -145,6 +145,9 @@ def prepare_args(self) -> ArgumentParser:
         arg_parser.add_argument("--mcp_enable", type=str,
                                 default=os.environ.get("AGENT_MCP_ENABLE", "false"),
                                 help="'true' if MCP protocol service should be enabled")
+        arg_parser.add_argument("--mcp_only", type=str,
+                                default=os.environ.get("AGENT_MCP_ONLY", "false"),
+                                help="'true' if only MCP protocol service will be run (no HTTP service)")
```
d1donlydfink, line 150: See comment above about mcp-only turned on with a typical single-boolean value manifest file.

neuro_san/service/mcp/processors/mcp_tools_processor.py
```diff
@@ -94,6 +94,10 @@ async def call_tool(self, request_id, metadata: Dict[str, Any], tool_name: str,
             # No such tool is found:
             return McpErrorsUtil.get_tool_error(request_id, f"Tool not found: {tool_name}")
         service: AsyncAgentService = service_provider.get_service()
+        if not service.is_mcp_tool():
+            # Service is not allowed to be called as MCP tool:
+            return McpErrorsUtil.get_tool_error(request_id, f"Service not available as MCP tool: {tool_name}")
+
```
andreidenissov-cog, line 100: Only allow MCP-designated agent networks to be called as MCP tools.

neuro_san/service/main_loop/server_main_loop.py
```diff
@@ -191,6 +194,10 @@ def parse_args(self):
         # Do we to enable MCP service?
         if args.mcp_enable.lower() != "true":
             server_status.mcp_service.set_requested(False)
+        if args.mcp_only.lower() == "true":
+            server_status.mcp_service.set_requested(True)
+            # Disable HTTP service if MCP only is requested
+            server_status.http_service.set_requested(False)
```
andreidenissov-cog, line 201: If we request "MCP only" mode, "plain" HTTP API is not needed.

neuro_san/service/main_loop/server_main_loop.py
```diff
@@ -145,6 +145,9 @@ def prepare_args(self) -> ArgumentParser:
         arg_parser.add_argument("--mcp_enable", type=str,
                                 default=os.environ.get("AGENT_MCP_ENABLE", "false"),
                                 help="'true' if MCP protocol service should be enabled")
+        arg_parser.add_argument("--mcp_only", type=str,
+                                default=os.environ.get("AGENT_MCP_ONLY", "false"),
+                                help="'true' if only MCP protocol service will be run (no HTTP service)")
         return arg_parser
```
andreidenissov-cog, line 151: Add another config parameter to only enable MCP protocol, without native neuro-san API.

neuro_san/service/http/server/http_server.py
```diff
@@ -187,18 +190,21 @@ def make_app(self, requests_limit: int, logger: EventLoopLogger):
...
```
andreidenissov-cog, line 208: Enable all neuro-san handlers if HTTP API is requested.

neuro_san/internals/graph/persistence/manifest_dict_config_filter.py
andreidenissov-cog: It's basically deciding what is more convenient - stating what exactly we want vs stating what exactly we don't want. What if given "legacy" manifest file we only want to MCP expose 5 networks out of 20?
We'll have to rewrite the whole thing putting "mcp: False" in full form dictionaries for the whole manifest.
Anyway, I don't have a strong opinion here - it's just an internal flag flip.

neuro_san/service/generic/async_agent_service.py
```diff
@@ -125,6 +125,12 @@ def get_request_timeout_seconds(self) -> float:
         """
         return self.agent_network_provider.get_agent_network().get_request_timeout_seconds()
 
+    def is_mcp_tool(self) -> bool:
+        """
+        :return: True if the agent served by this service is an MCP tool
+        """
+        return self.agent_network_provider.get_agent_network().is_mcp_tool()
```
d1donlydfink, line 132: FWIW: "MCP tool" might not be the best long-term name for this. What you have is fine for now, though, to get going.
We're not really dealing with "tools" at this level of the service, we're merely trying to serve something up.
Maybe is_mcp_agent() instead or is_served_by_mcp()? or maybe just is_mcp()?

PR #593, opened 2025-11-24
Logging of next neuro-san manifest update time.

andreidenissov-cog:
This PR implements what 1C team has asked for: when doing a check for potentially updated manifest file,
log the time for next update check.

### DIFF
--- a/neuro_san/service/watcher/registries/registry_storage_updater.py
+++ b/neuro_san/service/watcher/registries/registry_storage_updater.py
@@ -14,9 +14,11 @@
 # limitations under the License.
 #
 # END COPYRIGHT
+import time
 from typing import Any
 from typing import Dict
 
+import datetime
 from logging import getLogger
 from logging import Logger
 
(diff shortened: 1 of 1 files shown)


neuro_san/service/watcher/registries/registry_storage_updater.py
```diff
@@ -77,6 +79,15 @@ def start(self):
...
 
+    def log_next_update_time(self):
+        """
+        Log the time for next manifest file update check
+        """
+        # Log the NEXT update time
+        next_time = time.time() + self.get_update_period_in_seconds()
+        next_time_fmt: str = datetime.datetime.fromtimestamp(next_time).isoformat()
+        self.logger.info("Next time manifest will be checked for updates: %s", next_time_fmt)
+
```
andreidenissov-cog, line 90: Self-explanatory: compute next time (approx, but good enough) when "update_storage" will be called,
and log it in standard ISO format.

neuro_san/service/watcher/registries/registry_storage_updater.py
```diff
@@ -87,6 +98,7 @@ def update_storage(self):
         modified, added, deleted = self.observer.reset_event_counters()
 
         if modified == added == deleted == 0:
+            self.log_next_update_time()
             # Nothing happened - go on observing
```
andreidenissov-cog, line 102: If there is no update this time, still log the time for next check.

PR #624, opened 2025-12-11
Some build tweaks

d1donlydfink:
* Use a .coveragerc instead of piling the coverage needs on the pytest command line
* This makes test.yml more repo-agnostic
* Add config to not have tests.yml run more than once per PR checkin after a PR is opened.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,23 @@
+[run]
+# Include specific source so we can tell when entire directories are not measured
+# XXX It'd sure be nice to have a way to make this repo-agnostic.
+source = 
+    neuro_san
+
+# Omit specific files or directories from the coverage measurement
+omit =
+    # Exclude the entire 'tests' directory
+    tests/*
+    # Exclude everything in /usr which usually also means Python standard library
+    /usr/*
+
+# Do not measure coverage for the Python standard library
+cover_pylib = false
+
+[report]
+# Omit specific files or directories from the coverage measurement
+omit =
+    # Exclude the entire 'tests' directory
+    tests/*
+    # Exclude everything in /usr which usually also means Python standard library
+    /usr/*
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -8,6 +8,7 @@ on:
     branches:
       - "**"
   pull_request:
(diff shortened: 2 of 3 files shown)


.github/workflows/tests.yml
```diff
@@ -8,6 +8,7 @@ on:
     branches:
       - "**"
   pull_request:
+    types: [opened, reopened]
```
d1donlydfink, line 11: Add config to not have tests.yml run more than once per PR checkin after a PR is opened.

.github/workflows/tests.yml
```diff
@@ -47,7 +48,7 @@ jobs:
         shell: bash
         run: |
           build_scripts/server_start.sh
-          pytest --verbose -m "not integration and not smoke" --timer-top-n 10 -n auto --cov=neuro_san
+          pytest --verbose -m "not integration and not smoke" --timer-top-n 10 -n auto --cov-config=.coveragerc --cov
```
d1donlydfink, line 51: Use a .coveragerc instead of piling the coverage needs on the pytest command line
This makes tests.yml more repo-agnostic.

In the end, the only thing that makes this repo's tests.yml special is the call to ./build_scripts/server_start.sh

.coveragerc
```diff
@@ -0,0 +1,23 @@
+[run]
+# Include specific source so we can tell when entire directories are not measured
+# XXX It'd sure be nice to have a way to make this repo-agnostic.
+source = 
+    neuro_san
```
d1donlydfink, line 5: Have the source point to the entire neuro_san package.
When this gets copy/pasted to other repos, a few things will happen:
* When copying the new tests.yml, it will prompt the new repo to copy this file
* It should prompt the copier to notice that this is where they need to customize, not in tests.yml

.gitignore
```diff
@@ -36,10 +36,9 @@ site/
 # macOS folder metadata
 .DS_Store
 
-# Temporary integrtion test files
+# Temporary integration test files
 tmp_*
 .coverage
-.coverage*
```
d1donlydfink, line 42: This line is not necessary and covers up any changes to .coveragerc configuration.

PR #627, opened 2025-12-15
Use llm_config.hocon in place of llm_config field

Noravee:
To enable users to modify `llm_config` field more easily since most networks use the same config. This can be done by using `include` in the network hocon files.

If everyone agrees on this format, it will be implemented on neuro-san-studio as well to fix https://github.com/cognizant-ai-lab/neuro-san-studio/issues/325.

### DIFF
--- a/neuro_san/session/async_direct_agent_session.py
+++ b/neuro_san/session/async_direct_agent_session.py
@@ -21,7 +21,7 @@ from typing import Generator
 from typing import List
 
 from asyncio import Future
-import contextlib
+from contextlib import suppress
 from copy import copy
 import logging
 
@@ -217,7 +217,8 @@ class AsyncDirectAgentSession(AsyncAgentSession):
                     response_dict["response"] = message
                     yield response_dict
         finally:
-            with contextlib.suppress(Exception):
+            # Release resources without exceptions
+            with suppress(Exception):
                 await chat_session.delete_resources()
 
     def reset(self):
--- a/neuro_san/session/direct_agent_session.py
+++ b/neuro_san/session/direct_agent_session.py
@@ -20,8 +20,8 @@ from typing import Dict
 from typing import Generator
 from typing import List
 
-import asyncio
-import contextlib
+from asyncio import Future
+from contextlib import suppress
 from copy import copy
 
 from leaf_common.asyncio.async_to_sync_generator import AsyncToSyncGenerator
@@ -175,9 +175,9 @@ class DirectAgentSession(AgentSession):
         # This might take a few minutes, which can be longer than some
         # sockets stay open.
         asyncio_executor: AsyncioExecutor = self.invocation_context.get_asyncio_executor()
-        future: asyncio.Future = asyncio_executor.submit(self.request_id, chat_session.streaming_chat,
-                                                         user_input, self.invocation_context, sly_data,
-                                                         chat_context)
+        future: Future = asyncio_executor.submit(self.request_id, chat_session.streaming_chat,
+                                                 user_input, self.invocation_context, sly_data,
+                                                 chat_context)
         # Ignore the future. Live in the now.
         _ = future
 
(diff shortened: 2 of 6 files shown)


neuro_san/registries/llm_config.hocon
```diff
@@ -0,0 +1,32 @@
...
+        #
+        # | LLM Provider     | `class` Value   | Links for model names |
+        # |------------------|-----------------|-----------------------|
+        # | Amazon Bedrock   | `bedrock`       | https://docs.aws.amazon.com/bedrock/latest/userguide/model-ids.html |
+        # | Anthropic        | `anthropic`     | https://docs.anthropic.com/claude/docs/models-overview |
+        # | Azure OpenAI     | `azure_openai`  | https://learn.microsoft.com/azure/ai-services/openai/concepts/models |
+        # | Google Gemini    | `gemini`        | https://ai.google.dev/gemini-api/docs/models |
+        # | NVIDIA           | `nvidia`        | https://build.nvidia.com/explore/discover |
+        # | Ollama           | `ollama`        | https://ollama.com/library |
+        # | OpenAI           | `openai`        | https://platform.openai.com/docs/models |
```
Noravee, line 17: Provide the table of `class` for supported LLM providers as well as the links for model names of each providers.

neuro_san/registries/llm_config.hocon
```diff
@@ -0,0 +1,32 @@
...
+        # Note:
+        # - For Ollama, the model must be downloaded locally before use.
+        # - Additional LLM parameters (e.g., `temperature`, `max_tokens`) can be configured below.
+        # - Agent-level control (e.g., `max_iterations`) can also be set here.
+        #
+        #   For more details, see:
+        #   - `neuro_san/internals/run_context/langchain/llms/default_llm_info.hocon`
+        #   - `docs/agent_hocon_reference.md`
+
+        "class": "openai",
```
Noravee, line 28: Use `class` here so that users can easily use any models from the supported providers.

neuro_san/registries/math_guy.hocon
```diff
@@ -19,6 +19,12 @@
 # https://github.com/cognizant-ai-lab/neuro-san/blob/main/docs/agent_hocon_reference.md
 
 {
+    # Load the shared LLM configuration from a single source of truth.
+    # This allows users to change the model in one file rather than
+    # modifying the configuration for each agent network.
+    # Note that the file path here is relative to the root level of the repo.
+    include "neuro_san/registries/llm_config.hocon"
```
Noravee, line 26: Use `include` for the `llm_config`.

neuro_san/registries/assess_failure.hocon
```diff
@@ -19,15 +19,17 @@
...
+
     # Optional metadata describing this agent network
     "metadata": {
         "description": "Used by Neuro-SAN Assessor in its test infrastruture to classify agent network failures",
         "tags": ["test_infrastructure"]
     },
-
-    "llm_config": {
-        "model_name": "gpt-4o",
-    },
```
d1donlydfink, line 30: This is a good start, but ...
What I don't like about this is that it removes the simpleness of the example itself.
Maybe there is a way to keep the llm_config definition in each one, but do something like:

```
"llm_config": {

     # We use the default_model_name definition from llm_config.hocon as the value, 
     # to allow convenience changing of all values from a single place for new users testing
     # things out, but your own agent network hocons can include a simple string.
     # For more details see:
     # https://github.com/cognizant-ai-lab/neuro-san/blob/main/docs/agent_hocon_reference.md#llm_config
    
    "model_name": ${default_model_name}
}
```

Can do similar thing with class as well, but emphasize in the comments that class can be optional.

What I don't want is people getting the first impression that only a single model can be used globally among the agent network itself or among the entire server.  Very fine line that only explaining our logic can help.

Sometimes the minimal solution is not the best solution, especially in teaching moments.

ofrancon (approved):
I actually see how people can easily copy `llm_config.hocon` to create a `gemini_config.hocon` or `anthropic_fast_config.hocon` and include them for specifc agents in their `.hocon` files.
So there could be multiple llm_config files, and more than one would be used in a single agent network.
I also like how most of the doc is living in the `llm_config.hocon` file now. Makes it easier to customize it.

PR #629, opened 2025-12-16
Update to leaf-server-common==0.1.23

d1donlydfink:
Takes new leaf-server-common that has fixes in response to opentelemetry API changes

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,7 +7,7 @@
 
 # In-house dependencies
 leaf-common>=1.2.32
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -7,7 +7,7 @@
 
 # In-house dependencies
 leaf-common>=1.2.32
-leaf-server-common>=0.1.22
+leaf-server-common>=0.1.23
```
d1donlydfink, line 10: That's all folks!

PR #631, opened 2025-12-16
Fix async issues : delete_resources() not awaited and AsyncDirectAgentSession not getting answers

d1donlydfink:
* Fix a "delete_resources() not awaited" warning that comes up in testing
* Fix AsyncDirectAgentSessions not getting messages in their queues, especially from external agents.
* Add skip_on_empty to the hocon test harnesses.  This is a nice little trick I found that allows smooth running if all the tests happen to be commented out. It's good templating for future repos.

I found these while attempting an agent hocon test case of a network that uses direct async agents for external agent calls.

@vince-leaf : This will allow tests of agent networks that call out to /external_agents to not require a server running.

### DIFF
--- a/tests/neuro_san/zzz_hocons/test_smoke_test_hocons.py
+++ b/tests/neuro_san/zzz_hocons/test_smoke_test_hocons.py
@@ -42,7 +42,7 @@ class TestSmokeTestHocons(TestCase):
         "music_nerd_pro/combination_responses_with_history_direct.hocon",
 
         # List more hocon files as they become available here.
-    ]))
+    ]), skip_on_empty=True)
     @pytest.mark.timeout(30)  # 30 seconds for this test
     @pytest.mark.smoke
     @pytest.mark.needs_server
@@ -70,7 +70,7 @@ class TestSmokeTestHocons(TestCase):
         "music_nerd_pro_llm_bedrock_claude/combination_responses_with_history_direct.hocon",
 
         # List more hocon files as they become available here.
-    ]))
+    ]), skip_on_empty=True)
     @pytest.mark.timeout(60)  # in seconds for this test
     @pytest.mark.smoke
     @pytest.mark.non_default_llm_provider
@@ -99,7 +99,7 @@ class TestSmokeTestHocons(TestCase):
         "music_nerd_pro_llm_ollama/combination_responses_with_history_http.hocon",
 
         # List more hocon files as they become available here.
-    ]))
+    ]), skip_on_empty=True)
     @pytest.mark.timeout(12 * 60)  # in mins for this test
     @pytest.mark.smoke
     @pytest.mark.needs_server
--- a/tests/neuro_san/zzz_hocons/test_integration_test_hocons.py
+++ b/tests/neuro_san/zzz_hocons/test_integration_test_hocons.py
@@ -43,7 +43,7 @@ class TestIntegrationTestHocons(TestCase):
         "mcp_deepwiki_dict/cail_neuro_san.hocon",
 
         # List more hocon files as they become available here.
-    ]))
+    ]), skip_on_empty=True)
     @pytest.mark.timeout(90)  # 90 seconds for this test
     @pytest.mark.integration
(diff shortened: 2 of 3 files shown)


neuro_san/session/async_direct_agent_session.py
```diff
@@ -217,7 +217,8 @@ async def streaming_chat(self, request_dict: Dict[str, Any]) -> Generator[Dict[s
                     response_dict["response"] = message
                     yield response_dict
         finally:
-            with contextlib.suppress(Exception):
+            # Release resources without exceptions
+            with suppress(Exception):
```
d1donlydfink, line 221: Make it look like the Direct class below as much as possible.

neuro_san/session/direct_agent_session.py
```diff
@@ -175,9 +175,9 @@ def streaming_chat(self, request_dict: Dict[str, Any]) -> Generator[Dict[str, An
         # This might take a few minutes, which can be longer than some
         # sockets stay open.
         asyncio_executor: AsyncioExecutor = self.invocation_context.get_asyncio_executor()
-        future: asyncio.Future = asyncio_executor.submit(self.request_id, chat_session.streaming_chat,
-                                                         user_input, self.invocation_context, sly_data,
-                                                         chat_context)
+        future: Future = asyncio_executor.submit(self.request_id, chat_session.streaming_chat,
+                                                 user_input, self.invocation_context, sly_data,
+                                                 chat_context)
```
d1donlydfink, line 180: import Future directly. It's the only thing we use from asyncio in here.

neuro_san/session/direct_agent_session.py
```diff
@@ -212,11 +212,12 @@ def streaming_chat(self, request_dict: Dict[str, Any]) -> Generator[Dict[str, An
...
-                # We are in a sync run-time environment,
-                # so we need to run the async delete_resources() method
-                # with the asyncio.run() helper.
-                asyncio.run(chat_session.delete_resources())
+            # Release resources without exceptions
+            with suppress(Exception):
+                # Cannot run as if in sync environment, so run async
+                # Use the asyncio_executor so as to not induce other async warnings
+                future: Future = asyncio_executor.submit(self.request_id, chat_session.delete_resources)
+                _ = future
```
d1donlydfink, line 220: This is one of the real fixes:
Do not use asyncio.run() because doing that will cause a warning about how delete_resources() is not awaited.  Instead use the asyncio_executor which is at hand.

neuro_san/session/session_invocation_context.py
```diff
@@ -213,4 +213,8 @@ def safe_shallow_copy(self) -> SessionInvocationContext:
         # with direct sessions.
         invocation_context.queue: AsyncCollatingQueue = AsyncCollatingQueue()
 
+        # We need a different journal in order for direct sessions to call external agents
+        # with direct sessions.
+        invocation_context.journal: Journal = MessageJournal(invocation_context.queue)
```
d1donlydfink, line 220: Oh boy this was a bear to find.
Create a new MessageJournal with the new queue we create here when safe_shallow_copy()-ing.

This prevents some very nasty non-answers when using AsyncDirectAgentSessions for external tools.

tests/neuro_san/zzz_hocons/test_smoke_test_hocons.py
```diff
@@ -42,7 +42,7 @@ class TestSmokeTestHocons(TestCase):
         "music_nerd_pro/combination_responses_with_history_direct.hocon",
 
         # List more hocon files as they become available here.
-    ]))
+    ]), skip_on_empty=True)
```
d1donlydfink, line 45: Add skip_on_empty to the hocon test harnesses. This is a nice little trick I found that allows smooth running if all the tests happen to be commented out. It's good templating for future repos.

neuro_san/session/direct_agent_session.py
andreidenissov-cog: I don't quite get it: was this warning affecting functionality? Or was it just warning?

PR #637, opened 2025-12-18
Upgrade langchain-openai to a minimum of 1.0.0 to match langchain and langchain-core

ofrancon:
Upgraded `langchain-openai` minimum version to `1.0.0`
It now matches the version requirements of `langchain` and `langchain-core` and should probably be kept in synch with them.

For reference, the image generation API has changed and versions prior to `langchain-openai` `1.0.0` won't work for  the image generation tool in Neuro SAN Studio (added in https://github.com/cognizant-ai-lab/neuro-san-studio/pull/492)


andreidenissov-cog (approved):
LGTM.

PR #639, opened 2025-12-18
NS-638: Fix conversion for ChatMessages.

andreidenissov-cog:
This PR fixes the crash encountered when using MCP tool from neuro-san agent.
The issue is: when doing ChatMessage conversion by means of ChatMessageConverter class,
we rely on deepcopy() method. But it turns out, some of our messages (especially coming back from Langchain MCP adaptors) could contain a lot of lower-level objects which are neither copy-able nor json serializable.
So the need to do a conversion in more careful manner, cutting out all "bad" sub-objects which will not go into final chat results sent back to a client.

### DIFF
--- a/neuro_san/service/generic/chat_message_converter.py
+++ b/neuro_san/service/generic/chat_message_converter.py
@@ -33,7 +33,11 @@ class ChatMessageConverter(DictionaryConverter):
         :param obj: The object (chat response) to be converted into a dictionary
         :return: chat response dictionary in format expected by clients
         """
+        # Do "the safe" copy, moving over only json serializable values.
+        # Note that values of ChatMessageType are passed through,
+        # because they are handled by a separate post-processing step.
         response_dict = self.to_json_safe(obj)
+        # This is where ChatMessageType enum values are converted:
         self.convert(response_dict)
         return response_dict
 


neuro_san/service/generic/chat_message_converter.py
```diff
@@ -34,7 +33,7 @@ def to_dict(self, obj: object) -> Dict[str, object]:
         :param obj: The object (chat response) to be converted into a dictionary
         :return: chat response dictionary in format expected by clients
         """
-        response_dict = copy.deepcopy(obj)
+        response_dict = self.to_json_safe(obj)
         self.convert(response_dict)
```
andreidenissov-cog, line 41: Do "the safe" copy, moving over only json serializable constructs.
Note that values of ChatMessageType are passed through, because they are handled
by a separate post-processing step.

neuro_san/service/generic/chat_message_converter.py
```diff
@@ -73,3 +76,38 @@ def from_dict(self, obj_dict: Dict[str, object]) -> object:
                 to return None.
         """
         raise NotImplementedError
+
+    JSON_PRIMITIVES = (str, int, float, bool, type(None))
+    SKIP_TYPES = (ChatMessageType,)
+
+    def to_json_safe(self, obj: Any):
```
d1donlydfink, line 83: Nice. Might be worthwhile to make a unit test for this and maybe eventually make its way down to leaf-common

d1donlydfink (approved):

neuro_san/service/generic/chat_message_converter.py
d1donlydfink: Curious though: What kinds of messages ended up being "offensive"?

I ask because in neuro_san/internals/messages, there is a notion of a TracedMessages, on which you can specify which fields should not be converted.  Maybe this ultimately belongs in that package somehow,
because that's where all the other ChatMessage-related stuff lives (This seems generic enough to not live in server-generic).

Issue #632, opened 2025-12-17
Remove temperature and max_tokens in default_config from default llm info


Noravee:
These parameters should come from class, model or user configs.

PR #644, opened 2026-01-06
Remove temperature and max_tokens in default_config from default llm info

Noravee:
#632 

- Remove temperature and max_tokens from default_config
- Remove default temp from doc

### DIFF
--- a/docs/agent_hocon_reference.md
+++ b/docs/agent_hocon_reference.md
@@ -214,9 +214,10 @@ see the [llm_info_hocon_reference](./llm_info_hocon_reference.md).
 
 #### temperature
 
-Pretty much any of the LLMs will take a floating-point temperature parameter as an argument.
-Roughly speaking, temperature is a number between 0.0 and 1.0 that indicates a relative amount of randomness
-in answers provided by the LLM.  By default this value is 0.7.
+Most LLMs accept a floating-point temperature parameter, except reasoning-focused models like gpt-5.
+In simple terms, temperature is a non-negative value that controls the randomness or creativity of the model’s outputs:
+higher values produce more diverse responses, while lower values make answers more deterministic.
+If not specified, the default temperature varies depending on the provider.
 
 #### Other LLM-specific Parameters
 
--- a/neuro_san/internals/run_context/langchain/llms/default_llm_info.hocon
+++ b/neuro_san/internals/run_context/langchain/llms/default_llm_info.hocon
@@ -1038,10 +1038,6 @@
 
         "model_name": "gpt-4o",             # The string name of the default model to use.
 
-        "temperature": 0.7,                 # The default LLM temperature (randomness) to use.
-                                            # Values are floats between 0.0 (least random) to
-                                            # 1.0 (most random).
-
         "prompt_token_fraction": 1.0,       # The fraction of total tokens (not necessarily words
                                             # or letters) to use for a prompt. Each model_name
                                             # has a documented number of max_tokens it can handle
@@ -1049,10 +1045,6 @@
                                             # which goes into the calculation involved in
                                             # get_max_prompt_tokens().
                                             # By default the value is 1.0.
-
-        "max_tokens": null,                 # The maximum number of tokens to use in
-                                            # computing prompt tokens. By default this comes from
-                                            # the model description in this class.
 
         # The following are more agent-level control as opposed to LLM control above.
 


neuro_san/internals/run_context/langchain/llms/default_llm_info.hocon
```diff
@@ -1038,10 +1038,6 @@
 
         "model_name": "gpt-4o",             # The string name of the default model to use.
 
-        "temperature": 0.7,                 # The default LLM temperature (randomness) to use.
```
Noravee, line 1041: The default temperature from `langchain-openai` used to be `0.7`. This was changed to `None` in newer version.
The temperature from `openai` ranges from 0.0-2.0, and default at `1.0`. 
See https://platform.openai.com/docs/api-reference/chat/create?utm_source=chatgpt.com#chat_create-temperature

docs/agent_hocon_reference.md
```diff
@@ -214,9 +214,10 @@ see the [llm_info_hocon_reference](./llm_info_hocon_reference.md).
 
 #### temperature
 
-Pretty much any of the LLMs will take a floating-point temperature parameter as an argument.
-Roughly speaking, temperature is a number between 0.0 and 1.0 that indicates a relative amount of randomness
-in answers provided by the LLM.  By default this value is 0.7.
+Most LLMs accept a floating-point temperature parameter, except reasoning-focused models like gpt-5.
+In simple terms, temperature is a non-negative value that controls the randomness or creativity of the model’s outputs:
+higher values produce more diverse responses, while lower values make answers more deterministic.
+If not specified, the default temperature varies depending on the provider.
```
Noravee, line 220: Modify the doc accordingly.

neuro_san/internals/run_context/langchain/llms/default_llm_info.hocon
```diff
@@ -1050,10 +1046,6 @@
                                             # get_max_prompt_tokens().
                                             # By default the value is 1.0.
 
-        "max_tokens": null,                 # The maximum number of tokens to use in
-                                            # computing prompt tokens. By default this comes from
-                                            # the model description in this class.
```
d1donlydfink, line 1055: If this is null, removing this seems unnecessary, as we will just use what is in the model description.
Thus we are removing a nice pointer to other information.

What problem are we trying to solve here?

neuro_san/internals/run_context/langchain/llms/default_llm_info.hocon
d1donlydfink: Hmm. This could very subtly alter the default probabalistic behavior of a lot of people's carefully tuned networks by removing this here.  

According to some rando threads from Googling, it seems the default value for temperature for some models (including openai) is 1.0.   In other model classes within this file we specifically set it to 0.7 (see "gemini" spec in this file).

I think we need to tread very carefully here, as this "just nuke it" approach is kind of ham-handed and doesn't empathize with the plights of existing deployments enough.

I will ask again here: What problem are we trying to solve?

If there are asserts that are being thrown with reasoning models (I think this is what I heard from last discussion), then maybe there is a safer course of action here with an eye for backwards compatibility.

Assuming that we are trying to avoid out-of-the-box asserts from simple hoc on-based model changes... What if we took an approach where we defined a new class for temperature-averse models for a given class?

Take openai as an example. We can define an "openai-base" class in this file with one set of args where temperature is not set, used by reasoning models from openai. We can also define another "openai-compatibility" class in this file where temperature is set.  This allows for a backwards compatibility path with nuance.

The work would not stop here, as azure-openai definitions and other model classes would need similar attention.

Thoughts?

docs/agent_hocon_reference.md
Noravee: Here are some thoughts I wanted to share regarding the default LLM configuration:

- **Provider defaults**: When users do not explicitly specify any LLM parameters, it may be preferable to fall back to the provider’s default parameter values.
- **Misconfigured fields**: In the past, we have encountered cases where users unintentionally used incorrect fields (for example, `use_model_name` instead of `model_name`). In such situations, the network runs with the default model, while the user believes the specified model is being used. This can lead to confusion and makes troubleshooting more difficult.
- **Agent-level parameters**: Some agent-level parameters, such as `verbose`, are currently included in the default LLM configuration, but they are not actually applied. The default LLM config is only used when instantiating the LLM, whereas these parameters are consumed during agent creation via a different configuration. As a result, updating these parameters in `default_llm_info_file.hocon` does not have any practical effect.

I completely understand the concern around backward compatibility. With that in mind, one possible compromise could be:

- Remove agent-level parameters from the default LLM configuration, since they are not currently used.
- Keep the remaining LLM-related parameters unchanged.
- Introduce warnings to clearly inform users when default model and parameter values are being applied.

Looking ahead, we could also consider marking these agent-level parameters as **deprecated**, with the intention of removing them in a future release (for example, `neurosan==0.8`).

I’d really appreciate your thoughts and feedback on this approach.

neuro_san/internals/run_context/langchain/llms/default_llm_info.hocon
ofrancon: I think we simply can't keep up with the default values providers decide to use. 
Temperature is not a thing anymore with "reasoning" models for OpenAI. See https://community.openai.com/t/temperature-in-gpt-5-models/1337133 for instance.
OpenAI used to use `0.7` but now they use `1` as a default.

The comment for this temperature param is also wrong:
```
# Values are floats between 0.0 (least random) to
                                            # 1.0 (most random).
```
Temperature for OpenAI is actually between `0` and `2`. See https://platform.openai.com/docs/api-reference/chat/create#chat_create-temperature

For Gemini 3, the doc strongly recommends setting default temperature to 1. So this default value of `0.7` is actually "corrupting" the model:
```
Changing the temperature (setting it below 1.0) may lead to unexpected behavior, such as looping or degraded performance, particularly in complex mathematical or reasoning tasks.
```
See https://ai.google.dev/gemini-api/docs/gemini-3#temperature

So I don't think the backward compatibility potential issue is going to be as much of a problem as the current one we have with latest models. In short, I think we should let LLM providers decide, for each model, which default temperature to use. They know better than us, and it's not a 'one size fits all' parameter.

But this is indeed a breaking change, so I like @Noravee 's idea of deprecating the `temperature` parameter for now and removing it only in a future non bug-fix release.

Issue #641, opened 2025-12-28
Test driver fails on Windows due to illegal characters in thinking_dir path


M-Elsaied:
#### **Description**

When running integration tests on Windows using `data_driven_agent_test_driver.py`, the process fails during the interaction phase. The framework attempts to create a "thinking" directory using a Unix-style path and a timestamp containing colons (`:`). Windows does not allow colons in folder names (except for drive designators), resulting in an `OSError: [WinError 123]`.

#### **Error Log**

```text
E OSError: [WinError 123] The filename, directory name, or volume label syntax is incorrect: '/tmp/agent_test/2025-12-28-13:17:37_industry_agent'
...
venv\Lib\site-packages\neuro_san\test\driver\data_driven_agent_test_driver.py:247: in interact
    os.makedirs(thinking_dir)

```

#### **Root Cause**

At this point from what I see, it's because

1. **Hardcoded Unix Path:** The path starts with `/tmp/`, which is not a standard root on Windows.
2. **Illegal Characters:** The timestamp format used to generate the directory name includes colons (`HH:MM:SS`).

#### **Environment**

* **OS:** Windows 11
* **Python Version:** 3.13.0
* **Neuro SAN Version:** 0.6.21

PR #653, opened 2026-01-13
NS-641 Remove colons from filename to fix hocon test cases on Windows

d1donlydfink:
Fixes worst part of #641 which had windows machines failing on the directory created because of colons in the file path

### DIFF
--- a/neuro_san/test/driver/data_driven_agent_test_driver.py
+++ b/neuro_san/test/driver/data_driven_agent_test_driver.py
@@ -232,7 +232,7 @@ Need at least {num_need_success} to consider {hocon_file} test to be successful.
 
         # Prepare the processor
         now = datetime.now()
-        datestr: str = now.strftime("%Y-%m-%d-%H:%M:%S")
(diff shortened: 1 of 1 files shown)


neuro_san/test/driver/data_driven_agent_test_driver.py
```diff
@@ -232,7 +232,7 @@ def interact(self, agent: str, session: AgentSession, interaction: Dict[str, Any
 
         # Prepare the processor
         now = datetime.now()
-        datestr: str = now.strftime("%Y-%m-%d-%H:%M:%S")
+        datestr: str = now.strftime("%Y-%m-%d-%H_%M_%S")
```
d1donlydfink, line 235: Assuming this will do the trick as we do not really have windows machines to test on.

vince-leaf: ChatGPS said this.
which is valid on Windows, Linux, and macOS.

PR #657, opened 2026-01-13
NS-640 Paralellize success_ratio test attempts and allow for AGENT_TEST_THINKING_BASIS env var

d1donlydfink:
* Parallelize running of test attempts into N threads, where N is defined in the denominator of the test case's `success_ratio`.  Interestingly, tests with several iterations do not necessarily take the same amount of time as a single test does. There is a few-second linear component added for each new iteration - unclear why this happens. Still it's better than sequential tests.
* Allow for AGENT_TEST_THINKING_BASIS env var setting of where the agent_test directory goes. This also allows Windows users an out.

### DIFF
--- a/neuro_san/internals/journals/tool_argument_reporting.py
+++ b/neuro_san/internals/journals/tool_argument_reporting.py
@@ -27,8 +27,9 @@ class ToolArgumentReporting:
     """
 
     # List of keys for policy objects that cannot be serialized in a message.
-    # These are set in AbstractClassActivation.
-    POLICY_OBJECT_KEYS: List[str] = ["reservationist", "progress_reporter"]
+    # The first two are set in AbstractClassActivation.
+    # The last one is injected by Langgraph when using MCP tools with langchain-mcp-adapters>=0.1.14.
+    POLICY_OBJECT_KEYS: List[str] = ["reservationist", "progress_reporter", "runtime"]
 
     @staticmethod
     def prepare_tool_start_dict(tool_args: Dict[str, Any],


neuro_san/test/driver/data_driven_agent_test_driver.py
```diff
@@ -102,27 +106,35 @@ def one_test(self, hocon_file: str):
 
         # Extract the second-to-last part of the path,the parent folder name.
         fixture_hocon_name = os.path.basename(os.path.dirname(hocon_file))
-        for index in range(num_iterations):
```
d1donlydfink, line 105: This for loop moves inside the `with ThreadPoolExecutor` block

neuro_san/test/driver/data_driven_agent_test_driver.py
```diff
@@ -102,27 +106,35 @@ def one_test(self, hocon_file: str):
...
-            assert_capture = AssertCapture(self.asserts_basis)
-            iteration_asserts.append(assert_capture)
+            futures: List[Future] = []
+            for iteration_index in range(num_iterations):
 
-            # Perform a single iteration of the test.
-            self.one_iteration(test_case, assert_capture, timeouts, fixture_hocon_name)
+                # Don't include an iteration index if there is only one iteration to do.
+                if num_iterations == 1:
+                    iteration_index = None
```
d1donlydfink, line 118: We are now passing the iteration_index down to the test iteration so that when we collect thinking files for a particular iteration, they can be routed to the right place on the filesystem.
However, when there is only a single iteration to be done, we need to send the signal (via None here) that we do not want an iteration suffix on the directory.

neuro_san/test/driver/data_driven_agent_test_driver.py
```diff
@@ -102,27 +106,35 @@ def one_test(self, hocon_file: str):
...
+                asserts: List[AssertionError] = assert_capture.get_asserts()
+                if len(asserts) > 0:
+                    # Not successful
+                    continue
+
+                num_successful += 1
+                if num_successful == num_need_success:
+                    # Don't look at more tests than we actually need to
+                    executor.shutdown(wait=False, cancel_futures=True)
+                    break
```
d1donlydfink, line 137: Shutdown when we reach our limit.

neuro_san/test/driver/data_driven_agent_test_driver.py
```diff
@@ -140,15 +152,35 @@ def one_test(self, hocon_file: str):
...
+        :param iteration_index: The index of this test iteration for the success_ratio
+        :return: An AssertCapture object for the iteration.
+        """
+        # Capture the asserts for this iteration and add it to the list for later
+        assert_capture = AssertCapture(self.asserts_basis)
+
+        # Perform a single iteration of the test.
+        self.one_iteration(test_case, assert_capture, timeouts, fixture_hocon_name, iteration_index)
+
+        return assert_capture
```
d1donlydfink, line 171: Have the AssertCapture be returned as the basis for the parallelization Future result for the test iteration.

neuro_san/test/driver/data_driven_agent_test_driver.py
```diff
@@ -196,7 +228,7 @@ def one_iteration(self, test_case: Dict[str, Any], asserts: AssertForwarder,
                     session.reset()
 
                 chat_context = self.interact(agent, session, interaction, chat_context, asserts,
-                                             timeouts, fixture_hocon_name)
+                                             timeouts, fixture_hocon_name, iteration_index)
```
d1donlydfink, line 231: Add plumbing for iteration_index

neuro_san/test/driver/data_driven_agent_test_driver.py
```diff
@@ -233,10 +267,15 @@ def interact(self, agent: str, session: AgentSession, interaction: Dict[str, Any
         # Prepare the processor
         now = datetime.now()
         datestr: str = now.strftime("%Y-%m-%d-%H_%M_%S")
-        thinking_file: str = f"/tmp/agent_test/{datestr}_agent.txt"
+        basis_dir: str = os.environ.get("AGENT_TEST_THINKING_BASIS", "/tmp/agent_test")
```
d1donlydfink, line 270: Get the basis of the agent_test results from an env var if that is set.
This allows these to move around, and also gives Windows folks an out if they need to add C: in front of stuff.

neuro_san/test/driver/data_driven_agent_test_driver.py
```diff
@@ -233,10 +267,15 @@ def interact(self, agent: str, session: AgentSession, interaction: Dict[str, Any
...
+        thinking_file: str = f"{basis_dir}/{datestr}_agent.txt"
+
         # Added fixture_hocon_name to thinking_dir
         # for better uniqueness and traceability across different test fixtures.
-        thinking_dir: str = f"/tmp/agent_test/{datestr}_{fixture_hocon_name}_agent"
+        index_suffix: str = ""
+        if iteration_index is not None:
+            index_suffix = f"_{iteration_index}"
```
d1donlydfink, line 277: Do not add an iteration_index if we get the signal of None that there is only a single iteration.

docs/test_case_hocon_reference.md
```diff
@@ -61,6 +61,7 @@ Single string values can be:
 | direct (default)| Connect directly to the agent via a neuro-san library call - no server required. |
 | http  |  Connect to the agent via a server via http |
 | https |  Connect to the agent via a server via secure http |
+| mcp   |  Connect to the agent via MCP protocol |
```
d1donlydfink: @andreidenissov-cog recently checked in the ability for AgentSessionFactory to use MCP sessions.
As it happens, this also comes for free as a new hocon test case connection, so document it.

neuro_san/test/driver/data_driven_agent_test_driver.py
andreidenissov-cog, line 311: Ugh, this one will not work for MCP sessions, because MCP session returns response in MCP standard format, which is different from native neuro-san. In CLI tool, I do different "unpacking" of chat results to make them suitable for  further processor.process_message() handling.
So something needs to be done.

andreidenissov-cog (changes requested):
See comment about chat results in MCP session.

neuro_san/test/driver/data_driven_agent_test_driver.py
d1donlydfink: OK. Let's file a ticket for that. This is good stuff for other people as it is if we disregard the new mcp functionality.

d1donlydfink: Filed: https://github.com/cognizant-ai-lab/neuro-san/issues/660

d1donlydfink: Removed mcp from test case docs until we get it right.

vince-leaf:
Assuming the number of available threads will be determined and used is handled by the OS.

neuro_san/test/driver/data_driven_agent_test_driver.py
d1donlydfink: This is a very good catch. Filed https://github.com/cognizant-ai-lab/neuro-san/issues/660

PR #666, opened 2026-01-16
NS-532 Let sly_data through MCP service connection.

andreidenissov-cog:
This is somewhat speculative PR to find a way for sly_data to go through in case we are using MCP connection.
Main idea: sly_data on the upstream direction (from agent network to a client) is already controlled 
by configuration "allow" block. So by default it would be blocked on neuro-san level already.
What PR does is lets sly data through MCP transport IF we allow it explicitly.
Tested by running cli client for "math_guy" agent with MCP connection.

### DIFF
--- a/neuro_san/service/mcp/processors/mcp_tools_processor.py
+++ b/neuro_san/service/mcp/processors/mcp_tools_processor.py
@@ -124,10 +124,6 @@ class McpToolsProcessor:
             async with asyncio.timeout(tool_timeout_seconds):
                 result_generator = service.streaming_chat(input_request, metadata)
                 async for result_dict in result_generator:
-
-                    print(f">>>>>>>>>>>>>>>>>>>>> MCP result_dict: {json.dumps(result_dict, indent=4)}")
-
-
                     partial_response, structure_data, sly_data = await self._extract_tool_response_part(result_dict)
                     if partial_response is not None:
                         response_text = response_text + partial_response


neuro_san/service/mcp/processors/mcp_tools_processor.py
```diff
@@ -119,15 +119,18 @@ async def call_tool(self, request_id, metadata: Dict[str, Any],
...
         response_text: str = ""
         response_structure: Dict[str, Any] = None
+        response_sly_data: Dict[str, Any] = None
         try:
             async with asyncio.timeout(tool_timeout_seconds):
                 result_generator = service.streaming_chat(input_request, metadata)
                 async for result_dict in result_generator:
-                    partial_response, structure_data = await self._extract_tool_response_part(result_dict)
+                    partial_response, structure_data, sly_data = await self._extract_tool_response_part(result_dict)
                     if partial_response is not None:
```
andreidenissov-cog, line 128: Add extracting potential sly data component from internal chat result.

neuro_san/service/mcp/processors/mcp_tools_processor.py
```diff
@@ -189,6 +194,8 @@ async def build_tool_call_result(
             # For backward compatibility, also add text version of structure:
             structure_str: str = f"```json\n{json.dumps(result_structure, indent=2)}\n```"
             result_text = result_text + structure_str
+        if result_sly_data is not None:
+            call_result["result"]["sly_data"] = result_sly_data
         call_result["result"]["content"][0]["text"] = McpRequestUtil.safe_message(result_text)
```
andreidenissov-cog, line 199: If present, put sly_data dictionary into MCP response structure.

neuro_san/service/mcp/processors/mcp_tools_processor.py
```diff
@@ -206,21 +213,23 @@ async def _get_tool_description(self, agent_name: str, metadata: Dict[str, Any])
...
         response_part_dict: Dict[str, Any] = response_dict.get("response", {})
         response_type: str = response_part_dict.get("type", "")
         if response_type == "AGENT_FRAMEWORK":
             text: str = response_part_dict.get("text", None)
             structure_data: Dict[str, Any] = response_part_dict.get("structure", None)
-            return text, structure_data
-        return None, None
+            sly_data: Dict[str, Any] = response_part_dict.get("sly_data", None)
+            return text, structure_data, sly_data
+        return None, None, None
```
andreidenissov-cog, line 232: Extract sly data from internally generated chat message.

neuro_san/session/mcp_chat_response_dictionary_converter.py
```diff
@@ -50,14 +50,21 @@ def to_dict(self, obj: object) -> Dict[str, object]:
...
+        content_dict: Dict[str, Any] = content_seq[0]
+        response_dict: Dict[str, Any] = {
             "response": {
                 "type": ChatMessageType.AGENT_FRAMEWORK.name,
-                "text": response.get("text", "")
+                "text": content_dict.get("text", ""),
+                # This is a workaround to ensure message will make it through
+                # client-side message filters
+                "chat_context": {}
             },
```
andreidenissov-cog, line 61: It turns out this is needed to pass through message filter if we want sly data to get processed.

neuro_san/session/mcp_chat_response_dictionary_converter.py
```diff
@@ -50,14 +50,21 @@ def to_dict(self, obj: object) -> Dict[str, object]:
...
             "mcp_response": chat_response
         }
+        sly_data_dict: Dict[str, Any] = result.get("sly_data", None)
+        if sly_data_dict:
+            response_dict["response"]["sly_data"] = sly_data_dict
+        return response_dict
```
andreidenissov-cog, line 67: Stick sly data into neuro-san formatted response.

neuro_san/service/mcp/processors/mcp_tools_processor.py
d1donlydfink: So, I think this is back to the previous argument of:

We cannot assume that an MCP consumer will only ever be a neuro-san session.
We have to assume that it is going back to a dumber LLM-based agent using MCP as a tool that is going to process the sly_data and give that to whatever LLM provider is out there.
We also cannot assume that just because sly_data is allowed out to a neuro_san client, that doesn't mean it does not contain PII or secrets.

For this reason, I don't think we should be sending sly_data back via MCP ever.
Perhaps this is worth adding a comment to the code so we don't try to do this a 3rd time.

Maybe the bigger question here is: Why do you want to do this for MCP?

If there were some other control for sly_data for non-neuro-san clients (a2a would have this problem too), then that could be off by default and selectively turned on. That would maybe be a manifest thing?

neuro_san/session/mcp_chat_response_dictionary_converter.py
d1donlydfink: ... but maybe we don't?

d1donlydfink (changes requested):
One biggie:


So, I think this is back to the previous argument of:

We cannot assume that an MCP consumer will only ever be a neuro-san session.
We have to assume that it is going back to a dumber LLM-based agent using MCP as a tool that is going to process the sly_data and give that to whatever LLM provider is out there.

For this reason, I don't think we should be sending sly_data back via MCP ever.
Perhaps this is worth adding a comment to the code so we don't try to do this a 3rd time.

Since you're trying again here, this makes me wonder what your motivation to include sly_data in an MCP response is?
Is there a test you are trying to make work?
If so, maybe we need to reframe the problem.

Maybe I am misunderstanding something.

PR #669, opened 2026-01-21
Add client-side local-time timestamps to events going into thinking files

### LINKED ISSUE #670: Timestamp in thinking file
As a user developing complex agentic systems, I want the logs/tingling_dir/ files to contain timestamp in order to debug system failures. It's key to determine 1/ if an agent was called, 2/ when the agent was called, 3/ which arguments the agent received + timestamp, and 4/ which results the agent returned + timestamp.

d1donlydfink:
An obviously necessary feature requested by @jcyn2004 

Output in thinking files now looks like this:

```
[SYSTEM] @ 2026-01-21 10:03:11:

You are a decision assistant.
...
```

The idea is that client-side timestamps is likely good enough for most cases.
Server-side timestamps could require a protocol change and innards manipulation.
For now this is the Easy (and still very useful) thing to do.
Sequence is at least preserved, which is really the most important aspect.

Fixes #670

### DIFF
--- a/neuro_san/client/thinking_file_message_processor.py
+++ b/neuro_san/client/thinking_file_message_processor.py
@@ -125,7 +125,7 @@ class ThinkingFileMessageProcessor(MessageProcessor):
                 # Retry with a uuid as file name.
                 # If this fails, there's no helping ya.
                 origin_filename = str(uuid.uuid4())
-                self._write_to_file(origin_filename, origin_str, message_type_str, use_origin, text)
+                self._write_to_file(origin_filename, origin_str, message_type_str, use_origin, text, timestamp)
 
                 # Squirell that uuid away so results continue to go to the same
                 # file over and over again.


neuro_san/client/thinking_file_message_processor.py
```diff
@@ -87,6 +90,7 @@ def write_message(self, response: Dict[str, Any], origin_str: str):
         :param origin_str: The string representing the origin of the message
         """
 
+        timestamp: float = time()
```
d1donlydfink, line 93: Take the timestamp as early as possible.
For now we are only doing client-side timestamps, as adding information would require a protocol change, and sequence received is probably good enough for most events.

neuro_san/client/thinking_file_message_processor.py
```diff
@@ -150,7 +155,8 @@ def _write_to_file(self, origin_filename: str, origin_str: str,
                 thinking.write(f"Agent: {origin_str}\n")
 
             # Write the message out
-            thinking.write(f"\n[{message_type_str}{use_origin}]:\n")
+            timestamp_str: str = strftime("%Y-%m-%d %H:%M:%S", localtime(timestamp))
+            thinking.write(f"\n[{message_type_str}{use_origin}] @ {timestamp_str}:\n")
```
d1donlydfink, line 159: Write out the timestamp in local time.

neuro_san/client/thinking_file_message_processor.py
jcyn2004: Thank you. This will help and accelerate development/test/debugging/release of agentic systems. I have file the following request: https://github.com/cognizant-ai-lab/neuro-san/issues/670

jcyn2004:
Thank you. This will help and accelerate development/test/debugging/release of agentic systems. I have file the following request: https://github.com/cognizant-ai-lab/neuro-san/issues/670

neuro_san/client/thinking_file_message_processor.py
jcyn2004: Thank you. This will help and accelerate development/test/debugging/release of agentic systems. I have file the following request: https://github.com/cognizant-ai-lab/neuro-san/issues/670

vince-leaf: It is about time to timestamp each item in the thinking file. This would help somewhat for debugging.

Having the time before the original message would give us quicker clarity.

Issue #688, opened 2026-01-29
Allow request metadata to come into the CodedTool


d1donlydfink:
A pattern we have been adopting lately is one where extra information/objects come in via the args that augment what comes in from the tool call from the agents.  Sometimes these just come for free and sometimes there is a specific "allow" block enabler.  Examples include: origin information, ProgressReporter, Reservationist.

There seems to be a legit need from Mo to access request metadata via coded tools.

PR #691, opened 2026-02-02
NS-688 pass request header metadata to sly_data

d1donlydfink:
Pass the dictionary of metadata that is in the HTTP request as its own field called "request_metadata" in sly_data.

Was able to verify manually using a modified math_guy calculator that:
* "request_metadata" does indeed show up as expected
* request_metadata does not go back to client by default as expected.

Addresses #688

### DIFF
--- a/neuro_san/internals/chat/data_driven_chat_session.py
+++ b/neuro_san/internals/chat/data_driven_chat_session.py
@@ -118,8 +118,11 @@ class DataDrivenChatSession(RunTarget):
 
         # Update sly data with metadata from the request headers so this information
         # can be made available in CodedTools, privately, without any leakage to LLMs.
+        # Only add this if it doesn't already exist w/rt what the user gave us.
         if self.sly_data.get("request_metadata") is None:
-            self.sly_data["request_metadata"] = invocation_context.get_metadata()
+            # Make a deep copy so the server's idea of the original metadata doesn't get modified
+            # should any CodedTool somehow get the idea to modify it.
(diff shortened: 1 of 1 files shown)


neuro_san/internals/chat/data_driven_chat_session.py
```diff
@@ -108,14 +108,22 @@ async def set_up(self, invocation_context: InvocationContext,
...
         if sly_data is not None:
             self.sly_data.update(sly_data)
 
+        # Update sly data with metadata from the request headers so this information
+        # can be made available in CodedTools, privately, without any leakage to LLMs.
+        # Only add this if it doesn't already exist w/rt what the user gave us.
+        if self.sly_data.get("request_metadata") is None:
+            # Make a deep copy so the server's idea of the original metadata doesn't get modified
+            # should any CodedTool somehow get the idea to modify it.
+            self.sly_data["request_metadata"] = copy.deepcopy(invocation_context.get_metadata())
```
d1donlydfink, line 125: This is one of those things that the hardest part is figuring out just where the easy 2 lines of code belongs for your feature.

Noravee (approved):
LGTM.

neuro_san/internals/chat/data_driven_chat_session.py
andreidenissov-cog: Will this metadata contain all original incoming request http headers?
I understand this is the goal, right?
Or: we'll be handling this by defining our "request metadata" list separately?

PR #693, opened 2026-02-03
NS-586 Tweak how test driver handles thinking files

d1donlydfink:
There had been a bug where no thinking file output was being generated by tests when AGENT_TEST_THINKING_BASIS was set.  This PR cleans up that and a couple other things:

* If AGENT_TEST_THINKING_BASIS env var is **not** set, then we **no longer** write any output files or directories at all.   This is a change from previous behavior where directories for tests were created even when the env var was not set.
* If AGENT_TEST_THINKING_BASIS is set, we now correctly output thinking files to a timestamped directory for the test
* For pytests, change the name of the directory to the name of the test
* For the Assessor, change the name of the directory to the basename of the the test hocon file
* The format of the directory timestamp has changed to YYYY-MM-DD_HH-mm-SS

To be clear about the **change in behavior**: If you want thinking files from your pytest runs and/or the assessor runs,
you need to explicitly do the following:
```
export AGENT_TEST_THINKING_BASIS=/tmp/agent_test
# ... or wherever
```

Addresses #586

### DIFF
--- a/neuro_san/test/assessor/assessor.py
+++ b/neuro_san/test/assessor/assessor.py
@@ -49,7 +49,8 @@ class Assessor:
 
         # Run the tests per the test case, collecting failure information in the AssertForwarder
         asserts = AssessorAssertForwarder()
-        test_name: str = os.path.basename(self.args.test_hocon)
+        test_name: str = None
+        test_name, _ = os.path.splitext(os.path.basename(self.args.test_hocon))
         driver = DataDrivenAgentTestDriver(asserts, test_name=test_name)
 
         print(f"Testing {self.args.test_hocon}:")


neuro_san/test/assessor/assessor.py
```diff
@@ -48,7 +49,9 @@ def main(self):
 
         # Run the tests per the test case, collecting failure information in the AssertForwarder
         asserts = AssessorAssertForwarder()
-        driver = DataDrivenAgentTestDriver(asserts)
+        test_name: str = None
+        test_name, _ = os.path.splitext(os.path.basename(self.args.test_hocon))
+        driver = DataDrivenAgentTestDriver(asserts, test_name=test_name)
```
d1donlydfink, line 54: Use the name of the test case hocon file as a basis for the output directories.

neuro_san/test/driver/data_driven_agent_test_driver.py
```diff
@@ -55,7 +55,7 @@ class DataDrivenAgentTestDriver:
 
     TEST_KEYS: List[str] = ["text", "structure", "sly_data"]
 
-    def __init__(self, asserts: AssertForwarder, fixtures: FileOfClass = None):
+    def __init__(self, asserts: AssertForwarder, fixtures: FileOfClass = None, test_name: str = None):
```
d1donlydfink, line 58: Pass a test_name to use as a basis for thinking file output.

neuro_san/test/driver/data_driven_agent_test_driver.py
```diff
@@ -265,36 +266,52 @@ def interact(self, agent: str, session: AgentSession, interaction: Dict[str, Any
...
-        # Create the directory anew
-        os.makedirs(thinking_dir)
-
-        input_processor = StreamingInputProcessor("", thinking_file, session, thinking_dir)
+        thinking_dir: str = None
+
+        # A reasonable default here is basis_dir = "/tmp/agent_test", but we
+        # don't want to write thinking files out if no one wants them.
+        basis_dir: str = os.environ.get("AGENT_TEST_THINKING_BASIS")
+        if basis_dir is not None and len(basis_dir) > 0:
```
d1donlydfink, line 274: Rework the logic for how to put together the directory for the thinking files for this test iteration.
1 - Don't output any thinking files if the AGENT_TEST_THINKING_BASIS is not set at all.

neuro_san/test/driver/data_driven_agent_test_driver.py
```diff
@@ -265,36 +266,52 @@ def interact(self, agent: str, session: AgentSession, interaction: Dict[str, Any
...
+            now = datetime.now()
+            datestr: str = now.strftime("%Y-%m-%d_%H-%M-%S")
```
d1donlydfink, line 276: 2 - New format for timestamp.

neuro_san/test/driver/data_driven_agent_test_driver.py
```diff
@@ -265,36 +266,52 @@ def interact(self, agent: str, session: AgentSession, interaction: Dict[str, Any
...
+
+            # Add a test name to thinking_dir
+            # for better uniqueness and traceability across different test fixtures.
+            use_name: str = self.test_name
+            if use_name is None:
+                use_name: str = fixture_hocon_name
```
d1donlydfink, line 282: 3 - Use the test_name that was passed in as the basis for the directory if that is available.

neuro_san/test/driver/data_driven_agent_test_driver.py
```diff
@@ -265,36 +266,52 @@ def interact(self, agent: str, session: AgentSession, interaction: Dict[str, Any
...
+
+            thinking_dir = f"{basis_dir}/{datestr}_{use_name}{index_suffix}"
+
+            # Remove any contents that might be there already.
+            # Writing over existing dir will just confuse output.
+            # Although it is unlikely that two tests run at the same time...
+            if os.path.exists(thinking_dir):
+                shutil.rmtree(thinking_dir)
+            # Create the directory anew
+            os.makedirs(thinking_dir)
```
d1donlydfink, line 297: 4 - Only remove/create the directory if we are actually going to put thinking_file output in there.
This will prevent feckless test turds from accumulating.

neuro_san/test/driver/data_driven_agent_test_driver.py
```diff
@@ -265,36 +266,52 @@ def interact(self, agent: str, session: AgentSession, interaction: Dict[str, Any
...
 
         # Prepare the request
         text: str = interaction.get("text")
         sly_data: str = interaction.get("sly_data")
+
         # By having level to MINIMAL avoid unnecesssary thinking file(s) created.
         # MAXIMAL set to have thinking files.
+        default_chat_filter: str = "MINIMAL"
+        if thinking_dir is not None:
+            default_chat_filter: str = "MAXIMAL"
```
d1donlydfink, line 311: When we are using thinking files, be sure to set the chat filter to MAXIMAL to actually get output to capture.

neuro_san/test/unittest/dynamic_hocon_unit_tests.py
```diff
@@ -81,7 +81,7 @@ def one_test_hocon(self, test_case: TestCase, test_name: str, test_hocon: str):
 
         # Set up the driver
         asserts = UnitTestAssertForwarder(test_case)
-        driver = DataDrivenAgentTestDriver(asserts)
+        driver = DataDrivenAgentTestDriver(asserts, test_name=test_name)
```
d1donlydfink, line 84: For unit tests, pass the name of the test as the test_name basis for thinking file output.

PR #707, opened 2026-02-05
NS-684 Make MCP server retain chat history.

andreidenissov-cog:
This PR fixes an issue with MCP neuro-san server not keeping agent chat context and therefore "forgetting" things during the chat. We use "structuredContent" protocol-specified option for sending chat context in MCP response.
Tested by running agent_cli client chatting to music_nerd agent.


neuro_san/session/mcp_chat_response_dictionary_converter.py
```diff
@@ -50,14 +50,18 @@ def to_dict(self, obj: object) -> Dict[str, object]:
...
         response: Dict[str, Any] = content_seq[0]
-        return {
+        final_response: Dict[str, Any] = {
             "response": {
                 "type": ChatMessageType.AGENT_FRAMEWORK.name,
                 "text": response.get("text", "")
             },
             "mcp_response": chat_response
         }
+        if structured_data is not None:
```
andreidenissov-cog, line 63: If "structuredContent" is present in MCP chat response, it may contain both chat context dictionary,
and structured data generated from agent response.
We pass both along as parts of neuro-san native response format.

neuro_san/service/mcp/processors/mcp_tools_processor.py
```diff
@@ -218,7 +218,8 @@ async def _extract_tool_response_part(
         response_type: str = response_part_dict.get("type", "")
         if response_type == "AGENT_FRAMEWORK":
             text: str = response_part_dict.get("text", None)
-            structure_data: Dict[str, Any] = response_part_dict.get("structure", None)
+            # For final response, there could be chat_context structured data we need to return:
+            structure_data: Dict[str, Any] = response_part_dict.get("chat_context", None)
             return text, structure_data
```
andreidenissov-cog, line 226: AGENT_FRAMEWORK chat message with our final "answer" may carry chat_context data structure,
which we pass along to be put into MCP response.

d1donlydfink: So then what happens to the "structure" field?
Cut off nose to spite face?

neuro_san/session/mcp_chat_response_dictionary_converter.py
```diff
@@ -50,14 +50,18 @@ def to_dict(self, obj: object) -> Dict[str, object]:
         content_seq: Sequence[Dict[str, Any]] = result.get("content", [])
         if len(content_seq) == 0:
             return empty
+        structured_data: Dict[str, Any] = result.get("structuredContent", None)
```
d1donlydfink, line 54: Is "structuredContent" key ours or MCPs?
If ours, then please use snake_case
If MCPs, then please add comment as to why it's not snake_case.

andreidenissov-cog: Commented - it is indeed part of MCP standard format.

neuro_san/service/mcp/processors/mcp_tools_processor.py
andreidenissov-cog: Good catch. Fixed that, we now aggregate both optional parts.

PR #710, opened 2026-02-05
NS-705 Temporarily pin leaf-common to 1.2.32

andreidenissov-cog:
Until I figure out the cause of unit tests failures.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,7 +6,7 @@
 # requirements-build.txt
 
 # In-house dependencies
-leaf-common>=1.2.32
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -6,7 +6,7 @@
 # requirements-build.txt
 
 # In-house dependencies
-leaf-common>=1.2.32
+leaf-common==1.2.32
 leaf-server-common>=0.1.23
```
andreidenissov-cog, line 10: Pin the version which is known to work.
Until unit test failures are root-caused.

PR #711, opened 2026-02-06
NS-410 Optional Neuro SAN Authorization interface w/ initial OpenFGA implementation

d1donlydfink:
As requested (directly or indirectly) by a number of people.

Create an Authorizer interface for optional policy-agnostic inclusion into a Neuro SAN server.
The default implementation is a NullAuthorizer which lets all requests through as has been happening since the beginning.

Note that Authorization - which asks "Does actor X have permission/relation Y on resource Z?"  -
is not to be confused with Authentication - which asks "Is this user who they say they are?".
Having this capability bumps up our street cred in the corporate deployment world.

The Authorizer asynchronous interface defines a few methods that must be implemented:
* authorize(actor, action, resource) -> bool    - used by HTTP handlers and MCP tool invocation.
* list(actor, action, resource) -> List[str] -   used by Concierge service
* grant(actor, relation, resource)  - will eventually be used by Reservations and authorize.py example in studio
* revoke(actor, relation, resource) - will eventually be used by Reservations and authorize.py example in studio
* query(actor, relation, resource) - unused, but included for completeness.

Authorizers must be used within an `async with` statement to properly bound their sessions with their implementation's external servers.

With this change we also offer an optional OpenFGA implementation, something which is a freely available
relations-based access control (ReBAC) model.  Just about all other authorization policies you can think of are a subset of ReBAC.

OpenFGA classes are lazily accessed so as not to introduce bloat into the main dependency set.
Should you want to use the OpenFGA implementation, you'll get a runtime message that says you need to install "openfga-sdk" if you want to go further.

There will be a corresponding plugin checkin to the neuro-san-studio repo which provides some testing policy and scripts for people to try it out manually.

Tested:
* always yes implementation (default)  through regular unit tests
* always no implementation - manually
* OpenFga implementation manually - with authorized users and without
* operation with and without openfga_sdk
* docker build and run without OpenFGA

Future work:
* I have not gone so far as to integrate Authorization into agent Reservations just yet.

### DIFF
--- a/neuro_san/deploy/Dockerfile
+++ b/neuro_san/deploy/Dockerfile
@@ -297,6 +297,8 @@ ENV AGENT_MCP_ONLY="false"
 #   neuro_san.internals.authorization.null.always_yes_authorizer.AlwaysYesAuthorizer (default)
 #   neuro_san.internals.authorization.null.always_no_authorizer.AlwaysNoAuthorizer
 #   neuro_san.internals.authorization.openfga.open_fga_authorizer.OpenFgaAuthorizer
+# Feel free to make your own as long as it derives from this interface:
+#   neuro_san.internals.authorization.interfaces.authorizer.Authorizer
 ENV AGENT_AUTHORIZER=""
 
 # The request metadata field to use as the actor id for authorization


neuro_san/deploy/Dockerfile
```diff
@@ -286,4 +286,51 @@ ENV AGENT_MCP_ENABLE="true"
 # while disabling neuro-san own http API.
 ENV AGENT_MCP_ONLY="false"
 
+#
+# Authorization
+#
```
d1donlydfink, line 291: A whole bunch of new env vars to control Authorization.

neuro_san/internals/authorization/factory/authorizer_factory.py
```diff
@@ -0,0 +1,56 @@
...
+    """
+
+    @staticmethod
+    def create_authorizer(auth_type: str = None) -> Authorizer:
+        """
+        Factory method for creating the appropriate Authorizer instance
+        :param auth_type: The type of authorization to use. If None (very common),
+                        we default to the value in the environment variable AGENT_AUTHORIZER.
+                        If that is not set, we default to NullAuthorizer.
+        """
```
d1donlydfink, line 37: Factory class for getting an Authorizer handle when needed.

neuro_san/internals/authorization/interfaces/abstract_authorizer.py
```diff
@@ -0,0 +1,173 @@
...
+
+from leaf_common.config.resolver import Resolver
+
+from neuro_san.service.authorization.interfaces.authorizer import Authorizer
+
+
+class AbstractAuthorizer(Authorizer):
+    """
+    Partial Authorizer implementation that allows for late-binding via a resolver.
+    """
```
d1donlydfink, line 32: Same as the Authorizer interface but with some niceties included like:
* a logger
* a Resolver for lazy binding

What you must do at least here is provide an authorize() and list() implementation.

neuro_san/service/interfaces/agent_authorizer.py
```diff
@@ -0,0 +1,50 @@
...
+from typing import List
+from typing import Tuple
+
+from neuro_san.service.generic.async_agent_service_provider import AsyncAgentServiceProvider
+
+
+class AgentAuthorizer:
+    """
+    Interface for authorizing agent specifics given metadata from a request.
+    """
```
d1donlydfink, line 30: This is the interface that the server uses to ask authorization questions it cares about.

This will grow when I get to Reservations work.

neuro_san/internals/authorization/interfaces/authorizer.py
```diff
@@ -0,0 +1,159 @@
...
+from typing import Any
+from typing import Dict
+from typing import List
+
+
+class Authorizer:
+    """
+    An interface for authorization.
+    This is based on what we need from what packages like OpenFGA or Oso provide.
+    """
```
d1donlydfink, line 29: Main Authorizer interface.  This is the interface that the AgentAuthorizer uses to call some Authorizer implementation abstractly.

We already have an AlwaysYes, AlwaysNo and OpenFga implementation.

Since this was lifted from a previous project we are fairly confident that other implementations like Oso will fit with this.

FWIW: This abstraction corresponds to a ReBAC (Relations Based Access Control) system, which is pretty much a superset of any other auth scheme.

neuro_san/service/interfaces/permission.py
```diff
@@ -14,22 +14,17 @@
...
-See class comment for details
-"""
-from neuro_san.service.generic.async_agent_service_provider import AsyncAgentServiceProvider
 
+from enum import Enum
 
-class AgentAuthorizer:
+
+# class syntax
+class Permission(Enum):
```
d1donlydfink, line 22: Diff is not relevant.
This is just an Enum to do things like CRUD permissions without copy/paste

neuro_san/internals/authorization/null/always_yes_authorizer.py
```diff
@@ -0,0 +1,121 @@
...
+from typing import List
+
+from neuro_san.service.authorization.interfaces.abstract_authorizer import AbstractAuthorizer
+
+
+class NullAuthorizer(AbstractAuthorizer):
+    """
+    Null implementation of the Authorizer interface that lets all requests through.
+    This gives us the behavior we have had since the beginning.
+    """
```
d1donlydfink, line 28: This gives us the authorization policy we already have - everything goes.

neuro_san/internals/authorization/openfga/open_fga_authorizer.py
```diff
@@ -0,0 +1,436 @@
...
+
+from neuro_san.service.authorization.interfaces.abstract_authorizer import AbstractAuthorizer
+from neuro_san.service.authorization.interfaces.authorizer import Authorizer
+from neuro_san.service.authorization.openfga.open_fga_store_cache import OpenFgaStoreCache
+
+
+class OpenFgaAuthorizer(AbstractAuthorizer):
+    """
+    AbstractAuthorizer implementation for Open FGA ("Fine Grained Authorization").
+    """
```
d1donlydfink, line 32: OpenFGA implementation for the Authorizer interface.
This code and its 2 helper classes were riffed from a previous project

neuro_san/service/http/handlers/base_request_handler.py
```diff
@@ -47,8 +47,6 @@ class BaseRequestHandler(RequestHandler):
     request_id_counter: AsyncAtomicCounter = AsyncAtomicCounter()
 
     # pylint: disable=attribute-defined-outside-init
-    # pylint: disable=too-many-arguments
-    # pylint: disable=too-many-positional-arguments
```
d1donlydfink, line 51: Remove several unnecessary pylint disables.

neuro_san/service/http/handlers/base_request_handler.py
```diff
@@ -137,12 +131,23 @@ async def get_service(self, agent_name: str, metadata: Dict[str, Any]) -> AsyncA
         :return: instance of AsyncAgentService if it is defined for this agent
                  None otherwise
         """
-        service_provider: AsyncAgentServiceProvider = self.agent_policy.allow(agent_name)
+        is_authorized: bool = False
+        service_provider: AsyncAgentServiceProvider = None
+        is_authorized, service_provider = await self.agent_policy.allow_agent(agent_name, metadata)
```
d1donlydfink, line 136: What was formerly allow() is now called allow_agent() on the policy object.
It now returns enough information as to whether or not we service the request, or return a 404 or 403.

neuro_san/service/http/handlers/concierge_handler.py
```diff
@@ -31,19 +32,27 @@ class ConciergeHandler(BaseRequestHandler):
...
         """
         Implementation of GET request handler for "concierge" API call.
         """
         metadata: Dict[str, Any] = self.get_metadata()
         self.application.start_client_request(metadata, "/api/v1/list")
         network_storage_dict: Dict[str, AgentNetworkStorage] = self.server_context.get_network_storage_dict()
         public_storage: AgentNetworkStorage = network_storage_dict.get("public")
+
+        # See what the authorizer says
+        allowed_agents: List[str] = await self.agent_policy.list_agents(metadata)
```
d1donlydfink, line 45: Call the list_agents() on the policy.  Note this is all asynchronous now.

neuro_san/service/http/handlers/concierge_handler.py
```diff
@@ -53,3 +62,30 @@ def get(self):
...
+        Remove agents which are not allowed.
+
+        :param allowed_agents: A list of agent names that are allowed by the authorization system
+        :param result_dict: A dictionary version of the ConciergeResponse
+                protobuf structure. Has the following keys:
+            "agents" - the sequence of dictionaries describing available agents
+        :return: A dictionary version of the ConciergeResponse
+                protobuf structure. Has the following keys:
+            "agents" - the sequence of dictionaries describing available agents
+        """
```
d1donlydfink, line 77: A helper method which takes the concierge output and squares it with what the auth policy says we are allowed to list given the user.

neuro_san/service/http/server/agent_authorization_policy.py
```diff
@@ -0,0 +1,118 @@
...
+        self.allowed_agents: Dict[str, AsyncAgentServiceProvider] = allowed_agents
+
+        # Only need to get these once
+        self.authorizer: Authorizer = AuthorizerFactory.create_authorizer()
+        self.actor_key: str = environ.get("AGENT_AUTHORIZER_ACTOR_KEY", "User")
+        self.actor_id_metadata_key: str = environ.get("AGENT_AUTHORIZER_ACTOR_ID_METADATA_KEY", "user_id")
+        self.resource_key: str = environ.get("AGENT_AUTHORIZER_RESOURCE_KEY", "AgentNetwork")
+        self.allow_relation: str = environ.get("AGENT_AUTHORIZER_ALLOW_RELATION", Permission.READ.value)
+
+    async def allow_agent(self, agent_name: str, metadata: Dict[str, Any]) -> Tuple[bool, AsyncAgentServiceProvider]:
```
d1donlydfink, line 54: Policy object which relieves HttpServer of the burden of having to do auth policy too.
these are now all async calls so as to call external authorization servers sanely within our environment if need be.

neuro_san/service/http/server/agent_authorization_policy.py
```diff
@@ -0,0 +1,118 @@
...
+
+        resource: Dict[str, Any] = {
+            "type": self.resource_key,
+            "id": agent_name
+        }
+
+        # Consult the authorizer
+        is_authorized: bool = False
+        async with self.authorizer as auth:
+            is_authorized = await auth.authorize(actor, self.allow_relation, resource)
```
d1donlydfink, line 77: Here the policy puts together the authorization question in the abstract, and then lets the Authorizer make the decisions.

neuro_san/service/http/server/http_server.py
```diff
@@ -63,7 +64,7 @@
 DEFAULT_REQUEST_LIMIT: int = 1000 * 1000
 
 
-class HttpServer(AgentAuthorizer, AgentStateListener):
+class HttpServer(AgentStateListener):
```
d1donlydfink, line 67: The server is no longer an AgentAuthorizer.

neuro_san/service/http/server/http_server.py
```diff
@@ -108,6 +109,7 @@ def __init__(self,
         self.forwarded_request_metadata: List[str] = forwarded_request_metadata.split(" ")
         self.logger = HttpLogger(self.forwarded_request_metadata)
         self.allowed_agents: Dict[str, AsyncAgentServiceProvider] = {}
+        self.authorization_policy: AgentAuthorizer = AgentAuthorizationPolicy(self.allowed_agents)
```
d1donlydfink, line 112: Create an AgentAuthorizer delegate.

neuro_san/service/http/server/http_server.py
```diff
@@ -277,7 +276,7 @@ def build_request_data(self) -> Dict[str, Any]:
             raise ValueError(f"Failed to load '{self.openapi_service_spec_path}'") from exc
 
         return {
-            "agent_policy": self,
+            "agent_policy": self.authorization_policy,
```
d1donlydfink, line 279: This allows the policy object to get to all the requests.

neuro_san/service/http/server/http_server.py
```diff
@@ -218,9 +220,6 @@ def make_app(self, requests_limit: int, logger: EventLoopLogger):
 
         return HttpServerApp(handlers, requests_limit, logger, self.forwarded_request_metadata)
 
-    def allow(self, agent_name) -> AsyncAgentServiceProvider:
-        return self.allowed_agents.get(agent_name)
```
d1donlydfink, line 222: Homie don't play that no more.
Plus, we add a list_agents() call on the new policy object anyway. Better that grows on its own someplace else.

neuro_san/internals/authorization/openfga/open_fga_store_cache.py
```diff
@@ -0,0 +1,131 @@
...
+    to have their own client.
+
+    In testing, however, we have the opposite: multiple stores for (likely)
+    single threaded tests so that they specifically do not stomp on anything
+    real code would use.
+    """
+
+    # A mapping of store names (like we get in the env var)
+    # to store ids (like we get back from the OpenFGA server to initialize clients with).
+    store_name_to_id: Dict[str, str] = {}
```
d1donlydfink, line 55: Worth noting that we are using class variables as a singleton-esque way to store the Open FGA notion of "store_id" (which is a hash code) given a more palettable "store_name".  We only really ever use one of these in production, but the previous abstraction allowed for more for testing purposes.

neuro_san/internals/authorization/openfga/open_fga_store_cache.py
```diff
@@ -0,0 +1,131 @@
...
+        # Lazy loading of OpenFgaInit class which directly uses OpenFga SDK types.
+        # pylint: disable=import-outside-toplevel
+        from neuro_san.service.authorization.openfga.open_fga_init import OpenFgaInit
+
+        if store_id is None:
+            # Note: Synchronous lock is required here
+            init = OpenFgaInit()
+            with OpenFgaStoreCache.lock:
+                store_id = await init.initialize_store(store_name)
+                OpenFgaStoreCache.store_name_to_id[store_name] = store_id
```
d1donlydfink, line 89: If we have never seen this store before we have to do some extra initialization with the OpenFGA server to be sure it has the policy we want.

neuro_san/internals/authorization/openfga/open_fga_authorizer.py
```diff
@@ -0,0 +1,436 @@
...
+    async def handle_special_user_writes(self, writes: List[Any], actor: Dict[str, Any], resource: Dict[str, Any]):
+        """
+        This method is called from grant() or revoke() to handle special cases where
+        the actor or resource is a special user.
+
+        :writes: A list of ClientTuples to be written
+        :actor: The actor dictionary with the keys "type" and "id"
+        :resource: The resource dictionary with the keys "type" and "id"
+        """
+        _ = writes, actor, resource
```
d1donlydfink, line 436: A useful out from previous project where we had the notion of special users that could do special things.

neuro_san/internals/authorization/openfga/open_fga_init.py
```diff
@@ -0,0 +1,283 @@
...
+        _ = open_fga_client
+
+    async def is_seen_auth_model(self, open_fga_client: OpenFgaClient, auth_model_id: str) -> bool:
+        """
+        Optional method to allow for self-updating authorization servers
+        :param auth_model_id: The auth model id to look for
+        :return: True if the auth model was found
+        """
+        _ = open_fga_client, auth_model_id
+        return False
```
d1donlydfink, line 283: Two useful methods from previous projects that allowed for storing seen auth model information in the auth store itself. Don't always need that, but we allow an out for it.

neuro_san/service/mcp/processors/mcp_tools_processor.py
```diff
@@ -102,10 +103,17 @@ async def call_tool(self, request_id, metadata: Dict[str, Any],
         # pylint: disable=too-many-arguments
         # pylint: disable=too-many-positional-arguments
 
-        service_provider: AsyncAgentServiceProvider = self.agent_policy.allow(tool_name)
+        is_authorized: bool = False
+        service_provider: AsyncAgentServiceProvider = None
+        is_authorized, service_provider = await self.agent_policy.allow_agent(tool_name, metadata)
```
d1donlydfink, line 116: MCP needs to ask auth policy too.

It's possible I might be missing something here w/rt Concierge and listings of agents + auth.
@andreidenissov-cog : Please chime in if you know better than I do.

neuro_san/session/abstract_http_service_agent_session.py
```diff
@@ -128,5 +128,7 @@ def help_message(self, path: str) -> str:
            provide any of the necessary certificates for this to work and any certs used will
            need to be well known.  If you're unfamiliar with this process, it's a big deal.
            Try regular http instead.
+        10. Is your server running running with an AGENT_AUTHORIZER?  If so check your
+            authorization policy/database to be sure the user has the correct permissions for the network.
```
d1donlydfink: A new thing to check for when the server returns "huh?" when requesting an agent.

README.md
```diff
@@ -40,7 +40,8 @@ Neuro SAN also offers:
     * data-driven test cases
     * the ability for LLMs to test your agent networks
     * an Assessor app which classifies the modes of failure for your agents, given a data-driven test case
-* MCP protocol API
+* MCP protocol API - Every Neuro SAN server can be an MCP Server.
+* per-user authorization for Agent Networks - optional implementations include: OpenFGA
```
d1donlydfink, line 44: Update the README to tout new capability.

neuro_san/session/abstract_http_service_agent_session.py
andreidenissov-cog: running running == running fast? :-)

neuro_san/service/http/server/agent_authorization_policy.py
```diff
@@ -0,0 +1,118 @@
...
+
+        # The network still needs to exist.
+        service_provider: AsyncAgentServiceProvider = self.allowed_agents.get(agent_name)
+        return is_authorized, service_provider
```
andreidenissov-cog, line 81: So is it possible to have (True, None)?
That is, agent is authorized, but we don't have it really?

neuro_san/service/authorization/factory/authorizer_factory.py
```diff
@@ -0,0 +1,56 @@
...
+        if auth_type is None:
+
+            auth_classname: str = environ.get("AGENT_AUTHORIZER")
+            if auth_classname is not None and len(auth_classname) > 0:
+                authorizer = ResolverUtil.create_instance(auth_classname,
+                                                          "AGENT_AUTHORIZER env var",
+                                                          Authorizer)
+            else:
+                authorizer = NullAuthorizer()
+        elif auth_type == "openfga":
```
andreidenissov-cog: Hmm. Looks like a security hole?
Make a mistake in AGENT_AUTHORIZER class name, intentional or not,
and our server is open for anything.
Maybe we need a special "Dr.NO" authorizer, which will forbid everything in this case?

neuro_san/service/http/server/agent_authorization_policy.py
d1donlydfink: Technically yes.
This would come as a race condition where the manifest on the filesystem was updated before the authorization database was updated. 
Given the ordering of determination of HTTP error codes for agent access,
I believe we'd send a 404 (no resource) over a 403 (no permission) in this situation.

neuro_san/service/authorization/factory/authorizer_factory.py
d1donlydfink: Interesting idea.

d1donlydfink: Added AlwaysNoAuthorizer for uncomplicated testing.

d1donlydfink: Also added a comment to note the fact that if the class is not found an exception will be raised within the ResolverUtil method.

neuro_san/session/abstract_http_service_agent_session.py
d1donlydfink: Fixed and made instructions more from an end-user perspective than a developer.

neuro_san/internals/authorization/null/always_no_authorizer.py
```diff
@@ -0,0 +1,121 @@
...
+from typing import Dict
+from typing import List
+
+from neuro_san.internals.authorization.interfaces.abstract_authorizer import AbstractAuthorizer
+
+
+class AlwaysNoAuthorizer(AbstractAuthorizer):
+    """
+    Implementation of the Authorizer interface that lets no requests through.
+    """
```
d1donlydfink, line 27: Per @andreidenissov-cog 's intriguing comment. This implementation never lets anything through.
Allows us to test behavior without the burden of setting up some big server deal.

neuro_san/service/mcp/processors/mcp_tools_processor.py
andreidenissov-cog, line 82: If we put an agent name through auth logic here, returned tool_dict could be None.
In this case, don't put it in tools_description list.

neuro_san/service/mcp/processors/mcp_tools_processor.py
```diff
@@ -196,9 +204,14 @@ async def build_tool_call_result(
         return call_result
 
     async def _get_tool_description(self, agent_name: str, metadata: Dict[str, Any]) -> Dict[str, Any]:
-        service_provider: AsyncAgentServiceProvider = self.agent_policy.allow(agent_name)
-        if service_provider is None:
+
+        is_authorized: bool = False
+        service_provider: AsyncAgentServiceProvider = None
+        is_authorized, service_provider = await self.agent_policy.allow_agent(agent_name, metadata)
+
```
andreidenissov-cog, line 219: So do we have the same policy for calling agents and for listing them?
Maybe it is how it should be, just wondering.

d1donlydfink: No - the idea of authorize() being different than list() is that each question can be answered with a single transaction to the authorization server

neuro_san/service/mcp/processors/mcp_tools_processor.py
d1donlydfink: Ah. This is where the (new) list() authorization should go. Thanks!

neuro_san/service/mcp/processors/mcp_tools_processor.py
```diff
@@ -63,9 +63,17 @@ async def list_tools(self, request_id, metadata: Dict[str, Any]) -> Dict[str, An
...
+        # See which agents the user has access to per authorization policy
+        authorized_agents: List[str] = await self.agent_policy.list_agents(metadata)
+
         public_storage: AgentNetworkStorage = self.network_storage_dict.get("public")
         tools_description: List[Dict[str, Any]] = []
         for agent_name in public_storage.get_agent_names():
+
+            # Skip agents that are not authorized
+            if agent_name not in authorized_agents:
+                continue
```
d1donlydfink, line 75: With this and the block above, we do not list agents in MCP listing that the user does not have access to.

neuro_san/service/mcp/processors/mcp_tools_processor.py
d1donlydfink: See additions above.

neuro_san/service/mcp/processors/mcp_tools_processor.py
d1donlydfink: Not entirely sure how to test this via agent_cli.
What I want to do is this:
```
python -m neuro_san.client.agent_cli --mcp --list
```
But what I get is this:
```
ValueError: session_type mcp is not understood
```

Filed https://github.com/cognizant-ai-lab/neuro-san/issues/736

PR #728, opened 2026-02-12
Add startup scripts for server and client on macOS/Linux and Windows

makkzone:
Added the startup scripts which acts as single script to run the application.

### DIFF
--- a/neuro_san/deploy/Dockerfile
+++ b/neuro_san/deploy/Dockerfile
@@ -116,6 +116,40 @@ ENV APP_ENTRYPOINT=${PACKAGE_DEPLOY}/entrypoint.sh
 
 ARG NEURO_SAN_VERSION
 
+
+#
+# API Keys for LLM Access
+#
+# Which env vars you need may vary depending on which LLM(s) you choose for your own setup.
+# See https://github.com/cognizant-ai-lab/neuro-san/blob/main/docs/agent_hocon_reference.md#llm_config
+# for details on different LLM providers' needs for out-of-the-box support.
+#
+# Note that you really really don't want to leak secrets by checking the actual values into source control.
+# These are best injected into the container via environment variables at runtime ala:
+#   docker run -e OPENAI_API_KEY=...
+#
+
+# OpenAI/ChatGPT access
+ENV OPENAI_API_KEY=""
+
+# Amazon Bedrock access
+ENV AWS_ACCESS_KEY_ID=""
+ENV AWS_SECRET_ACCESS_KEY=""
+
+# Anthropic/Claude access
+ENV ANTHROPIC_API_KEY=""
+
+# MicroSoft Azure/OpenAI access
+# OPENAI_API_KEY is a fallback for AZURE_OPENAI_API_KEY
+# ENV AZURE_OPENAI_API_KEY=""
+ENV AZURE_OPENAI_ENDPOINT=""
+
+# Google/Gemini access
+ENV GOOGLE_API_KEY=""
+
+# NVidia LLM access
+ENV NVIDIA_API_KEY=""
+
 #
 # Server configuration
 #


README.md
```diff
@@ -42,6 +42,23 @@ Neuro SAN also offers:
     * an Assessor app which classifies the modes of failure for your agents, given a data-driven test case
 * MCP protocol API
 
+## Quick Start
+
+**🚀 For the easiest way to get started, use our automated startup scripts!**
```
d1donlydfink: You should add items to the effect of:

These assume:
* you have Python 3.12 or better installed on your machine already.
* you have a virtual environment for python installed on your machine already.

startup/README.md
```diff
@@ -0,0 +1,136 @@
+# Neuro-SAN Startup Scripts
+
+This folder contains easy-to-use startup scripts for launching the Neuro-SAN server and client on different operating systems.
```
d1donlydfink: "the Neuro-SAN server and client" makes it sound like there will only ever be one.
That's not the intent with this repo.  Perhaps better:

This folder contains easy-to-use startup scripts for launching the Neuro SAN server for this repo and client on different operating systems.   We hope you will eventually create your own networks on your own servers for your own clients in your own repo. Enjoy!

startup/start-server.bat
```diff
@@ -0,0 +1,95 @@
...
+)
+echo.
+
+REM Check for API keys
+if not defined OPENAI_API_KEY (
+    echo WARNING: OPENAI_API_KEY environment variable is not set!
+    echo Some features may not work without API keys.
+    echo Set it with: set OPENAI_API_KEY=your-key-here
+    echo.
+)
```
d1donlydfink: Worth noting that OPENAI_API_KEY is not strictly required, and actually I believe there already exists much better LLM-agnostic help than this when the key is not set, which I think this kind of error checking would end up masking, therefore generating more questions than answers.  See https://github.com/cognizant-ai-lab/neuro-san/blob/main/neuro_san/internals/run_context/langchain/util/api_key_error_check.py#L79

... here and other places.

startup/README.md
d1donlydfink: FWIW: It's either neuro-san (like the repo) or Neuro SAN (for the acronym), never Neuro-SAN.
(We sometimes get blowback for even Neuro SAN 🤷 )

startup/README.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Quick Start
+
+### macOS / Linux
+
+**Start the server:**
+```bash
+chmod +x startup/start-server.sh
+./startup/start-server.sh
```
d1donlydfink: Seems like you should be able to check in a script with execution permissions with git already, so this type of thing shouldn't be necessary.

d1donlydfink: here and other places.

startup/README.md
```diff
@@ -0,0 +1,136 @@
...
+```
+
+### Server Mode (Connects to Running Server)
+```bash
+# macOS/Linux
+./startup/start-client.sh hello_world --http
+
+# Windows
+startup\start-client.bat hello_world --http
+```
```
d1donlydfink: What I worry about here is that you might start to obscure what agent_cli is doing with shell additions.
Perhaps worth a comment somewhere that if people need more control than this, to simply use agent_cli directly.

startup/README.md
```diff
@@ -0,0 +1,136 @@
...
+## Available Agents
+
+Based on the server logs, you can use any of these agents:
+- `hello_world` - Simple greeting agent (default)
+- `copy_cat` - Echo agent
+- `math_guy` - Math operations
+- `date_time` - Current date and time
+- `music_nerd` - Music-related queries
+- `gist` - Terse announcements
+- And many more...
```
d1donlydfink: I'm not a fan of having to maintain this list in documentation.  Makes this repo seem limiting.
What you really should consider doing is both:
* adding a --list functionality to the client so that people can see all the agents available
* Add a pointer to neuro-san-studio repo which is our real tutorial/example showcase.
FWIW: The agents that are in here are just enough to test the Neuro SAN engine, which is why we point noobs at neuro-san-studio.

Also: "gist" is not about terse announcements. It is specifically for the test infrastructure.

startup/README.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Environment Variables
+
+### Required
+- `OPENAI_API_KEY` - Your OpenAI API key
```
d1donlydfink: Not necessarily required if you are using  music_nerd_pro_llm_anthropic.hocon, for instance.

startup/README.md
```diff
@@ -0,0 +1,136 @@
...
+
+### Optional
+- `AGENT_HTTP_PORT` - Server port (default: 8080)
+- `AGENT_SERVER_NAME` - Server name for health reporting
+- Other provider API keys (AWS, Azure, Anthropic, etc.)
```
d1donlydfink: Perhaps provide a pointer to the neuro_san/deploy/Docker container, which is where we keep our documentation on all the env vars people can set.

startup/README.md
d1donlydfink: Also, to make it less confusing, maybe we should call this directory "quick-start", to make the mental link to the top-level README clearer.  I wouldn't want to leave the impression that this is what we expect cluster deployments to use.

startup/README.md
```diff
@@ -0,0 +1,136 @@
...
+
+### Setting Environment Variables
+
```
d1donlydfink: This section seems redundant with the above until one realizes you are teaching people how to set env vars on different OS-es.  Do we really need to do that level of education in this repo?

startup/README.md
```diff
@@ -0,0 +1,136 @@
...
+```bash
+pip install -r requirements.txt
+```
+
+## Server Endpoints
+
+Once the server is running:
+- **Server URL:** http://localhost:8080
+- **Health Check:** http://localhost:8080/health
+- **API endpoints:** See the OpenAPI spec for details
```
d1donlydfink: You should provide the curl command to get the OpenAPI spec from the server
and/or provide a link to the source.

startup/README.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Stopping the Server
+
+Press `Ctrl+C` in the terminal where the server is running.
```
d1donlydfink: Again, do we really need this level of total noob education in this repo?
Your audience either already knows this or will be asking "how do I open a terminal?"
Perhaps better suited for neuro-san-studio.

startup/start-server.sh
```diff
@@ -0,0 +1,81 @@
...
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+PROJECT_DIR="$(dirname "$SCRIPT_DIR")"
+
+cd "$PROJECT_DIR"
+
+echo "Project directory: $PROJECT_DIR"
+echo ""
+
+# Check if virtual environment exists
+if [ ! -d "venv" ] && [ ! -d ".venv" ]; then
```
vince-leaf: Assuming you are running in a fresh container or environment. Therefore, this if statement is always True.
If False, I suggest deleting it 1st, then creating a new venv. This will avoid dependencies on outdated versions and related issues.

quick-start/start-server.sh
```diff
@@ -0,0 +1,81 @@
...
+# Start the server
+echo "==============================================="
+echo "  Starting Neuro-SAN Server on port 8080"
+echo "==============================================="
+echo ""
+echo "Server will be available at: http://localhost:8080"
+echo "Press Ctrl+C to stop the server"
+echo ""
+
+python -m neuro_san.service.main_loop.server_main_loop
```
vince-leaf, line 85: I would suggest:
Add a port readiness check before and after to start_server.sh (waits for server to be ready after launch — would require backgrounding the process).

note:
If you run python -m neuro_san.service.main_loop.server_main_loop and the HTTP port (default 8080) is already bound by another server instance, the OS will refuse to bind the socket, and the process will abort with an Address already in use error. So the server already has implicit port conflict detection at the OS level. Weird behavior afterward.

makkzone (commented):
@d1donlydfink @vince-leaf Thank you for reviewing it.

I have made the changes accordingly.

I have seen some references to Neuro-SAN in existing code, I have updated them too.

d1donlydfink (approved):
Nice!
Thanks for the effort!

startup/README.md
d1donlydfink: List taken out of PR. Thanks.

startup/README.md
d1donlydfink: Thanks.

PR #739, opened 2026-02-18
Remove hvac from requirements.txt

d1donlydfink:


requirements.txt
```diff
@@ -6,7 +6,6 @@
 # requirements-build.txt
 
 # In-house dependencies
-# note: ticket issue #705 and switch back to latest leaf-common once we have a new release with the latest fixes and features #706
```
d1donlydfink, line 9: This is no longer the case. So remove comment.

requirements.txt
```diff
@@ -29,7 +28,6 @@ aiofiles>=25.1.0
 
 # These are transient dependencies needed by leaf-common
 ruamel.yaml>=0.18.6
-hvac>=1.1.0
```
d1donlydfink, line 32: hvac (Hashicorp Vault lib) is now a lazily loaded dependency of leaf-common.
neuro-san does not use it.

andreidenissov-cog (approved):
LGTM.

PR #742, opened 2026-02-20
NS-661 Refactor usage of temporary agent networks.

andreidenissov-cog:
This PR refactors how we manage temporary agent networks generated by neuro-san.
Regular storage updating logic used for other kinds of agent networks is not scalable on multi-nodes deployment level, especially when temp networks are generated en-masse.
Instead, we use some persistent storage (S3) as base and per-service local storage tables as in-memory cache mirroring this base.
Note that expiration logic (periodic polling and removing expired instances) is running independently on both levels: local cache and base storage. We never return an expired network to client query.


neuro_san/interfaces/reservation.py
```diff
@@ -51,6 +53,12 @@ def get_expiration_time_in_seconds(self) -> float:
         """
         return self.expiration_time_in_seconds
 
+    def is_expired(self) -> bool:
+        """
+        :return: Whether the reservation is expired or not.
+        """
+        return time.time() >= self.expiration_time_in_seconds
+
```
andreidenissov-cog: Convenience function for simpler checks.

neuro_san/internals/network_providers/fixed_agent_network_provider.py
```diff
@@ -0,0 +1,39 @@
...
+# END COPYRIGHT
+
+from typing import Dict
+
+from neuro_san.internals.graph.registry.agent_network import AgentNetwork
+from neuro_san.internals.interfaces.agent_network_provider import AgentNetworkProvider
+
+
+class FixedAgentNetworkProvider(AgentNetworkProvider):
+    """
```
andreidenissov-cog, line 23: AgentNetworkProvider which returns immutable instance of temp network.
We don't want extra redirection here which would allow dynamic replacement of agent network version.

neuro_san/service/utils/server_context.py
```diff
@@ -48,7 +49,7 @@ def __init__(self):
         self.network_storage_dict: Dict[str, AgentNetworkStorage] = {
             "protected": AgentNetworkStorage(),
             "public": AgentNetworkStorage(),
-            "temp": ExpiringAgentNetworkStorage()
+            "temp": ExpiringCachingAgentNetworkStorage()
         }
```
andreidenissov-cog, line 53: Type for our temp networks storage class now changes.

neuro_san/service/watcher/main_loop/storage_watcher.py
```diff
@@ -55,7 +55,7 @@ def __init__(self,
 
         self.storage_updaters: List[StorageUpdater] = [
             RegistryStorageUpdater(server_context.get_network_storage_dict(), watcher_config),
-            TempNetworkStorageUpdater(server_context.get_network_storage_dict(), watcher_config,
+            TempNetworkStorageUpdater(server_context.get_network_storage_dict(),
                                       server_context.get_queues())
         ]
```
andreidenissov-cog, line 57: No "watcher_config" in parameters list - our TempNetworkStorageUpdater is not really a "watcher".

neuro_san/internals/interfaces/expiring_reservations_storage.py
```diff
@@ -0,0 +1,67 @@
...
+from neuro_san.interfaces.reservation import Reservation
+
+
+class ExpiringReservationsStorage:
+    """
+    An interface for implementations of basic Reservations storage,
+    supporting addition Reservations in bulk and retrieval of individual Reservations,
+    as well as expiration of Reservations based on their lifetime.
+    """
+
```
andreidenissov-cog: New interface for classes storing Reservations, with methods to add and extract reservations,
set them to expired state AND set another ExpiringReservationsStorage instance
to serve as a persistent backup, so our instance will work as a front cache for Reservations requests.

neuro_san/service/main_loop/server_main_loop.py
```diff
@@ -256,6 +257,10 @@ def main_loop(self):
                           'AGENT_SERVICE_LOG_LEVEL')
             watcher = StorageWatcher(self.watcher_config, self.server_context)
             components_to_start.append(watcher)
+        # Another component to start is the temporary networks updater
+        temp_networks_updater: TempNetworkStorageUpdater =\
+            TempNetworkStorageUpdater(self.server_context.get_network_storage_dict(), self.server_context.get_queues())
+        components_to_start.append(temp_networks_updater)
```
andreidenissov-cog, line 262: TempNetworkStorageUpdater is not a "watcher" anymore, so create it separately
and add to the list of components to be started after our HTTP server starts running.

neuro_san/service/watcher/temp_networks/s3_reservations_storage.py
```diff
@@ -0,0 +1,298 @@
...
+
+    def set_base_storage(self, base_storage: ExpiringReservationsStorage):
+        """
+        Set a "base" storage to use as a source of truth for reservations.
+        This is optional, but if set, will be used as the source of truth for reservations
+        and the implementing class will act as a cache in front of it.
+        :param base_storage: An ExpiringReservationsStorage instance to use as a source of truth
+        """
+        raise NotImplementedError
+
```
andreidenissov-cog, line 224: For S3 based reservations storage, we don't have a backup place.

neuro_san/service/watcher/temp_networks/s3_reservations_storage.py
```diff
@@ -0,0 +1,298 @@
...
+class S3ExpiringReservationsStorage(AbstractExpiringReservationsStorage):
+    """
+    AWS S3-based implementation of ExpiringReservationsStorage.
+
+    Stores reservations as JSON objects in an S3 bucket, with each reservation
+    stored in its associated agent spec as metadata.
+    """
+
+    def __init__(self, bucket_name: str = "", prefix: str = "reservations/",
```
andreidenissov-cog, line 55: Implementation especially for S3 operations is mostly copied from S3ReservationsStorage class.
Difference is that we are running our own thread (owned by AbstractExpiringReservationsStorage)
which actively polls for expired Reservation objects.

neuro_san/service/watcher/temp_networks/s3_reservations_storage.py
```diff
@@ -0,0 +1,298 @@
...
+        except ClientError as exception:
+            # Handle various S3 access errors with specific messages
+            error_code: str = exception.response["Error"]["Code"]
+            if error_code == "404":
+                raise ValueError(f"S3 bucket '{self.bucket_name}' does not exist") from exception
+            if error_code == "403":
+                raise ValueError(f"Access denied to S3 bucket '{self.bucket_name}'") from exception
+            raise ValueError(f"Error accessing S3 bucket '{self.bucket_name}': {exception}") from exception
+        # We are good with S3 connection and bucket access at this point,
+        # let's start underlying logic:
```
andreidenissov-cog, line 113: Start AbstractExpiringReservationsStorage thread.

neuro_san/service/watcher/temp_networks/temp_network_storage_updater.py
```diff
@@ -31,107 +31,89 @@
...
 
 
-class TempNetworkStorageUpdater(AbstractStorageUpdater):
+class TempNetworkStorageUpdater(Startable):
     """
-    StorageUpdater implementation for temporary network updates.
+    StorageUpdater implementation for temporary network being generated by neuro-san service.
     """
+    # Size of thread pool for processing reservations queues.
+    THREAD_POOL_SIZE: int = 32
```
andreidenissov-cog, line 48: Diff here is rather confusing,
but what is being done is:
TempNetworkStorageUpdater responsibility now is to take queues of reservations arriving on "main" queue, and process them by AbstractAgentReservationist class member.
Processing is done by assigning a separate thread from local ThreadPool for reservations queue processing.
Temp networks storage is queried from network storage dictionary passed in as a constructor parameter.

neuro_san/service/watcher/temp_networks/temp_network_storage_updater.py
```diff
@@ -31,107 +31,89 @@
...
-            # Potentially create an external storage class
+            # Potentially create a "base" reservations storage class
             storage_class_name: str = environ.get("AGENT_EXTERNAL_RESERVATIONS_STORAGE", "")
-            external_storage: ReservationsStorage = ResolverUtil.create_instance(
+            external_storage: ExpiringReservationsStorage = ResolverUtil.create_instance(
                     storage_class_name,
                     "AGENT_EXTERNAL_RESERVATIONS_STORAGE env var",
-                    ReservationsStorage)
+                    ExpiringReservationsStorage)
             if external_storage is not None:
```
andreidenissov-cog, line 77: Here we create a "backup" storage if it is specified.
With success, link it to our "temp networks" storage we are processing.

neuro_san/service/watcher/temp_networks/temp_network_storage_updater.py
```diff
@@ -31,107 +31,89 @@
...
+        while True:
             async_collating_queue: AsyncCollatingQueue = self.incoming.sync_q.get()
-            self.queue_pool.add(async_collating_queue)
+            # We have a new queue to process.
+            # Submit this queue processing  task to our thread pool,
+            # IF we have a storage to put the results in.
+            # Since each queue is processed by its own thread,
+            # we avoid potential locking out of different queues received from "main" self.incoming
+            if self.reservationist is not None:
+                self.queue_processing_pool.submit(self.process_one_queue, async_collating_queue)
```
andreidenissov-cog, line 134: Submit the task to process individual reservations queue to our thread pool.
Note that this is done by a task already running in the same thread pool - "incoming queue task".

neuro_san/service/watcher/temp_networks/temp_network_storage_updater.py
```diff
@@ -142,14 +124,12 @@ def process_one_queue(self, async_collating_queue: AsyncCollatingQueue):
         janus_queue: Queue = async_collating_queue.get_queue()
 
         # See what has come over this particular queue
-        while janus_queue.sync_q.qsize() > 0:
-
```
andreidenissov-cog, line 146: That was a potential subtle timing problem here in a previous code version:
what if we submit several items to the queue, and then delay adding the rest of items to it?

neuro_san/internals/network_providers/expiring_caching_agent_network_storage.py
```diff
@@ -0,0 +1,190 @@
...
+    import AbstractExpiringReservationsStorage
+from neuro_san.internals.network_providers.fixed_agent_network_provider import FixedAgentNetworkProvider
+
+
+class ExpiringCachingAgentNetworkStorage(AbstractExpiringReservationsStorage, AgentNetworkStorage):
+    """
+    An AgentNetworkStorage instance where AgentNetworks are allowed to expire.
+    This implementation also allows for an optional "source" storage to be set,
+    which will be used as a "base" source of truth.
+    """
```
andreidenissov-cog: This is the class to contain our temporary agent networks (AgentNetworkStorage)
and manage lifetime of those networks reservations (AbstractExpiringReservationsStorage).

neuro_san/internals/network_providers/expiring_caching_agent_network_storage.py
```diff
@@ -0,0 +1,190 @@
...
+
+    def add_reservations(self, reservations_dict: Dict[Reservation, Dict[str, Any]],
+                         source: str = None):
+        """
+        Add a set of reservations for agent networks en-masse
+
+        :param reservations_dict: A mapping of Reservation -> agent network spec
+        :param source: A string describing where the deployment was coming from
+        """
+        if not reservations_dict:
```
andreidenissov-cog: Implementation is mostly reused from ExpiringAgentNetworkStorage class,
but if self.base_storage is set, we add through incoming reservations to this underlying storage first.

neuro_san/internals/network_providers/expiring_caching_agent_network_storage.py
```diff
@@ -0,0 +1,190 @@
...
+                 (None, None) otherwise
+        """
+        # For this class, this method is never used.
+        return None, None
+
+    def expire_reservations(self):
+        """
+        Remove Reservations that are expired
+        """
+
```
andreidenissov-cog: Again, reused from ExpiringAgentNetworkStorage class, but now this method is executed by internal thread in AbstractExpiringReservationsStorage.

neuro_san/internals/network_providers/expiring_caching_agent_network_storage.py
```diff
@@ -0,0 +1,190 @@
...
+            for agent_name in expired:
+                listener.agent_removed(agent_name, self)
+                self.logger.info("REMOVED network for agent %s", agent_name)
+
+    def get_agent_network_provider(self, agent_name: str) -> AgentNetworkProvider:
+        """
+        Get AgentNetworkProvider for a specific agent
+        :param agent_name: name of an agent
+        """
+        # Agents and their Reservations are always present or absent together.
```
andreidenissov-cog: This is where the "cache + persistent storage" is implemented:
check in-memory storage first, and if agent network is present,
check if it's already expired.

neuro_san/internals/network_providers/expiring_caching_agent_network_storage.py
```diff
@@ -0,0 +1,190 @@
...
+                    listener.agent_removed(agent_name, self)
+                    self.logger.info("REMOVED expired network for agent %s", agent_name)
+
+                # Now if we have a source storage,
+                # we rely on it to expire this reservation according to its own schedule.
+
+                # Our agent network no longer exists.
+                return None
+            # Reservation is still valid, so we can return the AgentNetworkProvider for this agent.
+            agent_network = self.agents_table[agent_name]
```
andreidenissov-cog: If agent network is good, wrap it into AgentNetworkProvider of special kind:
ensuring that agent network cannot be overwritten by modifications of storage table.

neuro_san/internals/network_providers/expiring_caching_agent_network_storage.py
```diff
@@ -0,0 +1,190 @@
...
+            return FixedAgentNetworkProvider(agent_network)
+        # We don't have a reservation for this agent,
+        # so check the source storage if we have one:
+        if self.base_storage is not None:
+            reservation, agent_network = self.base_storage.get_one_reservation(agent_name)
+            if reservation is not None and not reservation.is_expired():
```
andreidenissov-cog: If no luck with local storage, go for persistent one.
Note that we always check for expiration right before we return result to the method caller.

neuro_san/internals/network_providers/abstract_expiring_reservations_storage.py
```diff
@@ -0,0 +1,117 @@
...
+from typing import Optional
+from typing import Tuple
+
+import logging
+import time
+import threading
+
+from neuro_san.interfaces.reservation import Reservation
+from neuro_san.internals.interfaces.expiring_reservations_storage import ExpiringReservationsStorage
+from neuro_san.service.interfaces.startable import Startable
```
d1donlydfink: Can't have service imports in internals.
That's a setup for tangles and circular dependencies.
service can depend on internals but not the other way around.

d1donlydfink: FWIW: That is probably the reason why it feels weird to have threading stuff in internals.
Normally, it's the service that is controlling the threads.

neuro_san/internals/network_providers/expiring_caching_agent_network_storage.py
```diff
@@ -0,0 +1,190 @@
...
+        """
+        Extract a single reservation.
+
+        :param obj_key: unique key for the reservation
+        :return: Tuple of (reservation, agent data) if successful
+                 and reservation is not expired,
+                 (None, None) otherwise
+        """
+        # For this class, this method is never used.
+        return None, None
```
d1donlydfink: Maybe allow the NotImplementedError to enforce this?

neuro_san/internals/network_providers/expiring_caching_agent_network_storage.py
```diff
@@ -0,0 +1,190 @@
...
+        Get AgentNetworkProvider for a specific agent
+        :param agent_name: name of an agent
+        """
+        # Agents and their Reservations are always present or absent together.
+        reservation: Reservation = None
+        agent_network: AgentNetwork = None
+        if agent_name in self.reservations_table:
+            # We have a reservation for this agent, but we need to check is it still valid:
+            reservation = self.reservations_table[agent_name]
+            if reservation.is_expired():
```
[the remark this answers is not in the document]
d1donlydfink: Copilot is catching some good issues here.
If you did self.reservations_table.get(agent_name) and checked for None instead of what you've got here, you might be able to avoid the lock it suggests.

This would alleviate my concern of never using [] to access dictionary values because of the possibility of KeyErrors

neuro_san/internals/network_providers/expiring_caching_agent_network_storage.py
d1donlydfink: Another place where dictionary.get() would be better than [] access to avoid KeyError possibility.

neuro_san/internals/network_providers/expiring_caching_agent_network_storage.py
```diff
@@ -0,0 +1,190 @@
...
+            # Reservation is still valid, so we can return the AgentNetworkProvider for this agent.
+            agent_network = self.agents_table[agent_name]
+            return FixedAgentNetworkProvider(agent_network)
+        # We don't have a reservation for this agent,
+        # so check the source storage if we have one:
+        if self.base_storage is not None:
+            reservation, agent_network = self.base_storage.get_one_reservation(agent_name)
+            if reservation is not None and not reservation.is_expired():
+                # We have a valid reservation for this agent in the source storage,
+                # so return the AgentNetworkProvider for this agent.
```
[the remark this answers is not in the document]
d1donlydfink: Not bad Copilot.

neuro_san/internals/network_providers/fixed_agent_network_provider.py
d1donlydfink: This might not be overkill for networks that are getting updated on the filesystem, but it would be overkill for temp networks which are immutable given their network name.

Said another way: If temp networks are always immutable, why the special treatment?

neuro_san/service/utils/server_context.py
d1donlydfink: Does this mean that ExpiringAgentNetworkStorage should go away?

neuro_san/service/watcher/temp_networks/s3_expiring_reservations_storage.py
```diff
@@ -0,0 +1,298 @@
...
+from botocore.exceptions import NoCredentialsError
+
+from neuro_san.interfaces.reservation import Reservation
+from neuro_san.internals.interfaces.expiring_reservations_storage import ExpiringReservationsStorage
+from neuro_san.internals.network_providers.abstract_expiring_reservations_storage \
+    import AbstractExpiringReservationsStorage
+from neuro_san.internals.reservations.reservation_dictionary_converter import ReservationDictionaryConverter
+
+
+class S3ExpiringReservationsStorage(AbstractExpiringReservationsStorage):
```
d1donlydfink: From the Dockerfile:

```
# A reference to a Python class that can be used to store temporary agent references
# External to the server, so other replicated pods can pick them up.
# This class must have a no-args constructor and implement the
# neuro_san.internals.interfaces.ReservationsStorage interface.
# When not set, this defaults to a null implementation.
# A simple default implementation is provided by the class
# "neuro_san.service.watcher.temp_networks.s3_reservations_storage.S3ReservationsStorage"
ENV AGENT_EXTERNAL_RESERVATIONS_STORAGE=""
```
Where is the hook into the ReservationsStorage interface with this change?
How does this S3ExpiringReservationsStorage actually get instantiated optionally?

neuro_san/internals/network_providers/expiring_caching_agent_network_storage.py
d1donlydfink: Does this guy really need to be a Startable?
Who starts/stops him?


Moving the Startable-ness to S3ExpiringReservationsStorage might solve a lot of problems.

neuro_san/service/watcher/temp_networks/temp_network_storage_updater.py
d1donlydfink: Seems like we need to resolve the differences between the ReservationsStorage interface and the ExpiringReservationsStorage interface.

Also, if you are requiring the Expiring type now the Dockerfile will need to be changed.

neuro_san/service/watcher/main_loop/storage_watcher.py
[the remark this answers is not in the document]
andreidenissov-cog: This is done already.

neuro_san/internals/network_providers/expiring_caching_agent_network_storage.py
andreidenissov-cog: Done.

neuro_san/internals/network_providers/expiring_caching_agent_network_storage.py
andreidenissov-cog: Yes it does, because it owns its agent networks expiring thread.
And our policy is to separate constructor from actual starting of a component.
Who starts it - HTTP server does, it goes to the list of things to start by the server
after possible process forking. The same way as for other storage updaters.

neuro_san/internals/network_providers/abstract_expiring_reservations_storage.py
andreidenissov-cog: All our internals are built on threads (AsyncioExecutors) and after we create them, there is not that much control we have, except trying to cleanly shut them down.
So in neuro-san active components having their own activity are all around the place.
But you are right: in this sense Startable interface is misplaced being very core-level for the system.
So I moved it to internals/interfaces.

neuro_san/service/watcher/temp_networks/s3_expiring_reservations_storage.py
```diff
@@ -0,0 +1,298 @@
...
+
+        return expired
+
+    def expire_reservations(self):
+        """
+        Remove expired reservations from S3 storage.
+        """
+        self.logger.debug("Starting expiration process for S3 reservations")
+        print("Starting expiration process for S3 reservations")
+
```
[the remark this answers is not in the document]
andreidenissov-cog: Removed debug print.

neuro_san/service/watcher/temp_networks/s3_expiring_reservations_storage.py
```diff
@@ -0,0 +1,298 @@
...
+        for obj in response["Contents"]:
+            # Attempt to expire this reservation and increment counter if successful
+            if self.expire_one_reservation(obj["Key"], current_time):
+                expired_count += 1
+
+        if expired_count > 0:
+            self.logger.info("Expiration complete: removed %d expired reservations from S3", expired_count)
+            print(f"Expiration complete: removed {expired_count} expired reservations from S3")
+        else:
+            self.logger.debug("Expiration complete: removed no expired reservations from S3")
```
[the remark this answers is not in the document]
andreidenissov-cog: Removed debug prints.

neuro_san/internals/network_providers/expiring_caching_agent_network_storage.py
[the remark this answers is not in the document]
andreidenissov-cog: Fixed it. Good catch from Copilot.

neuro_san/internals/network_providers/abstract_reservations_storage.py
```diff
@@ -0,0 +1,117 @@
...
+from typing import Tuple
+
+import logging
+import time
+import threading
+
+from neuro_san.interfaces.reservation import Reservation
+from neuro_san.internals.interfaces.reservations_storage import ReservationsStorage
+from neuro_san.internals.interfaces.startable import Startable
+
```
andreidenissov-cog, line 30: Abstract class for implementing expiration management with background thread.
Renamed from AbstractExpiringReservationsStorage module.

neuro_san/service/http/server/http_server.py
```diff
@@ -48,7 +48,7 @@
 from neuro_san.service.interfaces.agent_authorizer import AgentAuthorizer
 from neuro_san.service.interfaces.agent_server import AgentServer
 from neuro_san.service.interfaces.event_loop_logger import EventLoopLogger
-from neuro_san.service.interfaces.startable import Startable
+from neuro_san.internals.interfaces.startable import Startable
 from neuro_san.service.mcp.handlers.mcp_root_handler import McpRootHandler
```
andreidenissov-cog, line 52: Here and below: Startable interface moved to core internals/interfaces.

neuro_san/service/watcher/temp_networks/s3_reservations_storage.py
```diff
@@ -99,16 +107,72 @@ def start(self):
             if error_code == "403":
                 raise ValueError(f"Access denied to S3 bucket '{self.bucket_name}'") from exception
             raise ValueError(f"Error accessing S3 bucket '{self.bucket_name}': {exception}") from exception
+        # We are good with S3 connection and bucket access at this point,
+        # let's start underlying logic:
+        super().start()
 
-    def set_sync_target(self, sync_target: ReservationsStorage):
+    def _is_retryable_client_error(self, err: ClientError) -> bool:
+        """
```
andreidenissov-cog, line 117: Bad diff, but what we do here is wrap our S3 operations with retries+exponential back-off logic.

neuro_san/service/watcher/temp_networks/s3_reservations_storage.py
```diff
@@ -148,7 +212,30 @@ def add_reservations(self, reservations_dict: Dict[Reservation, Any],
...
+        raise NotImplementedError
+
+    def _retrieve_object_with_retries(self, obj_key: str) -> Dict[str, Any]:
+        """
+        Helper method to retrieve an S3 object with retries.
+        :param obj_key: S3 object key to retrieve
+        :return: The parsed JSON content of the S3 object as a dictionary
+        :raises: ClientError if the object cannot be retrieved after retries
+                 JSONDecodeError if the content cannot be parsed as JSON
+        """
```
andreidenissov-cog, line 232: Helper function to get our Reservation data dictionary from S3 bucket with retries and all.
Will use it for expiration check and for item retrieval on "get_one_reservation" call.

neuro_san/service/watcher/temp_networks/s3_reservations_storage.py
```diff
@@ -157,22 +244,18 @@ def sync_one_reservation(self, obj_key: str) -> Tuple[Reservation, Any]:
...
-            agent_spec: Dict[str, Any] = loads(json_content)
+            agent_spec: Dict[str, Any] = self._retrieve_object_with_retries(obj_key)
             metadata: Dict[str, Any] = agent_spec.get("metadata")
 
             # Reconstruct the Reservation object from stored dictionary
             reservation_dict: Dict[str, Any] = metadata.get("reservation")
             reservation = self.converter.from_dict(reservation_dict)
+            # Reconstruct the AgentNetwork object using the agent spec dictionary
+            # and reservation ID - which is our agent name in this design
+            agent_network: AgentNetwork = AgentNetwork(agent_spec, reservation.get_reservation_id())
```
andreidenissov-cog, line 262: That fixes Copilot mentioned issue: we always return AgentNetwork instance.

neuro_san/service/watcher/temp_networks/temp_network_storage_updater.py
andreidenissov-cog: ExpiringReservationsStorage interface is renamed to be ReservationsStorage.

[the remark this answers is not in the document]
andreidenissov-cog: Done.

neuro_san/session/external_agent_session_factory.py
```diff
@@ -92,7 +109,8 @@ def create_session_from_location_dict(self, agent_location: Dict[str, str],
             # and potentially relieve the direct user of the burden of having to start a server
 
             agent_network_provider: AgentNetworkProvider = None
-            for network_storage in self.network_storage_dict.values():
+            for network_storage_name in self.get_networks_order(self.network_storage_dict):
+                network_storage = self.network_storage_dict.get(network_storage_name)
                 # Be sure we have something
```
andreidenissov-cog, line 114: That fixes weird unit tests failure, where we were starting with checking "temp" storage first,
and going off into long wait for S3 storage with subsequent test timeout.

neuro_san/service/main_loop/server_main_loop.py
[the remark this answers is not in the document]
andreidenissov-cog: It only created here now.

neuro_san/internals/network_providers/fixed_agent_network_provider.py
andreidenissov-cog: Temp networks are indeed immutable, until they expire.
So my thinking here: after we got some temp network for request execution, do we want it
to magically disappear while still processing request?
If not, we need to hold on to it. If we are OK with this chance, we can use AgentNetworkProvider as before.

neuro_san/service/utils/server_context.py
andreidenissov-cog: I've renamed this class back to ExpiringAgentNetworkStorage.

neuro_san/service/watcher/temp_networks/temp_network_storage_updater.py
```diff
@@ -31,107 +31,89 @@
...
+        while True:
             async_collating_queue: AsyncCollatingQueue = self.incoming.sync_q.get()
-            self.queue_pool.add(async_collating_queue)
+            # We have a new queue to process.
+            # Submit this queue processing  task to our thread pool,
+            # IF we have a storage to put the results in.
+            # Since each queue is processed by its own thread,
+            # we avoid potential locking out of different queues received from "main" self.incoming
+            if self.reservationist is not None:
+                self.queue_processing_pool.submit(self.process_one_queue, async_collating_queue)
```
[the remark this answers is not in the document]
andreidenissov-cog: OK this one I've put in.

neuro_san/service/watcher/temp_networks/temp_network_storage_updater.py
```diff
@@ -31,107 +31,89 @@
...
+        for independent processing of each queue.
         """
-        while self.incoming.sync_q.qsize() > 0:
+        self.logger.info("Waiting for new queues to be added for processing")
+        while True:
             async_collating_queue: AsyncCollatingQueue = self.incoming.sync_q.get()
-            self.queue_pool.add(async_collating_queue)
+            # We have a new queue to process.
+            # Submit this queue processing  task to our thread pool,
+            # IF we have a storage to put the results in.
```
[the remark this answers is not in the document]
andreidenissov-cog: This "incoming" queue lifetime is lifetime of our service,
so I guess it's OK to keep it around indefinitely.
At least in the current setup.

neuro_san/service/watcher/temp_networks/temp_network_storage_updater.py
```diff
@@ -142,14 +124,12 @@ def process_one_queue(self, async_collating_queue: AsyncCollatingQueue):
         janus_queue: Queue = async_collating_queue.get_queue()
 
         # See what has come over this particular queue
-        while janus_queue.sync_q.qsize() > 0:
-
+        while True:
             # Get an item off the queue
             queued_item: Dict[str, Any] = janus_queue.sync_q.get()
             if async_collating_queue.is_final_item(queued_item):
                 # We have exhausted this queue.
```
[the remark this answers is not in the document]
andreidenissov-cog: Hmm. We don't have any expectations about what and how pushes stuff into that queue,
no expected timeouts and such.
What can we do here except wait for explicit "final item"?

neuro_san/interfaces/reservation.py
```diff
@@ -51,6 +53,12 @@ def get_expiration_time_in_seconds(self) -> float:
         """
         return self.expiration_time_in_seconds
 
+    def is_expired(self) -> bool:
+        """
+        :return: Whether the reservation is expired or not.
+        """
+        return time.time() >= self.expiration_time_in_seconds
```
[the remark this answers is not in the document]
d1donlydfink: Reasonable request from Copilot

neuro_san/internals/interfaces/reservations_storage.py
```diff
@@ -54,3 +56,12 @@ def expire_reservations(self):
         Remove Reservations that are expired
         """
         raise NotImplementedError
+
+    def set_base_storage(self, base_storage: ReservationsStorage):
+        """
+        Set a "base" storage to use as a source of truth for reservations.
+        This is optional, but if set, will be used as the source of truth for reservations
+        and the implementing class will act as a cache in front of it.
+        :param base_storage: An ExpiringReservationsStorage instance to use as a source of truth
```
d1donlydfink: Seems like comment for param needs to be updated.
All that is required is another ReservationsStorage instance.

neuro_san/internals/network_providers/abstract_reservations_storage.py
```diff
@@ -0,0 +1,117 @@
...
+    def _run_loop(self):
+        while not self._stop_event.is_set():
+            # Using "monotonic" time allows us to avoid potential issues with system clock changes
+            start = time.monotonic()
+            try:
+                self.expire_reservations()
+            except Exception as exception:  # pylint: disable=broad-except
+                self._logger.info("Expiration cleanup failed: %s", exception)
+            elapsed = time.monotonic() - start
+            self._logger.debug("Expiration cleanup took %f seconds.", elapsed)
```
d1donlydfink: Do we want a name passed in for the instance so these logger messages can be more precise as to just what is being cleaned up, etc?

neuro_san/internals/network_providers/abstract_reservations_storage.py
```diff
@@ -0,0 +1,117 @@
...
+        self._logger = logging.getLogger(self.__class__.__name__)
+
+    def stop(self, timeout: Optional[float] = None):
+        """
+        Signal the worker to stop and wait for it with timeout.
+        """
+        self._stop_event.set()
+        if self._thread:
+            self._thread.join(timeout)
+            self._logger.debug("Expiration cleanup thread stopped.")
```
d1donlydfink: Not fond of thread business in internals, but work-in-progress.
Maybe there is a way to wrap one of these guys in a Startable and then Startable with its thread-y implications can move back to neuro_san.service.

Someday.

neuro_san/internals/network_providers/expiring_agent_network_storage.py
```diff
@@ -18,33 +18,53 @@
...
-        # We don't need one of these
+        self.base_storage = base_storage
+
+    def start(self):
+        """
+        Start this storage, which includes starting the expiration checking loop.
+        """
+        super().start()
+        if self.base_storage is not None:
+            self.base_storage.start()
```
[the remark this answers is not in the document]
d1donlydfink: Seems somewhat reasonable, except for the blanket Exception catch.
Seems like we'd want to be more specific if possible.

neuro_san/internals/network_providers/expiring_agent_network_storage.py
```diff
@@ -18,33 +18,53 @@
...
```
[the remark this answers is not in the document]
d1donlydfink: This doesn't sound quite correct to me.
@andreidenissov-cog any thoughts?

neuro_san/internals/network_providers/abstract_reservations_storage.py
```diff
@@ -0,0 +1,117 @@
...
+
+    def start(self):
+        self._thread = threading.Thread(target=self._run_loop, daemon=True)
+        self._thread.start()
+        self._logger.debug("Expiration cleanup thread started.")
```
d1donlydfink: Should we check for env var turning this stuff on before going along with the rest of start()?
Maybe not in this Abstract guy but other place(s) that are more concrete?
No need to have threads we don't need.
Or maybe this is done elsewhere already?

d1donlydfink: FWIW: This would read better if start() were before stop() in the file.

neuro_san/internals/network_providers/expiring_agent_network_storage.py
```diff
@@ -118,11 +149,54 @@ def expire_reservations(self):
...
+                # Now if we have a source storage,
+                # we rely on it to expire this reservation according to its own schedule.
+
+                # Our agent network no longer exists.
+                return None
+            # Reservation is still valid, so we can return the AgentNetworkProvider for this agent.
+            agent_network: AgentNetwork = self.agents_table.get(agent_name, None)
+            if agent_network is not None:
+                return FixedAgentNetworkProvider(agent_network)
+            return None
```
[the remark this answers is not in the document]
d1donlydfink: I don't think this is a concern given that the names are all globally unique and only one thread can do the adding and a different one doing the adding.
Thoughts @andreidenissov-cog ?

neuro_san/internals/network_providers/expiring_agent_network_storage.py
```diff
@@ -118,11 +149,54 @@ def expire_reservations(self):
...
+        # We don't have a reservation for this agent,
+        # so check the source storage if we have one:
+        if self.base_storage is not None:
+            reservation, agent_network = self.base_storage.get_one_reservation(agent_name)
+            if reservation is not None and not reservation.is_expired():
+                # We have a valid reservation for this agent in the source storage,
+                # so return the AgentNetworkProvider for this agent.
+                # Also cache this reservation and agent network locally:
+                with self.lock:
```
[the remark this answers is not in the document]
d1donlydfink: Similar non-concern - only 1 thread is going to be doing the updating here.
@andreidenissov-cog : Perhaps it's worth commenting where Copilot is coming up with these threading concerns?

neuro_san/internals/network_providers/fixed_agent_network_provider.py
```diff
@@ -0,0 +1,37 @@
...
+# END COPYRIGHT
+
+from neuro_san.internals.graph.registry.agent_network import AgentNetwork
+from neuro_san.internals.interfaces.agent_network_provider import AgentNetworkProvider
+
+
+class FixedAgentNetworkProvider(AgentNetworkProvider):
+    """
+    Class providing fixed immutable AgentNetwork for a given agent in the service scope.
+    """
```
d1donlydfink, line 25: How would this guy be used? Why is he necessary?
I ask because I don't see him used anywhere.

neuro_san/service/watcher/main_loop/storage_watcher.py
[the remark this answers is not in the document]
d1donlydfink: where?
Maybe leave a breadcrumb comment as to where this happens nowadays for temp networks.

neuro_san/service/watcher/temp_networks/s3_reservations_storage.py
```diff
@@ -148,7 +212,30 @@ def add_reservations(self, reservations_dict: Dict[Reservation, Any],
...
 
-    def sync_one_reservation(self, obj_key: str) -> Tuple[Reservation, Any]:
+    def set_base_storage(self, base_storage: ReservationsStorage):
+        """
+        Set a "base" storage to use as a source of truth for reservations.
+        This is optional, but if set, will be used as the source of truth for reservations
+        and the implementing class will act as a cache in front of it.
+        :param base_storage: An ReservationsStorage instance to use as a source of truth
+        """
+        raise NotImplementedError
```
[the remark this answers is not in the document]
d1donlydfink: Not a bad recommendation

neuro_san/service/watcher/temp_networks/s3_reservations_storage.py
```diff
@@ -324,39 +323,42 @@ def expire_one_reservation(self, obj_key: str, current_time: float) -> bool:
...
-        # List all reservation objects in S3 bucket with our prefix
-        response: Dict[str, Any] = self.s3_client.list_objects_v2(
-            Bucket=self.bucket_name,
-            Prefix=self.prefix
-        )
-
-        # Handle case where no reservations exist in S3
-        if "Contents" not in response:
-            self.logger.debug("No reservations found in S3 bucket for expiration")
-            return
```
d1donlydfink, line 348: This seemed reasonable logging. Why remove it?

neuro_san/service/watcher/temp_networks/s3_reservations_storage.py
```diff
@@ -192,82 +275,7 @@ def sync_one_reservation(self, obj_key: str) -> Tuple[Reservation, Any]:
             self.logger.error("JSON error processing reservation object %s during sync: %s",
                               obj_key, str(exception))
 
-        return reservation, agent_spec
-
-    def sync_reservations(self):
-        """
-        Sync reservations from S3 to the sync target (if set).
-        Only syncs objects modified after the last sync timestamp for efficiency.
-        """
```
d1donlydfink, line 201: FWIW: This class was largely vibe-coded for its S3 acrobatics. No love lost

neuro_san/service/watcher/temp_networks/s3_reservations_storage.py
```diff
@@ -324,39 +323,42 @@ def expire_one_reservation(self, obj_key: str, current_time: float) -> bool:
...
+        paginator = self.s3_client.get_paginator("list_objects_v2")
+        paginate_function = partial(paginator.paginate,
+                                    Bucket=self.bucket_name,
+                                    Prefix=self.prefix,
+                                    PaginationConfig={"PageSize": self.max_keys_per_page})
+        for page in self._do_with_retries(paginate_function):
+            # _do_with_retries here wraps the generator creation
+            contents = page.get("Contents", [])
+            for obj in contents:
+                yield obj["Key"]
```
d1donlydfink, line 350: Never use [] to access a dictionary. Always use get()

neuro_san/service/watcher/temp_networks/temp_network_storage_updater.py
d1donlydfink: Or is it?

neuro_san/service/watcher/temp_networks/temp_network_storage_updater.py
```diff
@@ -35,103 +35,90 @@
...
                     "AGENT_EXTERNAL_RESERVATIONS_STORAGE env var",
                     ReservationsStorage)
             if external_storage is not None:
-                self.reservations_storage.add(external_storage)
-
-            # The ultimate target for sync is the temp storage
-            for storage in self.reservations_storage:
-                storage.set_sync_target(temp_storage)
+                self.temp_storage.set_base_storage(external_storage)
+            self.reservationist = AbstractAgentReservationist({self.temp_storage})
```
d1donlydfink, line 79: Why do we need the dictionary in initializing this guy? - It's not even a proper dictionary.
We didn't before. (see below)

neuro_san/service/watcher/temp_networks/temp_network_storage_updater.py
```diff
@@ -35,103 +35,90 @@
...
 
 
-class TempNetworkStorageUpdater(AbstractStorageUpdater):
+class TempNetworkStorageUpdater(Startable):
     """
-    StorageUpdater implementation for temporary network updates.
+    StorageUpdater implementation for temporary network being generated by neuro-san service.
     """
+    # Size of thread pool for processing reservations queues.
+    THREAD_POOL_SIZE: int = 32
```
d1donlydfink, line 47: Do we really need this many?

neuro_san/service/watcher/temp_networks/temp_network_storage_updater.py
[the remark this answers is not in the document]
d1donlydfink: Nothing much.
The final item I believe is placed on any one inner queue when Reservationist.close() is called in DataDrivenChatSession.

~~But this is the outer queue and it exists as long as the server does I think.~~

Looking at this again, I think this queue is an inner queue from one request.
Yes?

neuro_san/service/watcher/temp_networks/temp_network_storage_updater.py
```diff
@@ -142,14 +124,12 @@ def process_one_queue(self, async_collating_queue: AsyncCollatingQueue):
...
+        while True:
             # Get an item off the queue
             queued_item: Dict[str, Any] = janus_queue.sync_q.get()
             if async_collating_queue.is_final_item(queued_item):
                 # We have exhausted this queue.
-                # No one needs to worry about it any more.
-                self.queue_pool.remove(async_collating_queue)
+                # No one needs to worry about it anymore.
                 async_collating_queue.close()
                 return
```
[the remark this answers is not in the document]
d1donlydfink: Nah.

neuro_san/service/watcher/temp_networks/s3_reservations_storage.py
```diff
@@ -30,30 +33,36 @@
...
+from neuro_san.internals.graph.registry.agent_network import AgentNetwork
 from neuro_san.internals.interfaces.reservations_storage import ReservationsStorage
+from neuro_san.internals.network_providers.abstract_reservations_storage \
+    import AbstractReservationsStorage
 from neuro_san.internals.reservations.reservation_dictionary_converter import ReservationDictionaryConverter
-from neuro_san.service.interfaces.startable import Startable
 
 
-class S3ReservationsStorage(ReservationsStorage, Startable):
+class S3ReservationsStorage(AbstractReservationsStorage):
```
d1donlydfink, line 47: It's good that this is not moving/renamed.
One less thing to break in existing deployment.

d1donlydfink: Feels a little weird that this guy is an AbstractReservationsStorage.
If I am correct in my interpretation of what is going on here:
* Is this guy only the writer to S3 now?
* Where/when is the reading from S3 happening?

d1donlydfink: Or maybe this guy is also doing the reading for now for compatibility and you are setting yourself up for read-through cache with these changes later?

neuro_san/interfaces/reservation.py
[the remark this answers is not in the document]
andreidenissov-cog: Fixed.

neuro_san/internals/interfaces/reservations_storage.py
andreidenissov-cog: Fixed.

neuro_san/internals/network_providers/abstract_reservations_storage.py
andreidenissov-cog: Moved stop() after start()

neuro_san/internals/network_providers/expiring_agent_network_storage.py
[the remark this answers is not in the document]
andreidenissov-cog: Our "base_storage" is abstract to us, so how can we be specific about possible exceptions?

neuro_san/internals/network_providers/expiring_agent_network_storage.py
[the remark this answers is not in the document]
andreidenissov-cog: Logic itself looks reasonable to me. Another question is that we never stop() expiration loop in "server" storage. And should we even start() it for "base" storage is another question.

neuro_san/internals/network_providers/expiring_agent_network_storage.py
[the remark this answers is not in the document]
andreidenissov-cog: I agree with that. Lock here is possible of course, but would be overkill.
Theoretically we can get delete-listeners called twice on some agent network X,
but there is not much harm I think.

neuro_san/internals/network_providers/expiring_agent_network_storage.py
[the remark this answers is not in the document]
andreidenissov-cog: Yep. And global uniqueness of TNW names makes it simpler.

neuro_san/internals/network_providers/fixed_agent_network_provider.py
andreidenissov-cog: It is used by ExpiringAgentNetworkStorage when we return NetworkProvider as a result of lookup.
My (speculative) reasoning is this: after we provided this TNW for some request execution, 
we don't want it to disappear on us after expiration, allowing to finish processing.
That's why we give "fixed" TNW instance, not extra redirect through modifiable table.

neuro_san/service/watcher/temp_networks/temp_network_storage_updater.py
andreidenissov-cog: Umm, not a huge difference resource-wise, but we have better chance not to block
when some new queue will drop in.

neuro_san/service/watcher/temp_networks/s3_reservations_storage.py
andreidenissov-cog: AbstractReservationsStorage is responsible for running expiration thread, this logic is shared between implementations. TNW-specific operations (expire_reservations and such) are still need to be implemented for each storage class.

neuro_san/service/watcher/temp_networks/s3_reservations_storage.py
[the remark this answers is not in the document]
andreidenissov-cog: OK no-op it is.

neuro_san/interfaces/reservation.py
```diff
@@ -51,6 +53,12 @@ def get_expiration_time_in_seconds(self) -> float:
         """
         return self.expiration_time_in_seconds
 
+    def is_expired(self) -> bool:
+        """
+        :return: Whether the reservation is expired or not.
+        """
+        return time.time() > self.expiration_time_in_seconds
```
d1donlydfink: I was looking to keep this a data-only class and not so much a policy class. Now there is a verb-barnacle on a noun-class.
Class comments clearly mention "data-structure" implying serializability, which policy methods complicate.

It seems like the only place this is used is within ExpiringAgentNetworkStorage.
Can we move this there as is_expired(Reservation)?

neuro_san/service/watcher/temp_networks/temp_network_storage_updater.py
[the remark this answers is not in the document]
d1donlydfink: OK I think I get what is happening here now - a single thread for processing any one inner queue and this is that thread's main loop.  It might be worth a name change on the method to something like "loop_on_one_inner_queue()" to make this clear.
Now that I think I understand better my worries return to issues of scaling up.
If we have one thread per inner queue, and our thread pool is only sized to 32, then we will only ever have 31 user requests (32 - one for the main task) to vibe code agent networks (for instance) at the same time. This seems like an artificial limitation. (Recall we want this to scale up ridiculously to potentially 1000s of users and a typical start-to-finish request time of AgentNetworkDesigner is a couple of minutes). I don't think adding an additional env variable for the thread pool size/user limitations would even cut it.

I guess I still don't understand the motivation to moving to one processing thread per inner-queue.
All these items are going to the same storage. Seems unnecessary to now be thread-limited.
If there were a single thread doing all external storage updates from all inner-queues (like before), then we have no such artificial limitation. What do you see as the gain?

I'd also feel better about the looping if we had a termination boolean that was checked instead of while-True with a return. Seems a better read structurally because we don't actually want to loop infinitely.

neuro_san/service/watcher/temp_networks/temp_network_storage_updater.py
```diff
@@ -35,103 +35,90 @@
...
+            # IF we have a storage to put the results in.
+            # Since each queue is processed by its own thread,
+            # we avoid potential locking out of different queues received from "main" self.incoming
+            if self.reservationist is not None:
+                self.queue_processing_pool.submit(self.process_one_queue, async_collating_queue)
+            else:
+                # We don't have a reservation functionality set up,
+                # so we can't do anything with this queue, but we should still close it to avoid leaks.
+                self.logger.warning("No reservationist available, closing incoming queue without processing")
+                async_collating_queue.close()
```
d1donlydfink, line 133: Another infinite loop that might not be necessary.
We were processing these on the watcher loop before and only processing what we needed to then exiting. If we call add_new_queues_to_processing() more than once then we fall into a trap. But it looks like we don't because this is only ever called from start().  Maybe rename this method to reflect it's a loop?

Also: How does this loop get properly stop()-ped? Shouldn't we:
*  have a boolean somewhere checking for a clean exit during the loop?
*  have a sleep() call so this doesn't hog CPU by looping too tightly?
* Update the "we have a new queue to process" comment?
* Simply not enter the infinite loop if there is no reservationist?

neuro_san/internals/interfaces/reservations_storage.py
```diff
@@ -54,3 +56,12 @@ def expire_reservations(self):
...
         raise NotImplementedError
+
+    def set_base_storage(self, base_storage: ReservationsStorage):
+        """
+        Set a "base" storage to use as a source of truth for reservations.
+        This is optional, but if set, will be used as the source of truth for reservations
+        and the implementing class will act as a cache in front of it.
+        :param base_storage: A ReservationsStorage instance to use as a source of truth
+        """
+        raise NotImplementedError
```
d1donlydfink: I don't think this set_base_storage() is necessary on this interface any more.
I believe below you have it so that you know you are dealing with temp_storage when doing the only call to it.

neuro_san/service/watcher/temp_networks/temp_network_storage_updater.py
```diff
@@ -35,103 +35,90 @@
...
-        if temp_storage is not None:
+        self.reservationist: AbstractAgentReservationist = None
+        self.temp_storage: ReservationsStorage = network_storage_dict.get("temp")
+        if self.temp_storage is not None:
             # If we don't have temp storage, we don't got nothin'
-            self.reservations_storage.add(temp_storage)
+            # Check that the temp storage instance is indeed an ExpiringReservationsStorage,
+            # since we need to be able to expire reservations in it.
+            if not isinstance(self.temp_storage, ReservationsStorage):
+                raise ValueError("Temp reservations storage must be an instance of ReservationsStorage")
```
d1donlydfink: I actually think you would want a stronger assertion here now.
I think you want to be sure it's an ExpiringAgentNetworkStorage.
It's possible you might have that in the #770 PR
It'd be clearer to move that little bit into this one if so.
Still, I think this check is likely to be unnecessary, because what else is it going to be? It's not user-configurable.

neuro_san/service/watcher/temp_networks/temp_network_storage_updater.py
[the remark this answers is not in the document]
d1donlydfink: It still makes sense to me to have a single thread looking at the outer-queue and processing and processing all the inner-queue reservations coming down for the in-memory only case of a single server.
We shouldn't need the artificial limitation of one thread per inner-queue for that case at all.

For the multi-server case, I think I see that maybe (guessing) you might be trying to protect the write-through-ness for each inner-queue. 

But consider this: It might make more sense to have one more thread doing the processing for all of the external storage writes for all inner-queues connected to the first aforementioned always-necessary in-memory update thread. This guy could have its own yet-another-queue to check (producer would be that other thread) and potentially batch up writes to the external storage: Think multiple rows to a database added at once when there's > 1 thing to store externally.

The rub there becomes just when/where/how to trigger the asyncio.Event that signals the initiating CodedTool that all is deployed.  Still seems doable tho since we do it already.

neuro_san/service/watcher/temp_networks/temp_network_storage_updater.py
```diff
@@ -35,103 +35,90 @@
...
 
-            # Potentially create an external storage class
+            # Potentially create a "base" reservations storage class
             storage_class_name: str = environ.get("AGENT_EXTERNAL_RESERVATIONS_STORAGE", "")
             external_storage: ReservationsStorage = ResolverUtil.create_instance(
                     storage_class_name,
                     "AGENT_EXTERNAL_RESERVATIONS_STORAGE env var",
                     ReservationsStorage)
```
d1donlydfink, line 76: Per your previous answer, it seems like you now want to be sure that this AGENT_EXTERNAL_RESERVATIONS_STORAGE guy is an AbstractReservationsStorage now, yes?

neuro_san/session/external_agent_session_factory.py
d1donlydfink: Seems indicative of a larger problem.
Feels like something to be done once and not over and over again.

At the very least add a code comment describing the unit test failure indicating why this is necessary.

neuro_san/service/watcher/temp_networks/temp_network_storage_updater.py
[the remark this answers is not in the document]
d1donlydfink: Also: Where does the external storage expiration get checked for removal now in all these changes?

neuro_san/interfaces/reservation.py
andreidenissov-cog: OK done.

neuro_san/internals/interfaces/reservations_storage.py
andreidenissov-cog: Yes, removed it from this interface.
We do have it in ExpiringAgentNetworkStorage.

neuro_san/service/watcher/temp_networks/temp_network_storage_updater.py
andreidenissov-cog: Missed that. Correct assertion is described in the comment above, but not implemented.
Fixed.

neuro_san/service/watcher/temp_networks/temp_network_storage_updater.py
andreidenissov-cog: Not really. What we do with "external" reservations storage here is more than covered by ReservationsStorage interface. We don't even need "expire_reservations" from it.
Maybe we don't need this method in the interface at all.
Cleaning up expired stuff, or not, is responsibility of each implementation.

neuro_san/service/watcher/temp_networks/temp_network_storage_updater.py
andreidenissov-cog: That's an interesting part.
To start with, the whole thing is predicated on us having "temp" network storage. If we do, we get into activity of setting up possible "base" storage, and we unconditionally create this AbstractAgentReservationist working with our temp storage.
Now, we are also given some black box input queue which will push some other queues with reservations on us.
Do we know how long or how frequently this queue will produce stuff? Not here.
Will it ever stop while service is running? We don't know.
So what else can we do except listen to it indefinitely and be ready to dispatch arriving queues for processing?

As for hogging CPU, I hope that self.incoming.sync_q.get() is something like internal event based,
and doesn't stupidly poll the queue in a tight loop. Would be strange.
Yes, if there is no "reservationist" == "no temp storage", we should not attempt to do anything at all
and just close incoming queue.

neuro_san/internals/network_providers/abstract_reservations_storage.py
andreidenissov-cog: Added name parameter to AbstractReservationsStorage constructor and updated logging.
I see storage names in the log output.

neuro_san/internals/network_providers/expiring_agent_network_storage.py
```diff
@@ -18,33 +18,84 @@
...
+        except Exception as exc:  # pylint: disable=broad-except
+            self.logger.error("%s: Failed to stop base storage: %s", self._name, exc)
+            raise
+
+    @staticmethod
+    def is_expired(reservation: Reservation) -> bool:
+        """
+        :return: Whether this reservation is expired or not.
+        """
+        return time.time() > reservation.get_expiration_time_in_seconds()
```
d1donlydfink, line 98: Thanks!

neuro_san/service/watcher/temp_networks/temp_network_storage_updater.py
d1donlydfink: > Will it ever stop while service is running? We don't know.

* The inner-queue's lifetime is as long as a request, and exists only if the request says it needs Reservations.
* The outer-queue's lifetime is as long as the service is alive.

d1donlydfink: >Yes, if there is no "reservationist" == "no temp storage", we should not attempt to do anything at all
and just close incoming queue.

Theoretically we can ensure that the queue doesn't exist and that this never gets called in that situation.
By what you are telling me, I think we might be doing that already.
Maybe it's worth a method comment about expectations so we don't rehash this?

neuro_san/internals/network_providers/expiring_agent_network_storage.py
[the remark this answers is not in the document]
d1donlydfink: Fair!

neuro_san/internals/network_providers/expiring_agent_network_storage.py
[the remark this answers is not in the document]
d1donlydfink: Probably not.
It's possible there's another layer of interface to hang this on.
Work in progress, can move later.

PR #770, opened 2026-02-27
NS-661 Fix temporary networks lookup across multiple server instances.

andreidenissov-cog:
This PR fixes the issue with TNW being used across multi-servers configurations,
when they have shared external storage.


neuro_san/internals/network_providers/expiring_agent_network_storage.py
```diff
@@ -198,5 +198,11 @@ def get_agent_network_provider(self, agent_name: str) -> AgentNetworkProvider:
                 with self.lock:
                     self.reservations_table[agent_name] = reservation
                     self.agents_table[agent_name] = agent_network
+                # Notify listeners about this state change:
+                # do it outside of internal lock
+                for listener in self.listeners:
+                    listener.agent_added(agent_name, self)
+                    self.logger.info("ADDED network for agent %s from %s", agent_name, "base storage")
+
```
andreidenissov-cog, line 237: Whenever we add a new agent network to internal registry, notify listeners - in this case our HTTP server maintaining a table of "allowed" agents.

neuro_san/internals/reservations/agent_reservation.py
```diff
@@ -50,6 +51,25 @@ def get_reservation_id(self) -> str:
         """
         return f"{self.prefix}{self.id}"
 
+    @staticmethod
+    def is_reservation_name(agent_name: str) -> bool:
+        """
```
andreidenissov-cog, line 56: Helper method to decide whether agent name does look like a temporary network.

neuro_san/service/watcher/temp_networks/s3_reservations_storage.py
```diff
@@ -239,15 +247,17 @@ def get_one_reservation(self, obj_key: str) -> Tuple[Reservation, Any]:
...
-        :param obj_key: S3 object key for the reservation
+        :param obj_key: reservation ID to retrieve (used to construct S3 object key)
         :return: Tuple of (reservation, agent_spec) if successful and not expired,
                  (None, None) otherwise
         """
         reservation: Reservation = None
         agent_network: AgentNetwork = None
+        # Construct the S3 object key for this reservation ID
+        s3_obj_key: str = self.get_obj_key_for_reservation(obj_key)
         try:
```
andreidenissov-cog, line 251: We have incoming agent reservation name, but what we need to work with S3 storage is S3 object key.

neuro_san/service/http/handlers/base_request_handler.py
```diff
@@ -137,20 +140,44 @@ async def get_service(self, agent_name: str, metadata: Dict[str, Any]) -> AsyncA
         service_provider: AsyncAgentServiceProvider = None
         is_authorized, service_provider = await self.agent_policy.allow_agent(agent_name, metadata)
 
-        if service_provider is None:
-            # Resource not found
-            self.set_status(404)
-            self.logger.error(metadata, "error: Invalid request path %s", self.request.path)
-            self.do_finish()
-            return None
-
```
andreidenissov-cog, line 146: This is moved down in the code.

neuro_san/service/http/handlers/base_request_handler.py
```diff
@@ -137,20 +140,44 @@ async def get_service(self, agent_name: str, metadata: Dict[str, Any]) -> AsyncA
...
         if not is_authorized:
             # Forbidden
             self.set_status(403)
             self.do_finish()
             return None
 
-        return service_provider.get_service()
+        if service_provider is not None:
+            return service_provider.get_service()
+
```
andreidenissov-cog, line 151: If agent network is found - fine, we are done.
Otherwise - check for possible TNW we haven't yet picked up from external stoarge.

neuro_san/service/http/handlers/base_request_handler.py
```diff
@@ -137,20 +140,44 @@ async def get_service(self, agent_name: str, metadata: Dict[str, Any]) -> AsyncA
...
+        # Does it look like a temporary network name?
+        if AgentReservation.is_reservation_name(agent_name):
+            # Get temporary network storage:
+            network_storage_dict: Dict[str, AgentNetworkStorage] = self.server_context.get_network_storage_dict()
+            temp_storage: AgentNetworkStorage = network_storage_dict.get("temp", None)
+            if temp_storage is None or not isinstance(temp_storage, ExpiringAgentNetworkStorage):
+                self.set_status(500)
+                self.logger.error(metadata, "error: Temporary network storage is not properly configured.")
+                self.do_finish()
+                return None
```
andreidenissov-cog, line 162: Temp network storage should be here in our service. If not - bail out.

neuro_san/service/http/handlers/base_request_handler.py
```diff
@@ -137,20 +140,44 @@ async def get_service(self, agent_name: str, metadata: Dict[str, Any]) -> AsyncA
...
+            # This call will trigger temp network lookup in external storage
+            # and if found, it will be registered in our service
+            # and become available for future requests until it expires.
+            # Note that "network_provider" result is not used directly here,
+            # but the call is necessary to trigger the "side effect" of lookup and registration of temp network.
+            network_provider = temp_storage.get_agent_network_provider(agent_name)
+            if network_provider is not None:
```
andreidenissov-cog, line 171: This call actually does the thing: look up TNW in the external storage (if we have it set),
and if found, put in in our internal context, so we know this agent network.

neuro_san/service/http/handlers/base_request_handler.py
```diff
@@ -137,20 +140,44 @@ async def get_service(self, agent_name: str, metadata: Dict[str, Any]) -> AsyncA
...
+                # Now that our service is aware of this temp network,
+                # we can query our policy again,
+                # and we already know that this agent has been authorized:
+                _, service_provider = await self.agent_policy.allow_agent(agent_name, metadata)
+                return service_provider.get_service()
```
andreidenissov-cog: And then "just" repeat the query. Recursive call, but never more than 2 deep.

neuro_san/internals/network_providers/expiring_agent_network_storage.py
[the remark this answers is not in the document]
andreidenissov-cog: There is a possible race between worker thread for get_agent_network_provider() and main server event loop, but looking at the logic, we should be safe.

neuro_san/service/http/handlers/base_request_handler.py
```diff
@@ -137,20 +140,49 @@ async def get_service(self, agent_name: str, metadata: Dict[str, Any]) -> AsyncA
...
+            # Note that "network_provider" result is not used directly here,
+            # but the call is necessary to trigger the "side effect" of lookup and registration of temp network.
+            # We delegate call execution to a thread to avoid blocking server event loop
+            # with potentially slow external storage access.
+            network_provider = await asyncio.to_thread(temp_storage.get_agent_network_provider, agent_name)
+            if network_provider is not None:
+                # Now that our service is aware of this temp network,
+                # we can query our policy again,
+                # and we already know that this agent has been authorized:
+                _, service_provider = await self.agent_policy.allow_agent(agent_name, metadata)
```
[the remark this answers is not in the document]
andreidenissov-cog: That's a fair point, and should be addressed in the future. Right now, I believe we are safe - our processing  logic should not cause any real problems.

neuro_san/service/http/handlers/base_request_handler.py
```diff
@@ -137,20 +140,49 @@ async def get_service(self, agent_name: str, metadata: Dict[str, Any]) -> AsyncA
...
+                # If service provider is still None,
+                # that means something went very wrong - bail out:
+                if service_provider is not None:
+                    return service_provider.get_service()
```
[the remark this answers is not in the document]
andreidenissov-cog: The same comment as above. FWIW, our operations on allowed_agents dictionary are pretty atomic and therefore thread-safe.

neuro_san/service/http/handlers/base_request_handler.py
d1donlydfink: Hmm. I suppose it makes sense that we might need to futz with ordering precedence here - 404 vs 403.  Is it better to return a 403 over a 404 when both are in play?

neuro_san/service/http/handlers/base_request_handler.py
d1donlydfink: We might want to let other people create their own implementations here, for instance for something other than AWS/S3 storage - maybe a database.  So checking for the specific class might not be quite right.

Would it be better to check for an env var being set instead?  and/or
Or be sure it derives from a specific interface in the docs?   and/or
Have a method on ReservationsStorage that answers the question you want here - that guy is a pure interface and even the current S3ReservationsStorage derives from him.

neuro_san/service/http/handlers/base_request_handler.py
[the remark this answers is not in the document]
d1donlydfink: Didn't we just do allow_agent() above before we got here?
Theoretically any Authorizer would need to carry the single source of truth for all pods.

neuro_san/service/http/handlers/base_request_handler.py
d1donlydfink: Does this also do the local cache lookup? Or did that already happen?
If it's already happening above, maybe worth a comment or 2 as to which is the temp cache lookup.

Also: Is it worth eventually having a thread pool for these? (doesn't have to be now)

neuro_san/service/http/server/http_server.py
```diff
@@ -232,6 +232,15 @@ def agent_added(self, agent_name: str, source: AgentStorageSource):
...
         """
         agent_network_provider: AgentNetworkProvider = source.get_agent_network_provider(agent_name)
+        # There are some timing scenarios where at this point agent network provider might not be available,
+        # even though we got a notification that agent is added.
+        # One example: temporary agent network expiration.
+        # In this case, log the message and finish processing.
+        if agent_network_provider is None:
+            self.logger.warning({},
+                                "Agent %s is not available though it's being added to the service.",
+                                agent_name)
```
d1donlydfink: Should this really be a warning?
I always think of warnings as something I should be able to do something about.
If there's nothing I can do it should either be suppressed and/or downgraded to an INFO.

Also the verbiage here is a little cryptic. Perhaps a little more explanation in what gets sent to the logs.
Think: devops would not really know what to do with this message as-is.

neuro_san/service/http/handlers/base_request_handler.py
andreidenissov-cog: Not really sure about this. But: we know our 403 right here, and can bail out quickly.
To get to actual 404, we'll need to go all the way down - to potentially querying base storage and all.
Still, not hard to flip that logic around if necessary.

neuro_san/service/http/handlers/base_request_handler.py
andreidenissov-cog: If you mean my check for ExpiringAgentNetworkStorage, this is our concrete implementation class for our service in-memory NetworkStorage, BUT for TNW specifically. This will not be re-created by other people outside of neuro-san. The whole thing is actually just a sanity check: in our service-level NetworkStorage table, what we put there under "temp" key - is it something we can work with?
If this check fails - something is seriously screwed OR we didn't specify that we want to work with TNW to begin with.

neuro_san/service/http/handlers/base_request_handler.py
andreidenissov-cog: temp_storage.get_agent_network_provider(agent_name) does the whole thing - local cache (ExpiringAgentNetworkStorage) lookup, goes to external storage if necessary, pulls stuff out, fills local cache. All is good, BUT: we don't just need NetworkProvider here, we need a service representing this agent network.
So we do a second go through self.agent_policy, where everything is set up already, to get ServiceProvider and then service instance proper.

andreidenissov-cog: Thread pool - maybe, but we don't do this quite right here. We only need to mess with worker thread for external storage related operations (slow, retries, timeouts and all). But this requires some additional refactoring. Not now.

neuro_san/service/http/handlers/base_request_handler.py
andreidenissov-cog, line 179: Theoretically this service may already tragically expire by that moment,
so the need for one final check.

neuro_san/service/http/handlers/base_request_handler.py
d1donlydfink: I guess what I'm saying is: If there is no positive entry in an Authorization system, it will tend to produce a negative result before even knowing existence.  

The question is: Do we want to shed light on authorization or existence first?  Do we want to favor sending a 403 Unauthorized more often, potentially casting aspersions on the Authorization system first (and prompting an engagement between user and administrator), rather than getting to the bottom of the issue with a preferred 404?

I think what you have is just fine for now, but we might want to revisit the ordering at some point in the future.

neuro_san/service/http/handlers/base_request_handler.py
d1donlydfink: I guess what I question is: If we always expect this to be an ExpiringAgentNetworkStorage, and no one can really override it, then why bother with the check?

Or maybe you are saying that the point of override is now ExpiringAgentNetworkStorage itself?
If so, I would slightly disagree with this approach.  If we expect the main implementation of ExpiringAgentNetworkStorage to be essential and the storage mechanism be swappable, then maybe what we need is a pure virtual interface for the storage retrieval mechanism delegation and that is what is overridable from within ExpiringAgentNetworkStorage for reads at least.  Perhaps the same storage delegation implementation sans server policy is also at play for the writes to make the implementation symmetrical?

Thoughts?

Again: This is fine for now, just trying to get a sense for what the expansion path looks like.

PR #781, opened 2026-03-05
Adding token cost of gpt-5.2

kaushik-cognizant:
At present we have moved to GPT-5.2. Current version of langchain-community still haven't add the cost of the gpt-5.2 models. Also, there seems to be too many PRs for the same so we don't know when they will update their code. 

See https://github.com/langchain-ai/langchain-community/pulls?q=is%3Apr+is%3Aopen+GPT-5.2

In the meantime, since we need to do live demos, I think we should add the cost to the default_llm_info.hocon so that we can show all the parameters.

### DIFF
--- a/neuro_san/internals/run_context/langchain/llms/default_llm_info.hocon
+++ b/neuro_san/internals/run_context/langchain/llms/default_llm_info.hocon
@@ -54,6 +54,41 @@
         "max_output_tokens": 100000,
         "knowledge_cutoff": "09/30/2023",
     },
+
+    "gpt-5.4-pro": {
+        "use_model_name": "gpt-5.4-pro-2026-03-05",
+    },
+    "gpt-5.4-pro-2026-03-05": {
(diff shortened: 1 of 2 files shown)


neuro_san/internals/run_context/langchain/llms/default_llm_info.hocon
```diff
@@ -68,6 +68,8 @@
         "context_window_size": 400000,
         "max_output_tokens": 128000,
         "knowledge_cutoff": "08/31/2025",
+        "price_per_1k_input_tokens": 0.00175,
+        "price_per_1k_output_tokens": 0.014,
```
kaushik-cognizant, line 108: Price value taken from https://developers.openai.com/api/docs/pricing/

Noravee: Thank you for adding the pricing.

PR #782, opened 2026-03-05
Add unit test for AbstractAsyncConfigRestorer

Noravee:
**Add unit tests for `AbstractAsyncConfigRestorer`**

Adds a comprehensive unit test suite for `AbstractAsyncConfigRestorer`, covering all public methods: `__init__`, `get_file_path`, `deserialize_file_contents`, `filter_config`, `restore`, and `async_restore` using fixture files (`valid.json`, `valid.hocon`, `invalid.json`, `invalid.hocon`).

**What's tested**
- Constructor argument storage and defaults
- File path resolution from explicit references, env vars, deprecated env vars, and precedence between them
- Deprecation warning logging when the deprecated env var is used
- Deserialisation of valid and invalid `.json` and `.hocon` files
- Error propagation: `ParseException` for malformed content, `ValueError` for unsupported extensions, `FileNotFoundError` for missing files when `must_exist=True`
- Silent `None` return for missing files when `must_exist=False`
- `filter_config` pass-through and error message content
- `filter_config` is called on every `restore` and `async_restore` invocation
- Env var lookup in both sync and async paths

### DIFF
--- a/tests/fixtures/invalid.json
+++ b/tests/fixtures/invalid.json
@@ -1,1 +1,3 @@
-{ not valid json @@@ }
+{ 
+    not valid json @@@
+}


tests/neuro_san/internals/persistence/test_abstract_async_config_restorer.py
```diff
@@ -0,0 +1,441 @@
...
+
+from neuro_san.internals.persistence.abstract_async_config_restorer import AbstractAsyncConfigRestorer
+
+
+# ---------------------------------------------------------------------------
+# Concrete subclass used by all tests
+# ---------------------------------------------------------------------------
+
+class ConcreteRestorer(AbstractAsyncConfigRestorer):
+    """Minimal concrete subclass – inherits all behaviour from the abstract base."""
```
d1donlydfink, line 40: Theoretically this should not be necessary.

tests/neuro_san/internals/persistence/test_abstract_async_config_restorer.py
```diff
@@ -0,0 +1,441 @@
...
+    path: Path = tmp_path / filename
+    path.write_text(json.dumps(content), encoding="utf-8")
+    return str(path)
+
+
+def write_hocon_file(tmp_path: Path, content: str, filename: str = "config.hocon") -> str:
+    """Write raw HOCON content to a temporary file and return its path."""
+    path: Path = tmp_path / filename
+    path.write_text(content, encoding="utf-8")
+    return str(path)
```
d1donlydfink: All of these methods should be part of the class.

tests/neuro_san/internals/persistence/test_abstract_async_config_restorer.py
```diff
@@ -0,0 +1,441 @@
...
+            r.restore(str(path))
+
+    # -----------------------------------------------------------------------
+    # async_restore
+    # -----------------------------------------------------------------------
+
+    def test_async_restore_returns_none_when_no_file_path(self) -> None:
+        """None is returned when no file path can be resolved."""
+        r: ConcreteRestorer = make_restorer(must_exist=False)
+        assert run(r.async_restore()) is None
```
d1donlydfink: Nice that async methods are tested too.

tests/neuro_san/internals/persistence/test_abstract_async_config_restorer.py
```diff
@@ -0,0 +1,441 @@
...
+# Constants
+# ---------------------------------------------------------------------------
+
+VALID_JSON: str = json.dumps({"key": "value", "nested": {"a": 1}})
+INVALID_JSON: str = "{ not valid json @@@ }"
+
+VALID_HOCON: str = 'key = "value"\nnested { a = 1 }'
+INVALID_HOCON: str = "{{{{{{{{{"
+
+VALID_DICT: Dict[str, Any] = {"key": "value", "nested": {"a": 1}}
```
d1donlydfink, line 47: Why all the mocking for these constants?
Why not just have some files with this stuff in it?
Then the tests would be less complicated and likely wouldn't need most of the mocking.

tests/neuro_san/internals/persistence/test_abstract_async_config_restorer.py
Noravee: The "Abstract" in the class name signals the intent that it should be subclassed. ConcreteRestorer is a one-liner that costs nothing. Happy to remove it if you prefers to keep things minimal, but I'd lean towards keeping it.

tests/neuro_san/internals/persistence/test_abstract_async_config_restorer.py
Noravee: Done

tests/neuro_san/internals/persistence/test_abstract_async_config_restorer.py
Noravee: Done.

Issue #783, opened 2026-03-05
Neuro-san server crashes if hocon file points to non existent substitution key


ofrancon:
Edit `neuro_san/registries/hello_world.hocon` to replace, for instance, the instructions by a substitution key. Except that this substitution does not exist. E.g.

```
            "instructions": ${non_existent_key_to_test_substitution},
```

Then start the server using:
```
 python -m neuro_san.service.main_loop.server_main_loop
```
It crashes with:
```
Traceback (most recent call last):
  File "/Users/user_id/workspace/neuro-san/venv/lib/python3.13/site-packages/pyhocon/config_parser.py", line 477, in _resolve_variable
    return True, config.get(variable)
                 ~~~~~~~~~~^^^^^^^^^^
  File "/Users/user_id/workspace/neuro-san/venv/lib/python3.13/site-packages/pyhocon/config_tree.py", line 236, in get
    return self._get(ConfigTree.parse_key(key), 0, default)
           ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/user_id/workspace/neuro-san/venv/lib/python3.13/site-packages/pyhocon/config_tree.py", line 176, in _get
    raise ConfigMissingException(
        u"No configuration setting found for key {key}".format(key='.'.join(key_path[:key_index + 1])))
pyhocon.exceptions.ConfigMissingException: 'No configuration setting found for key non_existent_key_to_test_substitution'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/Users/user_id/workspace/neuro-san/neuro_san/service/main_loop/server_main_loop.py", line 284, in <module>
    ServerMainLoop().main_loop()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/Users/user_id/workspace/neuro-san/neuro_san/service/main_loop/server_main_loop.py", line 228, in main_loop
    self.parse_args()
    ~~~~~~~~~~~~~~~^^
  File "/Users/user_id/workspace/neuro-san/neuro_san/service/main_loop/server_main_loop.py", line 206, in parse_args
    manifest_agent_networks: Dict[str, Dict[str, AgentNetwork]] = manifest_restorer.restore()
                                                                  ~~~~~~~~~~~~~~~~~~~~~~~~~^^
[...]

PR #785, opened 2026-03-05
#783 Catch pyhocon ConfigExceptions when restoring config files

ofrancon:
When trying to restore `.hocon` files `pyhocon` might throw `ConfigSubstitutionException` or other `ConfigExceptions`.

They were not caught and were crashing the `neuro-san` server (see #783)
Catching them now means the server:
- logs the error
- ignores this faulty hocon file
- and keeps going:

```
Parse error in registry item hello_world.hocon. Skipping. - Cannot resolve variable ${non_existent_key_to_test_substitution} (line: 62, col: 29)
manifest registry hello_world.hocon not found in /Users/754337/workspace/neuro-san/neuro_san/registries/manifest.hocon
```

### DIFF
--- a/neuro_san/internals/run_context/langchain/llms/default_llm_info.hocon
+++ b/neuro_san/internals/run_context/langchain/llms/default_llm_info.hocon
@@ -54,6 +54,41 @@
         "max_output_tokens": 100000,
         "knowledge_cutoff": "09/30/2023",
     },
+
+    "gpt-5.4-pro": {
+        "use_model_name": "gpt-5.4-pro-2026-03-05",
+    },
+    "gpt-5.4-pro-2026-03-05": {
+        "class": "openai",
+        "model_info_url": "https://developers.openai.com/api/docs/models/gpt-5.4-pro",
+        "modalities": {
+            "input": [ "text", "image" ],
+            "output": [ "text" ],
+        },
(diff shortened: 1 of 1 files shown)


neuro_san/internals/persistence/abstract_async_config_restorer.py
```diff
@@ -165,7 +166,7 @@ def deserialize_file_contents(self, file_path: str, file_contents: str) -> Dict[
         # Read the contents
         try:
             config = serialization.to_object(string_file)
-        except (ParseException, ParseSyntaxException, JSONDecodeError) as exception:
+        except (ParseException, ParseSyntaxException, JSONDecodeError, ConfigException) as exception:
```
ofrancon, line 169: Catching any pyhocon `ConfigException` in case pyhocon fails to restore the file.

andreidenissov-cog: Good thing to catch that.
Still any "other" exception potentially happening will fly out and crash our service?
Do we need some "catch-all" to protect against that?

d1donlydfink: I'd like to avoid catch-all exceptions whenever possible.
There's a reason pylint complains about them

PR #789, opened 2026-03-06
re-enabled test case

vince-leaf:
This PR re-enabled the smoke test on the anthropic test case. Also confirmed with using agent_cli on my local.

### DIFF
--- a/docs/agent_hocon_reference.md
+++ b/docs/agent_hocon_reference.md
@@ -621,9 +621,13 @@ Optional string specifying a Python class which implements the
 interface.
 
 <!-- pyml disable no-inline-html -->
+Implementations can be specified as a fully-qualified class name, as long as that is in your PYTHONPATH.
+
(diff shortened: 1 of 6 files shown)


tests/neuro_san/zzz_hocons/test_smoke_test_hocons.py
```diff
@@ -90,8 +90,7 @@ def test_hocon_with_server_default_llm(self, test_name: str, test_hocon: str):
         # These can be in any order.
         # Ideally more basic functionality will come first.
         # Barring that, try to stick to alphabetical order.
-        # Issue #762 and #763:
-        # "music_nerd_pro_llm_anthropic/combination_responses_with_history_direct.hocon",
+        "music_nerd_pro_llm_anthropic/combination_responses_with_history_direct.hocon",
```
vince-leaf, line 93: re-enabled test case

PR #797, opened 2026-03-11
NS-661 Optimize ServiceAgentReservationist queue creation for each instance.

andreidenissov-cog:
Simple optimization for ServiceAgentReservationist own processing queues.
Right now we create an empty queue in ServiceAgentReservationist constructor and immediately submit it to "main" queue of queues. This will result in some processing thread to latch on this empty queue and start blocking until something appears in it.
Instead, we now create the queue only when we have at least the first generated reservation item, so we'll not use up a thread from our pool for nothing.

Tested: running limited stress-test.

### DIFF
--- a/neuro_san/service/generic/agent_service.py
+++ b/neuro_san/service/generic/agent_service.py
@@ -251,8 +251,7 @@ class AgentService:
         # Create a reservationist
         reservationist: Reservationist = None
         if self.queues is not None:
-            reservationist = ServiceAgentReservationist()
-            self.queues.sync_q.put(reservationist.get_queue())
+            reservationist = ServiceAgentReservationist(self.queues)
 
         # Prepare
         factory = ExternalAgentSessionFactory(use_direct=False)
--- a/neuro_san/service/generic/async_agent_service.py
+++ b/neuro_san/service/generic/async_agent_service.py
@@ -249,8 +249,7 @@ class AsyncAgentService:
         # Create a reservationist for the occasion
         reservationist: Reservationist = None
         if self.queues is not None:
-            reservationist = ServiceAgentReservationist()
-            self.queues.sync_q.put(reservationist.get_queue())
+            reservationist = ServiceAgentReservationist(self.queues)
 
         # Prepare
         factory = ExternalAgentSessionFactory(use_direct=False)
--- a/neuro_san/service/generic/service_agent_reservationist.py
+++ b/neuro_san/service/generic/service_agent_reservationist.py
@@ -21,6 +21,8 @@ from typing import Union
 from asyncio import Event
 from asyncio import get_running_loop
 
+from janus import Queue
+
 from neuro_san.interfaces.reservation import Reservation
 from neuro_san.interfaces.reservationist import Reservationist
 from neuro_san.internals.chat.async_collating_queue import AsyncCollatingQueue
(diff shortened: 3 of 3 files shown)


neuro_san/service/generic/agent_service.py
```diff
@@ -251,8 +251,7 @@ def streaming_chat(self, request_dict: Dict[str, Any],
         # Create a reservationist
         reservationist: Reservationist = None
         if self.queues is not None:
-            reservationist = ServiceAgentReservationist()
-            self.queues.sync_q.put(reservationist.get_queue())
+            reservationist = ServiceAgentReservationist(self.queues)
```
andreidenissov-cog, line 255: Pass "main queue" as a constructor parameter to ServiceAgentReservationist.

neuro_san/service/generic/async_agent_service.py
```diff
@@ -249,8 +249,7 @@ async def streaming_chat(self, request_dict: Dict[str, Any],
         # Create a reservationist for the occasion
         reservationist: Reservationist = None
         if self.queues is not None:
-            reservationist = ServiceAgentReservationist()
-            self.queues.sync_q.put(reservationist.get_queue())
+            reservationist = ServiceAgentReservationist(self.queues)
```
andreidenissov-cog, line 253: Pass "main queue" as a constructor parameter to ServiceAgentReservationist.

neuro_san/service/generic/service_agent_reservationist.py
```diff
@@ -33,20 +35,18 @@ class ServiceAgentReservationist(Reservationist):
...
 
+        :param main_queue: The main Queue of AsyncCollatingQueues to which this instance
+                   will add its own AsyncCollatingQueue for deployment when it will be created.
         :param max_lifetime_in_seconds: The maximum lifetime allowed for any Reservation.
         """
         self.max_lifetime_in_seconds: float = max_lifetime_in_seconds
-        self.queue: AsyncCollatingQueue = AsyncCollatingQueue()
-
-    def get_queue(self) -> AsyncCollatingQueue:
-        """
```
andreidenissov-cog, line 46: This getter is no longer used.

neuro_san/service/generic/service_agent_reservationist.py
```diff
@@ -110,6 +110,11 @@ async def deploy(self,
                     queued_item["event"] = key
                     queued_item["event_loop"] = get_running_loop()
 
+                if self.queue is None:
+                    # We haven't created our queue yet, so we need to do that and put it on the main queue.
+                    self.queue = AsyncCollatingQueue()
+                    self.main_queue.sync_q.put(self.queue)
+
```
andreidenissov-cog, line 117: Create per-ServiceAgentReservationist reservations queue only when we have something to put into it.
And then submit it to the "main queue", so some processing thread can pick up this item ASAP.

neuro_san/service/generic/service_agent_reservationist.py
```diff
@@ -33,20 +35,18 @@ class ServiceAgentReservationist(Reservationist):
...
         self.max_lifetime_in_seconds: float = max_lifetime_in_seconds
-        self.queue: AsyncCollatingQueue = AsyncCollatingQueue()
-
-    def get_queue(self) -> AsyncCollatingQueue:
-        """
-        :return: The AsyncCollatingQueue belonging to this Reservationist
-        """
-        return self.queue
+        self.main_queue: Queue[AsyncCollatingQueue] = main_queue
+        self.queue: AsyncCollatingQueue = None
```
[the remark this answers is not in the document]
andreidenissov-cog: Good catch. Fixed.

neuro_san/service/generic/service_agent_reservationist.py
```diff
@@ -110,6 +110,11 @@ async def deploy(self,
                     queued_item["event"] = key
                     queued_item["event_loop"] = get_running_loop()
 
+                if self.queue is None:
+                    # We haven't created our queue yet, so we need to do that and put it on the main queue.
+                    self.queue = AsyncCollatingQueue()
+                    self.main_queue.sync_q.put(self.queue)
```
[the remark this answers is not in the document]
andreidenissov-cog: Used put_nowait here, but AFAIU, we cannot use async operations on this queue right here, since it is bound to another event loop.

neuro_san/service/generic/agent_service.py
d1donlydfink: This ServiceAgentReservationist is the instance that is handed to the CodedTools to do their business.
Now that you have handed the main queue-of-queues to that instance, you have also handed that main queue-of-queues to the coded tools. Now any coded tool can mess with the reservations system.

I believe until this change there was no other spot in the entirety of the system where any one request had the remotest ability to mess with any other request. Now you are introducing one.
You can never underestimate the hackiness of a python programmer.
"I have this reference here, that means I can do X!"

This is a security alert in the making.

d1donlydfink (changes requested):
In doing this optimization you are handing the CodedTools the ability to mess with the system as a whole.  There's a reason it was done this way.  

This was talked about in previous conversations. I don't understand what you are optimizing for other than your own agenda. Please explain.

PR #802, opened 2026-03-16
NS-661 Make external storage clean-up cycle configurable.

andreidenissov-cog:
This PR makes it possible to control how often we want to run check for expirations and clean-up for externally stored temp networks.
One of the options is to disable this check in our service completely, so as not to waste system resources on this and leave the task for some external tools or manual actions.

### DIFF
--- a/neuro_san/service/watcher/temp_networks/s3_reservations_storage.py
+++ b/neuro_san/service/watcher/temp_networks/s3_reservations_storage.py
@@ -77,9 +77,16 @@ class S3ReservationsStorage(AbstractReservationsStorage):
         try:
             expiration_check_period_seconds: float = float(envvar_value)
             self._check_interval_seconds = expiration_check_period_seconds
-        except ValueError:
-            self.logger.warning(
-                "Invalid value for %s, must be a number. Got: %s", envvar_name, envvar_value)
+        except ValueError as exc:
+            self.logger.error(
+                "Invalid value for %s, must be a number. Got: %s. "
+                "Please correct the environment variable or unset it.",
+                envvar_name,
+                envvar_value,
+            )
+            raise ValueError(
+                f"Invalid value for {envvar_name}: expected a numeric value, got {envvar_value!r}"
+            ) from exc
 
         # Configure bucket name from parameter or environment variable
         env_bucket: str = os.getenv("AGENT_RESERVATIONS_S3_BUCKET", "")


neuro_san/deploy/Dockerfile
```diff
@@ -307,6 +307,13 @@ ENV AGENT_EXTERNAL_RESERVATIONS_STORAGE=""
 # to use for cross-pod reservations storage.
 ENV AGENT_RESERVATIONS_S3_BUCKET=""
 
+# If AGENT_EXTERNAL_RESERVATIONS_STORAGE is set,
+# then this parameter specifies the period in seconds for checking the external storage for expired reservations.
+# Value of 0 means no periodic checks will be performed by the service,
+# in this case it is expected that the external storage itself will handle expired reservations,
+# or it will be done on demand by other tools.
+ENV AGENT_RESERVATIONS_EXTERNAL_STORAGE_CHECK_PERIOD_SECONDS="0"
+
```
andreidenissov-cog, line 316: Document new env var parameter for controlling clean-up cycle for temporary networks external storage.

neuro_san/internals/network_providers/abstract_reservations_storage.py
```diff
@@ -36,11 +36,12 @@ class AbstractReservationsStorage(ReservationsStorage, Startable):
     Specific logic for adding, retrieving, and expiring reservations is left to concrete implementations.
     """
 
-    def __init__(self, storage_name: str = "", check_expirations_interval_seconds: float = 60.0):
+    def __init__(self, storage_name: str = "", check_expirations_interval_seconds: float = 0.0):
         """
```
andreidenissov-cog, line 40: Make "no clean-up cycle" the default for external storage.
If we need it, we should specify this explicitly - clean-up takes CPU cycles.

neuro_san/internals/network_providers/abstract_reservations_storage.py
```diff
@@ -50,9 +51,13 @@ def __init__(self, storage_name: str = "", check_expirations_interval_seconds: f
...
 
     def start(self):
-        self._thread = threading.Thread(target=self._run_loop, daemon=True)
-        self._thread.start()
-        self._logger.debug("%s: Expiration cleanup thread started.", self._name)
+        if self._check_interval_seconds > 0:
+            self._thread = threading.Thread(target=self._run_loop, daemon=True)
+            self._thread.start()
+            self._logger.debug("%s: Expiration cleanup thread started with period %f sec.",
+                               self._name, self._check_interval_seconds)
```
andreidenissov-cog, line 58: Only start clean-up thread if requested.

neuro_san/service/watcher/temp_networks/s3_reservations_storage.py
```diff
@@ -216,12 +233,13 @@ def add_reservations(self, reservations_dict: Dict[Reservation, Any],
...
-                Body=json_body,
-                ContentType="application/json"
-            )
+
+            put_function = partial(self.s3_client.put_object,
+                                   Bucket=self.bucket_name,
+                                   Key=key,
+                                   Body=json_body,
+                                   ContentType="application/json")
+            self._do_with_retries(put_function)
```
andreidenissov-cog, line 250: Wrap up writing to S3 operation into our retry/backoff harness.

neuro_san/service/watcher/temp_networks/s3_reservations_storage.py
```diff
@@ -51,18 +51,43 @@ class S3ReservationsStorage(AbstractReservationsStorage):
...
                          check_expirations_interval_seconds=check_expirations_interval_seconds)
+        self.logger: Logger = getLogger(self.__class__.__name__)
+
+        # Check if expiration interval is set by environment variable,
+        # and adjust it if so (overriding the constructor parameter)
+        envvar_name: str = "AGENT_RESERVATIONS_EXTERNAL_STORAGE_CHECK_PERIOD_SECONDS"
+        envvar_value: str = os.getenv(envvar_name, "0")
+        try:
+            expiration_check_period_seconds: float = float(envvar_value)
+            self._check_interval_seconds = expiration_check_period_seconds
```
d1donlydfink, line 79: This is a superclass private variable from AbstractReservationStorage.
It's not entirely obvious from this class that only expire_reservations() gets called as part of its superclass loop. That's not bad, just had to go digging.

Makes me wonder if eventually the loop-iness/thread-iness can go back to the service side now that we are here.

PR #804, opened 2026-03-17
Summarization Middleware

Noravee:
**Updata 03/24/2026**
- Add flag `keep_summary_in_context` to control the outgoing chat history

Example — effect of ``keep_summary_in_context`` on chat history
(assuming ``keep=("messages", 2)`` and summarization is triggered):

incoming messages:
[SystemMessage, HumanMessage, AIMessage, HumanMessage, AIMessage, HumanMessage]

keep_summary_in_context=True:
chat history = [HumanMessage(summary), AIMessage(kept), HumanMessage(kept), AIMessage(new)]

keep_summary_in_context=False:
chat history = [HumanMessage, AIMessage, HumanMessage, AIMessage, HumanMessage, AIMessage(new)]

- Add unit test for `music_nerd_summarize`

**03/17/2026**

### Issue:
https://github.com/cognizant-ai-lab/neuro-san-studio/issues/702

### Overview

This draft PR demonstrates a possible approach for implementing chat summarization using middleware in Neuro-san.

The goal is to provide a concrete example that we can use to brainstorm, review, and align on the design before finalizing the implementation.
Once we reach agreement, this PR will be converted from draft to a full PR.

### Why not use the prebuilt middleware directly?

LangChain’s prebuilt SummarizationMiddleware assumes that it owns and mutates the chat state.

In Neuro-san, however:

- Chat history is managed outside of LangChain
- State updates must be synchronized with Neuro-san’s internal structures (e.g., chat history and journal)

Because of this mismatch, we cannot use the prebuilt middleware as-is and instead need a wrapper/adaptation layer.

### Changes in this draft

This PR introduces the following updates to support summarization:

- Pass chat history from RunContext → middleware factory
- Pass chat history and base journal from factory → middleware
    - Update chat history after summarization
    - Record summary events in the journal
- Add `NeuroSanSummarizationMiddleware`
    - Wraps LangChain’s middleware
    - Adapts its output to Neuro-san’s state management
- Add `music_nerd_summarize` example

### Notes

- This is intentionally a working example, not a finalized design
- Naming, structure, and responsibilities are all open for discussion

### DIFF
--- a/neuro_san/registries/music_nerd_summarize.hocon
+++ b/neuro_san/registries/music_nerd_summarize.hocon
@@ -105,7 +105,7 @@ You’re equal parts playlist curator, music historian, and pop culture mythbust
                         #
                         # keep_summary_in_context=False:
                         #     [HumanMessage, AIMessage, HumanMessage, AIMessage, HumanMessage, AIMessage(new)]
-                        "keep_summary_in_context": false,
+                        "keep_summary_in_context": true,
 
                         # Neuro-san-specific arguments:
                         # These are not part of LangChain. They are injected by Neuro-san at runtime
--- a/tests/fixtures/music_nerd_summarize/beatles_with_history_summarize.hocon
+++ b/tests/fixtures/music_nerd_summarize/beatles_with_history_summarize.hocon
@@ -1,0 +1,72 @@
+
+# Copyright © 2023-2026 Cognizant Technology Solutions Corp, www.cognizant.com.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# END COPYRIGHT
+
+# This file defines everything necessary for a data-driven test.
+# The schema specifications for this file are documented here:
+# https://github.com/cognizant-ai-lab/neuro-san/blob/main/docs/test_case_hocon_reference.md
+
+# This is a copy of tests/fixtures/music_nerd/beatles_with_history.hocon
+# This is to make sure that after the summarization the conversation can still be continue.
+{
+    # Describes what agent to test against.
+    "agent": "music_nerd_summarize",
+
+    # Timeout for the entire test.
+    # It's also possible to have a timeout_in_seconds on a per-interaction basis.
+    "timeout_in_seconds": 90,
+    
+    # success_ratio is useful for passing to the assessor to find out why
+    # a test is intermitently failing. Sometimes it's the agent prompt, but
+    # sometimes it's the test itself.
+    "success_ratio": "1/1",
+    
+    # Connect to the agent via a server
+    "connections": ["direct"],
+    
+    # Interactions are a series of dictionaries with request elements paired with
+    # descriptions of response checks.
+    "interactions": [
+        {
+            # This is what we send as input to streaming_chat()
+            "text": "Who did Yellow Submarine?",
+
+            # The response block treats how we are going to test what comes back
+            "response": {
+
+                # Text block says how we are going to examine the text of the response.
+                "text": {
+
+                    # Keywords says we are going to look for exact matches for each
+                    # element in a list of strings.  All elements need to show up
+                    # in order to pass.
+                    "keywords": ["Beatles"]
+                }
+            }
+        },
+        {
+            # This next interaction block tests the ability for an agent
+            # to continue its conversation with the context from the previous
+            # interaction.
+            "text": "Where were they from?",
+            "response": {
+                "text": {
+                    "keywords": ["Liverpool"]
+                }
+            }
+        }
+    ]
+}
--- a/tests/neuro_san/zzz_hocons/test_unit_test_hocons.py
+++ b/tests/neuro_san/zzz_hocons/test_unit_test_hocons.py
@@ -44,6 +44,7 @@ class TestUnitTestHocons(TestCase):
         "math_guy/forwarded_sly_data.hocon",
         "math_guy/deep_forwarded_sly_data.hocon",
         "music_nerd/beatles_with_history.hocon",
+        "music_nerd_summarize/beatles_with_history_summarize.hocon",
         "requests_get/get_example.hocon",
         "pii_middleware/jenny_phone.hocon",
 


neuro_san/internals/run_context/langchain/middleware/middleware_factory.py
```diff
@@ -173,6 +176,10 @@ def create_class(self, what_we_are_creating: str, create_type: Type[Any],
             args["origin"] = use_origin
         if "origin_str" in args:
             args["origin_str"] = Origination.get_full_name_from_origin(use_origin)
+        if "chat_history" in args:
+            args["chat_history"] = self.chat_history
+        if "base_journal" in args:
+            args["base_journal"] = self.base_journal
```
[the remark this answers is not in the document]
Noravee: Will be done in another PR.

neuro_san/middleware/neuro_san_summarization_middleware.py
```diff
@@ -0,0 +1,148 @@
...
+
+        # Record a marker in the journal indicating that a summary follows.
+        # The actual summarized content is written separately via the
+        # JournalCallbackHandler during LLM completion.
+        await self.base_journal.write_message(
+            message=AgentMessage("Here is a summary of the conversation to date:"),
+            # Exclude current middleware origin
+            origin=self.origin[:-1]
+        )
+
```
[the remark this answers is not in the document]
Noravee: Will add test once this is merged.

neuro_san/middleware/neuro_san_summarization_middleware.py
```diff
@@ -0,0 +1,148 @@
...
+from langchain.agents.middleware.summarization import DEFAULT_SUMMARY_PROMPT
+from langchain.agents.middleware.summarization import SummarizationMiddleware
+from langchain.agents.middleware.types import AgentState
+from langchain.agents.middleware.types import ContextT
+from langchain_core.language_models.chat_models import BaseChatModel
+from langchain_core.messages.base import BaseMessage
+from langchain_core.messages.utils import count_tokens_approximately
+from langchain_core.messages.utils import MessageLikeRepresentation
+from langgraph.runtime import Runtime
+
```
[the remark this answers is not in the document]
Noravee: Done.

neuro_san/registries/music_nerd_summarize.hocon
```diff
@@ -0,0 +1,118 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# END COPYRIGHT
+
+# The schema specifications for this file are documented here:
+# https://github.com/cognizant-ai-lab/neuro-san/blob/main/docs/agent_hocon_reference.md
```
d1donlydfink, line 19: How about a test for this?

neuro_san/middleware/neuro_san_summarization_middleware.py
```diff
@@ -0,0 +1,154 @@
...
+# to avoid importing private module-level constants (e.g.,
+# `_DEFAULT_MESSAGES_TO_KEEP`, `_DEFAULT_TRIM_TOKEN_LIMIT`), which are not
+# part of the public API and may change across LangChain versions.
+_DEFAULT_MESSAGES_TO_KEEP: int = 4
+_DEFAULT_TRIM_TOKEN_LIMIT: int = 3000
+
+
+class NeuroSanSummarizationMiddleware(SummarizationMiddleware):
+    """
+    Neuro-san-specific wrapper around LangChain's SummarizationMiddleware.
```
d1donlydfink: This is actually interesting.
Under most circumstances I would say "let this be in its own library", but because this specifically
has to do with neuro-san internals, this actually makes some sense.

neuro_san/internals/run_context/langchain/middleware/middleware_factory.py
d1donlydfink: This is not the right Journal to pass to the Middleware itself, as it has no origin information attached to it
and I don't like the idea of passing this base_journal directly to client code to mess with.
Too easy to screwup.

You'll notice the comment about ProgressReporter below - I had not gotten to doing the correct origin business for the progress reporter.  I was thinking about having middleware be its own point of origin whose origin parent was the calling agent.

Maybe in order to keep your own progress going with this here what you could do is this:
* Remove the base_journal case arg passing here
* Remove the journal reporting in your Middleware for now

When we get the correct journal plumbing in, then you can report from your middleware.
How does that sound?

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
```diff
@@ -241,7 +241,7 @@ def create_agent(self, prompt_template: ChatPromptTemplate, llm: BaseLanguageMod
 
         # Create any middleware instances that were specified, in the order they were specified.
         # This will be None for most simple situations.
-        middleware_factory = MiddlewareFactory(self.origin, self.invocation_context.get_journal())
+        middleware_factory = MiddlewareFactory(self.origin, self.chat_history, self.invocation_context.get_journal())
```
d1donlydfink: Looking at this again, this is probably not the right journal to pass to any Middleware.
See comment below.

neuro_san/middleware/neuro_san_summarization_middleware.py
```diff
@@ -0,0 +1,154 @@
...
+        # No summarization occurred
+        if not messages_dict:
+            return None
+
+        # Replace Neuro-san chat history with summarized messages
+        self.chat_history.clear()
+        # The first message is a LangChain-specific RemoveMessage instruction,
+        # which is not applicable in Neuro-san, so it is skipped.
+        summarized_messages: List[BaseMessage] = messages_dict.get("messages", [])
+        self.chat_history.extend(summarized_messages[1:])
```
d1donlydfink: It might be worth having a test to be sure the summarization keeps the messages that we expect.
For instance, I think we need there to be a SystemMessage and HumanMessage at the start at the very least.

Can there be more than 1 RemoveMessage?
Are there specific semantics around it that are worth commenting on?

neuro_san/middleware/neuro_san_summarization_middleware.py
```diff
@@ -0,0 +1,154 @@
...
+
+        # Record a marker in the journal indicating that a summary follows.
+        # The actual summarized content is written separately via the
+        # JournalCallbackHandler during LLM completion.
+        await self.base_journal.write_message(
+            message=AgentMessage("Here is a summary of the conversation to date:"),
+            # Exclude current middleware origin
+            origin=self.origin[:-1]
```
d1donlydfink: Similar deal with the origin.  I wouldn't use it directly.
Theoretically we should be passing an OriginatingJournal in here (which ideally would be origin-scoped to include the middleware as part of the origin) and you shouldn't need the separate origin information.

Maybe I can finish up my origin work next since you need it.

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
Noravee: Journal is actually not needed. The summary is automatically added to the calling agent journal by `JournalCallbackHandler.on_llm_end()`.

neuro_san/middleware/neuro_san_summarization_middleware.py
Noravee: Journal is actually not needed. The summary is automatically added to the calling agent journal by `JournalCallbackHandler.on_llm_end()`.

neuro_san/registries/music_nerd_summarize.hocon
Noravee: test added.

neuro_san/middleware/neuro_san_summarization_middleware.py
Noravee: The system message is not in the chat history.

There is only 1 RemoveMessage. The format is [remove message, summary, kept message].

neuro_san/middleware/neuro_san_summarization_middleware.py
```diff
@@ -0,0 +1,167 @@
...
+    (assuming ``keep=("messages", 2)`` and summarization is triggered):
+
+        incoming messages:
+            [HumanMessage, AIMessage, HumanMessage, AIMessage, HumanMessage]
+
+        keep_summary_in_context=True:
+            [HumanMessage(summary), AIMessage(kept), HumanMessage(kept), AIMessage(new)]
+
+        keep_summary_in_context=False:
+            [HumanMessage, AIMessage, HumanMessage, AIMessage, HumanMessage, AIMessage(new)]
```
Noravee, line 85: Add a flag `keep_summary_in_context` to choose the behavior of outgoing chat history.

neuro_san/registries/music_nerd_summarize.hocon
```diff
@@ -0,0 +1,133 @@
...
+                        #
+                        # incoming messages:
+                        #     [HumanMessage, AIMessage, HumanMessage, AIMessage, HumanMessage]
+                        #
+                        # keep_summary_in_context=True:
+                        #     [HumanMessage(summary), AIMessage(kept), HumanMessage(kept), AIMessage(new)]
+                        #
+                        # keep_summary_in_context=False:
+                        #     [HumanMessage, AIMessage, HumanMessage, AIMessage, HumanMessage, AIMessage(new)]
+                        "keep_summary_in_context": true,
```
Noravee, line 108: `keep_summary_in_context` default is `true`. That is the outgoing chat history should be like `[HumanMessage(summary), AIMessage(kept), HumanMessage(kept), AIMessage(new)]` instead of
`[HumanMessage, AIMessage, HumanMessage, AIMessage, HumanMessage, AIMessage(new)]`.

tests/fixtures/music_nerd_summarize/beatles_with_history_summarize.hocon
```diff
@@ -0,0 +1,72 @@
...
+# limitations under the License.
+#
+# END COPYRIGHT
+
+# This file defines everything necessary for a data-driven test.
+# The schema specifications for this file are documented here:
+# https://github.com/cognizant-ai-lab/neuro-san/blob/main/docs/test_case_hocon_reference.md
+
+# This is a copy of tests/fixtures/music_nerd/beatles_with_history.hocon
+# This is to make sure that after the summarization the conversation can still be continue.
```
Noravee, line 23: Add test for `music_nerd_summarize`. This is a copy of the test for `music_nerd`. This is to ensure that the conversation can be continued after summarization.

tests/fixtures/music_nerd_summarize/beatles_with_history_summarize.hocon
```diff
@@ -0,0 +1,72 @@
...
+        {
+            # This next interaction block tests the ability for an agent
+            # to continue its conversation with the context from the previous
+            # interaction.
+            "text": "Where were they from?",
+            "response": {
+                "text": {
+                    "keywords": ["Liverpool"]
+                }
+            }
```
ofrancon, line 69: Maybe we should have a mechanism in the test framework to check the number of messages in the chat context. Or check that some middleware was called: nothing in this test tells us anything was actually summarized.

PR #808, opened 2026-03-20
Catch middleware runtime errors

### LINKED ISSUE #799: Return error when middleware fails
Currently, it hangs when there are errors in the middleware.

Noravee:
Fix #799 

## Problem
Errors raised during middleware instantiation are currently handled in the middleware factory. However, this does not cover errors that occur during middleware execution at runtime.

## Change
Add error handling in `RunContextRunnable.invoke_agent_chain()` to catch exceptions raised while running middleware.

## Rationale
Middleware execution can fail for a wide range of reasons depending on implementation and runtime context. To ensure robustness, we catch a broad `Exception` at this stage rather than limiting to specific error types.

### DIFF
--- a/.github/workflows/smoke.yml
+++ b/.github/workflows/smoke.yml
@@ -75,7 +75,6 @@ jobs:
           AZURE_OPENAI_API_KEY: ${{ secrets.AZURE_OPENAI_API_KEY }}
           AZURE_OPENAI_ENDPOINT: ${{ vars.AZURE_OPENAI_ENDPOINT }}
           OPENAI_API_VERSION: ${{ vars.OPENAI_API_VERSION }}
-          AGENT_TEMPORARY_NETWORK_UPDATE_PERIOD_SECONDS: 5
           AGENT_TOOL_PATH: "./neuro_san/coded_tools"
           PYTHONPATH: ${{ env.PYTHONPATH }}:.
         run: |
--- a/docs/agent_hocon_reference.md
+++ b/docs/agent_hocon_reference.md
@@ -685,6 +685,9 @@ For more details on tool extension, see the [Toolbox Extension Guide](./toolbox_
 
 For more information on tool schema, see the [toolbox_info_hocon_reference](./toolbox_info_hocon_reference.md).
 
+Note that an agent using `toolbox` cannot have a `tools` field. A toolbox agent executes code directly
+rather than calling an LLM, so it cannot invoke other tools as downstream agents.
+
 Example networks using tools from toolbox:
 
 - [date_time_timezone.hocon](../neuro_san/registries/date_time_timezone.hocon) which uses predefined
--- a/neuro_san/deploy/Dockerfile
+++ b/neuro_san/deploy/Dockerfile
@@ -289,14 +289,6 @@ ENV AGENT_HTTP_SERVER_INSTANCES=1
 # Logging period is specified in seconds.
 ENV AGENT_HTTP_RESOURCES_MONITOR_INTERVAL=0
 
-# If this value is specified and >0,
-# it will enable dynamic temporary network updates of the server agents
-# allowing CodedTools to reserve temporary networks via the Reservationist
-# interface, if the agent network hocon file specifically allows it for the tool.
-# Update will be done every AGENT_TEMPORARY_NETWORK_UPDATE_PERIOD_SECONDS seconds;
-# if value is not specified or <= 0, no such dynamic updates will be executed.
-ENV AGENT_TEMPORARY_NETWORK_UPDATE_PERIOD_SECONDS=0
-
 # Optional URL describing how the server is to be referenced by the outside world.
 # This is useful when a server is behind a load-balancer as part of a larger cluster.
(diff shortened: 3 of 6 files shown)


neuro_san/internals/run_context/langchain/core/run_context_runnable.py
```diff
@@ -186,6 +187,13 @@ async def invoke_agent_chain(self, inputs: Dict[str, Any], runnable_config: Dict
                     retries = retries - 1
                     exception = value_error
                     backtrace = traceback.format_exc()
+            # pylint: disable=broad-exception-caught
+            except Exception as exception_error:
+                # This catches any errors from running middlewares and also error from exceeding the recursion_limit.
+                self.logger.warning("retrying from Exception")
```
ofrancon: Are you sure you don't want to log the full exception here? 
I guess it will be logged as an error once the number of retries has been reached, but it could still be useful to log a little more context. At least the source of the exception: "RunContextRunnable: retrying from exception".
Not sure how frequently that happens, but I think I'd like to know a little more when I see `retrying from Exception` warnings in my logs

Noravee: The detail of the error will get return as ai message. The conversation can still be continue after this.

neuro_san/internals/run_context/langchain/core/run_context_runnable.py
```diff
@@ -186,6 +187,13 @@ async def invoke_agent_chain(self, inputs: Dict[str, Any], runnable_config: Dict
...
+                retries = retries - 1
+                exception = exception_error
+                backtrace = traceback.format_exc()
 
         output: str = self.parse_chain_result(chain_result, exception, backtrace)
         return_message: BaseMessage = AIMessage(output)
```
Noravee, line 205: Results from LLM, exception, and backtrace are returned as ai message.

neuro_san/internals/run_context/langchain/core/run_context_runnable.py
```diff
@@ -186,6 +187,16 @@ async def invoke_agent_chain(self, inputs: Dict[str, Any], runnable_config: Dict
                     retries = retries - 1
                     exception = value_error
                     backtrace = traceback.format_exc()
+            # pylint: disable=broad-exception-caught
+            except Exception as exception_error:
```
d1donlydfink, line 193: Hmmm. What I don't like about this broad exception catching is that we actually retry them.
Seems like these would most likely be real errors that are non-retry-able.
What do you think @Noravee ?

Exceeding recursion_limit I can see being retryable tho.
Is there a specific exception for this?

Noravee: Good point — I agree that retrying on a broad `Exception` is not ideal since most of these are likely non-retryable and could mask real issues.

I’ll update the logic so that generic exceptions are surfaced without retrying, and we only retry on explicitly allowed cases.

For the recursion limit case, there is a specific exception in `LangGraph`. However, since `Neuro-San` doesn’t directly depend on LangGraph, I’d prefer to avoid introducing that dependency just for exception handling.

Also, retries don’t really help in this case because the recursion limit counter isn’t reset—so it will fail again almost immediately even after retrying.

I’ll handle the recursion limit case separately in a follow-up PR.

In the meantime, I’m thinking of handling retryable errors more generically (e.g., detecting via error type/message). Open to suggestions if you think there’s a cleaner approach.

d1donlydfink: We could concievably create our own RetryableException which wraps another exception.
This could allow CodedTools and Middleware to signal that a particular kind of exception is retryable and that could be caught here separately.

As for cataloging by type/message - I think except for langchain and llm provider errors we already look for, we would have another catalog of errors to maintain. Not sure we should go there.

ofrancon: I like the idea of a `RetryableException`. Not sure how practical it would be. We could also have a simple static Util `isRetryableException(Exception e)` that returns a boolean. The implementation could be a simple as a look up table on the class name. But we could make it more specific if needed, e.g. looking at the message itself for a given Class.

neuro_san/internals/run_context/langchain/core/run_context_runnable.py
```diff
@@ -197,6 +198,17 @@ async def invoke_agent_chain(self, inputs: Dict[str, Any], runnable_config: Dict
...
+                # This catches any errors from running middlewares and also error form exceeding the recursion_limit.
+                self.logger.error("Got exception in  %s. Error: %s",
```
ofrancon, line 195: Thanks for the mode detailed log.

neuro_san/internals/run_context/langchain/core/run_context_runnable.py
d1donlydfink: My 2 cents: The RetryableException definition on our part would allow CodedTools to contribute their own sense of what should be retried at the framework level before giving up.  Think: Talking to a flaky web server (like GitHub! :) )

PR #812, opened 2026-03-23
Upgrade leaf-common to 1.2.37

ofrancon:
Upgrading to leaf-common `1.2.37` to pick the fix for non-UTF-8 hocon files (see https://github.com/cognizant-ai-lab/leaf-common/pull/142)
That will fix https://github.com/cognizant-ai-lab/neuro-san-studio/issues/705

### DIFF
--- a/docs/agent_hocon_reference.md
+++ b/docs/agent_hocon_reference.md
@@ -81,8 +81,10 @@ Items in ***bold*** are essentials. Try to understand these first.
     - [middleware](#middleware)
         - [class](#class-2)
         - [args](#args-1)
+            - [chat_history](#chat_history)
             - [origin](#origin)
             - [origin_str](#origin_str)
+            - [progress_reporter](#progress_reporter)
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -6,7 +6,7 @@
 # requirements-build.txt
 
 # In-house dependencies
-leaf-common>=1.2.34
+leaf-common>=1.2.37
 leaf-server-common>=0.1.24
```
andreidenissov-cog, line 11: Shouldn't we pick up the latest leaf-common (which is 1.2.37) just by rebuilding virtual env?

andreidenissov-cog (approved):
LGTM.

requirements.txt
d1donlydfink: Yes. but I think what @ofrancon is trying to do is to _require_ the newest version when people upgrade to the next neuro-san version, which is totally fair.

PR #814, opened 2026-03-25
NS-805 Fix the issue with absent/bad API keys in neuro-san sessions.

andreidenissov-cog:
This PR addresses the issue where the problem with LLM provider API keys (absent or misspelled or otherwise bad)
leads to neuro-san request just hanging there without providing any feedback.
Exceptions are thrown in request execution event loop, and task dies,
but nobody notifies http handler in Tornado server event loop, because the only link between the two
is our async message queue.
The fix here is to construct an appropriate message with the failure reason, 
pit it in the message queue and finalize the request processing.

### DIFF
--- a/neuro_san/internals/run_context/langchain/core/run_context_runnable.py
+++ b/neuro_san/internals/run_context/langchain/core/run_context_runnable.py
@@ -165,6 +165,9 @@ class RunContextRunnable(NeuroSanRunnable):
                     chain_result = {
                         "output": message
                     }
+                    # Log the error with technical details for debugging purposes,
+                    # but we are returning a more user-friendly message to the client.
+                    self.logger.error("API KEY error detected: %s", str(api_error))
                     break
                 # Continue with regular retry logic:
                 self.logger.warning("retrying from %s", api_error.__class__.__name__)
--- a/neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
+++ b/neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
@@ -22,6 +22,9 @@ from typing import Set
 from typing import Type
 from typing import Tuple
 
+from logging import Logger
+from logging import getLogger
+
 import os
 
 from langchain_core.language_models.base import BaseLanguageModel
@@ -82,6 +85,7 @@ class DefaultLlmFactory(ContextTypeLlmFactory, LangChainLlmFactory):
         :param config: The config dictionary which may or may not contain
                        keys for the context_type and agent_llm_info_file
         """
+        self.logger: Logger = getLogger(self.__class__.__name__)
         self.llm_infos: Dict[str, Any] = {}
         self.overlayer = DictionaryOverlay()
         self.llm_factories: List[LangChainLlmFactory] = [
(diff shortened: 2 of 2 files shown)


neuro_san/internals/chat/data_driven_chat_session.py
```diff
@@ -252,7 +252,14 @@ async def run_it(self, inputs: AgentFrameworkMessage) -> AgentFrameworkMessage:
...
         if self.front_man is None:
-            await self.set_up(self.invocation_context, sly_data, chat_context)
+            try:
+                await self.set_up(self.invocation_context, sly_data, chat_context)
+            except ValueError as exc:
+                # This can happen if we have problems with LLM clients API keys:
+                # Construct a message to send back to the client with the error information.
+                message = AgentFrameworkMessage(content=str(exc))
+                await self.finalize_run(message)
+                return message
```
andreidenissov-cog, line 262: If API key is not set at all, this is where exception will be caught.
Exception will contain a generic client-oriented message:
"A value for the OPENAI_API_KEY environment variable must be correctly set..." etc.
We construct a message out of this string to carry it across - from request-specific event loop 
to http handler event loop.
And then we clean up this run.

neuro_san/internals/chat/data_driven_chat_session.py
```diff
@@ -290,6 +297,24 @@ async def run_it(self, inputs: AgentFrameworkMessage) -> AgentFrameworkMessage:
         # at the end of the tracing context.
         message = AgentFrameworkMessage(content=answer, chat_context=return_chat_context,
                                         sly_data=return_sly_data, structure=structure)
+        await self.finalize_run(message)
+
```
andreidenissov-cog, line 301: Clean-up steps are now factored out to "finalize_run" function.

neuro_san/internals/run_context/langchain/core/run_context_runnable.py
```diff
@@ -157,7 +157,15 @@ async def invoke_agent_chain(self, inputs: Dict[str, Any], runnable_config: Dict
...
-                    raise ValueError(message) from api_error
+                    # Construct a uniform message to return to the client
+                    # indicating that they likely have an API key problem,
+                    # rather than retrying and hitting the same error again.
+                    exception = None
+                    backtrace = None
+                    chain_result = {
+                        "output": message
+                    }
+                    break
```
andreidenissov-cog, line 171: There is no convenient spot up the call stack to catch and process this ValueError exception.
But: we already have all scaffolding here to wrap up the error message and send it across through
the message queue.

neuro_san/internals/chat/data_driven_chat_session.py
d1donlydfink: A couple of things here:

1) Seems like we need to be logging this on the server side.  

This link is where the error is being detected:
https://github.com/cognizant-ai-lab/neuro-san/blob/main/neuro_san/internals/run_context/langchain/util/api_key_error_check.py#L55

... but there are two places where that can be converted to a ValueError:

a) https://github.com/cognizant-ai-lab/neuro-san/blob/main/neuro_san/internals/run_context/langchain/core/run_context_runnable.py#L160
b) https://github.com/cognizant-ai-lab/neuro-san/blob/main/neuro_san/internals/run_context/langchain/llms/default_llm_factory.py#L339

I'm not exactly sure which is at play.

What really surprises me is that in no case is this being logged.  We should change that so devops can know they have key issues immediately.

2) There is a possibility that this ValueError catching here is both too broad and not broad enough.

a) Too broad in that other code signaling ValueErrors will get caught up in this (but perhaps that's OK)

b) Not broad enough in that yes, it is typical for a front man's LLMs to fairly homogenous with all the other agents' LLMs, but that is not always the case. That is, setting up the front man might not be the only place to report this, ultimately.

That said, for #2 this is progress to a main pain point, so let's go with it.

I do think we should figure out the right place to do actual logging of the nice error message within this PR.

PR #820, opened 2026-03-27
NS-816 Call external networks by correct invocation specified by Function() results

d1donlydfink:
* Get the way an external agent wants to be invoked (event vs chatbot) from the call to Function()
* Plumb that invocation from Function() results from LangChainOpenAIFunctionTool all the way to where the InvocationContext gets called with safe_shallow_copy() to pass that desired invocation in to the new copy.

I tested this by adding `"invocation": "event"` to agent_network_designer and watched it go to close to completion.
Even though with this manual test, the thing I am invoking is an event, the other external networks it pokes are not, and those seem to do OK enough - at least much more OK than they did before which was to return early with the previous behavior of inherited event-ness.  There might be more cases to get here, but what I have in this PR is real progress.

Was able to see this work by a combination of the following hacks:
* Commented out the stuff in SessionInvocationContext.close() that returns the AsyncioExecutor to the pool
* In AsyncDirectAgentSession.streaming_chat()'s finally block at the end I commented out chat_session.delete_resources() so the LlmResources/connections would still exist
* Do not close() the reservationist or sly_data in DataDrivenChatSession.finalize_run()

These are the kinds of things that will have to be finalized when the event processing is done.

### DIFF
--- a/neuro_san/deploy/Dockerfile
+++ b/neuro_san/deploy/Dockerfile
@@ -307,6 +307,13 @@ ENV AGENT_EXTERNAL_RESERVATIONS_STORAGE=""
 # to use for cross-pod reservations storage.
 ENV AGENT_RESERVATIONS_S3_BUCKET=""
 
+# If AGENT_EXTERNAL_RESERVATIONS_STORAGE is set,
+# then this parameter specifies the period in seconds for checking the external storage for expired reservations.
+# Value of 0 means no periodic checks will be performed by the service,
+# in this case it is expected that the external storage itself will handle expired reservations,
+# or it will be done on demand by other tools.
+ENV AGENT_RESERVATIONS_EXTERNAL_STORAGE_CHECK_PERIOD_SECONDS="0"
+
 # A hocon file with MCP servers information to be used by LangChainMcpAdapter
 # for connecting to external MCP servers with authentication and tool filtering.
 ENV MCP_SERVERS_INFO_FILE=""
--- a/neuro_san/internals/network_providers/abstract_reservations_storage.py
+++ b/neuro_san/internals/network_providers/abstract_reservations_storage.py
@@ -36,11 +36,12 @@ class AbstractReservationsStorage(ReservationsStorage, Startable):
     Specific logic for adding, retrieving, and expiring reservations is left to concrete implementations.
     """
 
-    def __init__(self, storage_name: str = "", check_expirations_interval_seconds: float = 60.0):
+    def __init__(self, storage_name: str = "", check_expirations_interval_seconds: float = 0.0):
         """
         Constructor
         :param storage_name: A string name for this storage, used for logging purposes.
         :param check_expirations_interval_seconds: The number of seconds between checks for expired reservations.
+                            If set to 0 or negative, the background thread will not be started.
         """
         super().__init__()
         self._thread: threading.Thread = None
@@ -50,9 +51,13 @@ class AbstractReservationsStorage(ReservationsStorage, Startable):
         self._name: str = storage_name
 
     def start(self):
-        self._thread = threading.Thread(target=self._run_loop, daemon=True)
-        self._thread.start()
-        self._logger.debug("%s: Expiration cleanup thread started.", self._name)
+        if self._check_interval_seconds > 0:
+            self._thread = threading.Thread(target=self._run_loop, daemon=True)
+            self._thread.start()
+            self._logger.debug("%s: Expiration cleanup thread started with period %f sec.",
+                               self._name, self._check_interval_seconds)
+        else:
+            self._logger.debug("%s: Expiration cleanup thread not started.", self._name)
 
     def stop(self, timeout: Optional[float] = None):
         """
--- a/neuro_san/service/watcher/temp_networks/s3_reservations_storage.py
+++ b/neuro_san/service/watcher/temp_networks/s3_reservations_storage.py
@@ -51,18 +51,43 @@ class S3ReservationsStorage(AbstractReservationsStorage):
     Stores reservations as JSON objects in an S3 bucket, with each reservation
     stored in its associated agent spec as metadata.
     """
+    # pylint: disable=too-many-instance-attributes
 
     def __init__(self, bucket_name: str = "", prefix: str = "reservations/",
-                 check_expirations_interval_seconds: float = 60.0):
+                 check_expirations_interval_seconds: float = 0.0):
         """
         Initialize S3 reservations storage.
 
         :param bucket_name: S3 bucket name (defaults to AGENT_RESERVATIONS_S3_BUCKET env var)
         :param prefix: S3 key prefix for reservation objects
-        :param check_expirations_interval_seconds: How often to check for expired reservations
+        :param check_expirations_interval_seconds: How often to check for expired reservations.
+                                    If 0 or negative, expiration checks are disabled.
         """
+        # Our default for check_expirations_interval_seconds is 0
+        # because S3 expiration check is generally a significant execution load,
+        # and we may want to run it externally on demand rather than on a fixed schedule inside the service.
         super().__init__(storage_name="s3_storage",
                          check_expirations_interval_seconds=check_expirations_interval_seconds)
+        self.logger: Logger = getLogger(self.__class__.__name__)
+
+        # Check if expiration interval is set by environment variable,
+        # and adjust it if so (overriding the constructor parameter)
+        envvar_name: str = "AGENT_RESERVATIONS_EXTERNAL_STORAGE_CHECK_PERIOD_SECONDS"
+        envvar_value: str = os.getenv(envvar_name, "0")
+        try:
+            expiration_check_period_seconds: float = float(envvar_value)
+            self._check_interval_seconds = expiration_check_period_seconds
+        except ValueError as exc:
+            self.logger.error(
+                "Invalid value for %s, must be a number. Got: %s. "
+                "Please correct the environment variable or unset it.",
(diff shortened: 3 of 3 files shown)


neuro_san/internals/graph/activations/calling_activation.py
```diff
@@ -212,9 +212,10 @@ async def make_one_tool_function_call(self, component_tool_call: ToolCall) -> Di
                 "tool_call_id" a string id representing the call to the tool itself
                 "output" a JSON string representing the output of the tool's function
         """
-        # Get the function args as a dictionary
+        # Get the essentials
         tool_name: str = component_tool_call.get_function_name()
         tool_arguments: Dict[str, Any] = component_tool_call.get_function_arguments()
+        invocation: str = component_tool_call.get_invocation()
```
d1donlydfink, line 218: When we get a ToolCall it now comes with an invocation string.
This is so we can pass along (quite a ways!) what we get from the Function() call.

neuro_san/internals/graph/activations/external_activation.py
```diff
@@ -153,7 +156,7 @@ async def build(self) -> BaseMessage:
         if self.session is None:
             invocation_context: InvocationContext = self.run_context.get_invocation_context()
             factory: AsyncAgentSessionFactory = invocation_context.get_async_session_factory()
-            self.session = factory.create_session(self.agent_url, invocation_context)
+            self.session = factory.create_session(self.agent_url, invocation_context, self.invocation)
```
d1donlydfink, line 159: pass along to create_session() so it knows how to treat the events

neuro_san/internals/graph/interfaces/agent_tool_factory.py
```diff
@@ -33,7 +35,9 @@ def create_agent_activation(self,
                                 parent_agent_spec: Dict[str, Any],
                                 name: str,
                                 sly_data: Dict[str, Any],
-                                arguments: Dict[str, Any]) -> CallableActivation:
+                                arguments: Dict[str, Any] = None,
+                                factory: AgentToolFactory = None,
+                                invocation: str = None) -> CallableActivation:
```
d1donlydfink, line 40: Add actual typing and new invocation arg. Had been missing the factory arg too. 🤷

neuro_san/internals/graph/registry/activation_factory.py
```diff
@@ -163,7 +165,7 @@ def create_agent_activation(self, parent_run_context: RunContext,
             allow_from_downstream: Dict[str, Any] = extractor.get("allow.from_downstream", empty)
 
             agent_activation = ExternalActivation(parent_run_context, factory, name, arguments, redacted_sly_data,
-                                                  allow_from_downstream)
+                                                  allow_from_downstream, invocation)
```
d1donlydfink, line 168: pass along the invocation so it knows how to treat the events

neuro_san/internals/graph/registry/agent_tool_registry.py
```diff
@@ -57,10 +59,12 @@ def create_agent_activation(self, parent_run_context: RunContext,
         :param sly_data: A mapping whose keys might be referenceable by agents, but whose
                  values should not appear in agent chat text. Can be an empty dictionary.
         :param arguments: A dictionary of arguments for the newly constructed agent
+        :param factory: A factory that will be used to create the agent tool
+        :param invocation: The invocation style of the activation.
         :return: The CallableActivation agent referred to by the name.
         """
         return self.factory.create_agent_activation(parent_run_context, parent_agent_spec,
-                                                    name, sly_data, arguments, self)
+                                                    name, sly_data, arguments, self, invocation)
```
d1donlydfink, line 67: pass along the invocation so it knows how to treat the events

neuro_san/internals/interfaces/async_agent_session_factory.py
```diff
@@ -24,7 +24,7 @@ class AsyncAgentSessionFactory:
     Creates asynchronous AsyncAgentSessions for external agents.
     """
 
-    def create_session(self, agent_url: str, invocation_context: Any) -> AsyncAgentSession:
+    def create_session(self, agent_url: str, invocation_context: Any, invocation: str = None) -> AsyncAgentSession:
```
d1donlydfink, line 27: This is an interface, so add the extra arg to be picked up by concrete subclasses.

neuro_san/internals/run_context/interfaces/tool_call.py
```diff
@@ -40,3 +40,12 @@ def get_function_name(self) -> str:
...
         raise NotImplementedError
+
+    def get_invocation(self) -> str:
+        """
+        :return: String describing how the tool wants to be invoked.
+                            Can be: "chatbot" - implies waiting for an answer.
+                                    "event" - implies no answer needed
+                                    None - implies chatbot
+        """
+        raise NotImplementedError
```
d1donlydfink, line 51: More interface tweaks for concrete subclasses to fill

neuro_san/session/external_agent_session_factory.py
```diff
@@ -117,7 +130,7 @@ def create_session_from_location_dict(self, agent_location: Dict[str, str],
                     break
 
             agent_network: AgentNetwork = agent_network_provider.get_agent_network()
-            safe_invocation_context: InvocationContext = invocation_context.safe_shallow_copy()
+            safe_invocation_context: InvocationContext = invocation_context.safe_shallow_copy(invocation)
```
d1donlydfink, line 133: This is kind of the endpoint - we are about to call an external agent and so we make a copy of the current invocation context, but now we pass the desired invocation carried all the way from the original Function() result.

neuro_san/session/session_invocation_context.py
```diff
@@ -229,4 +235,8 @@ def safe_shallow_copy(self) -> SessionInvocationContext:
         # to be sure that the messages are sent to the correct queue.
         invocation_context.journal: Journal = MessageJournal(invocation_context.queue)
 
+        # If an invocation was provided, use it
+        if invocation is not None:
+            invocation_context.effective_invocation = invocation
```
d1donlydfink, line 240: Here is where we actually do something with the data from Function()

neuro_san/internals/run_context/langchain/core/langchain_openai_function_tool.py
```diff
@@ -79,6 +79,8 @@ class LangChainOpenAIFunctionTool(BaseTool):
 
     tool_caller: Optional[ToolCaller] = None
 
+    invocation: Optional[str] = None
```
d1donlydfink, line 82: This is where things first get populated from the results of Function()
This class is essentially constructed with the results of Function() (for external networks) as kwargs.
So it "just" gets populated courtesy of Pydantic initialization magic.

neuro_san/internals/run_context/langchain/core/langchain_openai_function_tool.py
```diff
@@ -215,7 +217,7 @@ async def _arun(
             converter = PydanticArgumentDictionaryConverter()
             kwargs = converter.to_dict(kwargs)
 
-            initial_run = LangChainRun("tool_base", [""], self.name, kwargs)
+            initial_run = LangChainRun("tool_base", [""], self.name, kwargs, invocation=self.invocation)
```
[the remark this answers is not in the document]
d1donlydfink: This is where the initialization is being handled by Pydantic kwargs magic. Not going to do take the suggestion here.

neuro_san/internals/run_context/langchain/core/langchain_tool_call.py
```diff
@@ -40,10 +40,17 @@ def __init__(self, tool_name: str, args: Any, run_id: str):
...
+                            Can be: "chatbot" - implies waiting for an answer.
+                                    "event" - implies no answer needed
+                                    None - implies chatbot
         """
         self.tool_name: str = tool_name
         self.args = args
         self.id: str = f"tool_call_{run_id}_{uuid.uuid4()}"
+        self.invocation: str = invocation
+        if invocation is None:
+            self.invocation = "chatbot"
```
andreidenissov-cog, line 54: So is this kinda "end of the line" for invocation parameter going all the way down the stack?
I don't see it really used still?

d1donlydfink: More of a "resting stop" than going all the way down the stack.

What seems to happen is:
* Agent goes through a phase of asking what tools are available and how they will help/be called.
* To do this the `function` of the tools targets are read - either from local agent definition or external agent definition. What we care about here is external agents.
* `invocation` lives on the 'function` specification and during that tool-info collection phase gets stored here.
* Then the agent eventually decides what to actually do after gathering all that info.
* This structure is used for later invocation if agent decides to actually use a tool instance.
* and now it has enough information to know _how_ an external agent/tool wants to be invoked - chatbot or event.

The real end of the line is when the ExternalActivation decides to create a new session to eventually call the set up to ExternalAgentSessionFactory (bottom of PR).  In there the InvocationContext is largely copied, but how to treat the invocation - chatbot or event - of the new copy is now left as an argument to that guy, which comes from this data structure.

d1donlydfink: This actually brings up the question of:
Should we allow per-inner-agent specification of event vs chatbot?

I think for the foreseeable future the answer is no, since the machinery for how to handle event vs chatbot is at a per-request level.

If we get enough good reasons to revisit the inner-invocation switch, we can investigate then.
For now, I think leaving it as a per-network thing moves the ball enough to let people start playing with it.

PR #827, opened 2026-03-31
NS-826 Preserve metadata in Reservations

d1donlydfink:
@dsargent noticed that sample_queries were not coming back in the metadata from reservations.
Long story short: We were writing over the existing metadata when saving to S3.
Don't do that, add to it instead.

### DIFF
--- a/neuro_san/service/watcher/temp_networks/s3_reservations_storage.py
+++ b/neuro_san/service/watcher/temp_networks/s3_reservations_storage.py
@@ -202,10 +202,13 @@ class S3ReservationsStorage(AbstractReservationsStorage):
             # Build complete data structure containing reservation metadata,
             # the associated agent_spec, source information, and storage timestamp
             current_time: float = time.time()
-            agent_spec["metadata"] = {
+            new_metadata: Dict[str, Any] = {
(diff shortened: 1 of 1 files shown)


neuro_san/service/watcher/temp_networks/s3_reservations_storage.py
```diff
@@ -202,10 +202,13 @@ def add_reservations(self, reservations_dict: Dict[Reservation, Any],
...
             # the associated agent_spec, source information, and storage timestamp
             current_time: float = time.time()
-            agent_spec["metadata"] = {
+            new_metadata: Dict[str, Any] = {
                 "reservation": self.converter.to_dict(reservation),  # Serialized reservation object
                 "stored_at": current_time              # When stored in S3
             }
+            if agent_spec.get("metadata") is None:
+                agent_spec["metadata"] = {}
+            agent_spec["metadata"].update(new_metadata)
```
d1donlydfink, line 211: Do an update of the existing metadata dictionary with the new reservations metadata.

andreidenissov-cog (approved):
LGTM.

PR #829, opened 2026-03-31
Reduce noisy logging in coded tool class resolution

### LINKED ISSUE #824: Log when successfully resolve the coded tool class
There is warning when resolving failed but none when it succeed.

Noravee:
Fix #824 

**Update 2026-04-12:**
- Remove `Resolver` logger suppression — no longer needed since upstream leaf-common was updated to only log when `verbose=True`
- Change "Attempting to resolve..." and "Successfully resolved..." log level from `INFO` to `DEBUG`

**Original 2026-04-04:**
## Summary

- Reduce repetitive log noise during class resolution in `AbstractClassActivation` by logging only once at the start ("Attempting to resolve..."), once on success ("Successfully resolved..."), and once on failure (detailed error message)
- Suppress the `Resolver` logger (from `leaf_common`) during intermediate resolution attempts, since it logs at INFO for every failed import path — which is expected behavior during progressive resolution
- Extract resolution logic into `_attempt_resolve()` to separate the resolution strategy from logging concerns

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -15,46 +15,33 @@ jobs:
     runs-on: ubuntu-latest
     environment: pypi  # Match this to your configured GitHub Environment.
     permissions:
-      id-token: write #Mandatory for PyPi Trusted Publishing (configured at pypi.org)
+      id-token: write  # Mandatory for PyPI Trusted Publishing
 
     steps:
+      # v6.0.2
       - name: Checkout code
-        uses: actions/checkout@v4
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
 
-      - name: Set up Python
-        uses: actions/setup-python@v5
+      # 1.0.0
+      - name: Setup Python environment
+        uses: cognizant-ai-lab/build-common/actions/setup-python-env@455db3cc40e9ed6a0fb9bbaf8188888c1ae315c2
         with:
           python-version: '3.10'
-
-      - name: Install build tools
-        run: |
-          pip install --upgrade pip
-          pip install build
+          requirements-file: ''
+          requirements-build-file: ''
+          install-extras: 'build'
 
       - name: Build wheel and sdist
         run: python -m build
 
+      # v1.14.0
       - name: Publish to PyPI
-        uses: pypa/gh-action-pypi-publish@release/v1
+        uses: pypa/gh-action-pypi-publish@cef221092ed1bacb1cc03d23a2d87d1d172e277b
 
-      - name: Notify Slack on success
-        if: success()
-        uses: slackapi/slack-github-action@v1.24.0
+      # 1.0.0
+      - name: Notify Slack
+        if: ${{ !cancelled() }}
+        uses: cognizant-ai-lab/build-common/actions/slack-notify@455db3cc40e9ed6a0fb9bbaf8188888c1ae315c2
         with:
-          payload: |
-            {
-              "text": "✅ *Publish Action Passed* for `${{ github.repository }}` on `${{ github.ref_name }}`"
-            }
-        env:
-          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
-
-      - name: Notify Slack on failure
-        if: failure()
-        uses: slackapi/slack-github-action@v1.24.0
-        with:
-          payload: |
-            {
-              "text": "❌ *Publish Action Failed* for `${{ github.repository }}` on `${{ github.ref_name }}`"
-            }
-        env:
-          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+          status: ${{ job.status }}
+          webhook-url: ${{ secrets.SLACK_WEBHOOK_URL }}
--- a/.github/workflows/smoke.yml
+++ b/.github/workflows/smoke.yml
@@ -12,21 +12,16 @@ jobs:
   smoke-test:
     runs-on: ubuntu-latest
 
-    # ─── Job‑level ENV ─────────────────────────────────────────
-    # every step below will have these in its shell environment
     env:
       AWS_ACCESS_KEY_ID:     ${{ secrets.AWS_ACCESS_KEY_ID }}
       AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
       AWS_REGION:            ${{ vars.AWS_REGION }}
     
-    # 🚀 Ollama Service runs alongside main container
     services:
       ollama:
         image: ollama/ollama
         ports:
           - 11434:11434
-
-    # 🔧 Your actual job container
 
     # don't run this job in sync'd clones
     if: github.repository == 'cognizant-ai-lab/neuro-san'
@@ -34,22 +29,23 @@ jobs:
       image: python:3.12-slim
 
     steps:
+      # v4.3.1
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5
 
       - name: Check for NVIDIA GPU
         shell: bash
         run: |
           build_scripts/check_gpu.sh
       
-      - name: check available memory
+      - name: Check available memory
         shell: bash
         run: |
           build_scripts/check_memory.sh
 
       - name: Install dependencies
         run: |
-          apt-get update && apt-get install -y curl shellcheck
+          apt-get update && apt-get install --yes curl shellcheck
           pip install -r requirements-build.txt
           pip install -r requirements.txt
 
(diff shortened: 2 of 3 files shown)


neuro_san/internals/graph/activations/abstract_class_activation.py
```diff
@@ -266,10 +244,57 @@ def resolve_class(self, class_name: str, module_name: str):
...
+        """
+        Try to resolve the class through progressive resolution strategies.
 
-        return python_class
+        :param class_name: The name of the class to resolve
+        :param module_name: The module name containing the class
+        :param this_agent_tool_path: The root path from AGENT_TOOL_PATH plus the agent network name
+        :param this_agent_tool_path_parts: The agent tool path split into parts
+        :param agent_network_name_parts: The agent network name split into parts
+        :return: The resolved Python class, or None if not found
```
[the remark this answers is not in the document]
Noravee: Done

neuro_san/internals/graph/activations/abstract_class_activation.py
```diff
@@ -195,55 +196,32 @@ def resolve_class(self, class_name: str, module_name: str):
...
-            message = f"""
+        # Suppress the Resolver logger during intermediate attempts to avoid noisy logs.
+        # The Resolver always logs at INFO when it can't find a module, which is normal
+        # behavior during our progressive resolution strategy.
+        resolver_logger: Logger = getLogger("Resolver")
+        original_resolver_level: int = resolver_logger.level
+
+        self.logger.info("Attempting to resolve class `%s` in module `%s`", class_name, module_name)
+
+        resolver_logger.setLevel(WARNING)
```
Noravee: Suppress the Resovler logger during resolving attempt to avoid noisy logs.

neuro_san/internals/graph/activations/abstract_class_activation.py
```diff
@@ -195,55 +196,32 @@ def resolve_class(self, class_name: str, module_name: str):
...
+        try:
+            python_class = self._attempt_resolve(
+                class_name, module_name, this_agent_tool_path, this_agent_tool_path_parts, agent_network_name_parts
+            )
+        except (ValueError, AttributeError) as exception:
+            last_exception: Union[ValueError, AttributeError] = exception
+        finally:
+            resolver_logger.setLevel(original_resolver_level)
```
Noravee: Set the logger level back after finish resolving.

neuro_san/internals/graph/activations/abstract_class_activation.py
```diff
@@ -266,10 +244,57 @@ def resolve_class(self, class_name: str, module_name: str):
         the global module must not have the same name as the agent network.
 3. Is AGENT_TOOL_PATH findable from what is set for your PYTHONPATH?
 """
-            self.logger.error(message)
-            raise ValueError(message) from last_exception
+        self.logger.error(message)
+        raise ValueError(message) from last_exception
+
+    # pylint: disable=too-many-arguments,too-many-positional-arguments
+    def _attempt_resolve(self, class_name: str, module_name: str,
```
Noravee, line 240: Refactor the progressive resolution process.

neuro_san/internals/graph/activations/abstract_class_activation.py
```diff
@@ -195,55 +196,32 @@ def resolve_class(self, class_name: str, module_name: str):
...
+            python_class = self._attempt_resolve(
+                class_name, module_name, this_agent_tool_path, this_agent_tool_path_parts, agent_network_name_parts
+            )
+        except (ValueError, AttributeError) as exception:
+            last_exception: Union[ValueError, AttributeError] = exception
+        finally:
+            resolver_logger.setLevel(original_resolver_level)
+
+        if python_class is not None:
+            self.logger.info("Successfully resolved class `%s` in module `%s`", class_name, module_name)
```
Noravee: Log info when sucessful.

neuro_san/internals/graph/activations/abstract_class_activation.py
```diff
@@ -266,10 +244,57 @@ def resolve_class(self, class_name: str, module_name: str):
         the global module must not have the same name as the agent network.
 3. Is AGENT_TOOL_PATH findable from what is set for your PYTHONPATH?
 """
-            self.logger.error(message)
-            raise ValueError(message) from last_exception
+        self.logger.error(message)
```
Noravee, line 236: Only log error when we really fail to resolve class.

neuro_san/internals/graph/activations/abstract_class_activation.py
```diff
@@ -195,55 +196,32 @@ def resolve_class(self, class_name: str, module_name: str):
...
-                    self.logger.warning(warning_str)
-
-            agent_name: str = self.factory.get_name_from_spec(self.agent_tool_spec)
-            agent_tool_path: str = ".".join(this_agent_tool_path_parts[:-len(agent_network_name_parts)])
-            message = f"""
+        # Suppress the Resolver logger during intermediate attempts to avoid noisy logs.
+        # The Resolver always logs at INFO when it can't find a module, which is normal
+        # behavior during our progressive resolution strategy.
+        resolver_logger: Logger = getLogger("Resolver")
+        original_resolver_level: int = resolver_logger.level
```
d1donlydfink: I don't think this works because the logger for the Resolver is at more of a global scope - it's my understanding of how logging works anyway.
This setting/resetting would interfere with other requests.

Probably the better thing to do would be to pass some kind of arg into the Resolver that says whether we want logging or not (or level) (and it's on/warning by default)

neuro_san/internals/graph/activations/abstract_class_activation.py
dsargent: Messing with the logging level programmatically just seems wrong. Log levels shouldn't be configured at a DevOps level. The application should only decide, for each message, what level to log the message at.

DevOps config: "Only show messages at 'warn' or above"
App: "Log _this_ message at warn. Log _this_ message at info...etc"

I'm guessing you're trying to avoid modifying some downstream library here. But this hacky strategy is not the WAY IMO.

neuro_san/internals/graph/activations/abstract_class_activation.py
dsargent: Feels like more of a `debug` level log.

neuro_san/internals/graph/activations/abstract_class_activation.py
```diff
@@ -195,55 +196,32 @@ def resolve_class(self, class_name: str, module_name: str):
...
+
+        self.logger.info("Attempting to resolve class `%s` in module `%s`", class_name, module_name)
```
dsargent: Feels like should be debug level. This will get extremely chatty, even during normal operation.

neuro_san/internals/graph/activations/abstract_class_activation.py
dsargent: Most deployed systems should cruise at "warning" or "error" level logging, meaning only things at those levels or higher (if there is such a thing as higher) get logged.

Log spamming is an issue; it makes it harder to troubleshoot, not to mention we get charged by volume of logs for DataDog.

neuro_san/internals/graph/activations/abstract_class_activation.py
Noravee: Fixed.

neuro_san/internals/graph/activations/abstract_class_activation.py
Noravee: Fixed.

neuro_san/internals/graph/activations/abstract_class_activation.py
Noravee: Fixed.

neuro_san/internals/graph/activations/abstract_class_activation.py
Noravee: Fixed.

neuro_san/internals/graph/activations/abstract_class_activation.py
```diff
@@ -266,10 +234,57 @@ def resolve_class(self, class_name: str, module_name: str):
...
+            # The Resolver internally catches ImportError (and SyntaxError, etc.) in try_to_import_module()
+            # and converts them into messages. What it raises outward is:
+            # ValueError — when the module can't be found at all
+            # AttributeError — from getattr(found_module, class_name) when the module was found
+            # but the class doesn't exist in it
+            except (ValueError, AttributeError) as exception:
+                last_exception: Union[ValueError, AttributeError] = exception
+
+        # Re-raise the last exception so the caller knows resolution failed
+        raise last_exception
```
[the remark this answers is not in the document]
Noravee: It's a good defensive change. Both parts work together:

The None check — Currently resolve_class_in_module with raise_if_not_found=True (the default) should never return None. But if upstream Resolver behavior changes, returning None would silently propagate as a "successful" resolution and cause confusing errors downstream.

The fallback ValueError — This is a necessary consequence of adding the None check. If all iterations return None without raising, last_exception stays None, so raise last_exception would be a TypeError. The fallback gives a clear message instead.

With the current Resolver, neither path should ever be hit — but it's cheap insurance against upstream changes.

Noravee:
- Remove `Resolver` logger suppression — no longer needed since upstream leaf-common was updated to only log when `verbose=True`
- Change "Attempting to resolve..." and "Successfully resolved..." log level from `INFO` to `DEBUG`
- The refactored code passed the unit test

dsargent (approved):

PR #836, opened 2026-04-01
NS-661 Switch to using event loop for reservations queues processing.

andreidenissov-cog:
This PR switches TNW reservations processing from using fixed-size thread pool to a "regular" async event loop wrapped in our AsyncioExecutor instance. Event loop should provide better capability for scaling up.
Note that we are using no-wait polling on all queues with async sleeps in between to reduce mutual tasks blocking as much as possible.
Tested by 500x5 stress test: 500 agent_cli clients each generating 5 TNW reservations.

### DIFF
--- a/neuro_san/service/watcher/temp_networks/temp_network_storage_updater.py
+++ b/neuro_san/service/watcher/temp_networks/temp_network_storage_updater.py
@@ -26,7 +26,6 @@ from asyncio import sleep as asyncio_sleep
 from asyncio import CancelledError
 from logging import getLogger
 from logging import Logger
-import functools
 import queue as queue_mod
 
 from janus import Queue
@@ -146,8 +145,7 @@ class TempNetworkStorageUpdater(Startable):
                 return
 
             if self.reservationist is not None:
-                queue_processor = functools.partial(self.process_one_queue, async_collating_queue)
-                _ = self.executor.create_task(queue_processor(), "queue_processor")
+                _ = self.executor.create_task(self.process_one_queue(async_collating_queue), "queue_processor")
             else:
                 # We don't have a reservation functionality set up,
                 # so we can't do anything with this queue, but we should still close it to avoid leaks.


neuro_san/service/watcher/temp_networks/temp_network_storage_updater.py
```diff
@@ -106,50 +108,88 @@ def start(self):
...
+                continue
+            except SyncQueueShutDown:
+                # Explicit shutdown signal from Janus queue machinery
+                self.logger.info("SHUTDOWN signal for incoming queue")
+                return
+
             if self.reservationist is not None:
-                self.queue_processing_pool.submit(self.process_one_queue, async_collating_queue)
+                queue_processor = functools.partial(self.process_one_queue, async_collating_queue)
+                _ = self.executor.create_task(queue_processor(), "queue_processor")
```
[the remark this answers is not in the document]
andreidenissov-cog: Fixed.

neuro_san/service/watcher/temp_networks/temp_network_storage_updater.py
```diff
@@ -106,50 +107,87 @@ def start(self):
...
                 # We don't have a reservation functionality set up,
                 # so we can't do anything with this queue, but we should still close it to avoid leaks.
                 self.logger.warning("No reservationist available, closing incoming queue without processing")
                 async_collating_queue.close()
 
-    def process_one_queue(self, async_collating_queue: AsyncCollatingQueue):
+    async def process_one_queue(self, async_collating_queue: AsyncCollatingQueue):
         """
-        Processes a single AsyncCollatingQueue from the pool
+        Processes a single AsyncCollatingQueue
```
d1donlydfink, line 157: Interesting:
* One async task per inner-queue which is only done when the queue is empty.
* Have to ask the sync-q side in an asynchronous task if there's anything to do, otherwise async sleep.

Might be worth having this short assessment in the method comments here.

neuro_san/service/watcher/temp_networks/temp_network_storage_updater.py
```diff
@@ -106,50 +107,87 @@ def start(self):
...
 
         :param async_collating_queue: The AsyncCollatingQueue to process
         """
         janus_queue: Queue = async_collating_queue.get_queue()
+        idle_sleep_seconds: float = 0.5
```
d1donlydfink, line 162: Maybe send this to top-level class constant.
At some point we might want this configurable, but seems fine to start out with a "no" there.

neuro_san/service/watcher/temp_networks/temp_network_storage_updater.py
```diff
@@ -78,16 +80,15 @@ def __init__(self, network_storage_dict: Dict[str, AgentNetworkStorage],
...
 
-            # We use a thread pool to process the collection of tasks
+            # We use an AsyncioExecutor and its event loop to process the collection of tasks
             # this TempNetworkStorageUpdater instance is responsible for:
             # 1. Checking for new queues to process coming in self.incoming "main" queue
             # 2. Processing the individual queues that come in,
             #    which involves processing the items that come in on those queues
             #    and deploying reservations to the temp storage, which is done by self.reservationist.
-            self.queue_processing_pool = ThreadPoolExecutor(self.THREAD_POOL_SIZE)
+            self.executor = AsyncioExecutor()
```
d1donlydfink, line 89: Why assign to None up top when you do a real assignment here?
Seems like top declaration is unnecessary?

d1donlydfink (approved):

PR #838, opened 2026-04-03
Migrate integration.yml to build-common quality gate

donn-leaf:
## Summary

Replaces the inline integration test workflow with a `workflow_call` to the shared `_python-quality-gate.yml` reusable workflow from `build-common` (pinned to SHA `455db3c`, tag 1.0.0). Lint is skipped via `lint-command-override` since the original workflow only ran integration tests. All environment variables (`AGENT_TOOL_PATH`, `PYTHONPATH`, `AGENT_MANIFEST_FILE`) and secrets (`OPENAI_API_KEY`, `SLACK_WEBHOOK_URL`) are preserved via `test-command-override` and the `secrets:` block.

### Updates since last revision

- **Reverted all `server_start.sh` changes**: The `sudo` addition broke `smoke.yml` which runs inside `python:3.12-slim` (root, no `sudo`). Per reviewer feedback, the startup script is left untouched — only the `integration.yml` migration remains in this PR.

## Review & Testing Checklist for Human

- [ ] **Manual trigger verification (critical)**: The `integration.yml` workflow only fires on `schedule` and `workflow_dispatch`, so normal PR CI will **not** exercise it. Before merging, trigger a manual run on this branch via Actions → Integration → Run workflow to verify the integration tests pass end-to-end with the new quality gate.
- [ ] **Container → ubuntu-latest**: The old integration workflow ran inside `python:3.12-slim`; the quality gate uses `ubuntu-latest` with `actions/setup-python`. Verify integration tests do not depend on Debian-slim-specific behavior or system packages.
- [ ] **OPENAI_API_KEY availability**: Confirm the secret is injected into the `test-command-override` step by the quality gate. Integration tests will fail with auth errors if this is missing.

### Notes
- `pytest -s` was replaced with `pytest --capture=no` (equivalent, long-form flag convention).
- `lint-command-override: 'true'` runs a shell no-op to skip all lint steps cleanly; the original workflow had no lint.
- Slack notifications are now handled by the quality gate's built-in `slack-notify` action (different message format than the old emoji-prefixed payloads).
- The `pip freeze` debug step from the old workflow is not replicated.

Link to Devin session: https://app.devin.ai/sessions/ea86f39fe4a444329f017c796a09a5bb
Requested by: @donn-leaf

---

<a href="https://app.devin.ai/review/cognizant-ai-lab/neuro-san/pull/838" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open with Devin
  </picture>
</a>

### DIFF
--- a/docs/agent_hocon_reference.md
+++ b/docs/agent_hocon_reference.md
@@ -25,6 +25,7 @@ Items in ***bold*** are essentials. Try to understand these first.
         - [fallbacks](#fallbacks)
         - [temperature](#temperature)
         - [Other LLM-specific Parameters](#other-llm-specific-parameters)
+        - [Client-Provided API Keys](#client-provided-api-keys)
     - [***tools*** - list of agent/tool definitions](#tools)
     - [commondefs](#commondefs)
         - [replacement_strings](#replacement_strings)
@@ -50,6 +51,7 @@ Items in ***bold*** are essentials. Try to understand these first.
             - [properties](#properties)
             - [required](#required)
         - [sly_data_schema](#sly_data_schema)
+            - [llm_config](#llm_config-1)
         - [sly_data_output_schema](#sly_data_output_schema)
     - [***instructions*** - main system prompt for the agent](#instructions)
     - [***tools*** - list of other agents/tools that this agent may access](#tools-agents)
@@ -57,7 +59,7 @@ Items in ***bold*** are essentials. Try to understand these first.
         - [MCP Servers](#mcp-servers)
             - [Authentication](#authentication)
     - [***class*** - Python class name to invoke for Coded Tools](#class-1)
-    - [llm_config - agent-specific LLM configuration](#llm_config-1)
+    - [llm_config - agent-specific LLM configuration](#llm_config-2)
     - [command](#command)
     - [toolbox](#toolbox)
     - [args](#args)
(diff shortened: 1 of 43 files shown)


.github/workflows/integration.yml
```diff
@@ -12,50 +12,17 @@ jobs:
...
-        if: failure()
-        uses: slackapi/slack-github-action@v1.24.0
-        with:
-          payload: |
-            {
-              "text": "❌ *Integration Tests Failed* for `${{ github.repository }}` on `${{ github.ref_name }}`"
-            }
-        env:
-          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+    # 1.0.0
```
d1donlydfink, line 15: What does this version number refer to?
It's not clear from the context. Perhaps worth a code comment with a link to where this number comes from.

donn-leaf: The version number refers to the version of the github action in the "uses:" statement. In that statement, we use the git sha to pin _exactly_ to the desired version. Why you may ask? To avoid CodeQL warnings (see below).

We _could_ add the version as a comment at the end of the line, but the line is already too long.
We could also just not comment the version. What's your preference here @d1donlydfink 

3. CodeQL Alerts for Unpinned Actions 
CodeQL includes a specific security query, [actions-unpinned-tag](https://codeql.github.com/codeql-query-help/actions/actions-unpinned-tag/), which flags workflows using tags instead of SHAs. 

    Flag Purpose: To mitigate supply chain attacks where a malicious actor compromises an action repository and updates an existing tag to point to malicious code.
    Resolution: Replace the tag with a SHA as described above to clear the CodeQL alert.

PR #843, opened 2026-04-06
Fix setting up agent client timeout.

andreidenissov-cog:
This PR fixes timeout setting for our async client streaming http connection.
Previously, lower-level default timeout has been silently used,
so instead of "timeout never" we had "aiohttp connection will timeout after 300 seconds".
Found and fix is tested with Agent Network Designer stress-test.

### DIFF
--- a/neuro_san/session/async_http_service_agent_session.py
+++ b/neuro_san/session/async_http_service_agent_session.py
@@ -108,6 +108,9 @@ class AsyncHttpServiceAgentSession(AbstractHttpServiceAgentSession, AsyncAgentSe
             # To specify complete timeout value, we must use "total" parameter of ClientTimeout.
             # See https://docs.aiohttp.org/en/stable/client_reference.html#aiohttp.ClientTimeout for details.
             timeout: ClientTimeout = ClientTimeout(total=None)
+            # That will make sure that the connection will stay open until the (last) result is yielded,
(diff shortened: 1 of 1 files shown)


neuro_san/session/async_http_service_agent_session.py
```diff
@@ -105,9 +105,14 @@ async def streaming_chat(self, request_dict: Dict[str, Any]) -> Generator[Dict[s
...
-            timeout: ClientTimeout = None
+            # To specify complete timeout value, we must use "total" parameter of ClientTimeout.
+            # See https://docs.aiohttp.org/en/stable/client_reference.html#aiohttp.ClientTimeout for details.
+            timeout: ClientTimeout = ClientTimeout(total=None)
+            # That will make sure that the connection will stay open until the (last) result is yielded,
+            # which is what we want here.
+            # Not specifying "total" parameter will invoke lower-level aiohttp timeout, which is 300 seconds by default
             if self.streaming_timeout_in_seconds is not None:
-                timeout = ClientTimeout(self.streaming_timeout_in_seconds)
+                timeout = ClientTimeout(total=self.streaming_timeout_in_seconds)
```
d1donlydfink, line 115: Nice find!

d1donlydfink (approved):

PR #860, opened 2026-04-10
NS-790 sly_data user API keys for llm_config

d1donlydfink:
@Noravee is primary.

Allow the client to specify a special "llm_config" dictionary in the sly_data.
This dictionary has key/value pairs that correspond to llm_config keys (like "openai_api_key" and/or "anthropic_api_key") whose values are substituted in for any llm_config value that is "sly_data".

Example hocon snippet for llm_config.

```
    "llm_config": {
        "model_name": "gpt-5.2",

        # Indicates a user-specified OpenAI API key is required in order to operate.
        # If this key is not provided, the agent will not be able to operate.
        # If you have a list of fallbacks, you can specify that each of their models'
        # api keys are required as well.
        # You can also specify any other llm_config parameter should be specified in the
        # llm_config structure within sly_data, however these cases are much less common.
        "openai_api_key": "sly_data"
    },
```
... it should be possible to do this in the shared llm_config.hocon making all networks require user-provided API keys.

Later on in the front-man function spec you can tell the client that we really need their openai_api_key in order to function:

```
                # You can also optionally specify the schema needed for any sly_data input.
                # This is to allow generic clients to prompt for private-data inputs
                # that do not belong in the chat stream.
                "sly_data_schema": {
                    "type": "object",
                    "properties": {
                        # The following is an example of how to specify that sly_data is required
                        # to provide a single API key from the user for OpenAI.
                        # Concievably, this could be expanded to include multiple API keys,
                        # each of which may or may not be required for fallbacks, in which case
                        # the "required" list on this inner structure would be optional, as long
                        # as each of the API keys listed are themselves optional.
                        "llm_config": {
                            "type": "object",
                            "properties": {
                                "openai_api_key": {
                                    "type": "string",
                                    "description": "The User's OpenAI API key"
                                }
                            },
                            "required": ["openai_api_key"]
                        }
                    },
                    # Tell the client that llm_config is required.
                    "required": ["llm_config"]
                },
```

When the client supplies sly_data like this:

```
{
    "llm_config": {
        "openai_api_key": "<user-specific-api-key>"
    }
}
```
... everything functions nicely but _gets billed to the user's key_.

If this is not supplied, an error message in the test tells what is missing:
```
No fully-specified LLM found in llm_config or fallbacks.
LLM operation for this agent requires at least one of the following set in sly_data.llm_config:
openai_api_key
```

Feel free to chime in if y'all see some problems carrying forward with this tacit sly_data approach that I don't.

### DIFF
--- a/docs/agent_hocon_reference.md
+++ b/docs/agent_hocon_reference.md
@@ -235,14 +235,14 @@ see the [llm_info_hocon_reference](./llm_info_hocon_reference.md).
 
 #### Client-Provided API Keys
 
-It is possible to set any one of the api_keys described above as "sly_data" in the llm_config.
+It is possible to set any one of the API keys described above as "sly_data" in the llm_config.
 This will cause the system to look at the sly_data's llm_config dictionary for the appropriate
 api key(s) depending on the model provider. Note that key names are all-smalls versions of any
 environment variable names above.
 
-See also: [api_keys](#api_keys) in the [sly_data_schema](#sly_data_schema) section.
+See also: [llm_config](#llm_config-1) in the [sly_data_schema](#sly_data_schema) section.
 
-Example networks that advertise that their sly_data_schema needs external api_keys:
+Example networks that advertise that their sly_data_schema needs external API keys:
 
 - [music_nerd_pro_sly_api_key.hocon](../neuro_san/registries/music_nerd_pro_sly_api_key.hocon)
 
@@ -517,7 +517,7 @@ by having the user fill in their own API keys.
 
 Any key in this dictionary will be specific to a particular LLM (i.e. "openai_api_key" or "anthropic_api_key").
 
-Example networks that advertise that their sly_data_schema needs external api_keys:
+Example networks that advertise that their sly_data_schema needs external API keys:
 
 - [music_nerd_pro_sly_api_key.hocon](../neuro_san/registries/music_nerd_pro_sly_api_key.hocon)
 
--- a/docs/agent_hocon_reference.md
+++ b/docs/agent_hocon_reference.md
@@ -505,6 +505,7 @@ Example networks that advertise their sly_data_schema:
 
 There are select few tacit conventions supported for certain sly_data values that need to come from the client:
 
+<!--- pyml disable-next-line no-duplicate-heading -->
 ##### llm_config
 
 The sly_data dictionary can contain an optional `llm_config` key whose value is a dictionary
--- a/docs/agent_hocon_reference.md
+++ b/docs/agent_hocon_reference.md
@@ -23,9 +23,9 @@ Items in ***bold*** are essentials. Try to understand these first.
         - [***model_name*** - specifies the name of the default LLM to use](#model_name)
         - [class](#class)
         - [fallbacks](#fallbacks)
-        - [Client-Provided API Keys](#client-provided-api-keys)
         - [temperature](#temperature)
         - [Other LLM-specific Parameters](#other-llm-specific-parameters)
+        - [Client-Provided API Keys](#client-provided-api-keys)
     - [***tools*** - list of agent/tool definitions](#tools)
     - [commondefs](#commondefs)
         - [replacement_strings](#replacement_strings)
@@ -51,7 +51,7 @@ Items in ***bold*** are essentials. Try to understand these first.
             - [properties](#properties)
             - [required](#required)
         - [sly_data_schema](#sly_data_schema)
-            - [api_keys](#api_keys)
+            - [llm_config](#llm_config-1)
         - [sly_data_output_schema](#sly_data_output_schema)
     - [***instructions*** - main system prompt for the agent](#instructions)
     - [***tools*** - list of other agents/tools that this agent may access](#tools-agents)
@@ -59,7 +59,7 @@ Items in ***bold*** are essentials. Try to understand these first.
         - [MCP Servers](#mcp-servers)
             - [Authentication](#authentication)
     - [***class*** - Python class name to invoke for Coded Tools](#class-1)
-    - [llm_config - agent-specific LLM configuration](#llm_config-1)
+    - [llm_config - agent-specific LLM configuration](#llm_config-2)
     - [command](#command)
     - [toolbox](#toolbox)
     - [args](#args)
@@ -235,9 +235,10 @@ see the [llm_info_hocon_reference](./llm_info_hocon_reference.md).
 
 #### Client-Provided API Keys
 
-It is possible to set any one of the api_keys described above as "required" in the llm_config.
-This will cause the system to look at the sly_data for the appropriate api key(s) depending on
-the model provider.
+It is possible to set any one of the api_keys described above as "sly_data" in the llm_config.
+This will cause the system to look at the sly_data's llm_config dictionary for the appropriate
+api key(s) depending on the model provider. Note that key names are all-smalls versions of any
+environment variable names above.
 
 See also: [api_keys](#api_keys) in the [sly_data_schema](#sly_data_schema) section.
 
@@ -504,11 +505,11 @@ Example networks that advertise their sly_data_schema:
 
 There are select few tacit conventions supported for certain sly_data values that need to come from the client:
 
-##### api_keys
+##### llm_config
 
-The sly_data dictionary can contain an optional `api_keys` key whose value is a dictionary
+The sly_data dictionary can contain an optional `llm_config` key whose value is a dictionary
 containing per-user API keys to be filled in when an agent's llm_config has certain values
-specified as "required" for its values.
+specified as "sly_data" for its values.
 
 This enables any or all agent networks to potentially offload the costs operation onto the user
 by having the user fill in their own API keys.
--- a/docs/agent_hocon_reference.md
+++ b/docs/agent_hocon_reference.md
@@ -513,7 +513,7 @@ specified as "required" for its values.
 This enables any or all agent networks to potentially offload the costs operation onto the user
 by having the user fill in their own API keys.
 
-Any key in this dictionary will be specific to an particular LLM (i.e. "openai_api_key" or "anthropic_api_key").
+Any key in this dictionary will be specific to a particular LLM (i.e. "openai_api_key" or "anthropic_api_key").
 
 Example networks that advertise that their sly_data_schema needs external api_keys:
 
--- a/docs/agent_hocon_reference.md
+++ b/docs/agent_hocon_reference.md
@@ -23,6 +23,7 @@ Items in ***bold*** are essentials. Try to understand these first.
         - [***model_name*** - specifies the name of the default LLM to use](#model_name)
         - [class](#class)
         - [fallbacks](#fallbacks)
+        - [Client-Provided API Keys](#client-provided-api-keys)
         - [temperature](#temperature)
         - [Other LLM-specific Parameters](#other-llm-specific-parameters)
     - [***tools*** - list of agent/tool definitions](#tools)
@@ -50,6 +51,7 @@ Items in ***bold*** are essentials. Try to understand these first.
             - [properties](#properties)
             - [required](#required)
         - [sly_data_schema](#sly_data_schema)
+            - [api_keys](#api_keys)
         - [sly_data_output_schema](#sly_data_output_schema)
     - [***instructions*** - main system prompt for the agent](#instructions)
     - [***tools*** - list of other agents/tools that this agent may access](#tools-agents)
@@ -195,7 +197,7 @@ you will need to use a model that has been specifically trained for "tool use" f
 branches off work to any other agent/tool.  You can browse the `capabilities` section of the
 `default_llm_info.hocon` to be sure the llm you choose can use tools.
 
-You will need your own access key set as an environment variable in order
+The most common situation is one where you will need your own access key set as an environment variable in order
 to use LLMs from various providers.
 
 | LLM Provider  | API Key environment variable                   |
@@ -230,6 +232,18 @@ you can still use it by specifying the [class](#class) key directly, or by exten
 
 For complete information on adding your own llm models or providers to the default llm info,
 see the [llm_info_hocon_reference](./llm_info_hocon_reference.md).
+
+#### Client-Provided API Keys
+
+It is possible to set any one of the api_keys described above as "required" in the llm_config.
+This will cause the system to look at the sly_data for the appropriate api key(s) depending on
+the model provider.
+
+See also: [api_keys](#api_keys) in the [sly_data_schema](#sly_data_schema) section.
+
+Example networks that advertise that their sly_data_schema needs external api_keys:
+
+- [music_nerd_pro_sly_api_key.hocon](../neuro_san/registries/music_nerd_pro_sly_api_key.hocon)
 
 #### temperature
 
@@ -477,7 +491,7 @@ schema definition above.  Ideally there should be one [properties](#properties) 
 sly_data dictionary input key, and any absolutely necessary keys should be listed in the [required](#required)
 list.
 
-Note that it is not strictly necessary to advertise to the outside world the sly_data_schema that
+Note that it is not always strictly necessary to advertise to the outside world the sly_data_schema that
 your agent network requires, but doing so does allow generic clients to prompt for this extra
 information before sending any chat input.
 
@@ -487,6 +501,23 @@ definition, as sly_data itself is already visible to all other internal agents o
 Example networks that advertise their sly_data_schema:
 
 - [math_guy.hocon](../neuro_san/registries/math_guy.hocon)
+
+There are select few tacit conventions supported for certain sly_data values that need to come from the client:
+
+##### api_keys
+
+The sly_data dictionary can contain an optional `api_keys` key whose value is a dictionary
+containing per-user API keys to be filled in when an agent's llm_config has certain values
+specified as "required" for its values.
+
+This enables any or all agent networks to potentially offload the costs operation onto the user
+by having the user fill in their own API keys.
+
+Any key in this dictionary will be specific to an particular LLM (i.e. "openai_api_key" or "anthropic_api_key").
+
+Example networks that advertise that their sly_data_schema needs external api_keys:
+
+- [music_nerd_pro_sly_api_key.hocon](../neuro_san/registries/music_nerd_pro_sly_api_key.hocon)
 
 #### sly_data_output_schema
 
--- a/neuro_san/client/agent_cli.py
+++ b/neuro_san/client/agent_cli.py
@@ -323,14 +323,15 @@ All choices require an agent name.
         # How do we handle calls to external agents?
         group = arg_parser.add_argument_group(title="Local External Agents",
                                               description="How do handle calls to local /external agents?")
-        group.add_argument("--local_externals_direct", default=False, action="store_true",
+        # Set the default to True so that external tools are used on directly connection by default.
+        group.add_argument("--local_externals_direct", default=True, action="store_true",
                            help="""
 Have external tools that can be found in the local agent manifest use a
-direct connection instead of requiring a service to be stood up.
+direct connection instead of requiring a service to be stood up. (The default)
                            """)
         group.add_argument("--local_externals_service", dest="local_externals_direct", action="store_false",
                            help="""
-Have external tools that can be found in the local agent manifest use a service connection. (The default)
+Have external tools that can be found in the local agent manifest use a service connection.
                            """)
         self.arg_groups[group.title] = group
 
--- a/neuro_san/client/agent_cli.py
+++ b/neuro_san/client/agent_cli.py
@@ -209,8 +209,8 @@ Some suggestions:
 
             print(f"\nResponse from {state.get('origin_str')}:")
             print(f"{state.get('last_chat_response')}")
-            if state.get("sly_data") is not None:
-                pretty_sly: str = json.dumps(state.get('sly_data'), indent=4, sort_keys=True)
+            if state.get("returned_sly_data") is not None:
+                pretty_sly: str = json.dumps(state.get('returned_sly_data'), indent=4, sort_keys=True)
                 print(f"Returned sly_data is: {pretty_sly}")
 
             if self.args.tokens:
--- a/neuro_san/client/streaming_input_processor.py
+++ b/neuro_san/client/streaming_input_processor.py
@@ -124,6 +124,7 @@ class StreamingInputProcessor:
             "last_chat_response": last_chat_response,
             "user_input": None,
             "sly_data": sly_data,
+            "returned_sly_data": returned_sly_data,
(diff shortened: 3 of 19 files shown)


neuro_san/client/agent_cli.py
```diff
@@ -209,8 +209,8 @@ def loop_until_done(self, user_input: str, sly_data: Dict[str, Any],
 
             print(f"\nResponse from {state.get('origin_str')}:")
             print(f"{state.get('last_chat_response')}")
-            if state.get("sly_data") is not None:
-                pretty_sly: str = json.dumps(state.get('sly_data'), indent=4, sort_keys=True)
+            if state.get("returned_sly_data") is not None:
+                pretty_sly: str = json.dumps(state.get('returned_sly_data'), indent=4, sort_keys=True)
```
d1donlydfink, line 213: Was getting mixed signals from the client because it was spitting out that the returned sly_data actually had the api_keys in it, which it did not upon low-level inspection.  What had been happening is a report of a mix of old and returned sly_data, but it was advertising the output as what was returned.
This allows the state to be kept to report the right thing for return.

neuro_san/client/streaming_input_processor.py
```diff
@@ -124,6 +124,7 @@ def process_once(self, state: Dict[str, Any]) -> Dict[str, Any]:
             "last_chat_response": last_chat_response,
             "user_input": None,
             "sly_data": sly_data,
+            "returned_sly_data": returned_sly_data,
```
d1donlydfink, line 127: Add the new returned_sly_data field to the state.

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
```diff
@@ -204,25 +204,54 @@ def create_agent_with_fallbacks(self, instructions: str) -> Runnable:
...
+            if one_llm_resources is None:
+                # Nothing to use or report.
+                # Skip for now, a fallback might still be fulfilled.
+                continue
+
+            if isinstance(one_llm_resources, str):
+                # Report later on which required api_keys are missing
+                # Skip for now, a fallback might still be fulfilled.
+                required_api_keys.append(one_llm_resources)
+                continue
```
d1donlydfink, line 230: Accumulate complaints

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
```diff
@@ -204,25 +204,54 @@ def create_agent_with_fallbacks(self, instructions: str) -> Runnable:
...
 
-            if index == 0:
-                # The first agent is the one we want to be our main guy.
+            if first_llm:
+                # The first fully-specified agent is the one we want to be our main guy.
                 agent = one_agent
                 # For now. Could be problems with different providers w/ token counting.
                 self.llm_resources = one_llm_resources
+                # Anything that comes later will not be the first
+                first_llm = False
```
d1donlydfink, line 240: Now, it's not just index == 0 that could be the first LLM to add to the agent.
We could have a list of fallbacks, say 1st OpenAI, 2nd Anthropic, and the user only specified an Anthropic API key.

What we really want is the first fully-specified LLM in the fallbacks list.

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
```diff
@@ -204,25 +204,54 @@ def create_agent_with_fallbacks(self, instructions: str) -> Runnable:
...
             else:
                 # Anything later than the first guy is considered a fallback. Add it to the list.
                 chain_fallbacks.append(one_agent)
 
+        if agent is None:
+            error: str = "No fully-specified LLM found in llm_config or fallbacks."
+            if len(required_api_keys) > 0:
+                error += "\nLLM operation for this agent requires at least one of the following set in sly_data:\n"
+                error += ','.join(required_api_keys)
+            raise ValueError(error)
```
d1donlydfink, line 251: Spit out an error saying that nothing was fulfilled to spec to get the LLM going.
If we have information as to what api_keys were missing, send that as well, as we are expecting the client to provide at least some of this information.

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
```diff
@@ -161,36 +161,49 @@ def resolve_one_llm_factory(self, llm_factory_class_name: str, llm_info_file: st
...
                 a BaseLanguageModel (can be Chat or LLM) and all related resources
                 necessary for managing the model run-time lifecycle.
+                Can also return a string describing any required api_keys that are not provided.
                 Can raise a ValueError if the config's class or model_name value is
                 unknown to this method.
         """
-        full_config: Dict[str, Any] = self.create_full_llm_config(config)
+        full_config: Dict[str, Any] = self.create_full_llm_config(config, api_keys)
+        if full_config is None or isinstance(full_config, str):
+            return full_config
```
d1donlydfink, line 183: Return early if we got nothin'.

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
```diff
@@ -206,19 +219,23 @@ def create_full_llm_config(self, config: Dict[str, Any]) -> Dict[str, Any]:
 
             # Merge the defaults from llm_info with the user-defined config,
             # giving priority to values in config.
-            return self.overlayer.overlay(config_from_class_in_llm_info, config)
+            full_config = self.overlayer.overlay(config_from_class_in_llm_info, config)
+
+            # Get any required api keys into the full config.
+            full_config = self.replace_any_required_api_keys(full_config, api_keys)
+            return full_config
```
d1donlydfink, line 224: Call the new replace_any_required_api_keys() method, even on this default llm case.

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
```diff
@@ -206,19 +219,23 @@ def create_full_llm_config(self, config: Dict[str, Any]) -> Dict[str, Any]:
...
 
         default_config: Dict[str, Any] = self.llm_infos.get("default_config")
-        use_config = self.overlayer.overlay(default_config, config)
+        use_config: Dict[str, Any] = self.overlayer.overlay(default_config, config)
```
d1donlydfink, line 227: Add some type hints, here and below.

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
```diff
@@ -278,6 +298,45 @@ def get_chat_class_args(self, chat_class_name: str, use_model_name: str = None)
...
+        Get any required api keys into the config.
+        :param config: The fully specified llm config which is a product of
+                    _create_full_llm_config() above.
+        :param api_keys: A dictionary of api keys to replace in the config.
+            The keys in this dictionary are the names of the API keys (ala "OPENAI_API_KEY").
+            The values for the keys are the API keys themselves to be inserted into any llm configuration.
+            Can be None indiciating no API keys are provided at all and the system defaults will be used.
+        :return: The config with any required api keys replaced.
+            Can return None if any of the required api keys are not provided.
+        """
```
d1donlydfink, line 313: New method to complete the llm_config entries marked "required" with values provided in the api_keys dictionary.

neuro_san/registries/manifest.hocon
```diff
@@ -32,7 +32,8 @@
     "math_guy.hocon": true,
     "math_guy_passthrough.hocon": true,
     "music_nerd.hocon": true,
-    "music_nerd_pro_sly.hocon":true,
+    "music_nerd_pro_sly.hocon": true,
+    "music_nerd_pro_sly_api_key.hocon": true,
```
d1donlydfink, line 36: Add our new network. Fix the spacing in the entry above.

neuro_san/registries/music_nerd_pro_sly_api_key.hocon
```diff
@@ -0,0 +1,144 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# END COPYRIGHT
+{
+    # Optional metadata describing this agent network
+    "metadata": {
```
d1donlydfink, line 18: Riffed on from music_nerd_pro.

neuro_san/registries/music_nerd_pro_sly_api_key.hocon
```diff
@@ -0,0 +1,144 @@
...
+                # that do not belong in the chat stream.
+                "sly_data_output_schema": {
+                    "type": "object",
+                    "properties": {
+                        "running_cost": {
+                            "type": "float",
+                            "description": "The running cost of the operation for fake billing purposes"
+                        }
+                    },
+                    "required": ["running_cost"]
```
d1donlydfink, line 104: Also worth noting that we don't need to send the api_keys back to the client.
We are merely a consumer.

tests/fixtures/music_nerd_pro_sly_api_key/no_user_api_key.hocon
```diff
@@ -0,0 +1,63 @@
...
+    
+                    # Gist says we are going to ask an LLM about how the response
+                    # matches up to our description of acceptance criteria.
+                    # Each component of the list needs to pass muster in order to pass.
+                    "gist": ["""
+The text sample should say:
+1) that no fully-specified LLM was found
+AND
+2) list what should be set in sly_data
+"""]
```
d1donlydfink, line 58: Test that we get a message from the system back complaining as to how no api_keys were sent.

tests/fixtures/music_nerd_pro_sly_api_key/with_user_api_key.hocon
```diff
@@ -0,0 +1,86 @@
...
+    "interactions": [
+        {
+            # This is what we send as input to streaming_chat()
+            "text": "Who did Yellow Submarine?",
+
+            "sly_data": {
+                "api_keys": {
+                    "openai_api_key": ${OPENAI_API_KEY},
+                }
+            },
```
d1donlydfink, line 46: Send the api_keys as sly_data in this test.

The rest of the test is a standard Beatles interaction.  It should pass because the api_keys are properly configured.

tests/neuro_san/zzz_hocons/test_unit_test_hocons.py
```diff
@@ -44,6 +44,8 @@ class TestUnitTestHocons(TestCase):
         "math_guy/forwarded_sly_data.hocon",
         "math_guy/deep_forwarded_sly_data.hocon",
         "music_nerd/beatles_with_history.hocon",
+        "music_nerd_pro_sly_api_key/no_user_api_key.hocon",
+        "music_nerd_pro_sly_api_key/with_user_api_key.hocon",
```
d1donlydfink, line 48: Add the 2 new tests for client-provided api_keys.

docs/agent_hocon_reference.md
```diff
@@ -231,6 +233,18 @@ you can still use it by specifying the [class](#class) key directly, or by exten
...
+
+It is possible to set any one of the api_keys described above as "required" in the llm_config.
+This will cause the system to look at the sly_data for the appropriate api key(s) depending on
+the model provider.
+
+See also: [api_keys](#api_keys) in the [sly_data_schema](#sly_data_schema) section.
+
+Example networks that advertise that their sly_data_schema needs external api_keys:
+
+- [music_nerd_pro_sly_api_key.hocon](../neuro_san/registries/music_nerd_pro_sly_api_key.hocon)
```
d1donlydfink, line 247: Add some docs about client-provided api keys.

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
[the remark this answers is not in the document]
d1donlydfink: Addressed.

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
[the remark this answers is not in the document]
d1donlydfink: Addressed.

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
```diff
@@ -204,25 +205,54 @@ def create_agent_with_fallbacks(self, instructions: str) -> Runnable:
...
+        first_llm: bool = True
+        required_api_keys: Set[str] = set()
 
         # Go through the list of fallbacks in the config.
-        for index, fallback in enumerate(fallbacks):
+        for fallback in fallbacks:
 
             # Create a model we might use.
-            one_llm_resources: LangChainLlmResources = llm_factory.create_llm(fallback)
+            one_llm_resources: LangChainLlmResources | Set[str] = llm_factory.create_llm(fallback, api_keys)
```
d1donlydfink: We could get an LlmResources object, or now we could get a set of strings complaining about what was not fulfilled in the llm_config by the user.

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
```diff
@@ -161,36 +161,51 @@ def resolve_one_llm_factory(self, llm_factory_class_name: str, llm_info_file: st
 
         return llm_factory
 
-    def create_llm(
-            self,
-            config: Dict[str, Any]
-    ) -> LangChainLlmResources:
+    def create_llm(self, config: Dict[str, Any], api_keys: Dict[str, Any] = None) -> LangChainLlmResources | Set[str]:
```
d1donlydfink: We could returnan LlmResources object, or now we could return a set of strings complaining about what was not fulfilled in the llm_config by the user.

neuro_san/internals/interfaces/context_type_llm_factory.py
```diff
@@ -53,13 +53,18 @@ def load(self):
...
         :param config: The fully specified llm config from which the LLM instance
                     should be created.
+        :param api_keys: A dictionary of user API keys to use for user billing.
+                The keys in this dictionary are the names of the API keys (ala "OPENAI_API_KEY").
+                The values for the keys are the API keys themselves to be inserted into any llm configuration.
+                Can be None indiciating no API keys are provided at all and the system defaults will be used.
         :return: An llm instance native to the context type.
                 Can raise a ValueError if the config's class or model_name value is
                 unknown to this method.
+                Can return None if required api_keys are not provided.
```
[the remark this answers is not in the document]
d1donlydfink: Addressed.

neuro_san/internals/interfaces/context_type_llm_factory.py
```diff
@@ -53,13 +53,18 @@ def load(self):
...
-    def create_llm(self, config: Dict[str, Any]) -> Any:
+    def create_llm(self, config: Dict[str, Any], api_keys: Dict[str, Any] = None) -> Any:
         """
         Create an llm instance BaseLanguageModel from the fully-specified llm config.
         :param config: The fully specified llm config from which the LLM instance
                     should be created.
+        :param api_keys: A dictionary of user API keys to use for user billing.
+                The keys in this dictionary are the names of the API keys (ala "openai_api_key").
+                The values for the keys are the API keys themselves to be inserted into any llm configuration.
+                Can be None indiciating no API keys are provided at all and the system defaults will be used.
```
Noravee: Typo: indiciating.

Noravee (dismissed):
LGTM.

dsargent:
One thing I'm not clear about: is it possible to _override_? For example, an agent is configured to use GPT-5, but I want to use GPT-4o. Can I use `llm_config` to override the value configured on the server? This would be handy for clients. Or is it that if the config is set as "mandatory", the client has to supply it and cannot fall back to what is configured on the server?

neuro_san/internals/interfaces/context_type_llm_factory.py
```diff
@@ -53,13 +53,17 @@ def load(self):
         """
         raise NotImplementedError
 
-    def create_llm(self, config: Dict[str, Any]) -> Any:
+    def create_llm(self, config: Dict[str, Any], sly_data: Dict[str, Any] = None) -> Any:
```
d1donlydfink, line 56: Allow sly_data to be sent to augment the config of the LLM

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
```diff
@@ -161,36 +161,47 @@ def resolve_one_llm_factory(self, llm_factory_class_name: str, llm_info_file: st
 
         return llm_factory
 
-    def create_llm(
-            self,
-            config: Dict[str, Any]
-    ) -> LangChainLlmResources:
+    def create_llm(self, config: Dict[str, Any], sly_data: Dict[str, Any] = None) -> LangChainLlmResources | Set[str]:
```
d1donlydfink, line 164: Allow sly_data to be sent to augment the config of the LLM

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
```diff
@@ -238,12 +253,15 @@ def create_full_llm_config(self, config: Dict[str, Any]) -> Dict[str, Any]:
...
         # config we are going to use.
-        full_config: Dict[str, Any] = self.overlayer.overlay(default_config, config)
+        full_config = self.overlayer.overlay(default_config, config)
         full_config["class"] = chat_class_name
         full_config["model_name"] = llm_entry.get("use_model_name", use_model_name)
 
-        # Attempt to get a max_tokens through calculation
-        full_config["max_tokens"] = self.get_max_prompt_tokens(full_config)
+        # Get any required api keys into the full config.
+        full_config = self.replace_any_required_api_keys(full_config, sly_data)
```
d1donlydfink, line 261: Call the new replace_any_required_api_keys() method.

neuro_san/registries/music_nerd_pro_sly_api_key.hocon
```diff
@@ -0,0 +1,144 @@
...
+    "llm_config": {
+        "model_name": "gpt-5.2",
+
+        # Indicates a user-specified OpenAI API key is required in sly_data in order to operate.
+        # If this key is not provided, the agent will not be able to operate.
+        # If you have a list of fallbacks, you can specify that each of their models'
+        # api keys are required as well.
+        # You can also specify any other llm_config parameter should be specified in the
+        # llm_config structure within sly_data, however these cases are much less common.
+        "openai_api_key": "sly_data"
```
d1donlydfink, line 35: This is how you say "I need the API key from the user in order to function" in the llm_config.

neuro_san/registries/music_nerd_pro_sly_api_key.hocon
```diff
@@ -0,0 +1,144 @@
...
+                                "openai_api_key": {
+                                    "type": "string",
+                                    "description": "The User's OpenAI API key"
+                                }
+                            },
+                            "required": ["openai_api_key"]
+                        }
+                    },
+                    # Tell the client that api_keys is required.
+                    "required": ["llm_config"]
```
d1donlydfink, line 91: @dsargent : This is how you warn the client about how passing the llm_config API keys is a necessity via the Function() web API call.

For the llm_config entry in the function, note that this network needs a single API key for openai.

It's possible to create either/or situations where a list of fallbacks might contain an entry for OpenAI and another for Anthropic. If the user only needed one or the other and not both, the llm_config.required field might stay empty/unspecified, while the larger llm_config stays required.

neuro_san/internals/interfaces/context_type_llm_factory.py
d1donlydfink: Resolved

d1donlydfink:
@dsargent asks:

>One thing I'm not clear about: is it possible to override? For example, an agent is configured to use GPT-5, but I want to use GPT-4o. Can I use llm_config to override the value configured on the server? This would be handy for clients. Or is it that if the config is set as "mandatory", the client has to supply it and cannot fall back to what is configured on the server?

This is a very good question.
At this point, it's possible, but tricky and perhaps perilous.

My take on it goes like this:
* The agent network developer/server administrators have come to an agreement as to which model works best for the network and which is cost-effective for its operations.  It doesn't make sense to allow the dumb-ole-user to override that decision in every case simply by sending the right sly_data combo over the wire.
* You could set up a series of llm_config fallbacks such that:
    * the first model in the list has its "model" key tagged as coming from "sly_data".  If that isn't set at all in sly_data, the system would default to the next model in the list
    * the next model in the list could have its "model" key tagged as a default (say gpt-5.2) if sly_data.llm_config.model is not specified.

It's worth noting our "Security-by-Default" implementation here:  the hocon file for the agent_network needs to explicitly say "I am willing for the client to provide this key to the LLM". 

The peril comes in because the value sly_data.llm_config could specify:
* incorrect API keys - no key, no worky. 
* incorrect/inappropriate model is risky - maybe you get sora and and a big bill. Maybe you get an underpowered LLM that is just not up to the job
* temperature - could be safer and interesting, but the network developer needs to be able to take the risk - including whether or not specifying temperature will error out because it's not supported by a model.

@dsargent : It's possible we might have to come to some convention agreement so that you can detect the missing llm_config keys better. Happy for suggestions for a subsequent PR.

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
```diff
@@ -278,6 +296,47 @@ def get_chat_class_args(self, chat_class_name: str, use_model_name: str = None)
...
+            use_api_keys = sly_data.get("llm_config", use_api_keys)
+
+        # Loop through each of the config values and replace any "sly_data" values
+        # with their corresponding values from the llm_config dictionary.
+        required_set: Set[str] = set()
+        for key, value in config.items():
+            # If any value is "sly_data", replace it with the same value from the llm_config dictionary
+            small_value: str = str(value).lower()
+            if small_value == "sly_data":
+                # If we have a value in the sly_data.llm_config dictionary, use it,
```
d1donlydfink, line 325: Potential future-proofing here:
We are currently only accepting the single value "sly_data", which implies "Go look in sly_data.llm_config".

It is conceivable that we might want to get more specific as to where the value comes from, as long as it starts with "sly_data".   For instance, maybe someone doesn't like the conventions we have in place. We might in the future allow them to specify: "sly_data.OPENAI_API_KEY"  and that would mean that they expect the value to come from a higher-level key within sly_data.

I made some rearrangements of the code from last night to allow for the eventually so that if we go there, the changes would be more localized. This is why sly_data comes all the way down to this method in its entirety.

dsargent (approved):
Good start, as we discussed. Thanks for implementing this.

We can iterate on the "override" thing down the line as I think it would be valuable for scenarios like, "I'm a power user and I happen to know this agent would work better with GPT-5, even though it's configured for 4 on the server".

Noravee (approved):
LGTM.

PR #873, opened 2026-04-14
Fix returned sly_data reporting in agent_cli

d1donlydfink:
Was getting mixed signals from the client because it was spitting out that the returned sly_data actually had extra data in it, which it did not upon low-level inspection. What had been happening is a report of a mix of old and returned sly_data, but it was advertising the output as what was returned.

This PR allows the state to be kept to report the right thing for returned sly_data.

### DIFF
--- a/neuro_san/client/agent_cli.py
+++ b/neuro_san/client/agent_cli.py
@@ -209,8 +209,8 @@ Some suggestions:
 
             print(f"\nResponse from {state.get('origin_str')}:")
             print(f"{state.get('last_chat_response')}")
-            if state.get("sly_data") is not None:
-                pretty_sly: str = json.dumps(state.get('sly_data'), indent=4, sort_keys=True)
+            if state.get("returned_sly_data") is not None:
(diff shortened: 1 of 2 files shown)


neuro_san/client/agent_cli.py
```diff
@@ -209,8 +209,8 @@ def loop_until_done(self, user_input: str, sly_data: Dict[str, Any],
 
             print(f"\nResponse from {state.get('origin_str')}:")
             print(f"{state.get('last_chat_response')}")
-            if state.get("sly_data") is not None:
-                pretty_sly: str = json.dumps(state.get('sly_data'), indent=4, sort_keys=True)
+            if state.get("returned_sly_data") is not None:
+                pretty_sly: str = json.dumps(state.get('returned_sly_data'), indent=4, sort_keys=True)
```
d1donlydfink, line 213: Was getting mixed signals from the client because it was spitting out that the returned sly_data actually had extra data in it, which it did not upon low-level inspection. What had been happening is a report of a mix of old and returned sly_data, but it was advertising the output as what was returned.
This allows the state to be kept to report the right thing for return.

Noravee (approved):
LGTM.

PR #874, opened 2026-04-14
NS-863 Fix AgentFrameworkMessage chat_context default

d1donlydfink:
The MinimalMessageFilter had kinda required us to send an empty dictionary for the chat_context argument of the AgentFrameworkMessage.  There were 2 places in the code where we were not sending this and at least one of them was causing some problems in the tests.

The main situation here that was fixed was in the case where the network execution returned an exception of some kind.

This does _not_ solve the problem of final AgentFramework messages not showing up in thinking files.
That will be a separate investigation.

### DIFF
--- a/neuro_san/internals/chat/data_driven_chat_session.py
+++ b/neuro_san/internals/chat/data_driven_chat_session.py
@@ -168,9 +168,8 @@ class DataDrivenChatSession(RunTarget, LingeringResource):
             # This can happen if the user is trying to send a new message
             # while it is still working on a previous message that has not
             # yet returned.
-            empty: Dict[str, Any] = {}
             raw_messages: List[BaseMessage] = [
-                AgentFrameworkMessage(content="Patience, please. I'm working on it.", chat_context=empty)
+                AgentFrameworkMessage(content="Patience, please. I'm working on it.")
             ]
 
             logger: Logger = getLogger(self.__class__.__name__)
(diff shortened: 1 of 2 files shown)


neuro_san/internals/chat/data_driven_chat_session.py
```diff
@@ -257,8 +257,7 @@ async def run_it(self, inputs: AgentFrameworkMessage) -> AgentFrameworkMessage:
         # If we are invoked as an event, tell the caller that it's OK to disconnect early.
         # By definition, they are not expecting a custom response.
         if self.invocation_context.get_effective_invocation() == "event":
-            empty: Dict[str, Any] = {}
-            event_acknowledge = AgentFrameworkMessage(content="Event acknowledged", chat_context=empty)
+            event_acknowledge = AgentFrameworkMessage(content="Event acknowledged")
```
d1donlydfink, line 260: Don't need an empty dictionary chat_context any more.
That is now handled in the AgentFrameworkMessage class itself.

neuro_san/internals/messages/agent_framework_message.py
```diff
@@ -63,7 +63,13 @@ def __init__(self, content: Union[str, List[Union[str, Dict]]] = "",
         :param kwargs: Additional fields to pass to the superclass
         """
         super().__init__(content=content, trace_source=trace_source, **kwargs)
+
         self.chat_context: Dict[str, Any] = chat_context
+        if chat_context is None:
+            # Special case default so that the MinimalMessageFilter, which only
+            # looks for chat_context, does not fail to send the final AgentFrameworkMessage.
+            self.chat_context = {}
```
d1donlydfink, line 71: Since this has proven very easy to forget to initialize the chat_context correctly in client code,
do it here in the class.

neuro_san/internals/chat/data_driven_chat_session.py
d1donlydfink: There are actually 2 other places in DataDrivenChatSession which were not sending the equivalent of chat_context={}, and those were causing their own problems with the MinimalMessageFilter

neuro_san/internals/messages/agent_framework_message.py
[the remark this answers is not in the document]
d1donlydfink: Yup.

neuro_san/internals/chat/data_driven_chat_session.py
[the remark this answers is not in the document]
d1donlydfink: Yup

Noravee (approved):
LGTM.

deepsaia (approved):
Lgtm.

PR #890, opened 2026-04-20
NS-888 Use preferred, strict get_running_loop()

d1donlydfink:
Use stricter get_running_loop() instead of deprecated get_event_loop().
This avoids the warning message @vince-leaf saw with his pytest invocations for direct sessions.

Tested locally. All seems OK with and without events.

Note: The build is failing because of a build-common problem which @donn-leaf has a fix for, but just has to push.

### DIFF
--- a/neuro_san/session/session_invocation_context.py
+++ b/neuro_san/session/session_invocation_context.py
@@ -23,7 +23,7 @@ from typing import Dict
 from typing import List
 
 from asyncio import AbstractEventLoop
-from asyncio import get_event_loop
+from asyncio import get_running_loop
 from asyncio import run_coroutine_threadsafe
 from copy import copy
 from concurrent.futures import Future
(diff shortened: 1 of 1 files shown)


neuro_san/session/session_invocation_context.py
```diff
@@ -349,8 +349,7 @@ def _run_in_executor_until_complete(self, submitter_id: str, coroutine: Awaitabl
         other_loop: AbstractEventLoop = self.asyncio_executor.get_event_loop()
         loop: AbstractEventLoop = None
         try:
-            # Use get_event_loop() over get_running_loop(), as its more robust
-            loop = get_event_loop()
+            loop = get_running_loop()
```
d1donlydfink, line 352: Use stricter get_running_loop() instead of get_event_loop() which turns out to be deprecated.
The comment is a little misleading. The get_event_loop() source actually calls get_running_loop() itself, but also checks a policy object for another cached loop. That extra policy object cache check, I believe is irrelevant given the multi-event loop situation in the server.

andreidenissov-cog (approved):
Looks good to me.

PR #904, opened 2026-04-29
Update default LLM info file

Noravee:
- Add gemini 2.5 flash lite and 3.1 flash lite and remove 3.0 pro preview. See https://ai.google.dev/gemini-api/docs/models and https://ai.google.dev/gemini-api/docs/deprecations
- Remove bedrock sonnet 3.7. See https://docs.aws.amazon.com/bedrock/latest/userguide/model-lifecycle.html
- Modify `music_nerd_pro_llm_bedrock_claude` to use sonnet 4 instead since it used sonnet 3.7 from bedrock


ofrancon (approved):
lgtm

PR #912, opened 2026-05-05
905 add test coverage for s3 write functionality in neuro san

### LINKED ISSUE #905: Add test coverage for S3 write functionality in Neuro-san
Description:
We need to validate that Neuro-san can successfully write outputs to S3 as part of its execution flow. This should extend existing test coverage rather than introducing a completely new pattern.

Proposed Approach:

Reuse/extend the existing copy_cat test:
Current: in-memory → in-memory
New: in-memory → S3 (or S3-like target)
Ensure test verifies:
Successful write operation
Correct object structure/content in target storage
Avoid reliance on real AWS S3

Constraints / Considerations:

GitHub Actions does not have access to real S3
Use a mock S3 solution (e.g., LocalStack, MinIO, or similar)
Keep test compatible with existing CI pipeline

vince-leaf:
@andreidenissov-cog as primary reviewer.

- [X]  All 12 test files under tests/neuro_san/service/watcher/temp_networks/
- [X]  _test_base.py, conftest.py, and __init__.py files at each new package level
- [X]  docs/tests.md updated with both edits

All test cases ran between 1 and 2 seconds.
====================================================================================== 12 passed in 1.90s ======================================================================================

## Summary
 
Add a 12-test mock-based unit suite that exercises the write path
of `S3ReservationsStorage` end-to-end. No real AWS, no LocalStack
or moto, no new dependencies — just a small in-memory `FakeS3Client`
that implements the methods the storage actually uses
(`head_bucket`, `put_object`, `get_object`).
 
The suite verifies that the reservation feature works as expected
when reservations are written to and read back from S3-like
storage, and pins behavior that future refactors might silently
break.
 
## What's covered
 
| #   | File                              | Pins                                         |
|-----|-----------------------------------|----------------------------------------------|
| T1  | test_round_trip                   | Reservation round-trips via the public API   |
| T2  | test_user_authored_metadata       | User metadata is merged, not clobbered       |
| T3  | test_overwrite_duplicate_id       | Same-id writes overwrite (last-write-wins)   |
| T4  | test_object_key_format            | S3 key is `{prefix}{id}.json`                |
| T5  | test_multiple_reservations        | Batch entries are independent                |
| T6  | test_retry_on_throttling          | Retries succeed on transient errors          |
| T7  | test_no_retry_on_access_denied    | AccessDenied propagates without retry        |
| T8  | test_batch_partial_failure        | Earlier writes survive a later failure       |
| T9  | test_empty_batch_is_no_op         | `add_reservations({})` is a complete no-op   |
| T10 | test_custom_prefix                | Non-default prefix is honored                |
| T11 | test_spec_without_metadata        | Spec without `metadata` key initializes      |
| T12 | test_json_body_format             | On-disk JSON shape: spec at top, metadata    |
|     |                                   | injected, reservation id correct


neuro_san/service/watcher/temp_networks/s3_reservations_storage.py
```diff
@@ -213,8 +213,37 @@ def add_reservations(self, reservations_dict: Dict[Reservation, Any],
...
+        On-disk JSON schema written per reservation
+        (key = "<prefix><reservation_id>.json"):
+
+            {
+                "name":       <str>,    # Authored network name (HOCON "name")
+                "llm_config": <dict>,   # Optional LLM settings
+                "tools":      <list>,   # Agent definitions making up the network
+                ...                     # Any other top-level spec fields
+                "metadata": {
+                    ...                                          # User-authored keys (merged in)
```
vince-leaf, line 225: Added a sample JSON schema, which is expected on disk, for the reservation to write to S3. This is info that these test cases confirm. If any code changes, such as enhancements, are made to the schema info, it can be updated.

tests/neuro_san/service/watcher/temp_networks/_test_base.py
```diff
@@ -0,0 +1,204 @@
...
+    import S3ReservationsStorage
+
+
+# pylint: disable=invalid-name,unused-argument
+class FakeS3Client:
+    """
+    Minimal in-memory stand-in for a boto3 S3 client. Only implements the
+    methods that S3ReservationsStorage actually calls, storing object bodies
+    in a dict keyed by S3 object key. Method signatures use boto3's PascalCase
+    keyword arguments so the storage's call sites work unchanged.
```
vince-leaf: Added mock S3 class to be called by these test cases.

tests/neuro_san/service/watcher/temp_networks/_test_base.py
```diff
@@ -0,0 +1,204 @@
...
+                {"Error": {"Code": "NoSuchKey", "Message": "Not Found"}},
+                "GetObject",
+            )
+        return {"Body": io.BytesIO(self.objects[Key])}
+
+
+class S3ReservationsStorageTestBase(TestCase):
+    """
+    Base TestCase that exercises the reservation feature through
+    S3ReservationsStorage by injecting an in-memory FakeS3Client. No real
```
vince-leaf, line 44: Class to be called to inject our network metadata into mock S3 for all of these test cases.

tests/neuro_san/service/watcher/temp_networks/conftest.py
```diff
@@ -0,0 +1,39 @@
...
+client and do not call any LLM provider, so the project-wide skip
+condition does not apply. This conftest overrides that fixture with a
+no-op for tests in this subtree only.
+"""
+import pytest
+
+
+# pylint: disable=unused-argument
+@pytest.fixture(autouse=True)
+def configure_llm_provider_keys(request, monkeypatch):
```
vince-leaf, line 33: This will allow these test cases to run without requiring an OPENAI_API_KEY. Without it, pytest will force these test cases to be skipped.

tests/neuro_san/service/watcher/temp_networks/test_batch_partial_failure.py
```diff
@@ -0,0 +1,147 @@
...
+"""
+from unittest.mock import patch
+
+from botocore.exceptions import ClientError
+
+from tests.neuro_san.service.watcher.temp_networks._test_base \
+    import S3ReservationsStorageTestBase
+
+
+class TestBatchPartialFailure(S3ReservationsStorageTestBase):
```
vince-leaf, line 31: Add batch partial-failure test for S3ReservationsStorage
S3ReservationsStorage.add_reservations writes batch entries
sequentially in a Python for-loop with no atomic-batch
semantics. If a later entry's put_object fails, earlier
Successful writes remain in S3.
This test pins the current partial-success contract so a
future change introducing rollback (or accidentally swallowing
per-entry errors) requires an explicit update and surfaces
in code review.
Setup: 3 reservations in one batch; the third put_object
raises AccessDenied. Asserts that:
  - The error propagates with its original code preserved.
  - put_object is invoked exactly 3 times (no retry, no
    skipped entries, no continuation past failure).
  - The two earlier successes survive in S3.
  - The failed entry's key is NOT in S3

Restore patched put_object after partial-failure test
Document the partial-failure test's intent to leave the fake
S3 client unchanged across tests by adding an explicit
addCleanup that restores the original put_object bound method.
The base TestCase already hands each test a fresh
FakeS3Client in setUp, so this cleanup is a no-op in the
current code path. Keeping it makes the test correct under
a future refactor that shares a single FakeS3Client across
tests, and answers the equally-valid "do we need to restore?"
question from code review without requiring readers to trace
through the base class.

tests/neuro_san/service/watcher/temp_networks/test_custom_prefix.py
```diff
@@ -0,0 +1,102 @@
...
+"""
+from neuro_san.service.watcher.temp_networks.s3_reservations_storage \
+    import S3ReservationsStorage
+
+from tests.neuro_san.service.watcher.temp_networks._test_base \
+    import S3ReservationsStorageTestBase
+
+
+class TestCustomPrefix(S3ReservationsStorageTestBase):
+    """
```
vince-leaf, line 34: Add custom-prefix test for S3ReservationsStorage
S3ReservationsStorage accepts a configurable prefix (default
"reservations/") prepended to every reservation's S3 key.
Existing tests all use the default value, so a hardcoded prefix
regression in get_obj_key_for_reservation would silently slip
past CI and break only on production deploys with non-default
prefixes (multi-tenancy, environment separation, version
migration).
This test instantiates a storage with the prefix
"my-tenant/reservations-v2/" and verifies that:
  - The written object's key uses the custom prefix.
  - The object is NOT at the default-prefix key.
  - Exactly one object is written.

tests/neuro_san/service/watcher/temp_networks/test_empty_batch_is_no_op.py
```diff
@@ -0,0 +1,87 @@
...
+pass {} when pre-filtering yields no new reservations; the storage
+must handle that without crashing, without writing placeholder
+objects, and without making any S3 calls.
+"""
+from tests.neuro_san.service.watcher.temp_networks._test_base \
+    import S3ReservationsStorageTestBase
+
+
+class TestEmptyBatchIsNoOp(S3ReservationsStorageTestBase):
+    """
```
vince-leaf, line 29: Validate JSON body against schema in T12
Replace the per-field assertIn/assertEqual checks in T12 with
a single jsonschema.validate call against
RESERVATION_OBJECT_SCHEMA, plus one value-level assertEqual
on the reservation id under metadata.
The schema documents the on-disk shape S3ReservationsStorage
commits to in one declarative place: top-level agent_spec
fields (no wrap-in-envelope) and storage-injected
reservation+stored_at under "metadata". External consumers
read these objects directly from S3, so this becomes the
single source of truth for the format and is reusable from
any future test that needs the same shape pin.
jsonschema is already a runtime dependency
(requirements.txt:jsonschema>=4.19.0), so this introduces no
new dependency.

tests/neuro_san/service/watcher/temp_networks/test_json_body_format.py
```diff
@@ -0,0 +1,151 @@
...
+"""
+from json import loads
+
+from tests.neuro_san.service.watcher.temp_networks._test_base \
+    import S3ReservationsStorageTestBase
+
+
+class TestJsonBodyFormat(S3ReservationsStorageTestBase):
+    """
+    Pin the on-disk JSON shape produced by add_reservations. Reads
```
vince-leaf, line 89: Add JSON body format pin for S3ReservationsStorage
S3ReservationsStorage commits to a specific on-disk format: JSON
serialization, original agent_spec at the top level (no
wrap-in-envelope), storage-injected reservation+stored_at under
"metadata". External consumers read these objects directly and
Rely on the format being stable.
T1's round-trip would still pass if the read+write paths were
updated together, but the on-disk format silently changed. This
test reads the raw bytes from S3 and asserts on the parsed shape
independent of the read path, catching a format regression that
slips past the round-trip suite.

tests/neuro_san/service/watcher/temp_networks/test_multiple_reservations.py
```diff
@@ -0,0 +1,102 @@
...
+S3ReservationsStorage.add_reservations accepts a dict of multiple
+{Reservation: agent_spec} pairs. This module exercises the for-loop
+that iterates over those pairs.
+"""
+from tests.neuro_san.service.watcher.temp_networks._test_base \
+    import S3ReservationsStorageTestBase
+
+
+class TestMultipleReservations(S3ReservationsStorageTestBase):
+    """
```
vince-leaf, line 27: Add multi-reservation batch test for S3ReservationsStorage
Verify that a single add_reservations call with N>1 entries writes
one S3 object per entry and preserves each spec independently. The
existing tests only exercise N=1 per call, so the for-loop body is
never iterated more than once. This test fills that gap.
Catches bugs like a loop break after the first iteration, an
iteration that skips entries, or cross-contamination of the agent
specs between iterations.

tests/neuro_san/service/watcher/temp_networks/test_no_retry_on_access_denied.py
```diff
@@ -0,0 +1,129 @@
...
+from unittest.mock import patch
+
+from botocore.exceptions import ClientError
+
+from tests.neuro_san.service.watcher.temp_networks._test_base \
+    import S3ReservationsStorageTestBase
+
+
+class TestNoRetryOnAccessDenied(S3ReservationsStorageTestBase):
+    """
```
vince-leaf, line 34: Add no-retry-on-AccessDenied test for S3ReservationsStorage
S3ReservationsStorage._is_retryable_client_error must classify
AccessDenied (HTTP 403) is non-retryable, so callers see
permission errors immediately rather than after 8 wasted retries
attempts. None of the existing tests exercises the
non-retryable branch.
This test injects an AccessDenied ClientError on every
put_object call, and verifies that:
  - The error propagates with its original code preserved.
  - put_object is invoked exactly once (no retries).
  - S3 ends up with no objects.
  - get_one_reservation returns (None, None) for the failed id,
    confirming nothing leaked into any cache or fallback store.
Catches regressions where AccessDenied is added to the retryable
set or where _do_with_retries silently swallows ClientError.

tests/neuro_san/service/watcher/temp_networks/test_object_key_format.py
```diff
@@ -0,0 +1,80 @@
...
+"""
+S3ReservationsStorage.add_reservations writes each reservation to a
+specific S3 object key. This module pins that key format as a contract.
+"""
+from tests.neuro_san.service.watcher.temp_networks._test_base \
+    import S3ReservationsStorageTestBase
+
+
+class TestObjectKeyFormat(S3ReservationsStorageTestBase):
+    """
```
vince-leaf, line 26: Add object-key format test for S3ReservationsStorage
Pin the documented S3 object-key contract written by
add_reservations: "{prefix}{reservation_id}.json". Round-trip
tests cannot catch a bug where read and write both drift to a
new layout because they share the same helper; this test
locks in the literal layout from outside the storage.

tests/neuro_san/service/watcher/temp_networks/test_overwrite_duplicate_id.py
```diff
@@ -0,0 +1,101 @@
...
+"""
+S3ReservationsStorage.add_reservations is last-writer-wins for a given
+reservation id: a second call writes a new JSON blob to the same S3 key,
+replacing the first one.
+"""
+from tests.neuro_san.service.watcher.temp_networks._test_base \
+    import S3ReservationsStorageTestBase
+
+
+class TestOverwriteDuplicateId(S3ReservationsStorageTestBase):
```
vince-leaf, line 26: Note: **That test confirms that the server can override the existing reservation_id JSON.**

Add overwrite-on-duplicate-id test for S3ReservationsStorage
Verify that calling add_reservations twice with the same
reservation ID results in the second write fully replacing the
first in S3, and that exactly one S3 object exists per id (no
leaked duplicate keys, no merged payloads).
Documents the intentional last-writer-wins semantics of the
storage and guards against a future regression that "protects."
duplicate IDs by skipping the second put or merging payloads.

tests/neuro_san/service/watcher/temp_networks/test_retry_on_throttling.py
```diff
@@ -0,0 +1,113 @@
...
+
+from botocore.exceptions import ClientError
+
+from tests.neuro_san.service.watcher.temp_networks._test_base \
+    import S3ReservationsStorageTestBase
+
+
+class TestRetryOnThrottling(S3ReservationsStorageTestBase):
+    """
+    The earlier tests all run against a FakeS3Client that never fails,
```
vince-leaf, line 33: Add retry-on-throttling test for S3ReservationsStorage
S3ReservationsStorage routes every put_object call through
_do_with_retries, which retries on transient ClientError codes
(SlowDown, Throttling, 5xx, etc.) with exponential backoff. None
of the existing tests trigger a put failure, so the retry branch
is uncovered.
This test injects a one-shot ThrottlingException on the first
put_object attempt, and mocks time. Sleep to keep the test fast, and
verify that the retry succeeds and S3 ends up with exactly
one object at the documented key.
Catches regressions where retries are disabled, the
max_attempts cap is too low, or _is_retryable_client_error
misclassifies retryable error codes.

tests/neuro_san/service/watcher/temp_networks/test_round_trip.py
```diff
@@ -0,0 +1,120 @@
...
+"""
+Round-trip a single reservation through S3ReservationsStorage.
+"""
+from tests.neuro_san.service.watcher.temp_networks._test_base \
+    import S3ReservationsStorageTestBase
+
+
+class TestRoundTrip(S3ReservationsStorageTestBase):
+    """
+    Reservation feature works end-to-end against an S3-like backend:
```
vince-leaf, line 26: Add round-trip test for S3ReservationsStorage
Pin the baseline contract that a reservation was written via
add_reservations comes back equivalent via get_one_reservation.
The test verifies id, lifetime_in_seconds, and
expiration_time_in_seconds all survive serialization through
the in-memory FakeS3Client.
This is the foundation test that every other write-path test
in the suite builds on. If this test fails, the storage's
write-then-read invariant is broken.

tests/neuro_san/service/watcher/temp_networks/test_spec_without_metadata.py
```diff
@@ -0,0 +1,150 @@
...
+    agent_spec["metadata"].update(new_metadata)
+"""
+from json import loads
+
+from tests.neuro_san.service.watcher.temp_networks._test_base \
+    import S3ReservationsStorageTestBase
+
+
+class TestSpecWithoutMetadata(S3ReservationsStorageTestBase):
+    """
```
vince-leaf, line 38: Test storage initializes missing metadata key
When an agent_spec arrives at add_reservations without a
"metadata" key, S3ReservationsStorage initializes the key to
an empty dict before merging into the storage's bookkeeping
fields (reservation, stored_at). Registry HOCONs document
"metadata" is optional, so missing-metadata input is a real
production path.
This test pins that initialization branch: it constructs a
spec without a "metadata" key, calls add_reservations, and
verifies the write succeeds, the on-disk JSON has the
storage-injected metadata fields, and the reservation
round-trips via the public read API.
Catches a regression where the is-None guard is dropped
(would raise KeyError on .update) or where the missing-
metadata path silently no-ops the write.

tests/neuro_san/service/watcher/temp_networks/test_user_authored_metadata.py
```diff
@@ -0,0 +1,124 @@
...
+"""
+from typing import Any
+from typing import Dict
+
+from tests.neuro_san.service.watcher.temp_networks._test_base \
+    import S3ReservationsStorageTestBase
+
+
+class TestUserAuthoredMetadata(S3ReservationsStorageTestBase):
+    """
```
vince-leaf, line 30: Test storage merges user-authored metadata
When an agent_spec arrives at add_reservations with user-
authored metadata (description, tags, etc. from a registry
HOCON), S3ReservationsStorage must merge its bookkeeping
fields (reservation, stored_at) without clobbering the
caller's existing keys.
This test pins that merge contract: it constructs a spec
with a metadata block containing a user-authored description
and tags, calls add_reservations, then verifies the on-disk
JSON has both the user fields and the storage-injected
fields side by side.
Catches a regression where the storage replaces the metadata
dict instead of merging into it (would silently drop the user
description/tags on every write).

docs/tests.md
```diff
@@ -18,6 +18,10 @@ Many (but not absolutely all) tests will require an active OPENAI_API_KEY or oth
 from another LLM provider in order to run successfully.  Please be sure you have a basic
 agent network running in your environment as described in the top-level README.md to this repo.
 
+Some unit tests (for example, the S3 reservations storage suite under
+`tests/neuro_san/service/watcher/temp_networks/`) are fully mocked and run without any
+LLM, AWS, or external service keys.
```
vince-leaf, line 23: Added path to these test cases.

docs/tests.md
```diff
@@ -31,6 +35,29 @@ To run all basic unit tests:
 
 The -n auto allows the tests to run in parallel on available CPUs.
 
+### Mock-based unit tests for cloud-backed storage
+
+Some unit tests exercise components that talk to AWS or other cloud
+services in production but use an in-memory mock of the SDK client
+in tests.  These tests do not need any AWS credentials, do not pull
+in `moto` or `localstack`, and run anywhere a basic Python
+environment is available.
```
vince-leaf, line 44: Added info regarding tests, like mock, and how to run these test cases.

neuro_san/service/watcher/temp_networks/s3_reservations_storage.py
andreidenissov-cog: Nice document description 👍

tests/neuro_san/service/watcher/temp_networks/_test_base.py
andreidenissov-cog: Our general style policy: one class - one Python file. So please separate it into its own file.

tests/neuro_san/service/watcher/temp_networks/test_batch_partial_failure.py
```diff
@@ -0,0 +1,147 @@
...
+                )
+            return real_put(
+                Bucket=Bucket,
+                Key=Key,
+                Body=Body,
+                ContentType=ContentType,
+            )
+
+        self.fake_s3.put_object = fail_on_third
+
```
andreidenissov-cog, line 96: Just curious: do we need to restore this to "original"  self.fake_s3.put_object
after this particular test is done?

tests/neuro_san/service/watcher/temp_networks/test_json_body_format.py
```diff
@@ -0,0 +1,151 @@
...
+        # KeyError otherwise).
+        expected_key = f"reservations/{reservation_id}.json"
+        self.assertIn(
+            expected_key,
+            self.fake_s3.objects,
+            f"Expected object at {expected_key!r}; bucket has "
+            f"{list(self.fake_s3.objects)}.",
+        )
+        body: bytes = self.fake_s3.objects[expected_key]
+
```
andreidenissov-cog, line 127: We could possibly use some existing json structure validator (against schema) instead of bunch of asserts for each specific json member.

andreidenissov-cog (changes requested):

andreidenissov-cog:
One request: Our general style policy: one class - one Python file. So please separate it into its own file.
(this is about [tests/neuro_san/service/watcher/temp_networks/_test_base.py] file.)

tests/neuro_san/service/watcher/temp_networks/_test_base.py
vince-leaf: Done. Refactor to one class in a Python file. Added _fake_3_client.py

vince-leaf:
> One request: Our general style policy: one class - one Python file. So please separate it into its own file. (this is about [tests/neuro_san/service/watcher/temp_networks/_test_base.py] file.)

Done!

tests/neuro_san/service/watcher/temp_networks/test_batch_partial_failure.py
vince-leaf: Good catch/instinct — but in the current code path, no restore
is needed: S3ReservationsStorageTestBase.setUp() hands each test
a fresh FakeS3Client (self.fake_s3 = FakeS3Client()), so the
fail_on_third assignment only mutates the instance attribute on
THIS test's FakeS3Client, which is discarded at the end of the test. The
next test's setUp gives us a new instance whose put_object is
the original bound method.
That said, you're right that the explicit restore is more
defensive and removes the need to trace through the base class
to verify the no-leak claim. I've added an addCleanup right
after capturing real_put, with a comment explaining that it's a
no-op today and is here to keep the test correct under any
future refactor that turns self.fake_s3 into a shared object.

tests/neuro_san/service/watcher/temp_networks/test_json_body_format.py
vince-leaf: Done!

Good suggestion - jsonschema is already in requirements.txt so
This costs nothing. Updated T12 to validate the parsed body
against a RESERVATION_OBJECT_SCHEMA constant that documents the
on-disk shape, plus one value-level assertEqual to confirm the
reservation ID round-trips correctly. The schema becomes the
single source of truth for the format S3ReservationsStorage
commits to - any future test that needs to pin the same shape
can reuse it.

tests/neuro_san/service/watcher/temp_networks/_fake_s3_client.py
```diff
@@ -0,0 +1,78 @@
...
+"""
+import io
+
+from typing import Dict
+
+from botocore.exceptions import ClientError
+
+
+# pylint: disable=invalid-name,unused-argument
+class FakeS3Client:
```
vince-leaf, line 37: Drop it into tests/neuro_san/service/watcher/temp_networks/ next to _ test_base.py to one class - one Python file.
This is now a single class.

andreidenissov-cog (approved):
Looks great - very comprehensive test set.

PR #922, opened 2026-05-08
NS-917 Remove unnecessary sleep() in HTTP request handler.

andreidenissov-cog:
This PR:
1.  removes unconditional sleep() we have put in HTTP request handler to work around UX client problems in incoming messages decoding. These problems has now been fixed, so we were just wasting 0.3 seconds on each outgoing message;
2. modifies logic for incoming messages decoding in neuro-san http clients (both sync and async) to be more efficient and uniform for both modes.

### DIFF
--- a/neuro_san/service/http/handlers/base_request_handler.py
+++ b/neuro_san/service/http/handlers/base_request_handler.py
@@ -251,19 +251,6 @@ class BaseRequestHandler(RequestHandler):
         """
         try:
             await self.flush()
-            # What happens here: we have finished writing out one data item in our output stream,
-            # and we have flushed Tornado output.
-            # BUT: this does not guarantee in general that underlying TCP/IP transport
-            # will flush its own buffers, so low-level buffering is still possible.
-            # Result would be that several chat responses will be bunched together
-            # and received by a client as one data piece.
-            # If client is not ready for this, there will be problems.
-            # SO: this real wall clock delay here helps to encourage underlying transport
-            # to flush its own buffers - and we are good.
-            # Duration of delay is speculative and maybe could be adjusted.
-            # But best solution and reliable one: make client accept multiple data items
-            # in one "get" request - as it should when dealing with streaming service.
-            await asyncio.sleep(0.3)
             return True
         except tornado.iostream.StreamClosedError:
             self.logger.warning(self.get_metadata(), "Flush: client closed connection unexpectedly.")
--- a/neuro_san/service/http/handlers/openapi_publish_handler.py
+++ b/neuro_san/service/http/handlers/openapi_publish_handler.py
@@ -30,7 +30,7 @@ class OpenApiPublishHandler(BaseRequestHandler):
     Handler class for neuro-san OpenAPI service spec publishing"concierge" API call.
     """
 
-    def get(self):
+    async def get(self):
         """
         Implementation of GET request handler
         for "publish my OpenAPI specification document" call.
--- a/neuro_san/session/async_http_service_agent_session.py
+++ b/neuro_san/session/async_http_service_agent_session.py
@@ -126,19 +126,18 @@ class AsyncHttpServiceAgentSession(AbstractHttpServiceAgentSession, AsyncAgentSe
                     #               ... blah blah ...
                     #       but that could fail with ValueError("Chunk too big")
                     #       if a single line was too long.
-                    accumulator: bytes = b""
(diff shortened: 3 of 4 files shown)


neuro_san/service/http/handlers/base_request_handler.py
```diff
@@ -251,19 +251,6 @@ async def do_flush(self) -> bool:
...
-            # Result would be that several chat responses will be bunched together
-            # and received by a client as one data piece.
-            # If client is not ready for this, there will be problems.
-            # SO: this real wall clock delay here helps to encourage underlying transport
-            # to flush its own buffers - and we are good.
-            # Duration of delay is speculative and maybe could be adjusted.
-            # But best solution and reliable one: make client accept multiple data items
-            # in one "get" request - as it should when dealing with streaming service.
-            await asyncio.sleep(0.3)
             return True
```
andreidenissov-cog, line 254: Not needed any more. Yay!

neuro_san/service/http/handlers/openapi_publish_handler.py
```diff
@@ -30,7 +30,7 @@ class OpenApiPublishHandler(BaseRequestHandler):
     Handler class for neuro-san OpenAPI service spec publishing"concierge" API call.
     """
 
-    def get(self):
+    async def get(self):
         """
```
andreidenissov-cog, line 34: Side fix - our GETs are async.

neuro_san/session/async_http_service_agent_session.py
```diff
@@ -126,36 +126,36 @@ async def streaming_chat(self, request_dict: Dict[str, Any]) -> Generator[Dict[s
                     #               ... blah blah ...
                     #       but that could fail with ValueError("Chunk too big")
                     #       if a single line was too long.
-                    accumulator: bytes = b""
+                    accumulator: bytearray = bytearray(b"")
                     async for data in response.content.iter_chunked(max_chunk_size):
```
andreidenissov-cog, line 130: Use more efficient bytearrays instead of immutable "bytes" - less memory copying.

neuro_san/session/async_http_service_agent_session.py
```diff
@@ -126,36 +126,36 @@ async def streaming_chat(self, request_dict: Dict[str, Any]) -> Generator[Dict[s
...
                         index: int = accumulator.find(separator)
                         while index >= 0:
 
                             # Grab a single line
-                            line: bytes = accumulator[:index]
-                            unicode_line = line.decode("utf-8")
-                            if unicode_line.strip():    # Skip empty lines
-
+                            unicode_line: str = accumulator[:index].decode("utf-8").strip()
+                            if unicode_line:    # Skip empty lines
```
andreidenissov-cog, line 141: Decode incoming data bytes to UTF-8 string explicitly, with no reliance on external HTTP server settings.

neuro_san/session/async_http_service_agent_session.py
```diff
@@ -126,36 +126,36 @@ async def streaming_chat(self, request_dict: Dict[str, Any]) -> Generator[Dict[s
...
 
                             # Allow for case of multiple lines in one chunk
                             index = accumulator.find(separator)
 
                     # If there is anything left in the accumulator, yield it
                     if len(accumulator) > 0:
-                        result_dict = json.loads(accumulator.decode("utf-8"))
-                        yield result_dict
+                        unicode_line: str = accumulator.decode("utf-8").strip()
+                        if unicode_line:
```
andreidenissov-cog, line 156: Additional check that our data tail is not empty before json parsing.

neuro_san/session/http_service_agent_session.py
```diff
@@ -84,16 +84,54 @@ def streaming_chat(self, request_dict: Dict[str, Any]) -> Generator[Dict[str, An
...
         try:
             with requests.post(path, json=request_dict, headers=self.get_headers(),
                                stream=True,
                                timeout=self.streaming_timeout_in_seconds) as response:
                 response.raise_for_status()
 
-                for line in response.iter_lines(decode_unicode=True):
-                    if line.strip():  # Skip empty lines
-                        result_dict = json.loads(line)
+                # Iterate over the content stream as it comes in.
```
andreidenissov-cog, line 96: This is pretty much copy of the same code for AsyncHttpServiceAgentSession,
but for sync mode.

d1donlydfink (approved):

PR #937, opened 2026-05-15
Add run-time snapshot query to neuro-san service.

andreidenissov-cog:
This PR adds an additional endpoint for neuro-san service: 
/snapshot
which will query the current state of system run-time resources and return it as a JSON dictionary.
This query can be used for external server monitoring/testing and works in parallel with internal system resources logging.

### DIFF
--- a/neuro_san/deploy/Dockerfile
+++ b/neuro_san/deploy/Dockerfile
@@ -369,6 +369,11 @@ ENV AGENT_AUTHORIZER_RESOURCE_KEY=AgentNetwork
 # Typically "read" from CRUD-based permissions.
 ENV AGENT_AUTHORIZER_ALLOW_RELATION=read
 
+# Whether to enable run-time statistics endpoints for neuro-san server as a whole.
+# These endpoints can be used to get information about the server's current resource usage
+# and also to control run-time service profiler.
+ENV ENABLE_RUN_TIME_STATISTICS="false"
+
 # Below: Env vars specific to OpenFgaAuthorizer
 
 # Where the OpenFGA HTTP server is running for the OpenFgaAuthorizer
--- a/neuro_san/service/http/handlers/snapshot_handler.py
+++ b/neuro_san/service/http/handlers/snapshot_handler.py
@@ -1,0 +1,59 @@
+
+# Copyright © 2023-2026 Cognizant Technology Solutions Corp, www.cognizant.com.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# END COPYRIGHT
+"""
+See class comment for details
+"""
+from typing import Any, Dict, Optional
+import json
+import logging
+
+from http import HTTPStatus
+from tornado.web import RequestHandler
+
+from neuro_san.service.utils.service_resources import ServiceResources
+
+
+class SnapshotHandler(RequestHandler):
+    """
+    Handler class for executing system run-time snapshot query via HTTP API calls.
+    """
+
+    # pylint: disable=attribute-defined-outside-init
+    def initialize(self, **kwargs):
+        """
+        This method is called by Tornado framework to allow
+        injecting service-specific data into local handler context.
+        :param kwargs: dictionary of named parameters, including:
+            "http_port" - HTTP server port;
+        """
+        # Set up local members from kwargs dictionary passed in:
+        self.http_port: int = int(kwargs.pop("http_port"))
+        self.logger = logging.getLogger(self.__class__.__name__)
+
+    async def get(self):
+        """
+        Implementation of GET request handler for system snapshot query.
+        """
+        snapshot_dict: Dict[str, Any] = ServiceResources.get_snapshot_dict(self.http_port)
+        self.write(json.dumps(snapshot_dict, indent=4))
+        self.logger.info("Returned system snapshot: %s", json.dumps(snapshot_dict, indent=4))
+
+    def data_received(self, chunk):
+        """
+        This method is required to be implemented as part of RequestHandler subclass,
+        but is not used in our case as we do not expect any data in the request body.
+        """
--- a/neuro_san/service/http/handlers/snapshot_handler.py
+++ b/neuro_san/service/http/handlers/snapshot_handler.py
@@ -18,8 +18,10 @@
 See class comment for details
 """
 from typing import Any, Dict
+import os
 import json
 import logging
+from http import HTTPStatus
 
 from tornado.web import RequestHandler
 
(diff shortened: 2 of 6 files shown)


neuro_san/service/http/handlers/snapshot_handler.py
```diff
@@ -0,0 +1,58 @@
...
+from tornado.web import RequestHandler
+
+from neuro_san.service.utils.service_resources import ServiceResources
+
+
+class SnapshotHandler(RequestHandler):
+    """
+    Handler class for executing system run-time snapshot query via HTTP API calls.
+    """
+
```
andreidenissov-cog, line 35: New request handler - ismply calls ServiceResources.get_snapshot_dict() and publishes it.

neuro_san/service/http/server/http_server.py
```diff
@@ -223,6 +224,9 @@ def make_app(self, requests_limit: int, concurrent_requests_limit: int, logger:
         # Setup handler for profiler control
         handlers.append(("/profiler", ProfilerControlHandler))
 
+        # Setup handler for system snapshot query:
+        handlers.append(("/snapshot", SnapshotHandler, {"http_port": self.http_port}))
+
```
andreidenissov-cog, line 229: Add snapshot handler to our HTTP service.

neuro_san/service/http/server/resources_usage_logger.py
```diff
@@ -54,16 +54,8 @@ def log_resources_usage(self):
...
-        sock_classes = ServiceResources.classify_sockets(self.http_port)
-        log_dict: Dict[str, Any] = {
-            "soft_limit": soft_limit,
-            "hard_limit": hard_limit,
-            "file_descriptors": fd_dict,
-            "sockets": sock_classes
-        }
-        self.logger.info({}, "Used: %s", json.dumps(log_dict, indent=4))
+        snapshot_dict: Dict[str, Any] = ServiceResources.get_snapshot_dict(self.http_port)
+        self.logger.info({}, "Used: %s", json.dumps(snapshot_dict, indent=4))
```
andreidenissov-cog, line 58: Use ServiceResources.get_snapshot_dict() as a single functional point in both places:
in-service logging and external http query.

neuro_san/service/utils/service_resources.py
```diff
@@ -292,3 +292,43 @@ def classify_sockets(cls, server_port: int) -> Dict[str, Any]:
...
+        return mem_info.rss / (1024 * 1024), cls.max_memory_used_bytes / (1024 * 1024)
+
+    @classmethod
+    def get_snapshot_dict(cls, server_port: int) -> Dict[str, Any]:
+        """
+        Get a snapshot of current resource usage for logging or metrics.
+        :param server_port: server port to classify sockets
+        :return: dictionary with resource usage information
+        """
+        # Get used file descriptors:
```
andreidenissov-cog, line 336: Consolidate system run-time data in a single dictionary.

neuro_san/service/http/server/http_server.py
d1donlydfink: You might want to enable this + profiler only when a single env variable is set.
One could construe this information as a security leak at the operations level.

neuro_san/service/utils/service_resources.py
```diff
@@ -292,3 +292,43 @@ def classify_sockets(cls, server_port: int) -> Dict[str, Any]:
             "inbound_accepted": len(inbound_accepted_list),
             "outbound_tcp": cls.classify_outbound_sockets(outbound_list),
         }
+
+    # High watermark for memory usage in bytes since process start (for logging purposes)
+    max_memory_used_bytes: float = 0.0
```
d1donlydfink: Please add new variables at the top of the class where people expect them to be.
Going hunting in the middlings is no fun.

neuro_san/service/http/server/http_server.py
d1donlydfink: The endpoint name "/snapshot" doesn't really tell what we are snapshotting.
How about something like "/resource_utilization" instead?

d1donlydfink (changes requested):
* The name "/snapshot" doesnt' really convey what's going on
* We should put handlers for "/profiler" and whatever "/snapshot" turns out to be under an env var for explicit activation
* Put variable declarations at the top of the class.

neuro_san/service/http/server/http_server.py
andreidenissov-cog: OK, /resources_utilization it is. Long and awkward name, but who cares?
As for protecting profiler+snapshot data - use undocumented env vars as protection?
What we should really do is put "admin"-level operations under auth machinery protection,
which we have none at the moment anyway. But should move in that direction eventually, right?

neuro_san/service/utils/service_resources.py
andreidenissov-cog: Moved.

vince-leaf:
This is cool. This /snapshot endpoint already returns RSS, FDs, and sockets. My #933 soak test needs to poll that endpoint — I don't need to write any psutil code myself. 

@andreidenissov-cog, my intention for my #933 soak tests is also to check for CPU usage (CPU Leak).
Can we add cpu_percent and num_threads to get_snapshot_dict() so that the /snapshot endpoint returns everything in a single call?

neuro_san/service/utils/service_resources.py
```diff
@@ -292,3 +295,40 @@ def classify_sockets(cls, server_port: int) -> Dict[str, Any]:
             "inbound_accepted": len(inbound_accepted_list),
             "outbound_tcp": cls.classify_outbound_sockets(outbound_list),
         }
+
+    @classmethod
+    def get_memory_used_mbytes(cls) -> Tuple[float, float]:
```
vince-leaf, line 300: Add a new method get_cpu_and_threads()

neuro_san/service/http/server/http_server.py
d1donlydfink: Authorization is always optional.
Operational security is always not.

Basically the way things are now, everyone always has an endpoint to see the inner workings of the server.
Env var documented in Dockerfile whose default is off seems most reasonable way to go.
Turning this on is an admin-level decision, auth or no auth.
@donn-leaf : We will probably want this on in our own dev and staging environments. Maybe not in prod.

neuro_san/deploy/Dockerfile
```diff
@@ -369,6 +369,11 @@ ENV AGENT_AUTHORIZER_RESOURCE_KEY=AgentNetwork
 # Typically "read" from CRUD-based permissions.
 ENV AGENT_AUTHORIZER_ALLOW_RELATION=read
 
+# Whether to enable run-time statistics endpoints for neuro-san server as a whole.
+# These endpoints can be used to get information about the server's current resource usage
+# and also to control run-time service profiler.
+ENV ENABLE_RUN_TIME_STATISTICS="false"
+
```
andreidenissov-cog, line 376: Add an environment variable to control access to collecting run-time server statistics.

neuro_san/service/utils/service_resources.py
```diff
@@ -292,3 +295,60 @@ def classify_sockets(cls, server_port: int) -> Dict[str, Any]:
...
+        """
+        p = psutil.Process()
+        mem_info = p.memory_info()
+        cls.max_memory_used_bytes = max(cls.max_memory_used_bytes, mem_info.rss)
+        # Return memory sizes in megabytes
+        return mem_info.rss / (1024 * 1024), cls.max_memory_used_bytes / (1024 * 1024)
+
+    @classmethod
+    def get_cpu_load(cls) -> float:
+        """
```
andreidenissov-cog, line 314: New function to get estimated CPU load - in 0..100% range.

d1donlydfink (approved):
Added @donn-leaf so he knows about the new env var.

PR #942, opened 2026-05-15
Add mock LLM http server to neuro-san test framework.

andreidenissov-cog:
This PR adds a simple mock LLM server to neuro-san test infrastructure.
Model emulated is of "OpenAI" type, that is satisfies OpenAI API.
README.md describes how to use this server.


tests/mock_llm_server/chat_completions_handler.py
```diff
@@ -0,0 +1,254 @@
...
+                completion_id, created, model, {"content": token}))
+            if self.state.stream_token_delay > 0:
+                await asyncio.sleep(self.state.stream_token_delay)
+        await self._send_event(self._chunk_envelope(
+            completion_id, created, model, {}, finish_reason="stop"))
+
+    async def _send_event(self, payload: Dict[str, Any]) -> None:
+        """Write one `data: {json}\\n\\n` SSE frame and flush it to the client."""
+        serialized_payload = json.dumps(payload, separators=(",", ":"))
+        self.write(f"data: {serialized_payload}\n\n")
```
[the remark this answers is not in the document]
andreidenissov-cog: This response should be safe, since this is pure SSE message for streaming mode.

tests/mock_llm_server/README.md
```diff
@@ -0,0 +1,257 @@
+# Mock LLM Server
+
```
andreidenissov-cog, line 2: Docs for how to use this mock LLM server.

tests/mock_llm_server/mock_llm_server.py
```diff
@@ -0,0 +1,139 @@
...
+        args.max_latency = max(args.min_latency, args.max_latency)
+        args.stream_token_delay = max(0.0, args.stream_token_delay)
+        try:
+            asyncio.run(cls.run(args))
+        except KeyboardInterrupt:
+            logging.info("mock-llm shutting down")
+
+
+if __name__ == "__main__":
+    MockLlmServer.main()
```
d1donlydfink, line 139: So this is a whole server with its own CLI for test purposes that will get installed on any neuro-san installation.  This actually could open up a whole new security hole because now there's another executable that could be started.  Is that what we really want?

Maybe either:
* Move this to tests (as opposed to neuro_san/test) so this at least does not get distributed with neuro-san itself
* Move to its own repo so people can install only if they want it.

Judging by the imports in all the new files, this is completely separable from neuro-san itself.

d1donlydfink (changes requested):

tests/mock_llm_server/mock_llm_server.py
andreidenissov-cog: OK, I think separate repo is kinda overkill at this point. So I moved it to tests/

tests/mock_llm_server/chat_completions_handler.py
```diff
@@ -0,0 +1,254 @@
...
```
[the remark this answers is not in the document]
andreidenissov-cog: Already commented on this - in SSE mode, this serialized json is treated as purely data,
so will not be interpreted. JSON serialization provides necessary protection by symbols escaping.

d1donlydfink (approved):

tests/mock_llm_server/chat_completions_handler.py
```diff
@@ -0,0 +1,254 @@
...
+        model: str,
+        delta: Dict[str, Any],
+        finish_reason: Optional[str] = None,
+    ) -> Dict[str, Any]:
+        """Build one OpenAI chat.completion.chunk envelope."""
+        return {
+            "id": completion_id,
+            "object": "chat.completion.chunk",
+            "created": created,
+            "model": html.escape(str(model)),
```
vince-leaf, line 161: @andreidenissov-cog 

One minor bug?  I noticed: the HTML.escape() calls on tool names and the model name in chat_completions_handler.py might or will cause issues.

Since this is a JSON API (not an HTML page), html.escape ends up mangling the values—for example, & becomes &amp; in the JSON payload. Any OpenAI-compatible client parsing that JSON would see corrupted tool names and fail to match them. json.dumps already handles safe JSON serialization, so the extra escaping isn't needed here.

example:

Without htm.escape
"name": "search_q&a_tool"  
With html.escape (broken)
"name": "search_q&amp;a_tool"

tests/mock_llm_server/chat_completions_handler.py
```diff
@@ -0,0 +1,254 @@
...
+    def _chat_completion_envelope(
+        model: str,
+        message: Dict[str, Any],
+        finish_reason: str,
+    ) -> Dict[str, Any]:
+        return {
+            "id": f"chatcmpl-{uuid.uuid4().hex[:24]}",
+            "object": "chat.completion",
+            "created": int(time.time()),
+            "model": html.escape(str(model)),
```
vince-leaf, line 241: Same comment above related to the HTML.escape()

tests/mock_llm_server/chat_completions_handler.py
```diff
@@ -0,0 +1,254 @@
...
+        # Opening chunk: role + tool_call skeleton with the function name.
+        await self._send_event(self._chunk_envelope(completion_id, created, model, {
+            "role": "assistant",
+            "content": None,
+            "tool_calls": [{
+                "index": 0,
+                "id": tool_call_id,
+                "type": "function",
+                "function": {
+                    "name": html.escape(tool_name),
```
vince-leaf, line 190: Same comment related to the HTML.escape()

PR #947, opened 2026-05-18
NS-930 Remove OpenAI dependency

d1donlydfink:
We should depend on langchain-openai and its versions, not OpenAI directly.

As of now, at the end of this exercise, we get openai==2.37.0

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -36,7 +36,6 @@ langchain>=1.2.0,<2.0
 langchain-core>=1.2.5,<2.0
 langchain-classic>=1.0.0,<2.0
 langchain-community>=0.4,<1.0
-openai>=1.54.1,<2.0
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -36,7 +36,6 @@ langchain>=1.2.0,<2.0
 langchain-core>=1.2.5,<2.0
 langchain-classic>=1.0.0,<2.0
 langchain-community>=0.4,<1.0
-openai>=1.54.1,<2.0
```
d1donlydfink, line 39: Remove direct OpenAI dependency. Let langchain-openai (below) guide what version is pulled in.

Noravee (approved):
LGTM.

andreidenissov-cog (approved):
LGTM.

PR #949, opened 2026-05-19
Try next fallback when LLM construction raises ValueError

### LINKED ISSUE #948: LLM fallback loop aborts on construction error, never tries later fallbacks
## Summary

When an agent network specifies multiple fallbacks in its `llm_config`, a `ValueError` raised while constructing an earlier fallback (e.g. missing API key env var, invalid credentials) propagates out of the loop and prevents later fallbacks from being tried.

## Reproduce

```json
"llm_config": {
    "fallbacks": [
        { "class": "anthropic", "model_name": "claude-sonnet-4-6" },
        { "class": "openai",    "model_name": "gpt-5.2" }
    ]
}
```
Run with only `OPENAI_API_KEY` set in the environment (no `ANTHROPIC_API_KEY`, no `sly_data`).

**Expected**: Anthropic fails to construct, the OpenAI fallback is used, the agent answers.

**Actual**: Anthropic construction raises ValueError (re-raised from `API_KEY_ERRORS` in `default_llm_factory.py:397-405`. The fallback loop in `langchain_run_context.py:217` does not catch it, so the loop aborts and the OpenAI fallback is never attempted.

## Root cause

The fallback loop handles two non-fatal outcomes per iteration (`None` and `Set[str]-of-missing-keys`), but treats a construction `ValueError` as fatal. There is no semantic difference from the user's perspective — both mean "this fallback isn't usable, try the next one."

The current workaround is to mark each provider's API key as `sly_data` in the hocon, which causes the factory to return a set before construction is attempted (bypassing the `ValueError` path) — but this forces all keys through `sly_data` and prevents straightforward env-var-only configs.

Noravee:
The fallback loop in `LangChainRunContext` handled two non-fatal outcomes per iteration (`None` and `Set[str]-of-missing-keys`), but treated a `ValueError` from `create_llm` as fatal — aborting the loop and preventing later fallbacks from being attempted.

This meant a config like:

```json
  "fallbacks": [
    { "class": "anthropic", "model_name": "claude-sonnet-4-6" },
    { "class": "openai",    "model_name": "gpt-5.2" }
  ]
```

would fail outright when `ANTHROPIC_API_KEY` was unset, even if `OPENAI_API_KEY` was available — because `ChatAnthropic` construction raised an API-key error (re-raised as `ValueError` by `DefaultLlmFactory`) and the openai fallback was never tried.

## Changes

- `LangChainRunContext.create_agent_with_fallbacks`: Wrap `create_llm` in `try/except ValueError`, record the message in a `construction_errors` list, and continue to the next fallback. When no fallback succeeds, the final raised `ValueError` includes the collected construction errors under a clear section header ("The following errors occurred while constructing LLMs:") so the operator can see why each candidate failed.

- `DefaultLlmFactory`: demote API-key error log from `ERROR` to `INFO`. Previously, a per-attempt API-key failure was logged at `ERROR` level — which made sense when an API-key failure was always fatal. Now that the fallback loop can recover from such a failure by trying the next fallback, leaving the log at `ERROR` produced noisy/misleading `"API KEY error detected: …"` lines even on the recovered (successful) path. Demoting to `INFO` keeps the technical details available for debugging without alarming operators when a fallback transparently saves the day. The all-fallbacks-failed case still surfaces a clear, aggregated `ValueError` to the caller, so visibility of true failures isn't lost.

- Added unit tests in `tests/neuro_san/internals/run_context/langchain/core/test_langchain_run_context.py`

Fix #948

### DIFF
--- a/neuro_san/internals/run_context/langchain/core/langchain_run_context.py
+++ b/neuro_san/internals/run_context/langchain/core/langchain_run_context.py
@@ -257,6 +257,7 @@ class LangChainRunContext(RunContext):
                 error += "of the following set in sly_data.llm_config:\n"
                 error += "\n".join(sorted(required_llm_config)) + "\n"
             if len(construction_errors) > 0:
+                error += "\nThe following errors occurred while constructing LLMs:\n"
                 error += "\n".join(construction_errors) + "\n"
             raise ValueError(error)
 
--- a/neuro_san/internals/run_context/langchain/core/langchain_run_context.py
+++ b/neuro_san/internals/run_context/langchain/core/langchain_run_context.py
@@ -214,10 +214,18 @@ class LangChainRunContext(RunContext):
         required_llm_config: Set[str] = set()
 
         # Go through the list of fallbacks in the config.
+        construction_errors: List[str] = []
         for fallback in fallbacks:
 
             # Create a model we might use.
-            one_llm_resources: LangChainLlmResources | Set[str] = llm_factory.create_llm(fallback, sly_data)
+            # If construction fails (e.g. missing API key in env), record the error and
+            # try the next fallback rather than aborting the whole loop.
+            try:
+                one_llm_resources: LangChainLlmResources | Set[str] = llm_factory.create_llm(fallback, sly_data)
+            except ValueError as exception:
+                construction_errors.append(str(exception))
+                continue
+
             if one_llm_resources is None:
                 # Nothing to use or report.
                 # Skip for now, a fallback might still be fulfilled.
@@ -248,6 +256,8 @@ class LangChainRunContext(RunContext):
                 error += "\nLLM operation for this agent requires at least one "
                 error += "of the following set in sly_data.llm_config:\n"
                 error += "\n".join(sorted(required_llm_config)) + "\n"
+            if len(construction_errors) > 0:
+                error += "\n".join(construction_errors) + "\n"
             raise ValueError(error)
 
         if len(chain_fallbacks) > 0:
--- a/neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
+++ b/neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
@@ -405,16 +405,44 @@ class DefaultLlmFactory(ContextTypeLlmFactory, LangChainLlmFactory):
                     # fallback path, while still being on by default in most deployments
                     # so the trail isn't lost.
                     #
+                    # Scope of this check: only failures that fire at LLM *construction*
+                    # time land here — i.e. the key is missing (None) or of the wrong
+                    # data type (e.g. list/dict supplied via sly_data where a string is
+                    # expected). A key that is structurally a string but rejected by the
+                    # provider (wrong/expired/over-quota) doesn't fail construction; it
+                    # fails at *runtime* on the first API call, and is handled by the
+                    # analogous check in RunContextRunnable.invoke_agent_chain()
+                    # (see run_context_runnable.py).
+                    #
                     # The user-friendly `message` we raise is still reported to the user.
-                    # Two cases to consider:
+                    # The raised ValueError is collected by the fallback loop in
+                    # LangChainRunContext.create_agent_with_fallbacks() and, if no
+                    # fallback succeeds, aggregated into a final ValueError surfaced
+                    # to the caller. Two cases to consider:
+                    #
                     #   a) Env var missing / misconfigured — server-side fix.
                     #      The aggregated ValueError lands in server output (the operator
                     #      sees it), and this log line provides the technical detail to
-                    #      diagnose it.
+                    #      diagnose it. The surfaced text looks like:
+                    #
+                    #          No fully-specified LLM found in llm_config or fallbacks.
+                    #          The following errors occurred while constructing LLMs:
+                    #
+                    #          A value for the OPENAI_API_KEY environment variable must be
+                    #          correctly set in the neuro-san server or run-time environment
+                    #          in order to use this agent network.
+                    #          ...
+                    #
                     #   b) sly_data was supposed to supply the key — client-side fix.
                     #      The aggregated ValueError surfaces to the chat client so the
                     #      end user (or calling system) knows to provide the key in
-                    #      sly_data.llm_config.
+                    #      sly_data.llm_config. The surfaced text looks like:
+                    #
+                    #          No fully-specified LLM found in llm_config or fallbacks.
+                    #          LLM operation for this agent requires at least one of the
+                    #          following set in sly_data.llm_config:
+                    #          anthropic_api_key
+                    #
                     self.logger.info("API KEY error detected: %s", str(exception))
                     raise ValueError(message) from exception
                 found_exception = exception
--- a/neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
+++ b/neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
@@ -401,7 +401,7 @@ class DefaultLlmFactory(ContextTypeLlmFactory, LangChainLlmFactory):
                 if message is not None:
                     # Log the error with technical details for debugging purposes,
                     # but we are returning a more user-friendly message to the client.
-                    self.logger.error("API KEY error detected: %s", str(exception))
+                    self.logger.debug("API KEY error detected: %s", str(exception))
                     raise ValueError(message) from exception
                 found_exception = exception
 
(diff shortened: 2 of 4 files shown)


neuro_san/internals/run_context/langchain/core/langchain_run_context.py
```diff
@@ -248,6 +256,8 @@ def create_agent_with_fallbacks(self, instructions: str) -> Runnable:
                 error += "\nLLM operation for this agent requires at least one "
                 error += "of the following set in sly_data.llm_config:\n"
                 error += "\n".join(sorted(required_llm_config)) + "\n"
+            if len(construction_errors) > 0:
+                error += "\n".join(construction_errors) + "\n"
```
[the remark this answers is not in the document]
Noravee: Done.

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
```diff
@@ -214,10 +214,18 @@ def create_agent_with_fallbacks(self, instructions: str) -> Runnable:
...
 
             # Create a model we might use.
-            one_llm_resources: LangChainLlmResources | Set[str] = llm_factory.create_llm(fallback, sly_data)
+            # If construction fails (e.g. missing API key in env), record the error and
+            # try the next fallback rather than aborting the whole loop.
+            try:
+                one_llm_resources: LangChainLlmResources | Set[str] = llm_factory.create_llm(fallback, sly_data)
+            except ValueError as exception:
+                construction_errors.append(str(exception))
+                continue
```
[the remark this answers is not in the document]
Noravee: Done.

Noravee, line 227: Catch the missing API key in env error as ValueError and add to the list.

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
```diff
@@ -248,6 +256,9 @@ def create_agent_with_fallbacks(self, instructions: str) -> Runnable:
                 error += "\nLLM operation for this agent requires at least one "
                 error += "of the following set in sly_data.llm_config:\n"
                 error += "\n".join(sorted(required_llm_config)) + "\n"
+            if len(construction_errors) > 0:
+                error += "\nThe following errors occurred while constructing LLMs:\n"
+                error += "\n".join(construction_errors) + "\n"
             raise ValueError(error)
```
Noravee, line 262: Combine the list and report error if all llm construction failed.

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
```diff
@@ -401,7 +401,7 @@ def create_llm_resources(self, config: Dict[str, Any]) -> LangChainLlmResources:
                 if message is not None:
                     # Log the error with technical details for debugging purposes,
                     # but we are returning a more user-friendly message to the client.
-                    self.logger.error("API KEY error detected: %s", str(exception))
+                    self.logger.debug("API KEY error detected: %s", str(exception))
```
Noravee: Change the log level of API key error to debug.

Note that if it is the real error, it will still get report to the user.

d1donlydfink: Debug logging will be turned off by default for most deployments, so these will get lost most of the time.
When users have problems with their own keys (even via sly_data), it will be helpful for a sysadmin to correlate users failures.
It can also be helpful for a sysadmin to gather data on how people are using/failing to use the system.

Maybe:
1) log at INFO level
2) Add a comment as to how "it will still get reported to the user" since it is being demoted and that is still very important.  There are 2 reporting cases to consider:
        a) The normal case where env var is missing or not set correctly - should appear in server output
        b) The special case where sly_data is supposed to give they key - should appear in user chat output.

tests/neuro_san/internals/run_context/langchain/core/test_langchain_run_context.py
```diff
@@ -0,0 +1,212 @@
...
+from neuro_san.internals.run_context.langchain.core.langchain_run_context import LangChainRunContext
+
+
+# Both markers needed so conftest.py skips the OPENAI_API_KEY requirement:
+# these tests fully mock the LLM factory, so no real provider key is required.
+pytestmark = [pytest.mark.non_default_llm_provider, pytest.mark.ollama]
+
+
+class TestLangChainRunContext:
+    """Tests for the LangChainRunContext class."""
```
d1donlydfink: First of all, several million points out of ten for creating these tests.
However:
The great thing about tests is that their name often reflects what is being tested,
and most often the scope of the class involved defines what is being tested thoroughly.

That is now not necessarily the case here.
It would be ideal to break the functionality out of LangChainRunContext into its own separate class,
however I concede that would be tricky and perhaps out of scope of this PR.

The name of this class implies that all of LangChainRunContext is being tested and it most certainly is not because it is one of the system's more complex beast.  Instead we can rename the test/file to be something like TestLangChainContextFallbacks to reflect the slice of functionality that is being tested.
(Frankly, testing all the functionality of LangChainRunContext would be too big for one file anyway)

d1donlydfink (commented):
* log level in DefaultLlmFactory -> info
* Comments there as to 2 user notification scenarios now that log has been downgraded.
* rename test to TestLangChainRunContextFallbacks.

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
d1donlydfink: You should at least manually test the messaging in both of these scenarios.

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
d1donlydfink: Are we sure the construction errors will never actually contain the value for the key itself?

Noravee: The text reaching the chat client is safe since we create it in `ApiKeyErrorCheck.check_for_api_key_exception()`. 

However, the leak can come from logging of pydantic error (type check e.g. passing a list instead of a string).

Create #951 to keep this PR stay scoped.

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
Noravee: Log level is now set to `info` and comments are added.

This has been tested manually and the unit test also cover it.

Note that the comment also add that the api key check in during LLM construction get triggered by missing the key or wrong data type only. If the key is wrong, it will get triggered by the check in runtime (runnable run context).

tests/neuro_san/internals/run_context/langchain/core/test_langchain_run_context.py
Noravee: Good point — agreed that the current name implies broader coverage than the test actually provides. Renamed the class (and file) to `TestLangChainContextLlmFallbacks` / `test_langchain_context_llm_fallbacks.py` to reflect that this is specifically the LLM-fallback slice of `LangChainRunContext.create_agent_with_fallbacks()`, not the class as a whole.

tests/neuro_san/internals/run_context/langchain/core/test_langchain_run_context_llm_fallbacks.py
```diff
@@ -0,0 +1,217 @@
...
+from unittest.mock import MagicMock
+
+import pytest
+
+from neuro_san.internals.run_context.langchain.core.langchain_run_context import LangChainRunContext
+
+
+# Both markers needed so conftest.py skips the OPENAI_API_KEY requirement:
+# these tests fully mock the LLM factory, so no real provider key is required.
+pytestmark = [pytest.mark.non_default_llm_provider, pytest.mark.ollama]
```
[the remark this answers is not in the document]
Noravee: warning-only, not blocking

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
```diff
@@ -399,9 +399,51 @@ def create_llm_resources(self, config: Dict[str, Any]) -> LangChainLlmResources:
...
+                    #      end user (or calling system) knows to provide the key in
+                    #      sly_data.llm_config. The surfaced text looks like:
+                    #
+                    #          No fully-specified LLM found in llm_config or fallbacks.
+                    #          LLM operation for this agent requires at least one of the
+                    #          following set in sly_data.llm_config:
+                    #          anthropic_api_key
+                    #
+                    self.logger.info("API KEY error detected: %s", str(exception))
                     raise ValueError(message) from exception
```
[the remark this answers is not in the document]
Noravee: See #951

d1donlydfink (approved):

PR #954, opened 2026-05-20
NS-916 Explicitly disable LLM-level streaming.

andreidenissov-cog:
This PR disables LLM-level streaming (per-token streaming logic) in all LLM clients we are constructing.
It should not affect LLM functionality we are actually using in neuro-san service,
but gives some performance gain, relatively modest though.


neuro_san/internals/run_context/langchain/llms/azure_llm_policy.py
```diff
@@ -149,7 +149,12 @@ def create_llm(self, config: Dict[str, Any], model_name: str, client: Any) -> Ba
             logprobs=config.get("logprobs"),
             top_logprobs=config.get("top_logprobs"),
             logit_bias=config.get("logit_bias"),
-            streaming=True,  # streaming is always on. Without it token counting will not work.
+            # Disable streaming: neuro-san does not consume per-token chunks from the model,
+            # and disabling streaming reduces per-request LangChain pipeline overhead. Token
+            # usage is still reported via llm_output["token_usage"] from _agenerate().
+            # We pass streaming explicitly so langchain_core._should_stream() recognizes it
+            # as opted-out even when a streaming-aware callback is attached to the run manager.
+            streaming=False,
```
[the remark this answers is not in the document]
andreidenissov-cog: stream_options removed.

neuro_san/internals/run_context/langchain/llms/openai_llm_policy.py
```diff
@@ -141,7 +141,13 @@ def create_llm(self, config: Dict[str, Any], model_name: str, client: Any) -> Ba
...
             top_logprobs=config.get("top_logprobs"),
             logit_bias=config.get("logit_bias"),
-            streaming=True,  # streaming is always on. Without it token counting will not work.
+            # Disable streaming: neuro-san does not consume per-token chunks from the model,
+            # and disabling streaming reduces per-request LangChain pipeline overhead. Token
+            # usage is still reported via llm_output["token_usage"] from _agenerate().
+            # We pass streaming explicitly (rather than relying on the False default) so that
+            # langchain_core._should_stream() recognizes it as opted-out even when a
+            # streaming-aware callback handler is attached to the run manager.
+            streaming=False,
```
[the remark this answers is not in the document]
andreidenissov-cog: set stream_usage=False.

neuro_san/internals/run_context/langchain/llms/azure_llm_policy.py
```diff
@@ -111,12 +111,6 @@ def create_llm(self, config: Dict[str, Any], model_name: str, client: Any) -> Ba
         :param client: The web client to use (if any)
         :return: A BaseLanguageModel (can be Chat or LLM)
         """
-        model_kwargs: Dict[str, Any] = {
-            "stream_options": {
-                "include_usage": True
-            }
-        }
-
```
andreidenissov-cog, line 119: Not needed anymore for tokens counting.

neuro_san/internals/run_context/langchain/llms/azure_llm_policy.py
```diff
@@ -184,10 +185,7 @@ def create_llm(self, config: Dict[str, Any], model_name: str, client: Any) -> Ba
             openai_api_type=self.get_value_or_env(config, "openai_api_type",
                                                   "OPENAI_API_TYPE", client),
 
-            model_version=config.get("model_version"),
-
-            # Needed for token counting
-            model_kwargs=model_kwargs,
+            model_version=config.get("model_version")
```
andreidenissov-cog, line 188: model_kwargs not needed now.

neuro_san/internals/run_context/langchain/llms/anthropic_llm_policy.py
```diff
@@ -65,9 +65,14 @@ def create_llm(self, config: Dict[str, Any], model_name: str, client: Any) -> Ba
...
-            streaming=True,  # streaming is always on. Without it token counting will not work.
-            # Set stream_usage to True in order to get token counting chunks.
-            stream_usage=True,
+            # Disable streaming: neuro-san does not consume per-token chunks from the model,
+            # and disabling streaming reduces per-request LangChain pipeline overhead.
+            # Token usage is collected from AIMessage.usage_metadata in LlmTokenCallbackHandler.
+            # We set streaming explicitly (rather than relying on the False default) so that
+            # langchain_core._should_stream() recognizes it as opted-out
+            # even when a streaming-aware callback is attached.
+            streaming=False,
```
andreidenissov-cog, line 74: Here and below for other models - update comment to describe what is actually done for tokens counting.

d1donlydfink: I am OK with this if @Noravee is - he has been master of token counting.

d1donlydfink (dismissed):
I'm OK with this if @Noravee is.

neuro_san/internals/run_context/langchain/llms/anthropic_llm_policy.py
Noravee: I think this comment is incorrect. `BaseChatModel._should_stream()` decides whether to stream through the following steps:

1. If the model doesn't implement `_stream` → False
2. If async and neither `_astream` nor `_stream` is implemented → False
3. If streaming is disabled, which includes: `disable_streaming=True`, `tool_calling` with tools passed, a falsy `stream` kwarg, or `streaming=False` **explicitly set in the constructor** (checked via `model_fields_set`) → False
4. If a truthy `stream` kwarg is passed at call time → True
5. If `streaming=True` was explicitly set in the constructor → True
6. If a callback handler that extends `_StreamingCallbackHandler` is attached → True

For point 6, as far as I can tell only `LogStreamCallbackHandler` and `_AstreamEventsCallbackHandler` implement `_StreamingCallbackHandler` — both are internal LangChain classes attached automatically by `.astream_log()` and `.astream_events()`. Our `LlmTokenCallbackHandler(AsyncCallbackHandler)` does not extend `_StreamingCallbackHandler`, so it will never trigger streaming.

So there is no need to set explicitly `streaming=False` since the default is also `False`. However, for OpenAI, AzureOpenAI, Anthropic, Gemini, and Bedrock — these all define `streaming` as a proper Pydantic field, so passing `streaming=False` correctly lands in `model_fields_set` and opts out via step 3.

However, for Ollama and Nvidia, `streaming` is not a defined model field on their classes — passing it causes it to go into `model_kwargs` instead, so it never appears in `model_fields_set` and has no effect on `_should_stream()` at all. Nvidia even emits a warning: "streaming was transferred to model_kwargs". For these two providers the parameter is a no-op either way.

Noravee: So I think if you want you can leave `streaming=False` for OpenAI, AzureOpenAI, Anthropic, Gemini, and Bedrock, but remove it for Ollama and Nvidia or you can remove them all for consistency,

andreidenissov-cog: Great analysis! So OK I've removed this parameter for Ollama and Nvidia models,
left it for all the rest as a visual reminder that we **do not** want streaming at this level.

Noravee (approved):
LGTM.

PR #959, opened 2026-05-21
NS-956 Slightly better langfuse reporting

d1donlydfink:
Slightly better langfuse reporting. Simpler. Uses the ContextVar like the plugin does.

The parentage of chains is still not good, but you can start to figure out what belongs to what when you look at the timeline view.   RunnableWithFallbacks (which should probably be attached to RunContextRunnalbe) and LangGraph (which should probably be attached to RunnableWithFallbacks) are still attaching themselves to the main observation from the TracingContext parent and not to what called them.

I've spent enough time on this for now. There are still some tricks to play with setting explicit parent_run_ids, but they get hackier and wackier.
Might pick this up again later when there isn't as many other things that need attention.

With this the plugin should not be necessary any more.

### DIFF
--- a/neuro_san/internals/run_context/langchain/tracing/langfuse_tracing_context.py
+++ b/neuro_san/internals/run_context/langchain/tracing/langfuse_tracing_context.py
@@ -18,14 +18,15 @@ from __future__ import annotations
 
 from typing import Any
 from typing import Dict
-from typing import List
 from typing import Type
 
+from contextvars import ContextVar
 from datetime import datetime
 from socket import gethostname
 
 from langchain_core.callbacks.base import BaseCallbackHandler
 from langchain_core.runnables.base import Runnable
+from langchain_core.tracers.context import register_configure_hook
 
 from leaf_common.config.resolver_util import ResolverUtil
 
@@ -38,6 +39,31 @@ class LangfuseTracingContext(LangChainTracingContext):
     """
     TracingContext implementation for runs that use Langfuse.
     """
+
+    @staticmethod
+    def register():
+        """
+        Globally egister the Langfuse CallbackHandler, if available
+        """
+
+        context_var = ContextVar("langfuse_handler", default=None)
+
+        # See if we can create a new langfuse handler instance.
+        callback_handler_type: Type[BaseCallbackHandler] = None
+        callback_handler_type = ResolverUtil.create_type("langfuse.langchain.CallbackHandler",
+                                                         raise_if_not_found=False)
+        if callback_handler_type is not None:
+
+            callback_handler: BaseCallbackHandler = callback_handler_type()
+
+            context_var.set(callback_handler)
+            register_configure_hook(context_var, inheritable=True)
+
+        return context_var
+
+    # Global context variable for the langfuse callback handler.
+    # Do the register() once at class load time.
+    HANDLER_CONTEXT_VAR: ContextVar = register()
 
     def __init__(self, run_target: RunTarget,
                  config: Dict[str, Any],
@@ -52,7 +78,6 @@ class LangfuseTracingContext(LangChainTracingContext):
         super().__init__(run_target=run_target, config=config)
 
         self.parent_context: LangfuseTracingContext = parent_context
-        self.callback_handler: BaseCallbackHandler = None
         self.main_span: Any = None
 
         # See if we can get a langfuse handler instance.
@@ -84,6 +109,47 @@ If you didn't mean to use langfuse for observability, you can do this:
         :param inputs: The inputs to the chain
         :param runnable_config: The config for the runnable
         """
+        # According to langfuse docs, this should be safe for use in async code.
+        if self.main_span is not None:
+            with self.main_span:
+                await super().ainvoke(chain, inputs, runnable_config)
+        else:
+            await super().ainvoke(chain, inputs, runnable_config)
+
+    def create_main_span(self, runnable_config: Dict[str, Any]):
+        """
+        Create the main span for the run
+        :param runnable_config: The config for the runnable
+        """
+
+        if self.main_span is not None:
+            return
+
+        if self.parent_context is not None:
+            return
+
+        run_name: str = runnable_config.get("run_name")
+
+        # No need to ResolverUtil absolutely everything, but we still need to locally import
+        # for the rest of the system to behave without langfuse installed.
+
+        # pylint: disable=import-outside-toplevel
+        from langfuse import get_client
+
+        # Get the langfuse client
+        langfuse_client: Any = get_client()
+        self.main_span = langfuse_client.start_as_current_observation(as_type="agent", name=run_name)
+
+    def augment_config(self, runnable_config: Dict[str, Any]) -> Dict[str, Any]:
+        """
+        Augment the configuration however the implementation sees fit (if at all).
+        :param runnable_config: The config for the runnable
+        :return: The augmented config
+        """
+        self.create_main_span(runnable_config)
+
+        runnable_config["neuro_san_tracing_context"] = self
+
         # Get the user_id for the trace
         empty: Dict[str, Any] = {}
         request_metadata: Dict[str, Any] = runnable_config.get("metadata", empty)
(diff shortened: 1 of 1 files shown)


neuro_san/internals/run_context/langchain/tracing/langfuse_tracing_context.py
```diff
@@ -39,6 +40,35 @@ class LangfuseTracingContext(LangChainTracingContext):
...
+
+            # Register the callback handler via the context variable
+            context_var.set(callback_handler)
+            register_configure_hook(context_var, inheritable=True)
+
+        return context_var
+
+    # Global context variable for the langfuse callback handler.
+    # Do the register() once at class load time.
+    HANDLER_CONTEXT_VAR: ContextVar = register()
```
d1donlydfink, line 70: Do a static register of a global context variable to keep the Langfuse CallbackHandler in the callbacks list for any langchain trace.  This is more or less what the studio-level plugin does and it gets more consistent reporting results this way. So keep that.

As a result, this should mean we don't really need the plugin any more.

neuro_san/internals/run_context/langchain/tracing/langfuse_tracing_context.py
```diff
@@ -51,13 +81,16 @@ def __init__(self, run_target: RunTarget,
...
+
+        # Keep a session_id for any child TracingContext to use in its langfuse config for the run.
+        self.session_id: str = None
 
         # See if we can get a langfuse handler instance.
-        handler_type = ResolverUtil.create_type("langfuse.langchain.CallbackHandler", raise_if_not_found=False)
-        if handler_type is None:
+        # handler_type = ResolverUtil.create_type("langfuse.langchain.CallbackHandler", raise_if_not_found=False)
+        # if handler_type is None:
+        if self.HANDLER_CONTEXT_VAR.get() is None:
```
d1donlydfink, line 93: Check the context variable to see if we have anything as the cue for the error message. All the ResolverUtil stuff has already been done and if that failed, then we need the error message.

neuro_san/internals/run_context/langchain/tracing/langfuse_tracing_context.py
```diff
@@ -68,6 +101,18 @@ def __init__(self, run_target: RunTarget,
...
+
+        # No need to ResolverUtil absolutely everything, but we still need to locally import
+        # for the rest of the system to behave without langfuse installed.
+        # pylint: disable=import-outside-toplevel
+        from langfuse import Langfuse
+        from langfuse import get_client
+        from opentelemetry.util._decorator import _AgnosticContextManager
+
+        self.langfuse_client: Langfuse = get_client()
+        self.main_span: _AgnosticContextManager[Any] = None
```
d1donlydfink, line 114: Keep track of some langfuse entities in the instance.

neuro_san/internals/run_context/langchain/tracing/langfuse_tracing_context.py
```diff
@@ -84,25 +129,12 @@ async def ainvoke(self, chain: Runnable, inputs: Any, runnable_config: Dict[str,
...
-        session_id: str = f"{request_id}@{now.strftime('%Y-%m-%d-%H:%M:%S.%f')}"
-        hostname: str = gethostname()
-        session_id: str = f"{session_id}@{hostname}"
-
-        # pylint: disable=import-outside-toplevel
-        from langfuse import propagate_attributes
-
-        # According to langfuse docs, this should be safe for use in async code.
-        with self.main_span:
-            propagate_attributes(user_id=user_id, session_id=session_id)
```
d1donlydfink, line 105: Above: We don't need to use propagate_attributes any more if we use the langfuse_* keys in the metadata config for the run. This construction has effectively moved to augment_config() below.

neuro_san/internals/run_context/langchain/tracing/langfuse_tracing_context.py
```diff
@@ -84,25 +129,12 @@ async def ainvoke(self, chain: Runnable, inputs: Any, runnable_config: Dict[str,
...
+        if self.main_span is not None:
+            # We have a main span. Use it as the context.
+            # pylint: disable=not-context-manager
+            with self.main_span:
+                await super().ainvoke(chain, inputs, runnable_config)
+        else:
             await super().ainvoke(chain, inputs, runnable_config)
```
d1donlydfink, line 138: We still want to use the main_span as a context if we have one.
We leave the policy as to when to create a main_span in the create_main_span() method below.

neuro_san/internals/run_context/langchain/tracing/langfuse_tracing_context.py
```diff
@@ -112,74 +144,72 @@ def create_main_span(self, runnable_config: Dict[str, Any]):
...
-            # Get the langfuse client
-            langfuse_client: Any = get_client()
-            self.main_span = langfuse_client.start_as_current_observation(as_type="span", name=run_name)
-        else:
-            # Get the langfuse client
-            langfuse_client: Any = get_client()
-            self.main_span = langfuse_client.start_as_current_observation(as_type="agent", name=run_name)
+        # This "agent" type gets us the nice little icon in the langfuse UI
+        # According to langfuse docs, this should be safe for use in async code.
+        self.main_span = self.langfuse_client.start_as_current_observation(as_type="agent", name=run_name)
```
d1donlydfink, line 162: We were doing the same thing regardless of parent context.
Now we just do the same thing with the langfuse_client member variable.

neuro_san/internals/run_context/langchain/tracing/langfuse_tracing_context.py
```diff
@@ -112,74 +144,72 @@ def create_main_span(self, runnable_config: Dict[str, Any]):
...
 
-    def maybe_create_handler(self, runnable_config: Dict[str, Any]):
+    def augment_config(self, runnable_config: Dict[str, Any]) -> Dict[str, Any]:
```
d1donlydfink, line 164: Not the greatest diff. 

maybe_create_handler() goes away because we have the global thing via context var going on.
Instead we now set up langfuse_* keys in augment_config().

neuro_san/internals/run_context/langchain/tracing/langfuse_tracing_context.py
```diff
@@ -112,74 +144,72 @@ def create_main_span(self, runnable_config: Dict[str, Any]):
...
-        if self.callback_handler not in callbacks:
-            callbacks.append(self.callback_handler)
-        runnable_config["callbacks"] = callbacks
+            # Create a session_id for the trace.
+            self.session_id: str = f"{run_name}@{request_id}@{hostname}@{now_str}"
 
-        runnable_config["neuro_san_tracing_context"] = self
+        request_metadata["langfuse_user_id"] = user_id
+        request_metadata["langfuse_session_id"] = self.session_id
+        runnable_config["metadata"] = request_metadata
```
d1donlydfink, line 198: Use langfuse_* keys in the runnable config metadata.  This more consistently gets down to langchain sub-traces instead of propagate_attributes().

neuro_san/internals/run_context/langchain/tracing/langfuse_tracing_context.py
```diff
@@ -112,74 +144,72 @@ def create_main_span(self, runnable_config: Dict[str, Any]):
...
+        """
+        Get the parent session id.
+        We want the to be consistent for any depth of trace in the request.
+
+        :return: The parent session id
+        """
+        if self.parent_context is not None:
+            return self.parent_context.get_parent_session_id()
+
+        return self.session_id
```
d1donlydfink, line 212: We want the langfuse_session_id to be consistent for any sub-traces we set data for.
Traverse back to the parent to see what had been determined when it was created.

neuro_san/internals/run_context/langchain/tracing/langfuse_tracing_context.py
```diff
@@ -112,74 +144,72 @@ def create_main_span(self, runnable_config: Dict[str, Any]):
...
     async def flush(self):
         """
         Flush the tracing context.
         """
-        # pylint: disable=import-outside-toplevel
-        from langfuse import get_client
-        langfuse_client: Any = get_client()
-        langfuse_client.flush()
+        if self.langfuse_client is not None:
+            self.langfuse_client.flush()
```
d1donlydfink, line 219: Use the client we stored as member.

ofrancon (approved):
lgtm.

PR #964, opened 2026-05-22
NS-736 Add --list option for MCP client connection.

andreidenissov-cog:
NS-736 Allow --list option in agent_cli client for MCP connection.


andreidenissov-cog:
This feature doesn't seem all that important anymore. Closing. Keeping my local branch just in case.

PR #967, opened 2026-05-22
Add parameterized load test script for mock LLM server

### LINKED ISSUE #933: Add memory and resource leak detection for soak testing
## Description

The mock LLM server (PR #931) combined with resource metrics (#932) enables load testing with visibility, but we still cannot distinguish between high steady-state memory usage and an actual memory leak. During AND load testing, pods OOM-crashed at 15-20 concurrent users — we need to determine if this is expected per-request memory (which drops when load drops) or a leak (memory grows indefinitely regardless of load).

A **soak test** runs moderate, steady traffic for hours (vs minutes in a regular load test) to surface slow degradation: memory leaks, FD leaks, thread pool exhaustion, or connection pool growth.

## Problem

- No way to track memory RSS over time to distinguish steady-state usage from a leak
- No per-request memory delta to identify which requests don't release memory
- No visibility into file descriptor, socket, or thread count trends over long periods
- No way to pinpoint *where* in the code a leak originates

### 1. Soak test logging mode
New CLI flag `--soak-test-log <filepath>` that writes periodic resource snapshots to a JSON-lines file:
```json
{"timestamp": "2026-05-12T16:30:00Z", "memory_rss_mb": 1240, "memory_vms_mb": 2100, "fd_count": 84, "thread_count": 12, "active_requests": 8, "total_requests_served": 1523}
```
Default interval: every 10 seconds, configurable via `--soak-test-interval <seconds>`.

### 2. Per-request memory delta
Log RSS before and after each `/v1/chat/completions` request:
```json
{"request_id": "vince-3-req-7", "rss_before_mb": 1200, "rss_after_mb": 1215, "delta_mb": 15, "active_requests_after": 0}
```
When `active_requests` drops to 0 but RSS doesn't return to baseline, that signals a leak.

### 3. Resource trend tracking in MockState
Track over time:
- Memory RSS high-water mark and baseline (RSS at server start)
- File descriptor count (via existing `ServiceResources.classify_fds()`)
- Thread count (`psutil.Process().num_threads()`)
- Open socket count
- Total requests served (cumulative counter)

### 4. `/debug/tracemalloc` endpoint
Disabled by default, enabled with `--enable-debug-endpoints`:
- Starts `tracemalloc` on server boot
- Returns top 20 memory allocation sites with file/line and size
- Pinpoints *where* in the code memory is being allocated but not freed

### 5. `/debug/gc` endpoint
Same opt-in flag:
- Returns `gc.get_stats()`, `gc.get_count()`, and number of objects per generation
- Shows whether garbage collector is keeping up or objects accumulating

### 6. Leak detection summary at shutdown
When the server is stopped (Ctrl+C), print a summary:
```
=== Soak Test Summary ===
Duration: 3600s
Total requests served: 4521
Memory RSS: baseline=320MB, peak=1890MB, final=1450MB
FD count: baseline=24, peak=142, final=38
Thread count: baseline=4, peak=16, final=6
Verdict: POSSIBLE LEAK (final RSS 4.5x baseline after 0 active requests)
```

## Acceptance Criteria

- [ ] `--soak-test-log` writes periodic resource snapshots to JSON-lines file
- [ ] Per-request memory delta logged (before/after RSS)
- [ ] FD count, thread count, socket count tracked alongside memory
- [ ] `/debug/tracemalloc` returns top allocation sites when enabled
- [ ] `/debug/gc` returns garbage collector stats when enabled
- [ ] Shutdown summary compares final vs baseline resources and flags potential leaks
- [ ] Debug endpoints disabled by default (no performance impact in normal use)
- [ ] Soak test with 10 concurrent users over 1 hour produces analyzable trending data

## Test Plan

1. Start neuro-san server pointing to mock LLM with `--soak-test-log metrics.jsonl --enable-debug-endpoints`
2. Run Locust: 10 users, `wait_time=between(5, 10)`, 1 hour
3. After test, analyze `metrics.jsonl` — RSS should be stable (flat trend), not growing
4. If RSS trends upward, hit `/debug/tracemalloc` to identify top allocators
5. Verify: when load stops and active_requests=0, RSS returns to near-baseline

## Context

- Depends on #932 (resource metrics on mock LLM server) for the underlying `psutil` collection methods
- Related to #927 (memory-aware request guard) — soak test determines if OOM is from leak vs steady-state
- `psutil` already a dependency; `tracemalloc` and `gc` are Python stdlib
- Load test post-mortem: OOM at 15-20 concurrent (3GB) — need to determine root cause

vince-leaf:
Description:

Adds load_test.py — a standalone script for load-testing the neuro-san server with the mock LLM server, with resource leak detection.

Features:

CLI-driven configuration via argparse (--agent, --num-requests, --max-workers, --num-rounds, --prompt, --sly-data, --settle-time, --host, --port)
Agent-specific presets (math_guy, hello_world, chat_mock_llm_echo) auto-fill prompt and sly-data; unknown agents require explicit --prompt
Fail-fast validation — aborts immediately if neuro-san server or mock LLM server processes aren't running (local mode)
Remote server support — when --host is not localhost, psutil monitoring is auto-disabled and the script reports pass/fail + timing only
Per-round resource monitoring via psutil (RSS, FDs, threads, connections, CPU, children)
Leak analysis table with per-round deltas (RSS Delta, Threads Start → End, Thread Delta)
Overall results summary (total pass/fail, total time, avg time per request)
Usage:

**Defaults**: math_guy, 5 rounds, 10 requests, 10 workers
python load_test.py
 
100 concurrent requests over 3 rounds
python load_test.py --num-requests 100 --max-workers 100 --num-rounds 3
 
Different agent (preset auto-fills prompt)
python load_test.py --agent hello_world
 
Remote server
python load_test.py --host 172.31.11.243 --port 8080
 
Unknown agent (requires explicit --prompt)
python load_test.py --agent my_custom_agent --prompt "test input" --no-sly-data

**Dependencies**: psutil
Note: This script uses psutil directly for external process monitoring. The neuro-san server itself has built-in resource monitoring via ServiceResources (neuro_san/service/utils/service_resources.py). A future improvement could expose a /resources HTTP endpoint on the server, allowing the load test script to query resource usage remotely without needing psutil locally — which would also enable full monitoring in remote mode.


tests/load_tests/load_test_mock_llm_service.py
```diff
@@ -0,0 +1,518 @@
...
+"""
+
+import argparse
+import subprocess
+import sys
+import time
+from concurrent.futures import ThreadPoolExecutor
+from typing import Any, Dict, List, Optional, Tuple
+
+import psutil
```
vince-leaf, line 57: I can remove this dependency in a future release and use service_resources.py

tests/load_tests/load_test_mock_llm_service.py
```diff
@@ -0,0 +1,518 @@
+#!/usr/bin/env python
+"""
+Load-test script for neuro-san server using the mock LLM service.
+
```
vince-leaf, line 4: In the future, I can refactor some of these methods into classes and share them with other new load_test files.

tests/load_tests/load_test_mock_llm_service.py
```diff
@@ -0,0 +1,518 @@
...
+    },
+    "hello_world": {
+        "prompt": "Greet developers that wrote their very first program",
+        "sly_data": None,
+    },
+    "chat_mock_llm_echo": {
+        "prompt": "Hello, testing the mock LLM",
+        "sly_data": None,
+    },
+}
```
d1donlydfink: Feels like these should eventually be references to test hocon files.
We could then use the DataDrivenAgentTestDriver to be the unit of work.

tests/load_tests/load_test_mock_llm_service.py
```diff
@@ -0,0 +1,518 @@
...
+
+
+def parse_args():
```
d1donlydfink: Make all of these methods go into a class.

tests/load_tests/load_test_mock_llm_service.py
```diff
@@ -0,0 +1,518 @@
...
+        help="Prompt text to send to the agent. "
+             "Auto-filled from preset if agent is known.",
+    )
+    parser.add_argument(
+        "--sly-data",
+        type=str,
+        default=None,
+        help="JSON sly_data string. Auto-filled from preset if agent is known. "
+             "Use --no-sly-data to omit.",
+    )
```
d1donlydfink: A lot of these CLI args wouldn't be necessary if we used the existing test infrastructure.
That can be the next step (not for this PR)

d1donlydfink (changes requested):
Now: Put all the methods into a class. You will probably find that a bunch of commonly passed method arguments can be member variables. And this begins to make this framework more extensible.

Next: Use the DataDrivenAgentTestDriver as the basis for your unit of work so we can use the existing test hocon files as drivers.

FWIW: It's good you start this out under tests.  It's possible we might want to eventually (not now) move this to neuro_san/tests as something that ships.  The downside is that once something moves under neuro_san directory CheckMarx has much higher level of scrutiny because then it's also installed on every server (this is why not now).
This is fodder for a neuro-san repo breakup.

tests/load_tests/load_test_mock_llm_service.py
```diff
@@ -0,0 +1,552 @@
...
+from typing import Any, Dict, List, Optional, Tuple
+
+import psutil
+
+logging.basicConfig(level=logging.INFO, format="%(message)s")
+logger = logging.getLogger(__name__)
+
+
+class MockLlmLoadTest:
+    """Load test runner for the neuro-san server using mock LLM service."""
```
d1donlydfink, line 64: I think this is actually a good start on a load test framework - mock or not.
Might eventually want to make mock-or-not a command line option (and rename the class+file at that point)

vince-leaf:
> Now: Put all the methods into a class. You will probably find that a bunch of commonly passed method arguments can be member variables. And this begins to make this framework more extensible.

@d1donlydfink, It was on my local, did not push the changes. It's now there.

tests/load_tests/load_test_mock_llm_service.py
```diff
@@ -0,0 +1,552 @@
...
+(RSS, FDs, threads, connections), and prints a per-round summary
+with an overall leak analysis.
+
+Prerequisites:
+    1. Mock LLM server running (Terminal 1):
+       python -m tests.mock_llm_server.mock_llm_server --port 8888
+
+    2. Neuro-san server running with OPENAI_API_BASE (Terminal 2):
+       export OPENAI_API_BASE=http://localhost:8888/v1
+       python -m neuro_san.service.main_loop.server_main_loop
```
d1donlydfink, line 16: FWIW: You are already using subprocesses. What's a couple more?
You could eventually start these within this app, but good to start simple like you have.

tests/load_tests/load_test_mock_llm_service.py
```diff
@@ -0,0 +1,552 @@
...
+    python tests/load_tests/load_test_mock_llm_service.py --host 172.31.11.243 --port 8080
+"""
+
+import argparse
+import logging
+import subprocess
+import sys
+import time
+from concurrent.futures import ThreadPoolExecutor
+from typing import Any, Dict, List, Optional, Tuple
```
d1donlydfink: Nit: I generally like 1 import per line. I am old fashioned. Easier for grepping tasks to find code dependencies.

vince-leaf:
> FWIW: It's good you start this out under tests. It's possible we might want to eventually (not now) move this to neuro_san/tests as something that ships. The downside is that once something moves under neuro_san directory CheckMarx has much higher level of scrutiny because then it's also installed on every server (this is why not now).
> This is fodder for a neuro-san repo breakup.

Yes, I was thinking of the long line. I need to refactor the Dependencies: psutil 1st.

tests/load_tests/load_test_mock_llm_service.py
vince-leaf: > Make all of these methods go into a class.
Done, forgot to push. Take a look now.

tests/load_tests/load_test_mock_llm_service.py
vince-leaf: > Feels like these should eventually be references to test hocon files. We could then use the DataDrivenAgentTestDriver to be the unit of work.

As 1st, I went with this route as an added test hocon, but then I wanted more generic code in this PR and to add more unit testing on demand. This is thinking I was going to talk to you in our upcoming 1-on-1.

tests/load_tests/load_test_mock_llm_service.py
vince-leaf: Ok Done

tests/load_tests/load_test_mock_llm_service.py
vince-leaf: Done added --auto-start

tests/load_tests/load_test_mock_llm_service.py
vince-leaf: Done — split from typing import Any, Dict, List, Optional, Tuple into one per line.

vince-leaf:
Done in this PR:
✓ All methods in a class (MockLlmLoadTest)
✓ One import per line
✓ --auto-start to launch servers as subprocesses
✓ Dictionary .get() access throughout
✓ logger instead of print()
✓ OPENAI_API_BASE server validation with port mismatch detection
✓ Failure log file (/tmp/load_test_*.log)
✓ Boxed env var reminder at end of test
✓ Helpful error messages with setup instructions

vince-leaf:
**Next PR / Future work**:
Replace psutil dependency with service_resources.py
Use DataDrivenAgentTestDriver as basis for test unit of work (Dan's suggestion)
Mock-or-not as a CLI option (Dan's suggestion)
Eventually move to neuro_san/tests (after CheckMarx considerations)

1. [#968 — Replace psutil with service_resources.py](https://github.com/cognizant-ai-lab/neuro-san/issues/968)
2. [#969 — Use DataDrivenAgentTestDriver](https://github.com/cognizant-ai-lab/neuro-san/issues/969)
3. [#970 — Mock-or-not CLI option](https://github.com/cognizant-ai-lab/neuro-san/issues/970)
4. [#971 — Move to neuro_san/tests](https://github.com/cognizant-ai-lab/neuro-san/issues/971)
5. [#972 — Reduce static methods](https://github.com/cognizant-ai-lab/neuro-san/issues/972)

d1donlydfink (approved):
Not bad.
A little odd to see so many static methods, but it'll do.

vince-leaf:
> Not bad. A little odd to see so many static methods, but it'll do.

I will add a ticket above and look into this.

PR #973, opened 2026-05-23
Limiting Dynamic Class Instantiation via Environment Variable in AuthorizerFactory

tomaioo:
## Summary

Security: Dynamic Class Instantiation via Environment Variable in AuthorizerFactory

## Problem

**Severity**: `Critical` | **File**: `neuro_san/internals/authorization/factory/authorizer_factory.py:L33`

The AuthorizerFactory.create_authorizer() method uses ResolverUtil.create_instance() to dynamically instantiate a class based on the AGENT_AUTHORIZER environment variable. This is a code injection vulnerability where an attacker who can control the environment variable can cause arbitrary code execution. The method reads the class name from an environment variable and passes it directly to a reflection-based instantiation mechanism without validation.

## Solution

Implement a whitelist of allowed authorizer class names or use a registry pattern instead of dynamic class loading from environment variables. If dynamic loading is required, validate the class name against a strict allowlist before instantiation.

## Changes

- `neuro_san/internals/authorization/factory/authorizer_factory.py` (modified)

### DIFF
--- a/neuro_san/internals/authorization/factory/authorizer_factory.py
+++ b/neuro_san/internals/authorization/factory/authorizer_factory.py
@@ -30,6 +30,13 @@ class AuthorizerFactory:
     Factory class for getting the appropriate Authorizer instance
     """
 
+    ALLOWED_AUTHORIZERS = {
+        "neuro_san.internals.authorization.jwt.jwt_authorizer.JwtAuthorizer",
+        "neuro_san.internals.authorization.jwt.jwt_issuer.JwtIssuer",
+        "neuro_san.internals.authorization.null.always_no_authorizer.AlwaysNoAuthorizer",
+        "neuro_san.internals.authorization.null.always_yes_authorizer.AlwaysYesAuthorizer",
+    }
+
     @staticmethod
     def create_authorizer() -> Authorizer:
         """
@@ -39,6 +46,8 @@ class AuthorizerFactory:
 
         auth_classname: str = environ.get("AGENT_AUTHORIZER")
         if auth_classname is not None and len(auth_classname) > 0:
+            if auth_classname not in AuthorizerFactory.ALLOWED_AUTHORIZERS:
+                raise ValueError(f"Unauthorized authorizer class: {auth_classname}")
             # Note that this will raise an exception if the class is not found
             authorizer = ResolverUtil.create_instance(auth_classname, "AGENT_AUTHORIZER env var", Authorizer)
         else:


d1donlydfink (changes requested):
Thank you @tomaioo for your contribution. I appreciate your taking the time to improve the neuro-san source base.
However, I must say, I believe you are missing the point of the AGENT_AUTHORIZER env var.

How else are people supposed to extend for their own authorization solutions (like the one provided in the plugin for OpenFGA in this very neuro-san distribution for instance)?    Because you did not include the OpenFGA implementation (documented fully in neuro-san-studio), this leads me to believe that you haven't done all of your homework. 

We are not in the business of writing or affirming every authorization plugin for every authorization scheme out there. The idea is extensibility.

As for security, if the class itself does not exist in the PYTHONPATH, the ResolverUtil will throw an exception and the system will not proceed at all.  The matter of locking down who has access to the setter of AGENT_AUTHORIZER is a matter for dev-ops for the deployed cluster - that is, keeping any attacker from being able to set this configuration variable and keeping the attacker from being able to install any malicious class that could be substituted on the filesystem of the deployment.

Following this logic to the extreme, would this imply that every environment variable (like the ones that describe the AGENT_MANIFEST_FILE or AGENT_TOOL_PATH) needs to have enumerated values for validity checks? 
Do we then need to validate PYTHONPATH itself with enumerated values?

Also: In your PR summary you worry about injection of arbitrary code execution.
Do you also propose that neuro-san hard-code a list of CodedTools that can be executed by agent networks?

Reducto ad absurdum.
Perhaps I don't understand something and you can describe in detail the scenario you fear here that cannot be solved by proper devops practices.

Also: There is no blessed class for JwtAuthorizer or JwtIssuer in neuro_san.internals.
Looks like you are building in a hack for your own purposes in your own fork in an undisclosed location. 
Not a fan of forward declarations for a _limited set_ of other libraries that are not already direct dependencies.

Also: JWT is usually something used for _Authentication_ (the idea of verifying a user's identity) which is very separate from  _Authorization_ (the process that asks the question "can actor X have permission Y on resource Z?")
Normally we leave _Authentication_ to an external system like a load-balancer for a cluster.
I'm not aware of _Authorization_ systems that use JWT. 
I'm curious what you had in mind?
See https://github.com/cognizant-ai-lab/neuro-san/pull/976 for a clarification of the scope of the Authorizer interface

Please attempt to convince me of the necessity of this change. So far I don't see it.
I do perhaps see confusion and conflation, though.

PR #978, opened 2026-05-26
Nest agent_chain spans under RunContextRunnable in traces

Noravee:
## Issue
`NeuroSanRunnable` inherits from `RunnablePassthrough`, whose `ainvoke` calls `afunc=run_it` raw and then wraps a no-op `aidentity` in `_acall_with_config`. That meant the span named `"RunContextRunnable"` wrapped the identity function (~0ms, empty), while `run_it` itself ran outside any span this runnable owned. As a result, child spans created inside `run_it` (notably the `agent_chain` / `LangGraph`) inherited `parent_run_id` from the surrounding `RunnableSequence` and showed up in the trace as siblings of `RunContextRunnable` rather than its children.

## Fix
Override `ainvoke` to call `_acall_with_config(self.afunc, ...)` directly so `run_it` itself runs inside the wrap. `run_it` now gets a real `run_manager`, `var_child_runnable_config` carries this runnable's `parent_run_id`, and `agent_chain` spans nest under `RunContextRunnable` in the trace.

Note that there is no functional change to neuro-san's agent execution — purely a tracing
fix. `run_it` is invoked with the same arguments, returns the same value, and journaling / token counting are unaffected.

### DIFF
--- a/neuro_san/internals/run_context/langchain/core/run_context_runnable.py
+++ b/neuro_san/internals/run_context/langchain/core/run_context_runnable.py
@@ -32,6 +32,8 @@ from langchain_core.language_models.base import BaseLanguageModel
 from langchain_core.messages.ai import AIMessage
 from langchain_core.messages.base import BaseMessage
 from langchain_core.runnables.base import Runnable
+from langchain_core.runnables.config import ensure_config
+from langchain_core.runnables.config import merge_configs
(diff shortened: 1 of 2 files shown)


d1donlydfink (approved):
Great low-level mojo.
LGTM AFAICT

PR #979, opened 2026-05-26
Fix Claude Opus 4.7 temperature 400 error

shrushtiimehta:
## Summary

Claude Opus 4.7 rejects `temperature`, `top_k`, and `top_p` parameters with a 400 error. The Python `langchain-anthropic` SDK does not strip these automatically (unlike the TypeScript version), so any value passed to the `ChatAnthropic` constructor is forwarded to the API.

This PR:
1. Sets `"temperature": null` in the global `default_config` so it no longer injects `0.7` into every model's config.
2. Removes `temperature`, `top_k`, and `top_p` from the `ChatAnthropic` constructor in `anthropic_llm_policy.py` (newer Anthropic models reject them).
3. Bumps `langchain-anthropic` minimum to `>=1.4.3` for full Opus 4.7 feature support.
4. Updates doc comments that referenced the old `0.7` default.

### DIFF
--- a/neuro_san/internals/run_context/langchain/llms/default_llm_info.hocon
+++ b/neuro_san/internals/run_context/langchain/llms/default_llm_info.hocon
@@ -600,10 +600,6 @@
         # https://platform.claude.com/docs/en/about-claude/pricing
         "price_per_1k_input_tokens": 0.005,
         "price_per_1k_output_tokens": 0.025,
-        # The Anthropic API returns a 400 error when temperature, top_k, or top_p is sent
-        # for this model. temperature is actively injected by default_config (0.7); top_k
-        # and top_p are already null but included here as a safety measure.
-        "deprecated_params": ["temperature", "top_k", "top_p"],
     }
 
 
@@ -1080,9 +1076,8 @@
                 # Note that we can only supply arguments that are "Just Data" in nature.
                 # See https://python.langchain.com/api_reference/anthropic/chat_models/langchain_anthropic.chat_models.ChatAnthropic.html#langchain_anthropic.chat_models.ChatAnthropic
                 "max_tokens": null,
-                "temperature": null,
-                "top_k": null,
-                "top_p": null,
+                # temperature, top_k, and top_p are not passed to ChatAnthropic;
+                # newer models (Opus 4.7+) reject them outright with a 400 error.
                 "default_request_timeout": null,
                 "max_retries": 2,
                 "stop_sequences": null,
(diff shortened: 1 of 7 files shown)


Noravee (changes requested):
Good catch, but the fix is already in `langchain-anthropic==1.4.1`.

I think we should upgrade it both in `neuro-san` and `neuro-san-studio`.

shrushtiimehta:
> Good catch, but the fix is already in `langchain-anthropic==1.4.1`.
> 
> I think we should upgrade it both in `neuro-san` and `neuro-san-studio`.

Claude Opus 4.7 rejects temperature/top_k/top_p with a 400 error, and the Python langchain-anthropic SDK doesn't strip them automatically. We still need them for haiku/sonnet. Our fix adds a deprecated_params list to the model entry that nulls out these values before they reach the API, and we bumped langchain-anthropic to >=1.4.3 (since it currently resolves to 1.4.3) for full Opus 4.7 feature support.

Noravee:
The issue here is that, unlike `ChatOpenAI`, `ChatAnthropic` does not skip temperature automatically for certain models.

neuro_san/internals/run_context/langchain/llms/default_llm_info.hocon
```diff
@@ -1259,9 +1259,10 @@
 
         "model_name": "gpt-5.2",           # The string name of the default model to use.
 
-        "temperature": 0.7,                 # The default LLM temperature (randomness) to use.
+        "temperature": null,                 # The default LLM temperature (randomness) to use.
```
Noravee: One thing to consider is that if a user has been using non-thinking models without explicitly setting a temperature, this change could affect the model’s behavior. We may want to display a warning such as:

Since no temperature was specified, the default value of 0.7 will be used.
This default may not work well for some models, such as Anthropic’s Claude Opus 4.7.
The default value of 0.7 will be deprecated in version 0.x.x. Going forward, the default temperature will instead follow the provider or model-specific default value.

What do you think? @shrushtiimehta @d1donlydfink @ofrancon

shrushtiimehta: Temperature has become much less impactful with newer models. We could deprecate it and let model providers choose sensible defaults. I don’t think enough models rely on it to warrant a deprecation warning, release notes could be enough.

d1donlydfink: 6 months ago I would have differed with you because the temperature deprecation was relatively new at the time and there could have been compatibility issues.  Nowadays, I think this approach is reasonable.
As long as we retain the ability to set the temperature when it matters for the models it matters to, then I think we are good.

d1donlydfink (dismissed):

Noravee (approved):
LGTM.

PR #984, opened 2026-05-27
Add ParametersSchemaNetworkValidator for hocon validation

### LINKED ISSUE #690: extra curly brace in hocon dies silently
**Describe the bug**
Loading the attached hocon file is successful on the server and produces no errors upon load or when we send it a user query, but the hocon has a bad syntax (extra curly in the second agent).

[loopy_manager_silent_killer.hocon.zip](https://github.com/user-attachments/files/25751699/loopy_manager_silent_killer.hocon.zip)

**To Reproduce**
1. place the hocon in your registry and add it to your manifest and run the server
2. Note there are no load errors in the logs
3. run any query against the agent network (e.g., "hello") and note that it freezes with no logs or errors

**Expected behavior**
Should create a clear error message upon loading and should not load the agent network

**Versions (please complete the following information):**
 - OS: [e.g. macOS]
 - Python version: [e.g. 3.12]
 - neuro-san version [e.g. 0.6.27]

shrushtiimehta:
## Summary

Adds `ParametersSchemaNetworkValidator` — a `CompositeDictionaryValidator` that validates each tool's `function.parameters` block via two independent, composable phases:

**Phase 1 — `PydanticParametersNetworkValidator`:** Runs `BaseModelDictionaryConverter.from_dict()` (the same pipeline used at tool-creation time) to catch unrecognized types, malformed structures, and anything that would crash at runtime. Imports `BaseModelDictionaryConverter` directly — no interface indirection.

**Phase 2 — `SemanticParametersNetworkValidator`:** Catches things pydantic is silent about:
- Nested `parameters` key (the headline bug from [studio#690](https://github.com/cognizant-ai-lab/neuro-san-studio/issues/690)) — pydantic treats this as just another property.
- `required` entries referencing undefined `properties` — pydantic silently ignores these.

Tests use HOCON fixtures via `AgentNetworkRestorer` (same filter chain as production), split into independent folders per validator (`pydantic_parameters/`, `semantic_parameters/`).

Also fixes real `required` typos in `assess_failure.hocon` and `gist.hocon` (`"sample"` → `"text_sample"`) caught by the new validator.

Fixes [cognizant-ai-lab/neuro-san-studio#690](https://github.com/cognizant-ai-lab/neuro-san-studio/issues/690)


neuro_san/internals/validation/network/manifest_network_validator.py
```diff
@@ -45,6 +47,7 @@ def __init__(self, external_network_names: List[str] = None, mcp_servers: List[s
             UnreachableNodesNetworkValidator(),
             # No ToolBoxNetworkValidator yet.
             ToolNameNetworkValidator(),
+            ParametersShapeNetworkValidator(),
```
shrushtiimehta: Registers the new validator in the composite chain. Catches the nested-parameters bug from studio#690 at network load time.

neuro_san/internals/validation/network/parameters_shape_network_validator.py
```diff
@@ -0,0 +1,196 @@
...
+        function_block: Any = agent_spec.get("function")
+        if isinstance(function_block, dict) and "parameters" in function_block:
+            value: Any = function_block.get("parameters")
+            return cls._EXPLICIT_NULL if value is None else value
+        if "parameters" in agent_spec:
+            value = agent_spec.get("parameters")
+            return cls._EXPLICIT_NULL if value is None else value
+        return None
+
+    def _validate_parameters(self, agent_name: str, params: Any) -> List[str]:
```
d1donlydfink: I would expect this to be recursive, since object definitions can have other object definitions inside them.
Yet I don't see that. Am I missing something?

d1donlydfink: Also: Instead of rolling your own, you might consider using JSON validation from Pydantic itself, which is what all this business is based on in the first place I believe.

It's a little tricky to find the right thing via google searches, as Pydantic has a lot to do with validating JSON description of an object against an object schema, but it feels like there could be some layer in there that would validate the JSON that other object-specific JSON would validate against.

neuro_san/internals/validation/network/parameters_shape_network_validator.py
```diff
@@ -0,0 +1,196 @@
...
+            found.append(path)
+
+        properties: Any = schema.get("properties")
+        if isinstance(properties, dict):
+            for prop_name, prop_schema in properties.items():
+                found.extend(
+                    cls._find_nested_parameters_keys(
+                        prop_schema, f"{path}.properties.{prop_name}"
+                    )
+                )
```
d1donlydfink: What I don't see here is looking for arrays which is another class of field that can have an object/properties definition nested within it. (This is something that a Pydantic-layer would get you for free).    For an example see: https://github.com/cognizant-ai-lab/neuro-san/blob/main/neuro_san/registries/esp_decision_assistant.hocon#L227

tests/neuro_san/internals/validation/network/test_parameters_shape_network_validator.py
```diff
@@ -0,0 +1,294 @@
...
+                    "parameters": {
+                        "type": "object",
+                        "properties": {
+                            "name": {"type": "string"},
+                        },
+                        "required": ["name"],
+                    },
+                },
+            }],
+        }
```
d1donlydfink: In a future PR, consider moving these tests to hocon files so the test bases can be more easily updated.

d1donlydfink (changes requested):
Now:
* Recursion?
* Look for Pydantic to do most of this for you
* Arrays
Future:
* Move tests to consume hocon to make it easier to add more test cases (with comments!).

shrushtiimehta:
> Now:
> 
> * Recursion?
> * Look for Pydantic to do most of this for you
> * Arrays
>   Future:
> * Move tests to consume hocon to make it easier to add more test cases (with comments!).

I've updated the implementation to address these points. Pydantic is able to do most of the work — Phase 1 runs the same BaseModelDictionaryConverter.from_dict() pipeline used at tool-creation time, catching unrecognized types, malformed structures, and anything that would crash at runtime. Phase 2 only handles two things pydantic is silent about: the nested parameters key bug from studio#690 (pydantic treats it as just another property) and required entries referencing undefined properties (pydantic silently ignores these). Both phases recurse into nested object properties and array items via a shared _iter_subschemas() helper, so errors at any depth are caught — including inside arrays, where unresolved string items references from missing commondefs are sanitized to a permissive dict before pydantic sees them.

neuro_san/internals/interfaces/schema_conversion_validator.py
```diff
@@ -0,0 +1,40 @@
...
+
+class SchemaConversionValidator:
+    """
+    Interface for validating that a parameters schema can be converted
+    into the model representation used at tool-creation time.
+
+    Implementations live in the run_context layer where the concrete
+    conversion logic (e.g. pydantic) already resides.  The validation
+    layer depends only on this interface, not on the converter itself.
+    """
```
d1donlydfink: Why do we need this new interface when the existing DictionaryValidator interface would suffice?
I'm guessing we probably don't need this interface at all if we use what we already have.
DictionaryValidator is already very very generic. So much so that I've been considering pushing it down to leaf-common

neuro_san/internals/graph/persistence/registry_manifest_restorer.py
```diff
@@ -42,6 +42,8 @@
 from neuro_san.internals.graph.registry.agent_network import AgentNetwork
 from neuro_san.internals.interfaces.agent_name_mapper import AgentNameMapper
 from neuro_san.internals.interfaces.storage_class import StorageClass
+from neuro_san.internals.run_context.langchain.core.pydantic_schema_conversion_validator import \
```
d1donlydfink: It seems like you go out of your way to make PydanticSchemaNetworkValidator abstract, but that comes at a cost of everyone else having to import run_context/langchain specifics everywhere.
You'll notice that all of the places that you inject run_context.langchain dependencies have been free of this kind of specificity for a long time.

neuro_san/internals/graph/persistence/registry_manifest_restorer.py
```diff
@@ -171,7 +173,8 @@ def restore_one_manifest(self, manifest_file: str) -> Dict[str, Dict[str, AgentN
             network_name: str = self.agent_mapper.filepath_to_agent_network_name(agent_filepath)
             validator = ManifestNetworkValidator(
                 external_network_names,
-                network_name=network_name)
+                network_name=network_name,
+                schema_validator=PydanticSchemaConversionValidator())
```
d1donlydfink: I'm seeing this in a lot of places. Maybe the question to ask is: "When do we not want the schema_validator to be PydanticSchemaConversionValidator?"

neuro_san/internals/validation/network/parameters_schema_network_validator.py
```diff
@@ -0,0 +1,230 @@
...
+                errors.append(
+                    f"{display_name}: 'parameters' must be object, "
+                    f"got {type(params).__name__}"
+                )
+                continue
+
+            # Phase 1: Structural validation via injected converter
+            if self.schema_validator is not None:
+                for msg in self.schema_validator.try_convert(params):
+                    errors.append(f"{display_name}: {msg}")
```
d1donlydfink: Observation:  If schema_validator were a DictionaryValidator, this class could be created by composition.
The phase 1 could be an optional schema_validator (will discuss details of that a bit later) and the phase 2 could be its own DictionaryValidator performing what phase 2 does.  These can easily be fitted together with an existing CompositeDictionaryValidator and (more importantly) unit tested independently of each other.

neuro_san/internals/validation/network/parameters_schema_network_validator.py
```diff
@@ -0,0 +1,230 @@
...
+    mistakes at any depth are caught.
+    """
+
+    # Sentinel returned by _locate_parameters meaning "the key is present but
+    # explicitly null". The caller flags this rather than silently skipping.
+    # object() is identity-safe - cannot collide with any real config value.
+    _EXPLICIT_NULL: Any = object()
+
+    def __init__(self, network_name: str = None,
+                 schema_validator: SchemaConversionValidator = None):
```
d1donlydfink: What type of other SchemaConversionValidators do you envision at this point?
I can see a None value skipping this phase, but is there actually a cosmic 3rd option that would be passed in here? 

I ask because maybe this should be a boolean (OMG I can't believe I just said that).
If this were a boolean, then all the import of run_context.langchain implementation details would be consolidated to this one class.  It's possible you might find that the boolean would be on all the time and then be unnecessary.

I still don't like run_context.langchain details being imported at this level,
but at least there'd be a single place to repair that.  (Though I just looked at BaseModelDictionaryConverter and it's not really langchain-y but more pydantic-y which is points in its favor for moving out of run_context.langchain area - we don't need to do that in this PR tho.)

Thoughts?

tests/fixtures/semantic_parameters/bad_required.hocon
```diff
@@ -0,0 +1,37 @@
...
+    "tools": [
+        {
+            "name": "agent_a",
+            "function": {
+                "parameters": {
+                    "type": "object",
+                    "properties": {
+                        "text_sample": {"type": "string"}
+                    },
+                    "required": ["sample"]
```
d1donlydfink, line 31: This is cool that you can make all these kinds of tests.

neuro_san/internals/validation/network/parameters_schema_network_validator.py
shrushtiimehta: There is no 2nd or 3rd option required. We would ideally want to run pydantic always. The abstraction was there to keep validation/ free of run_context.langchain imports, but in practice it just moved the import burden to callers.

neuro_san/registries/assess_failure.hocon
```diff
@@ -80,7 +80,7 @@ are known to fail against the given acceptance_criteria.
                             }
                         }
                     },
-                    "required": ["sample", "acceptance_criteria"]
+                    "required": ["text_sample", "acceptance_criteria"]
```
d1donlydfink, line 83: Why are these 2 hocons changing? Seems like they should be orthogonal to your change.

d1donlydfink (changes requested):
* It's very cool that you can add so many great test cases simply by adding new hocon test fixtures.
* SchemaConversionValidator probably doesn't need to exist. DictionaryValidator should suffice.
* Ask the questions:
    When do we not want a PydanticSchemaConversionValidator?
    What else would be use?
  It's possible that this abstraction should just be a boolean flag if there are times we don't want it.
  ... and that would reduce the propagation of run_context.langchain implementation details all over the place
* Strongly consider breaking the new Validator up into two separate Validators - phase 1 and phase 2
  and assemble them using the CompositeDictionaryValidator. This will encapsulate testing of them better.
* Do not change assess_failure and gist hocons.  They are essential to the test system and should be orthogonal
  to your change.

neuro_san/registries/assess_failure.hocon
shrushtiimehta: The variable doesn't exist, seems to be a typing error. required references "sample" but the actual property is "text_sample". Until now it was ignored. Since now we are adding a validator, the hocon would now not be loaded.

neuro_san/internals/validation/network/parameters_schema_network_validator.py
d1donlydfink: Ah, the case of not being able to sweep complexity under a rug. It always moves somewhere because it has to exist otherwise stuff wouldn't work at all - a law of nature.

shrushtiimehta: I'll remove the interface and have ParametersSchemaNetworkValidator import BaseModelDictionaryConverter directly. Callers go back to just passing network_name. In another PR, we could move it out.

neuro_san/internals/graph/persistence/registry_manifest_restorer.py
shrushtiimehta: Ideally we would always want it to be PydanticSchemaConversionValidator.

neuro_san/internals/validation/network/parameters_schema_network_validator.py
shrushtiimehta: yes makes sense, I will split it.

neuro_san/registries/assess_failure.hocon
d1donlydfink: Ah. I see. Your validator caught the errors. Nice.
Objection retracted!

neuro_san/internals/validation/network/pydantic_parameters_network_validator.py
```diff
@@ -0,0 +1,178 @@
...
+from copy import deepcopy
+from logging import getLogger
+from logging import Logger
+from typing import Any
+from typing import Dict
+from typing import List
+from typing import Optional
+
+from neuro_san.internals.run_context.langchain.core.base_model_dictionary_converter import \
+    BaseModelDictionaryConverter
```
d1donlydfink, line 25: Nice. The wonkiness of this import gets relegated to this one file.

neuro_san/internals/validation/network/pydantic_parameters_network_validator.py
```diff
@@ -0,0 +1,178 @@
...
+        error messages identify the right tool.
+        """
+        if agent_name:
+            return agent_name
+        if isinstance(agent_spec, dict):
+            function_block: Any = agent_spec.get("function")
+            if isinstance(function_block, dict):
+                fn_name: Any = function_block.get("name")
+                if fn_name:
+                    return fn_name
```
d1donlydfink: Why do we need to have the function block to have a name?
Or is this something we add in at a base level to satisfy Pydantic? 
If so, please add a comment, as it's not in the spec and it's not obvious why we want this.
If not, please remove.

neuro_san/internals/validation/network/pydantic_parameters_network_validator.py
```diff
@@ -0,0 +1,178 @@
...
+        return "<unnamed>"
```
d1donlydfink: Seems like something that is unnamed would be its own error, yes?

neuro_san/internals/validation/network/pydantic_parameters_network_validator.py
```diff
@@ -0,0 +1,178 @@
...
+        return result
+
+    @classmethod
+    def _replace_string_items(cls, schema: Any) -> None:
+        """Recursively replace string ``items`` values with ``{type: string}``."""
+        if not isinstance(schema, dict):
+            return
+        items: Any = schema.get("items")
+        if isinstance(items, str):
+            schema["items"] = {"type": "string"}
```
d1donlydfink: Maybe some comments as to why this is needed.

neuro_san/internals/validation/network/semantic_parameters_network_validator.py
```diff
@@ -0,0 +1,186 @@
...
+        """
+        if agent_name:
+            return agent_name
+        if isinstance(agent_spec, dict):
+            function_block: Any = agent_spec.get("function")
+            if isinstance(function_block, dict):
+                fn_name: Any = function_block.get("name")
+                if fn_name:
+                    return fn_name
+        return "<unnamed>"
```
d1donlydfink: Gee this looks familiar. Please factor these common items out into a helper class.
Copy/paste of policy is the enemy.

d1donlydfink (changes requested):
* Factor out common code to a helper class - these will be largely static methods
* Add comments to your policy helpers as to why we are doing things that aren't really in the agent spec or why we are doing certain substitutions.

neuro_san/internals/validation/network/pydantic_parameters_network_validator.py
shrushtiimehta: No, pydantic doesn't need this. It's for display for error messages.

neuro_san/internals/validation/network/pydantic_parameters_network_validator.py
shrushtiimehta: Yes, will remove it.

neuro_san/internals/validation/network/pydantic_parameters_network_validator.py
d1donlydfink: ok, so now that there are comments, I can now understand that what this likely is is a hack to make a single case for esp_decision_assistant commondefs pass.  

What is probably missing is a pass on the substitution that commondefs does.
To that end, there are at least 2 filters that live in neuro_san/internals/graph/filters that need to be put on the network to actually do the substitutions for commondefs.  

In fact, probably all the filters need to be done on any network spec coming through the validation wringer.  The order is defined in AgentNetworkRestorer.filter_config() and that method itself should probably be factored out to be its own composite method separate from AgentNetworkRestorer so that the Validation stuff can take its own separate crack at it.

neuro_san/internals/validation/network/semantic_parameters_network_validator.py
```diff
@@ -0,0 +1,151 @@
...
+        """
+        properties: Any = schema.get("properties")
+        if isinstance(properties, dict):
+            for prop_name, prop_schema in properties.items():
+                if isinstance(prop_schema, dict):
+                    yield prop_schema, f"{path}.properties.{prop_name}"
+
+        items: Any = schema.get("items")
+        if isinstance(items, dict):
+            yield items, f"{path}.items"
```
d1donlydfink, line 104: More items.  "items" I don't think is part of pedantic specifications. I think it's an actual field name.
Seeing this implies case-specific hackery.

d1donlydfink (changes requested):
Problems with "items" specificity emerge.

* Probably need to do the same filter_config() pass as happens in AgentNetworkRestorer before or during validation
*  Another items-specific block in SemanticBlahBlahValidation

neuro_san/internals/validation/network/semantic_parameters_network_validator.py
shrushtiimehta: items is the standard JSON Schema keyword for defining the element type of an array, it's not a user-defined field name. _iter_subschemas() needs to recurse into items the same way it recurses into properties to catch nested errors at any depth.

d1donlydfink (approved):

PR #990, opened 2026-05-28
Improve HOCON parse/validation error logging

shrushtiimehta:
## Summary

Improves error reporting for HOCON parse and validation failures in `RegistryManifestRestorer` and `AbstractAsyncConfigRestorer`.

**Key changes:**

- **RegistryManifestRestorer**: Validation errors joined into a single greppable log line (previously JSON-dumped with indentation). Extracted `_log_parse_error` helper that collapses multi-line pyparsing exceptions into one line. Converted `FileNotFoundError` handlers to lazy `%s` formatting. Removed redundant "not found" log that duplicated what the exception handler already reported. Removed unused `manifest_file`/`manifest_key` params from `process_one_agent_network`.
- **AbstractAsyncConfigRestorer**: Parse-error message now includes the original exception text inline and collapsed to one line (previously said "see the accompanying exception above" which was unhelpful).

## Impact

Before:

```
[2026-05-28 14:35:44 PDT] INFO     NeuroSan - manifest registry basic/loopy_manager_silent_killer.hocon has validation errors. Skipping. Errors: [                                                   
[2026-05-28 14:35:44 PDT] INFO     NeuroSan: Agent 'loopy_manager' references an unrecognized URL or path tool '/industry/cpg_agents_olivier'. Expected an external agent or network name starting   
                                   with '/' (e.g. '/bank_ops'), an MCP server, an http(s):// URL, or an MCP endpoint. Available URLs: ['/basic/advanced_calculator',                                 
                                   ........
                                   '/agent_network_designer', '/agent_network_editor', '/agent_network_instructions_editor', '/agent_network_query_generator', '/agent_network_test_generator',      
                                   '/agent_network_architect']                                                                                                                                       
[2026-05-28 14:35:44 PDT] INFO     NeuroSan - ]
```

After:

```
[2026-05-28 14:38:18 PDT] ERROR    NeuroSan - Validation error in manifest registry basic/loopy_manager_silent_killer.hocon: Agent 'loopy_manager' references an unrecognized URL or path tool       
                                   '/industry/cpg_agents_olivier'. Expected an external agent or network name starting with '/' (e.g. '/bank_ops'), an MCP server, an http(s):// URL, or an MCP      
                                   endpoint. Available URLs: ['/basic/advanced_calculator', '/basic/book_recommender_multiple_llm_configs', '/basic/coding_assistant',                               
                                   '/basic/coffee_finder_advanced', '/basic/coffee_finder', '/basic/hello_world', '/basic/internal_communication_skill', '/basic/job_guessing_skill',                
                                   ........
                                   '/agent_network_query_generator', '/agent_network_test_generator', '/agent_network_architect']. Skipping.                                                
```

### DIFF
--- a/neuro_san/internals/graph/persistence/registry_manifest_restorer.py
+++ b/neuro_san/internals/graph/persistence/registry_manifest_restorer.py
@@ -237,8 +237,8 @@ class RegistryManifestRestorer(Restorer):
         if agent_network is not None:
             validation_errors: List[str] = validator.validate(agent_network.get_config())
             if len(validation_errors) > 0:
-                joined_errors: str = "; ".join(str(err) for err in validation_errors)
-                self.logger.error("Validation error in manifest registry %s: %s. Skipping.",
+                joined_errors: str = "\n".join(str(err) for err in validation_errors)
+                self.logger.error("Validation error in manifest registry %s:\n%s\nSkipping.",
                                   agent_filepath, joined_errors)
                 return
 
@@ -309,19 +309,18 @@ class RegistryManifestRestorer(Restorer):
 
     def _log_parse_error(self, manifest_key: str, exception: Exception):
         """
-        Log a single-line ERROR for a parse failure.
+        Log an ERROR for a parse failure.
 
         Prefers the underlying parser exception (line/column info) over the
-        wrapper raised in AbstractAsyncConfigRestorer. The raw __str__ on a
-        pyparsing exception can include newlines; collapse to one line so the
-        message stays greppable.
+        wrapper raised in AbstractAsyncConfigRestorer.  Newlines are preserved
+        so that stack trace structure remains readable.
         """
         use_exception: Exception = exception
         if exception.__cause__ is not None:
             use_exception = exception.__cause__
 
-        detail: str = " ".join(str(use_exception).split())
-        self.logger.error("Parse error in registry item %s. Skipping. - %s", manifest_key, detail)
+        detail: str = "\n".join(str(use_exception).splitlines())
+        self.logger.error("Parse error in registry item %s. Skipping. -\n%s", manifest_key, detail)
 
     def restore(self, file_reference: str = None) -> Dict[str, Dict[str, AgentNetwork]]:
         """
--- a/neuro_san/internals/graph/persistence/registry_manifest_restorer.py
+++ b/neuro_san/internals/graph/persistence/registry_manifest_restorer.py
@@ -21,7 +21,6 @@ from typing import Sequence
 from typing import Union
 
 import os
-import json
 import logging
 
 from json.decoder import JSONDecodeError
@@ -176,8 +175,7 @@ class RegistryManifestRestorer(Restorer):
                 agent_network = self.restore_one_agent_network(manifest_dir, agent_filepath, manifest_key)
 
             self.process_one_agent_network(agent_network, usable_network, agent_filepath,
-                                           manifest_file, manifest_key, manifest_dict,
-                                           validator, agent_networks)
+                                           manifest_dict, validator, agent_networks)
 
         return agent_networks
 
@@ -219,38 +217,34 @@ class RegistryManifestRestorer(Restorer):
                 agent_network = await self.async_restore_one_agent_network(manifest_dir, agent_filepath, manifest_key)
 
             self.process_one_agent_network(agent_network, usable_network, agent_filepath,
-                                           manifest_file, manifest_key, manifest_dict,
-                                           validator, agent_networks)
+                                           manifest_dict, validator, agent_networks)
 
         return agent_networks
 
     # pylint: disable=too-many-arguments,too-many-positional-arguments
     def process_one_agent_network(self, agent_network: AgentNetwork, usable_network: bool, agent_filepath: str,
-                                  manifest_file: str, manifest_key: str, manifest_dict: Dict[str, Any],
+                                  manifest_dict: Dict[str, Any],
                                   validator: ManifestNetworkValidator,
                                   agent_networks: Dict[str, Dict[str, AgentNetwork]]):
         """
         :param agent_network: The agent network to process.
         :param usable_network: Is this agent network usable?
         :param agent_filepath: The filepath to the agent definition file.
-        :param manifest_file: The manifest file.
-        :param manifest_key: The manifest key.
         :param manifest_dict: The manifest dictionary.
         :param validator: The validator.
         :param agent_networks: The accumulated agent networks
         """
         if agent_network is not None:
-
             validation_errors: List[str] = validator.validate(agent_network.get_config())
             if len(validation_errors) > 0:
-                self.logger.error("manifest registry %s has validation errors. Skipping. Errors: %s",
-                                  agent_filepath,
-                                  json.dumps(validation_errors, indent=4, sort_keys=True))
-                agent_network = None
+                joined_errors: str = "; ".join(str(err) for err in validation_errors)
+                self.logger.error("Validation error in manifest registry %s: %s. Skipping.",
+                                  agent_filepath, joined_errors)
                 return
 
         if usable_network and agent_network is None:
-            self.logger.error("manifest registry %s not found in %s", manifest_key, manifest_file)
+            # restore_one_agent_network caught a FileNotFoundError or parse
+            # exception and already logged the failure; nothing more to do.
             return
 
         network_name: str = self.agent_mapper.filepath_to_agent_network_name(agent_filepath)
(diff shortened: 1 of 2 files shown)


neuro_san/internals/graph/persistence/registry_manifest_restorer.py
```diff
@@ -219,38 +217,34 @@ async def async_restore_one_manifest(self, manifest_file: str) -> Dict[str, Dict
...
         """
         if agent_network is not None:
-
             validation_errors: List[str] = validator.validate(agent_network.get_config())
             if len(validation_errors) > 0:
-                self.logger.error("manifest registry %s has validation errors. Skipping. Errors: %s",
-                                  agent_filepath,
-                                  json.dumps(validation_errors, indent=4, sort_keys=True))
-                agent_network = None
+                joined_errors: str = "; ".join(str(err) for err in validation_errors)
```
d1donlydfink, line 240: If we are joining a whole bunch of errors to be on one log line, let's make multiple errors findable with "\n" as part of the join delimiter.  This should still come out as a single log line, but it will be more apparent that there are multiple problems within the single hocon file.

neuro_san/internals/graph/persistence/registry_manifest_restorer.py
```diff
@@ -219,38 +217,34 @@ async def async_restore_one_manifest(self, manifest_file: str) -> Dict[str, Dict
...
+                self.logger.error("Validation error in manifest registry %s: %s. Skipping.",
+                                  agent_filepath, joined_errors)
```
d1donlydfink, line 242: Similarly, we probably want something like:

```
"Validation error in manifest registry %s: \n%s.\nSkipping."
```
... so that each element of the log line is immediately apparent when it is output.
If you have to scroll to see the "skipping", some vary salient information is being lost.

neuro_san/internals/graph/persistence/registry_manifest_restorer.py
```diff
@@ -219,38 +217,34 @@ async def async_restore_one_manifest(self, manifest_file: str) -> Dict[str, Dict
...
                 return
 
         if usable_network and agent_network is None:
-            self.logger.error("manifest registry %s not found in %s", manifest_key, manifest_file)
+            # restore_one_agent_network caught a FileNotFoundError or parse
+            # exception and already logged the failure; nothing more to do.
```
d1donlydfink, line 247: How does this error condition manifest itself now if we are not specifically logging it here?

neuro_san/internals/graph/persistence/registry_manifest_restorer.py
```diff
@@ -313,24 +298,31 @@ async def async_restore_one_agent_network(self, manifest_dir: str, agent_filepat
...
+        Prefers the underlying parser exception (line/column info) over the
+        wrapper raised in AbstractAsyncConfigRestorer. The raw __str__ on a
+        pyparsing exception can include newlines; collapse to one line so the
+        message stays greppable.
+        """
+        use_exception: Exception = exception
+        if exception.__cause__ is not None:
+            use_exception = exception.__cause__
+
+        detail: str = " ".join(str(use_exception).split())
```
d1donlydfink, line 323: Join() with "\n" so that you don't lose the immediate effect of what calls what in the stack trace.
It should still all go to a single log message - not multiple.

neuro_san/internals/persistence/abstract_async_config_restorer.py
```diff
@@ -168,11 +168,10 @@ def deserialize_file_contents(self, file_path: str, file_contents: bytes) -> Dic
...
         except (ParseException, ParseSyntaxException, JSONDecodeError, ConfigException) as exception:
-            message: str = f"""
-There was an error parsing {self.file_purpose} file "{file_path}".
-See the accompanying exception (above) for clues as to what might be
-syntactically incorrect in that file.
-"""
+            detail: str = " ".join(str(exception).split())
+            message: str = (
+                f'Error parsing {self.file_purpose} file "{file_path}": {detail}'
+            )
```
d1donlydfink: Use double quotes for strings in code. Use single quotes to delimit strings inside strings.
You have this reversed.

Also: ... "\n{detail}"

Also: Probably don't need parentheses.

Also: I actually think your new text is a little too terse. At the very least indicate that there is some kind of syntax/parse error going on.   What were you optimizing for here by going shorter?

neuro_san/internals/persistence/abstract_async_config_restorer.py
```diff
@@ -168,11 +168,10 @@ def deserialize_file_contents(self, file_path: str, file_contents: bytes) -> Dic
         try:
             config = serialization.to_object(bytes_file)
         except (ParseException, ParseSyntaxException, JSONDecodeError, ConfigException) as exception:
-            message: str = f"""
-There was an error parsing {self.file_purpose} file "{file_path}".
-See the accompanying exception (above) for clues as to what might be
-syntactically incorrect in that file.
-"""
+            detail: str = " ".join(str(exception).split())
```
d1donlydfink, line 171: Join() with "\n" to avoid scrolling for most salient stack trace information.

d1donlydfink (changes requested):
See PR comments.

I see the intent, but there are places where you are necessarily introducing extra scrolling in log messages to find the most salient information.  Everything has a UI. Including log messages. 

And don't forget that neuro-san-studio "rich" logging is not the only place that logs come out - people connect their logs to log capturing frameworks like DataDog and stuff should be readable and instantly grockable in those environments as well.  And then there's also plain old non-rich command line.

neuro_san/internals/graph/persistence/registry_manifest_restorer.py
shrushtiimehta: `restore_one_agent_network` already catches `FileNotFoundError` and parse exceptions and logs them with specifics before returning None. By the time `process_one_agent_network` sees agent_network is None, the failure is already logged. The old message was also misleading — it said "not found" even for parse errors.

neuro_san/internals/persistence/abstract_async_config_restorer.py
shrushtiimehta: Fixed quote style and parentheses. The longer messages told the user to "see the accompanying exception above" but now we embed the actual exception detail directly in the message ({detail}), making it self-contained

neuro_san/internals/graph/persistence/registry_manifest_restorer.py
shrushtiimehta: The StreamToLogger class in leaf_common splits any multi-line stdout write on newlines and re-logs each line as a separate call at logging.INFO. As a result, a single logger.error("...\n...") becomes multiple log entries, where only the first line is logged at ERROR and the remaining lines appear as INFO entries with their own timestamps.
This means that filtering by ERROR in Datadog would only capture the first line, potentially missing the actual error details. That's why "; " was used as the delimiter, it keeps the entire message in a single log entry at the correct log level.
I could change this behavior there, but that would affect all consumers of StreamToLogger, so I'm not sure we want to make that change globally.

neuro_san/internals/graph/persistence/registry_manifest_restorer.py
shrushtiimehta: Same StreamToLogger constraint.

neuro_san/internals/graph/persistence/registry_manifest_restorer.py
shrushtiimehta: Same StreamToLogger constraint.

neuro_san/internals/persistence/abstract_async_config_restorer.py
shrushtiimehta: Same StreamToLogger constraint

neuro_san/internals/graph/persistence/registry_manifest_restorer.py
d1donlydfink: That sounds like a good bug in StreamToLogger or the usage of it.
The basic idea behind StreamToLogger (if I recall correctly!) is to get stuff that code spits out to stdout or stderr into formatted log messages.  

This makes me wonder if this is a problem in all log sinks or only some (studio logger, DataDog, command line etc.).
What I don't want to do is focus on the problems of studio logger at the expense of the others.

If this really is only a problem with the studio logger (which I suspect it might be), then that means that the compress-to-one-line solution here is not really the end-all solution. It would then be at heart a logging configuration issue for the Rich logger. 

For instance, I can see here:
https://github.com/cognizant-ai-lab/neuro-san-studio/blob/main/neuro_san_studio/plugins/log_bridge/process_log_bridge.py#L664
... that the log bridge specifically spits stuff out to the console, which could engage StreamToLogger as INFO messages.

PR #998, opened 2026-05-29
Recurse into list elements in PydanticArgumentDictionaryConverter.to_dict

### LINKED ISSUE #997: PydanticArgumentDictionaryConverter.to_dict doesn't recurse into lists — array items reach CodedTools as pydantic objects
When a CodedTool's parameter schema declares an array of objects, the
list elements arrive at `async_invoke` as dynamically-generated pydantic
BaseModel instances (named e.g. `<field>_component`) instead of plain
dicts. Calling `.get()` on them raises `AttributeError`.

Noravee:
Fix #997 

Extracted per-value conversion into a small `_convert_value` helper that handles three cases: dict / pydantic BaseModel (existing behavior), list (new — recurses into each element), primitives (passthrough). The top-level loop now delegates to it.

### DIFF
--- a/neuro_san/internals/run_context/langchain/core/pydantic_argument_dictionary_converter.py
+++ b/neuro_san/internals/run_context/langchain/core/pydantic_argument_dictionary_converter.py
@@ -18,6 +18,7 @@ from __future__ import annotations
 
 from typing import Any
 from typing import Dict
+from typing import List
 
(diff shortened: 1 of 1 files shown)


andreidenissov-cog (approved):
Looks good to me.

d1donlydfink (approved):

PR #1005, opened 2026-05-31
Expand is_mcp_tool to accept canonical MCP server URIs

### LINKED ISSUE #1001: Expand the MCP server URL definition
This triggers https://github.com/cognizant-ai-lab/neuro-san-studio/issues/1106.
Currently only check that the url ends with `mcp` or `mcp/`.

Should be using
https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#canonical-server-uri

Noravee:
The old check (URL starts with "https://mcp" or ends with "/mcp" or "/mcp/") was too narrow and rejected valid MCP server URLs like "https://deepwiki.com/mcp/free", forcing users to fall back to the dict-form tool reference.

Replace the substring check with a proper validation against the MCP canonical server URI spec
(https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#canonical-server-uri):
  - require an http/https scheme (case-insensitive),
  - require a host,
  - reject URIs containing a fragment.

To distinguish MCP server URIs from Neuro SAN external agent URLs (which share the same shape), keep the "mcp" marker but accept it either as a host label (e.g. "mcp.example.com") or as any path segment (e.g. "/mcp", "/mcp/free", "/server/mcp"), not just at the start of the URL or the end of the path.

Also bump a tool-creation failure log from info to warning in base_tool_factory so the failure is visible at default log levels.

Fix #1001

### DIFF
--- a/neuro_san/internals/run_context/langchain/core/base_tool_factory.py
+++ b/neuro_san/internals/run_context/langchain/core/base_tool_factory.py
@@ -262,7 +262,7 @@ class BaseToolFactory:
             message: str = f"Failed to create Agent/tool '{name}': {tool_creation_exception}"
             agent_message = AgentMessage(content=message)
             await self.journal.write_message(agent_message)
-            self.logger.info(message)
+            self.logger.warning(message)
             return None
 
     def create_function_tool(self, function_json: Dict[str, Any], name: str) -> BaseTool:
--- a/neuro_san/internals/run_context/utils/external_agent_parsing.py
+++ b/neuro_san/internals/run_context/utils/external_agent_parsing.py
@@ -116,7 +116,8 @@ class ExternalAgentParsing:
         :return: True if this is an MCP tool reference
         """
         # Support both str and dict format;
-        # str format: "https://mcp.deepwiki.com/mcp" or "http://localhost:8000/mcp/"
+        # str format: a canonical MCP server URI, e.g. "https://mcp.deepwiki.com/mcp",
+        #             "http://localhost:8000/mcp/", or "https://deepwiki.com/mcp/free"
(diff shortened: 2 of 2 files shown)


neuro_san/internals/run_context/utils/external_agent_parsing.py
```diff
@@ -128,10 +129,33 @@ def is_mcp_tool(tool_ref: Union[str, Dict[str, Any]]) -> bool:
...
 
-        if isinstance(tool_ref, str):
-            return (tool_ref.startswith("https://mcp") or tool_ref.endswith(("/mcp", "/mcp/")))
-
-        return False
+        if not isinstance(tool_ref, str):
+            return False
+
+        # Validate against the MCP canonical server URI rules:
+        # https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#canonical-server-uri
```
Noravee, line 136: Follow the latest spec for MCP server.

neuro_san/internals/run_context/langchain/core/base_tool_factory.py
```diff
@@ -262,7 +262,7 @@ async def create_toolbox_tool(self, toolbox: str, agent_spec: Dict[str, Any], na
             message: str = f"Failed to create Agent/tool '{name}': {tool_creation_exception}"
             agent_message = AgentMessage(content=message)
             await self.journal.write_message(agent_message)
-            self.logger.info(message)
+            self.logger.warning(message)
```
Noravee, line 265: If failed to create the mcp tool, change log from info to warning.

d1donlydfink (approved):

PR #1008, opened 2026-06-01
Fix three latent bugs in ConnectivityReporter.assemble_tool_list

### LINKED ISSUE #957: Bug in ConnectivityReporter.assemble_tool_list
```python
if isinstance(args_tools, Dict):
            args_tools = args_tools.values()  # this line is not a list.

        if isinstance(args_tools, List):
            # Add to our list in order, but without repeats
            for tool in args_tools:
                if tool not in tool_set:
                    tool_set.add(tool)
                    tool_list.append(tool)

        return tool_list
```

Noravee:
Fix #957 

Replace the inline tool-list assembly with calls to the shared `AbstractNetworkValidator.coerce_tools`, `coerce_args_tools`, and `remove_dictionary_tools` helpers introduced for the validation chain. Same external contract — ordered list of unique agent-name strings, traditional `tools` first then `args.tools` — but correct on three previously-broken inputs:

  1. `args.tools` as a dict: `dict.values()` returned a dict_values view that failed `isinstance(args_tools, List)`, silently dropping all coded-tool down-chains from the connectivity report.
  2. `tools` as a string: the for-loop iterated the string character-by-character, emitting individual letters as fake tool names.
  3. dict element in `tools` (e.g., inline MCP server config): adding the unhashable dict to the dedup set raised `TypeError: unhashable type: 'dict'` instead of being filtered.

Add tests for each of the three scenarios.

### DIFF
--- a/neuro_san/internals/chat/connectivity_reporter.py
+++ b/neuro_san/internals/chat/connectivity_reporter.py
@@ -25,6 +25,7 @@ from neuro_san.internals.interfaces.context_type_toolbox_factory import ContextT
 from neuro_san.internals.run_context.factory.master_toolbox_factory import MasterToolboxFactory
 from neuro_san.internals.run_context.interfaces.agent_network_inspector import AgentNetworkInspector
 from neuro_san.internals.run_context.utils.external_agent_parsing import ExternalAgentParsing
+from neuro_san.internals.validation.network.abstract_network_validator import AbstractNetworkValidator
 
 
 class ConnectivityReporter:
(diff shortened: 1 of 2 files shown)


neuro_san/internals/chat/connectivity_reporter.py
```diff
@@ -182,37 +183,23 @@ def assemble_tool_list(agent_spec: Dict[str, Any]) -> List[str]:
...
 
-        extractor = DictionaryExtractor(agent_spec)
+        # Combine the traditional `tools` field with `args.tools` (the coded-tool
+        # convention). coerce_* keep the result tolerant to malformed shapes;
+        # remove_dictionary_tools drops non-string entries (e.g., inline MCP
+        # configs) so the dedup set sees only agent-name strings.
+        combined: List[Any] = (
+            AbstractNetworkValidator.coerce_tools(agent_spec) +
+            AbstractNetworkValidator.coerce_args_tools(agent_spec)
+        )
```
d1donlydfink, line 193: The parens make it look like this will be a Tuple, but it won't because of Pythonic parsing.
Would be more straightforward to set combined successively.

neuro_san/internals/chat/connectivity_reporter.py
```diff
@@ -182,37 +183,23 @@ def assemble_tool_list(agent_spec: Dict[str, Any]) -> List[str]:
...
-        if isinstance(args_tools, Dict):
-            args_tools = args_tools.values()
-
-        if isinstance(args_tools, List):
-            # Add to our list in order, but without repeats
-            for tool in args_tools:
-                if tool not in tool_set:
-                    tool_set.add(tool)
-                    tool_list.append(tool)
-
```
d1donlydfink, line 215: Nice consolidation

d1donlydfink (approved):

PR #1028, opened 2026-06-05
Make the "Invoking" journal event diagnostic when a tool has no name

thilo11:
## Problem

`JournalingCallbackHandler.on_tool_start` builds the journaled invocation label straight from `serialized.get("name")`:

```python
message = AgentMessage(content=f"Invoking: `{agent_name}` with:", ...)
```

When the serialized tool carries no name, this emits an uninformative `Invoking: `` with:` (empty backticks) plus a null `invoked_agent_name` — a decorative event with no indication of what was being invoked.

## Change

Fall back to a `<unnamed tool>` placeholder for the human-readable label when the name is missing, so the event stays diagnostic. The raw value (the real name, or `None`) is still carried unchanged in `structure.invoked_agent_name`, so downstream consumers are unaffected.

Behavior-preserving for the normal (named) case — only the missing-name path changes.

## Tests

Added `test_journaling_callback_handler.py`:

- named tool → label reports the name verbatim;
- missing name → label falls back to `<unnamed tool>` while `invoked_agent_name` stays `None`.

### DIFF
--- a/neuro_san/internals/run_context/langchain/journaling/journaling_callback_handler.py
+++ b/neuro_san/internals/run_context/langchain/journaling/journaling_callback_handler.py
@@ -162,8 +162,13 @@ class JournalingCallbackHandler(AsyncCallbackHandler):
             "params": tool_start_dict.get("tool_args")
         }
 
+        # Fall back to a placeholder label when the serialized tool has no name,
+        # so the journaled event stays diagnostic rather than an empty "Invoking: ``".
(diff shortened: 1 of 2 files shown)


d1donlydfink (approved):
Thanks for the contribution (however simple).
Several million points out of 10 for adding tests.

LGTM. 

It's possible at some point we want to consolidate in a set of constants for our unknown entities for standardized reporting. Don't need to go there now.

PR #1037, opened 2026-06-09
NS-1036 Silently remove fallbacks with no API_KEY set for them. Only error when nothing in fallbacks works.

d1donlydfink:
* Funnel all create_llm() calls in LangChainRunContext to create_llm_with_fallbacks()
* Have that create_llm_with_fallbacks() method return a dictionary with different error lists when nothing in the fallback list works at all.
* Have LangChainRunContext do the reporting of API Keys not working as well instead of lower. The idea is these can be deferred in a forward-looking fallback list scenario that @ofrancon shrewdly wants to entertain.
* Fix fallout

What @ofrancon wants to be able to do is to specify an llm_config like this:
```
    "llm_config": {
        "fallbacks": [
            {
                # Try OpenAI first
                "model_name": "gpt-5.2",
            },
            {
                # Fall back to Anthropic Claude if OpenAI is unavailable.
                "model_name": "claude-sonnet",
            },
            {
                # Fall back to Google Gemini if both OpenAI and Anthropic are unavailable.
                "model_name": "gemini-3-flash",
            },
        ]
    }
```
and not have to have every single key set for every LLM provider.
This is quite reasonable.   

Still, there are questions we need to answer about what a config looks like for devs vs our own production needs, but that is not for this PR. This PR is to enable the mechanism so that we can debate how to employ it.

With this PR, the system will take the subset of that llm_config fallbacks specification that will remotely work given the configuration and use that for the basis of agents.

@donn-leaf : There is a question for you about a reporting trade-off within this PR. Perhaps there is an env var afoot or an llm_config key to add to control preferences. Suggestions welcome.

Tested the above config manually with a bunch of different key configurations in hello_world.

### DIFF
--- a/neuro_san/internals/run_context/langchain/core/langchain_run_context.py
+++ b/neuro_san/internals/run_context/langchain/core/langchain_run_context.py
@@ -240,6 +240,7 @@ class LangChainRunContext(RunContext):
                 error += "\n".join(construction_errors) + "\n"
             raise ValueError(error)
 
+        self.llm_resources = main_llm_resources
         agent: Runnable = self.create_agent(instructions, main_llm_resources.get_model())
 
         return agent
--- a/neuro_san/internals/run_context/langchain/core/langchain_run_context.py
+++ b/neuro_san/internals/run_context/langchain/core/langchain_run_context.py
@@ -243,7 +243,8 @@ class LangChainRunContext(RunContext):
 
             api_key_errors: List[str] = error_dict.get("api_key_errors", [])
             if len(api_key_errors) > 0:
-                error += "\nThe following errors occurred while using API keys:\n"
+                error += "\nLLM operation for this agent requires certain API keys to be set as environment variables."
+                error += "\nThe following errors occurred while looking for LLM API keys:\n"
                 error += "\n    ".join(api_key_errors)
 
             raise ValueError(error)
--- a/neuro_san/internals/run_context/langchain/core/langchain_run_context.py
+++ b/neuro_san/internals/run_context/langchain/core/langchain_run_context.py
@@ -17,7 +17,6 @@
 from typing import Any
 from typing import Dict
 from typing import List
-from typing import Set
 from typing import Union
 
 import json
@@ -224,17 +223,18 @@ class LangChainRunContext(RunContext):
         llm_factory: ContextTypeLlmFactory = self.invocation_context.get_llm_factory()
         sly_data: Dict[str, Any] = self.tool_caller.get_sly_data()
 
-        main_llm_resources: LangChainLlmResources | List[Set[str]] = \
+        main_llm_resources: LangChainLlmResources | Dict[str, Any] = \
             llm_factory.create_llm_with_fallbacks(self.llm_config, sly_data)
 
-        if isinstance(main_llm_resources, list):
+        if isinstance(main_llm_resources, dict):
             error: str = "No fully-specified LLM found in llm_config or fallbacks."
-            required_llm_config: Set[str] = main_llm_resources.pop(0)
+            error_dict: Dict[str, Any] = main_llm_resources
+            required_llm_config: List[str] = error_dict.get("required_llm_config_errors", [])
             if len(required_llm_config) > 0:
                 error += "\nLLM operation for this agent requires at least one "
                 error += "of the following set in sly_data.llm_config:\n"
                 error += "\n".join(sorted(required_llm_config)) + "\n"
-            construction_errors: Set[str] = main_llm_resources.pop(0)
+            construction_errors: List[str] = error_dict.get("construction_errors", [])
             if len(construction_errors) > 0:
                 error += "\nThe following errors occurred while constructing LLMs:\n"
                 error += "\n".join(construction_errors) + "\n"
--- a/neuro_san/internals/run_context/langchain/core/langchain_run_context.py
+++ b/neuro_san/internals/run_context/langchain/core/langchain_run_context.py
@@ -223,70 +223,23 @@ class LangChainRunContext(RunContext):
         # Get the factory we will use
         llm_factory: ContextTypeLlmFactory = self.invocation_context.get_llm_factory()
 
-        # Prepare a list of fallbacks.  By default, the llm_config itself is a single-entry fallback list.
-        fallbacks: List[Dict[str, Any]] = [self.llm_config]
-        fallbacks = self.llm_config.get("fallbacks", fallbacks)
+        main_llm_resources: LangChainLlmResources | List[Set[str]] = \
+            llm_factory.create_llm_with_fallbacks(self.llm_config, self.sly_data)
 
-        # Get the sly data to see if there are any optional user llm_config (like API keys) to use
-        sly_data: Dict[str, Any] = self.tool_caller.get_sly_data()
-
-        # Initialize a list of chain fallbacks. This may or may not get filled.
-        chain_fallbacks: List[Runnable] = []
-        first_llm: bool = True
-        required_llm_config: Set[str] = set()
-
-        # Go through the list of fallbacks in the config.
-        construction_errors: List[str] = []
-        for fallback in fallbacks:
-
-            # Create a model we might use.
-            # If construction fails (e.g. missing API key in env), record the error and
-            # try the next fallback rather than aborting the whole loop.
-            try:
-                one_llm_resources: LangChainLlmResources | Set[str] = llm_factory.create_llm(fallback, sly_data)
-            except ValueError as exception:
-                construction_errors.append(str(exception))
-                continue
-
-            if one_llm_resources is None:
-                # Nothing to use or report.
-                # Skip for now, a fallback might still be fulfilled.
-                continue
-
-            if isinstance(one_llm_resources, set):
-                # Report later on which required llm_config are missing
-                # Skip for now, a fallback might still be fulfilled.
-                required_llm_config.update(one_llm_resources)
-                continue
-
-            one_agent: Runnable = self.create_agent(instructions, one_llm_resources.get_model())
-
-            if first_llm:
-                # The first fully-specified agent is the one we want to be our main guy.
-                agent = one_agent
-                # For now. Could be problems with different providers w/ token counting.
-                self.llm_resources = one_llm_resources
-                # Anything that comes later will not be the first
-                first_llm = False
-            else:
-                # Anything later than the first guy is considered a fallback. Add it to the list.
-                chain_fallbacks.append(one_agent)
-
-        if agent is None:
+        if isinstance(main_llm_resources, list):
             error: str = "No fully-specified LLM found in llm_config or fallbacks."
+            required_llm_config: Set[str] = main_llm_resources.pop(0)
             if len(required_llm_config) > 0:
                 error += "\nLLM operation for this agent requires at least one "
                 error += "of the following set in sly_data.llm_config:\n"
                 error += "\n".join(sorted(required_llm_config)) + "\n"
+            construction_errors: Set[str] = main_llm_resources.pop(0)
             if len(construction_errors) > 0:
                 error += "\nThe following errors occurred while constructing LLMs:\n"
                 error += "\n".join(construction_errors) + "\n"
             raise ValueError(error)
 
-        if len(chain_fallbacks) > 0:
-            # Set up fallbacks.
-            # See https://python.langchain.com/docs/how_to/tools_error/#tryexcept-tool-call
-            agent = agent.with_fallbacks(chain_fallbacks)
+        agent: Runnable = self.create_agent(instructions, main_llm_resources.get_model())
 
         return agent
 
--- a/neuro_san/internals/run_context/langchain/core/langchain_run_context.py
+++ b/neuro_san/internals/run_context/langchain/core/langchain_run_context.py
@@ -229,15 +229,23 @@ class LangChainRunContext(RunContext):
         if isinstance(main_llm_resources, dict):
             error: str = "No fully-specified LLM found in llm_config or fallbacks."
             error_dict: Dict[str, Any] = main_llm_resources
-            required_llm_config: List[str] = error_dict.get("required_llm_config_errors", [])
-            if len(required_llm_config) > 0:
+
+            required_sly_data: List[str] = error_dict.get("required_sly_data_errors", [])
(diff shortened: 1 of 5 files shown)


neuro_san/internals/run_context/langchain/core/langchain_run_context.py
```diff
@@ -222,71 +221,27 @@ def create_agent_with_fallbacks(self, instructions: str) -> Runnable:
...
-                agent = one_agent
-                # For now. Could be problems with different providers w/ token counting.
-                self.llm_resources = one_llm_resources
-                # Anything that comes later will not be the first
-                first_llm = False
-            else:
-                # Anything later than the first guy is considered a fallback. Add it to the list.
-                chain_fallbacks.append(one_agent)
+        main_llm_resources: LangChainLlmResources | Dict[str, Any] = \
+            llm_factory.create_llm_with_fallbacks(self.llm_config, sly_data)
```
d1donlydfink, line 227: Use the fallbacks logic that has been pushed into the DefaultLlmFactory.
Also, its return type has now been changed to a dictionary to better distinguish between different kinds of errors that can arise with fallbacks.

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
```diff
@@ -222,71 +221,27 @@ def create_agent_with_fallbacks(self, instructions: str) -> Runnable:
...
-        if agent is None:
+        if isinstance(main_llm_resources, dict):
             error: str = "No fully-specified LLM found in llm_config or fallbacks."
+            error_dict: Dict[str, Any] = main_llm_resources
+            required_llm_config: List[str] = error_dict.get("required_llm_config_errors", [])
             if len(required_llm_config) > 0:
                 error += "\nLLM operation for this agent requires at least one "
                 error += "of the following set in sly_data.llm_config:\n"
                 error += "\n".join(sorted(required_llm_config)) + "\n"
+            construction_errors: List[str] = error_dict.get("construction_errors", [])
```
d1donlydfink, line 239: Use the different kinds of errors reported by the fallbacks mechanism.

tests/neuro_san/internals/run_context/langchain/core/test_langchain_run_context_llm_fallbacks.py
```diff
@@ -53,7 +54,7 @@ def _make_run_context(llm_config: Dict[str, Any], create_llm_side_effect) -> Lan
         run_context.llm_config = llm_config
         run_context.llm_resources = None
 
-        llm_factory = MagicMock()
+        llm_factory = DefaultLlmFactory()
```
d1donlydfink, line 57: These tests count the number of times create_llm() is called.
Can't use a MagicMock object for that any more because the create_llm_with_fallbacks() policy is used all the time now.  Instead we create the actual object and replace the create_llm() method with a MagicMock for the desired test effects.  This was tricksy to discover the right way of doing this without rewriting the tests.

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
```diff
@@ -636,7 +651,12 @@ def create_llm_with_fallbacks(self, config: Dict[str, Any],
 
         if main_llm_resources is None:
             # Return all errors
-            return required_llm_config.union(construction_errors)
+            return {
+                "api_key_errors": list(api_key_errors),
+                "construction_errors": list(construction_errors),
+                "required_sly_data_errors": list(required_sly_data),
+            }
```
d1donlydfink, line 659: Return a dictionary with distinct error type lists.

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
```diff
@@ -613,7 +618,17 @@ def create_llm_with_fallbacks(self, config: Dict[str, Any],
...
-                construction_errors.append(str(exception))
+                if exception.__cause__ is not None:
+                    cause: Exception = exception.__cause__
+                    message: str = ApiKeyErrorCheck.check_for_api_key_exception(cause)
+                    if message is not None:
+                        # Make sure the message is fit for public consumption with no
+                        # explicit secrets in the text.
+                        message = ApiKeyErrorCheck.get_safe_log_message(exception)
+                        api_key_errors.add(message)
+                        continue
```
d1donlydfink, line 630: API Key errors will be rethrown as ValueErrors with a "from cause".  Check for that so we can report them separately if we need to.

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
```diff
@@ -222,71 +221,36 @@ def create_agent_with_fallbacks(self, instructions: str) -> Runnable:
...
-            if first_llm:
-                # The first fully-specified agent is the one we want to be our main guy.
-                agent = one_agent
-                # For now. Could be problems with different providers w/ token counting.
-                self.llm_resources = one_llm_resources
-                # Anything that comes later will not be the first
-                first_llm = False
-            else:
-                # Anything later than the first guy is considered a fallback. Add it to the list.
-                chain_fallbacks.append(one_agent)
```
d1donlydfink, line 273: We don't need this as a chain of fallback _agents_. Instead we now use the model set up as a chain of fallback _models_.

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
```diff
@@ -194,7 +194,7 @@ def create_full_llm_config(self, config: Dict[str, Any], sly_data: Dict[str, Any
         :return: The fully specified config with defaults filled in.
                 Can also return a set of strings describing any required sly_data API keys that are not provided.
         """
-        full_config: Dict[str, Any] = None
+        full_config: Dict[str, Any] | Set[str] = None
```
d1donlydfink, line 197: Further clarification for typing.

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
```diff
@@ -222,71 +221,42 @@ def create_agent_with_fallbacks(self, instructions: str) -> Runnable:
...
+                    # Note: We are assuming the errors have already been filtered for
+                    # not dispensing secrets with ApiKeyErrorCheck.get_safe_log_message()
+                    self.logger.error("API KEY error detected: %s", api_key_error)
+
             raise ValueError(error)
 
-        if len(chain_fallbacks) > 0:
-            # Set up fallbacks.
-            # See https://python.langchain.com/docs/how_to/tools_error/#tryexcept-tool-call
-            agent = agent.with_fallbacks(chain_fallbacks)
```
d1donlydfink, line 289: Now the main_llm_resources' model has the fallbacks attached directly to it.
We do not need to do agent.with_fallbacks() any more.

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
```diff
@@ -222,71 +221,42 @@ def create_agent_with_fallbacks(self, instructions: str) -> Runnable:
...
+        self.llm_resources = main_llm_resources
+        agent: Runnable = self.create_agent(instructions, main_llm_resources.get_model())
```
d1donlydfink, line 259: The model passed in here is the primary guy that made it through the gauntlet with any secondary guys as ordered fallbacks.

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
```diff
@@ -222,71 +221,42 @@ def create_agent_with_fallbacks(self, instructions: str) -> Runnable:
...
+            if len(api_key_errors) > 0:
+                error += "\n\nLLM operation for this agent requires at least one LLM provider API key"
+                error += " to be set as an environment variable."
+                error += "\nThe following errors occurred while looking for LLM API keys:\n"
+                error += "\n    ".join(api_key_errors)
+
+                for api_key_error in api_key_errors:
+                    # Note: We are assuming the errors have already been filtered for
+                    # not dispensing secrets with ApiKeyErrorCheck.get_safe_log_message()
+                    self.logger.error("API KEY error detected: %s", api_key_error)
```
d1donlydfink, line 254: Note we are now only giving these API Key errors en-masse when _nothing_ in the fallback list satisfies a go on using any of the LLMs in the fallback list.   Idea here is to reduce noise in the use case that @ofrancon is moving towards with an llm_config like this:

```
    "llm_config": {
        "fallbacks": [
            {
                # Try OpenAI first
                "model_name": "gpt-5.2",
            },
            {
                # Fall back to Anthropic Claude if OpenAI is unavailable.
                "model_name": "claude-sonnet",
            },
            {
                # Fall back to Google Gemini if both OpenAI and Anthropic are unavailable.
                "model_name": "gemini-3-flash",
            },
        ]
    }
```

In this scenario, having any one of these guys will be good enough, but we don't want everyone bothered by the extra error messages.

The down-side is that this will hide problems in setting production keys where you want to know all the failures up-front.  Not sure how to handle both of those worlds just yet.

@donn-leaf : What does your ideal look like?

neuro_san/internals/run_context/langchain/core/run_context_runnable.py
```diff
@@ -76,7 +75,10 @@ class RunContextRunnable(NeuroSanRunnable):
     # is able to use JSON schema definitions to validate fields.
     agent_chain: Runnable
 
-    primary_llm: BaseLanguageModel
+    # This needs to be a Runnable because nowadays the primary LLM could be
+    # be a BaseLanguageModel with fallbacks attached to it, which ends up
+    # being a Runnable.
+    primary_llm: Runnable
```
d1donlydfink, line 81: System will hang silently if the typing is not correct for something with fallbacks.

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
```diff
@@ -443,8 +443,6 @@ def create_llm_resources(self, config: Dict[str, Any]) -> LangChainLlmResources:
                     #          following set in sly_data.llm_config:
                     #          anthropic_api_key
                     #
-                    self.logger.error("API KEY error detected: %s",
-                                      ApiKeyErrorCheck.get_safe_log_message(exception))
```
d1donlydfink, line 447: This error logging is moved to langchain_run_context.

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
```diff
@@ -571,18 +569,20 @@ def sanitize_keys(self, d: Dict[str, Any]) -> Dict[str, Any]:
         """
         return {self.strip_outer_quotes(k): v for k, v in d.items()}
 
+    # pylint: disable=too-many-branches
     def create_llm_with_fallbacks(self, config: Dict[str, Any],
                                   sly_data: Dict[str, Any] = None,
-                                  num_fallbacks: int = None) -> LangChainLlmResources | Set[str]:
+                                  num_fallbacks: int = None) -> LangChainLlmResources | Dict[str, Any]:
```
d1donlydfink, line 575: Return a dictionary with distinct error type lists.
This is more straightforward than cryptic List[0] and needing to know what 0 means.

neuro_san/internals/run_context/utils/activation_capsule.py
```diff
@@ -124,8 +124,11 @@ def create_chat_model(self, llm_config: Dict[str, Any], sly_data: Dict[str, Any]
         llm_resources = llm_factory.create_llm_with_fallbacks(llm_config, sly_data, num_fallbacks)
         if llm_resources is None:
             raise ValueError("Unable to create LLM from llm_config (missing required configuration and/or sly_data).")
-        if isinstance(llm_resources, set):
-            required = "\n".join(sorted(llm_resources))
+        if isinstance(llm_resources, dict):
+            # For now, do not worry about the "construction_errors" or "api_key_errors" keys.
+            # These are very likely to already be handled by the front man.
+            required_sly_data: List[str] = llm_resources.get("required_sly_data_errors", [])
+            required = "\n".join(sorted(required_sly_data))
```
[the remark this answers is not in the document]
d1donlydfink: Not worrying about that now, per the comment.

neuro_san/internals/run_context/langchain/core/langchain_run_context.py
donn-leaf: In the ideal deployment world, when the server starts up, if the primary llm provider fails, i'd want that to be fatal. That would almost always indicate I'm missing a key or something. While possible, it seems low-probability that an llm provider would happen to be suffering downtime during a server start.

d1donlydfink: Makes sense. Thanks for the feedback @donn-leaf.
Maybe we want a key in the config block for each llm which controls the error reporting behavior in this case.
Maybe something like:

```
"llm_config": {
        "fallbacks": [
            {
                # Try OpenAI first
                "model_name": "gpt-5.2",
                "error_early": true
            },
            {
                # Fall back to Anthropic Claude if OpenAI is unavailable.
                "model_name": "claude-sonnet",
                "error_early": true
            },
            {
                # Fall back to Google Gemini if both OpenAI and Anthropic are unavailable.
                "model_name": "gemini-3-flash",
                "error_early": true
            },
        ]
}
```
The true/false-ness of error_early could even be set by env var, cuz hocon.

... this can be for a separate PR.

Thoughts?

ofrancon: In production, before starting the server, we could run this command to list and check the LLM keys:
```
python -m neuro-san-studio check-llm-keys
```
or, if neuro-san-studio is pip installed (we're not doing that yet in the Dockerfile, but maybe we could)
```
ns check-llm-keys
```

We should also run the following command to check the LLM config:
```
python -m neuro-san-studio check-config
```
That will go through each fallback model and make 1 call to check it works. If it reports any error we should probably NOT attempt starting the Neuro SAN server.

We can enhance these commands to make them more script friendly if needed (e.g. return specific exit codes)

ofrancon (approved):
lgtm, thanks!

d1donlydfink:
@ofrancon says:
> In production, before starting the server ...

This is fine for our use case, but not everyone will want to take on neuro-san-studio as a dependency in their own deployments.

deepsaia (approved):
Lgtm.

Noravee (approved):
LGTM.

PR #1040, opened 2026-06-10
Parse text block past thinking/reasoning in chain output

### LINKED ISSUE #1039: Agent output is empty when LLM returns thinking/reasoning content blocks before text
## Problem

`RunContextRunnable.parse_chain_result` assumes that when an `AIMessage.content`
is a list of content blocks, the first block is always the text block:

```python
if isinstance(output, list):
    output = output[0].get("text", "")
```

This is no longer true. Providers like Anthropic (extended thinking) and other
LangChain-standard chat models can return a list where non-text blocks
(thinking, reasoning, tool_use, etc.) precede the text block. In those
cases output[0].get("text", "") returns "" and the user-facing response is
empty even though the model produced a valid answer.

Noravee:
Anthropic and other providers can return a list of content blocks where `"thinking"/"reasoning"` blocks precede the `"text"` block. The old code took `output[0]` blindly and lost the actual text. Scan the list for the `"text"`-typed block and fall back to `""` if none is present so the downstream `error_detector` always receives a string.

Tested by running cli on music nerd pro (openai, anthropic, gemini), math guy passthrough, esp, intranet.

Fix #1039

### DIFF
--- a/neuro_san/internals/run_context/langchain/core/run_context_runnable.py
+++ b/neuro_san/internals/run_context/langchain/core/run_context_runnable.py
@@ -327,12 +327,17 @@ class RunContextRunnable(NeuroSanRunnable):
                 # We generally want the content of any single AIMessage we found from above
                 output = ai_message.content
 
-        # In general, output is a string. but output from Anthropic can either be
-        # a single string or a list of content blocks.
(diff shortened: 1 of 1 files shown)


ofrancon (approved):
lgtm

PR #1042, opened 2026-06-10
Coerce HOCON list triggers to tuples in summarization middleware

### LINKED ISSUE #1041: NeuroSanSummarizationMiddleware fails on langchain >=1.3 with HOCON-configured trigger/keep
## Problem

`NeuroSanSummarizationMiddleware` cannot be constructed from a HOCON agent
config on langchain `>=1.3`. Example failure from CI:

```
Problems creating AgentMiddleware instance of
  neuro_san.middleware.neuro_san_summarization_middleware.NeuroSanSummarizationMiddleware:
  Unsupported trigger item type: <class 'list'>
```

HOCON has no tuple type, so a config like

```hocon
"trigger": [["messages", 3]],
"keep": ["messages", 1],
```

deserializes to a Python `list` containing `list`s. langchain's
`SummarizationMiddleware` 1.3+ tightened its trigger validation to require
`Mapping` or `tuple` items and now raises `TypeError` on list items
(langchain/agents/middleware/summarization.py, `_validate_subject`).

Older langchain (1.2.x) used plain tuple-unpacking, which silently accepted
lists — so the bug only surfaces in CI / environments that have upgraded
langchain. Both versions are within our current pin (`langchain>=1.2.0,<2.0`),
making this a silent regression for users on a newer minor version.

## Repro

Running the existing unit test on langchain >=1.3:

```bash
pytest tests/neuro_san/zzz_hocons/test_unit_test_hocons.py \
  -k music_nerd_summarize_beatles_with_history_summarize
```

fails with the trigger validation error above. The same test passes on
langchain 1.2.x.

Noravee:
HOCON has no tuple type, so configs like `trigger: [["messages", 3]]` deserialize to nested lists. langchain>=1.3 SummarizationMiddleware tightened validation and now rejects list items with `Unsupported trigger item type: <class 'list'>`, breaking NeuroSanSummarizationMiddleware construction even though older langchain versions silently accepted them.

Coerce list-shaped `trigger` items and list-shaped `keep` to tuples before delegating to super().__init__. Dict (Mapping) clauses pass through unchanged so the forward-compatible `{"messages": 3}` form keeps working. Adds unit tests covering list coercion, dict pass-through, tuple pass-through, and the `trigger=None` default.

Fix #1041

### DIFF
--- a/neuro_san/middleware/neuro_san_summarization_middleware.py
+++ b/neuro_san/middleware/neuro_san_summarization_middleware.py
@@ -116,6 +116,13 @@ class NeuroSanSummarizationMiddleware(SummarizationMiddleware):
         :param trim_tokens_to_summarize: Maximum tokens to keep when preparing messages for
                 the summarization call. (default: _DEFAULT_TRIM_TOKEN_LIMIT)
         """
+        # HOCON has no tuple type, so callers configure trigger/keep as JSON arrays.
+        # langchain>=1.3 rejects list items in `trigger`, requiring tuples or dicts.
+        # Coerce list shapes back to tuples so both old and new langchain accept them.
+        if isinstance(trigger, list):
+            trigger = [tuple(item) if isinstance(item, list) else item for item in trigger]
+        if isinstance(keep, list):
+            keep = tuple(keep)
         super().__init__(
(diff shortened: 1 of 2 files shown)


andreidenissov-cog (approved):
Looks good to me.

d1donlydfink (approved):

PR #1044, opened 2026-06-10
NS-916 Configurable LLM streaming mode

andreidenissov-cog:
Make LLM streaming mode configurable through general llm_config part of agent network definition.
This way for any agent network we can make a choice:
with streaming, we lose some run-time network performance, BUT hopefully we improve LLM provider reaction time to client disconnect, and so saving ourselves tokens not spent;
without streaming, run-time performance is improved, but we risk that even if client aborts its request, inference on LLM side will continue until end - tokens spent for nothing.


tests/neuro_san/internals/run_context/langchain/llms/test_llm_factory.py
```diff
@@ -15,52 +15,17 @@
...
-        #           its constructor and you care about delete_resources() cleanup.
-        #
-        # See neuro_san.internals.run_context.langchain.llms.llm_policy.LlmPolicy and some
-        # of the base implementations near there for more details/examples.
-
-        class_to_llm_policy_type: Dict[str, Type[LlmPolicy]] = {
-            "test-openai": OpenAILlmPolicy
-        }
-        super().__init__(class_to_llm_policy_type)
+        _ = CustomLlmFactory()
```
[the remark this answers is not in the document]
andreidenissov-cog: The test does exactly what was done before.

neuro_san/internals/run_context/langchain/llms/anthropic_llm_policy.py
```diff
@@ -65,14 +65,15 @@ def create_llm(self, config: Dict[str, Any], model_name: str, client: Any) -> Ba
...
-            stream_usage=False,  # Don't even include token usage in the streaming response, since we're not streaming
+            # Streaming is configurable via the "streaming" key in llm_config; defaults
+            # to False so existing agents keep their long-standing non-streaming behavior.
+            # We pass streaming explicitly (rather than relying on LangChain's default) so
+            # that langchain_core._should_stream() picks up the configured value even when
+            # a streaming-aware callback is attached. Token usage is collected from
+            # AIMessage.usage_metadata in LlmTokenCallbackHandler regardless of streaming
+            # mode; stream_usage tracks streaming so usage frames flow only when streaming.
+            streaming=self.is_streaming(config),
+            stream_usage=self.is_streaming(config),
```
d1donlydfink: Why not simply do config.get("streaming", False) not unlike the way we do the other config variables?

neuro_san/internals/run_context/langchain/llms/llm_policy.py
```diff
@@ -119,6 +119,32 @@ async def delete_resources(self):
...
+                 or a non-zero integer for the "streaming" key; False otherwise (including when absent).
+        """
+        value = config.get("streaming", False)
+        if isinstance(value, bool):
+            return value
+        if isinstance(value, str):
+            return value.strip().lower() in ("true", "yes")
+        if isinstance(value, int):
+            return value != 0
+        return False
```
d1donlydfink: Why do we need this policy method to account for string values?

config.get("streaming", False) should be good enough, right?
If not, this boolean-or-from-string value detection should probably be moved out to a static utility method so the same policy can be reused in other places for consistency.

tests/neuro_san/internals/run_context/langchain/llms/test_llm_factory.py
```diff
@@ -15,52 +15,17 @@
...
-
-class TestLlmFactory(StandardLangChainLlmFactory):
+class TestLlmFactory:
     """
-    Test Factory class for LLM operations
+    Test creating custom Factory class for LLM operations
     """
 
-    def __init__(self):
+    def test_llm_factory(self):
```
d1donlydfink, line 26: I don't understand why what was here before wasn't good enough for the test?
Feels like this shouldn't have needed to change given what you were doing in the code above.

neuro_san/internals/run_context/langchain/llms/default_llm_info.hocon
```diff
@@ -1139,7 +1139,7 @@
                 "logprobs": null,
                 "top_logprobs": null,
                 "logit_bias": null,
-                # streaming is always on. Without it token counting will not work.
+                # streaming defaults to off; set "streaming": true in llm_config to enable.
```
Noravee, line 1195: I think we should also set `"streaming": false` here as well.

neuro_san/internals/run_context/langchain/llms/llm_policy.py
```diff
@@ -119,6 +119,32 @@ async def delete_resources(self):
...
+        constructor (typically as the ``streaming`` kwarg, and in some cases
+        also ``stream_usage``). The default is False, which preserves the
+        long-standing neuro-san behavior of using non-streaming HTTP requests
+        even when a streaming-aware callback handler is attached to the run
+        manager.
+
+        :param config: The fully specified llm config
+        :return: True if the config sets a truthy value for the "streaming"
+                 key; False otherwise (including when the key is absent).
+        """
```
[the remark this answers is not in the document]
andreidenissov-cog: Already fixed.

tests/neuro_san/internals/run_context/langchain/llms/test_llm_policy_streaming.py
```diff
@@ -0,0 +1,75 @@
...
+"""
+from neuro_san.internals.run_context.langchain.llms.llm_policy import LlmPolicy
+
+
+class TestLlmPolicyStreaming:
+    """
+    Verifies that LlmPolicy.is_streaming() correctly reads the "streaming"
+    key from the llm config, defaulting to False when absent, and coerces
+    arbitrary truthy / falsy values to a clean bool.
+    """
```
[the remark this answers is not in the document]
andreidenissov-cog: Fixed.

tests/neuro_san/internals/run_context/langchain/llms/test_llm_factory.py
andreidenissov-cog: Previously, this test was not executed at all, because of explicitly defined constructor.
The whole refactoring is about making this LLM constructor actually execute.

neuro_san/internals/run_context/langchain/llms/llm_policy.py
```diff
@@ -119,6 +119,30 @@ async def delete_resources(self):
...
+        value = config.get("streaming", False)
+        if isinstance(value, bool):
+            return value
+        if isinstance(value, str):
+            return value.strip().lower() == "true"
+        return False
```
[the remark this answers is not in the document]
andreidenissov-cog: Already fixed.

neuro_san/internals/run_context/langchain/llms/llm_policy.py
andreidenissov-cog: Are we validating our LLM configuration(s) to only use strict bool values like True and False?
If not, then "streaming": "no" will evaluate to streaming == True,
and we have to be more inclusive with our checking.

neuro_san/internals/run_context/langchain/llms/default_llm_info.hocon
andreidenissov-cog: This is done.

neuro_san/internals/run_context/langchain/llms/llm_policy.py
```diff
@@ -119,6 +119,25 @@ async def delete_resources(self):
...
+        also ``stream_usage``). The default is False, which preserves the
+        long-standing neuro-san behavior of using non-streaming HTTP requests
+        even when a streaming-aware callback handler is attached to the run
+        manager.
+
+        :param config: The fully specified llm config
+        :return: True if the config sets a truthy value for the "streaming"
+                 key; False otherwise (including when the key is absent).
+        """
+        return bool(config.get("streaming", False))
```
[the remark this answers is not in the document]
andreidenissov-cog: Fixed some commits ago.

Noravee (approved):
LGTM.

d1donlydfink (approved):

PR #1058, opened 2026-06-16
NS-1059 sly_data_schema merging with global basis for BYOK

d1donlydfink:
The reason this PR looks 20+ files scary is because:
* I added a bunch of missing `__init__.py` files to test directories, which uncovers ...
* a lot of pylint errors that are now fixed in this PR
... this isn't a big deal and is not the main event

The main event is really only 3 files:
* Add a notion of a global sly_data_schema to be specially merged with the tool.function.sly_data_schema
* Add a test for the new special merging.
* Add docs for this

The new global sly_data_schema definition should cover _most_ of @ofrancon 's wishes about BYOK being totally contained in the llm_config.hocon file. 

... But the devil is in the details and not all cases will be covered here:
1) There are still cases of allowing sly_data fields to be sent to other external networks (as AND does) that will have to be treated on a case-by-case basis.  Luckily that is only 3 rather complex networks in neuro-san-studio. 
2) It's also possible that AND hocon creation might need to be modified to account for BYOK as well when there are references to external networks in what is created.

For a preview of what BYOK will look like in neuro-san-studio, see this PR:
https://github.com/cognizant-ai-lab/neuro-san-studio/pull/1156

Also manually tested AND with santa.


neuro_san/internals/graph/filters/defaults_config_filter.py
```diff
@@ -38,10 +38,12 @@ class DefaultsConfigFilter(ConfigFilter):
...
     DEFAULTS_MAPPING: Dict[str, str] = {
-        # A value of None implies using the source key as the same destination
-        # key in the tool as well.
+        # A value of None implies using a default dictionary where the source key as the same destination
+        # key in the tool as well and it applies to all tools.
         "llm_config": None,
-        "llm_config.verbose": "verbose",
+        "llm_config.verbose": {
+            "dest_key": "verbose",
+        },
```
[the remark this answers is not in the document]
d1donlydfink: Addressed.

neuro_san/internals/graph/filters/defaults_config_filter.py
```diff
@@ -81,37 +95,112 @@ def filter_config(self, basis_config: Dict[str, Any]) \
...
+
+                if tool_front_man_only and not is_front_man:
+                    # Skip this one.
+                    continue
 
                 tool_value = tool_extractor.get(tool_dest_key)
                 if tool_value is None:
                     # If the tool does not have a value, use the basis_value whole cloth
-                    tool[tool_dest_key] = deepcopy(basis_value)
+                    self.set_tool_value(tool, tool_dest_key, basis_value)
```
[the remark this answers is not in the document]
d1donlydfink: Addressed.

neuro_san/internals/graph/filters/defaults_config_filter.py
```diff
@@ -81,37 +95,112 @@ def filter_config(self, basis_config: Dict[str, Any]) \
...
 
                 elif isinstance(tool_value, Dict) and isinstance(basis_value, Dict):
-                    # Do a dictionary overlay with basis_value defaults as what is
-                    # overrideable. The values of tool_value are favored.
-                    tool[tool_dest_key] = overlayer.overlay(basis_value, tool_value)
+                    # Merge the dictionaries
+                    union_fields: List[str] | str | None = tool_dest_dict.get("union_fields")
+                    self.merge_dictionaries(tool, tool_dest_key, basis_value, tool_value, union_fields)
```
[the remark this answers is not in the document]
d1donlydfink: addressed.

neuro_san/internals/graph/filters/defaults_config_filter.py
```diff
@@ -81,37 +95,112 @@ def filter_config(self, basis_config: Dict[str, Any]) \
...
+
+        if union_fields is not None:
+            if isinstance(union_fields, str):
+                union_fields = [union_fields]
+            for one_field in union_fields:
+                basis_field: List[str] = basis_value.get(one_field, [])
+                tool_field: List[str] = tool_value.get(one_field, [])
+                # The merged value is the union of the two sets with no duplicates
+                one_field_key: str = f"{tool_dest_key}.{one_field}"
+                self.set_tool_value(tool, one_field_key, list(set(basis_field) | set(tool_field)))
```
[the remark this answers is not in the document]
d1donlydfink: Addressed.

neuro_san/internals/graph/filters/defaults_config_filter.py
```diff
@@ -37,11 +37,13 @@ class DefaultsConfigFilter(ConfigFilter):
 
     # A mapping of source keys for defaults at the top level to destination
     # keys on the specific tool where the top-level defaults (if any) should be copied.
-    DEFAULTS_MAPPING: Dict[str, str] = {
-        # A value of None implies using the source key as the same destination
-        # key in the tool as well.
+    DEFAULTS_MAPPING: Dict[str, Dict[str, Any]] = {
+        # A value of None implies using a default dictionary where the source key as the same destination
+        # key in the tool as well and it applies to all tools.
```
d1donlydfink, line 42: New typing. Instead of a None or a string value, we now have a dictionary value that could have a bunch of other direction in it.

neuro_san/internals/graph/filters/defaults_config_filter.py
```diff
@@ -51,8 +53,21 @@ class DefaultsConfigFilter(ConfigFilter):
         "max_attempts": None,
         "error_formatter": None,
         "error_fragments": None,
+        "sly_data_schema": {
+            "dest_key": "function.sly_data_schema",
+            "front_man_only": True,
+            "union_fields": ["required"],
+        },
```
d1donlydfink, line 60: Here is a fully-specified dictionary example for the new translation of sly_data_schema.
We can designate that this defaults replacement:
* Have a compound.deep.key for dest_key
* Be only applicable to the front_man
* Have fields that require special union-ing merge behavior.

neuro_san/internals/graph/filters/defaults_config_filter.py
```diff
@@ -51,8 +53,21 @@ class DefaultsConfigFilter(ConfigFilter):
...
     }
 
+    def __init__(self):
+        """
+        Constructor
+        """
+        super().__init__()
+        self.overlayer = DictionaryOverlay()
```
d1donlydfink, line 68: Make only one of these per instance.

neuro_san/internals/graph/filters/defaults_config_filter.py
```diff
@@ -80,38 +95,129 @@ def filter_config(self, basis_config: Dict[str, Any]) \
...
-        overlayer = DictionaryOverlay()
+        basis_extractor = DictionaryExtractor(result_config)
 
         # Loop through all the tools making additions.
         tools = result_config.get("tools")
-        for tool in tools:
+
+        idx: int = 0
+        tool: Dict[str, Any] = None
+        for idx, tool in enumerate(tools):
```
d1donlydfink, line 105: Add some typing and keep track of the index to detect the front man.

neuro_san/internals/graph/filters/defaults_config_filter.py
```diff
@@ -80,38 +95,129 @@ def filter_config(self, basis_config: Dict[str, Any]) \
...
 
+            # Assume the front-man is the first tool in the list.
+            # This is largely consistent with AgentNetwork.find_front_man().
+            is_front_man: bool = idx == 0
+
             # Loop through all the keys in the defaults mapping.
-            for basis_source_key, tool_dest_key in self.DEFAULTS_MAPPING.items():
+            basis_source_key: str = None
+            tool_dest_dict: Dict[str, Any] = None
+            for basis_source_key, tool_dest_dict in self.DEFAULTS_MAPPING.items():
```
d1donlydfink, line 115: Keep better track of typing for the loop variables.

neuro_san/internals/graph/filters/defaults_config_filter.py
```diff
@@ -80,38 +95,129 @@ def filter_config(self, basis_config: Dict[str, Any]) \
...
 
                 # Account for semantics outlined in default_mapping above
-                if tool_dest_key is None:
-                    tool_dest_key = basis_source_key
+                if tool_dest_dict is None:
+                    tool_dest_dict = {
+                        "dest_key": basis_source_key,
+                        "front_man_only": False,
+                        "union_fields": None,
+                    }
```
d1donlydfink, line 128: Some defaults for when the entry is None.

neuro_san/internals/graph/filters/defaults_config_filter.py
```diff
@@ -80,38 +95,129 @@ def filter_config(self, basis_config: Dict[str, Any]) \
...
+
+                tool_dest_key = tool_dest_dict.get("dest_key", basis_source_key)
+                tool_front_man_only = tool_dest_dict.get("front_man_only", False)
+
+                if tool_front_man_only and not is_front_man:
+                    # Skip this one.
+                    continue
```
d1donlydfink, line 135: Skip if the field is only for the front man and we are not looking at the front man.

neuro_san/internals/graph/filters/defaults_config_filter.py
```diff
@@ -80,38 +95,129 @@ def filter_config(self, basis_config: Dict[str, Any]) \
...
 
                 tool_value = tool_extractor.get(tool_dest_key)
                 if tool_value is None:
                     # If the tool does not have a value, use the basis_value whole cloth
-                    tool[tool_dest_key] = deepcopy(basis_value)
+                    self.set_tool_value(tool, tool_dest_key, deepcopy(basis_value))
```
d1donlydfink, line 140: Use a common set_tool_value() method to allow for compound.deep.keys.

neuro_san/internals/graph/filters/defaults_config_filter.py
```diff
@@ -80,38 +95,129 @@ def filter_config(self, basis_config: Dict[str, Any]) \
...
 
-                elif isinstance(tool_value, Dict) and isinstance(basis_value, Dict):
-                    # Do a dictionary overlay with basis_value defaults as what is
-                    # overrideable. The values of tool_value are favored.
-                    tool[tool_dest_key] = overlayer.overlay(basis_value, tool_value)
+                elif isinstance(tool_value, dict) and isinstance(basis_value, dict):
+                    # Merge the dictionaries
+                    union_fields: List[str] | str | None = tool_dest_dict.get("union_fields")
+                    self.merge_dictionaries(tool, tool_dest_key, basis_value, tool_value, union_fields)
```
d1donlydfink, line 145: Move what happens in this block to a new merge_dictionaries() methods to avoid pylint complaining about too many branches.

neuro_san/internals/graph/filters/defaults_config_filter.py
```diff
@@ -80,38 +95,129 @@ def filter_config(self, basis_config: Dict[str, Any]) \
...
+        Merges the basis and tool dictionaries
+        :param tool: The tool dictionary itself
+        :param tool_dest_key: The destination key in the tool dictionary
+        :param basis_value: The dictionary value from the global basis
+        :param tool_value: The dictionary value from the tool
+        :param union_fields: The fields to union. Currently can not be a multi-part key.
+        """
+
+        # overrideable. The values of tool_value are favored.
+        self.set_tool_value(tool, tool_dest_key, self.overlayer.overlay(basis_value, tool_value))
```
d1donlydfink, line 167: Use the new set_tool_value() for the overlayer doing the initial merge.

neuro_san/internals/graph/filters/defaults_config_filter.py
```diff
@@ -80,38 +95,129 @@ def filter_config(self, basis_config: Dict[str, Any]) \
...
+
+        # Take care of special merge cases where lists are unioned
+        if union_fields is not None:
+
+            # Case of unioning a single field
+            if isinstance(union_fields, str):
+                union_fields = [union_fields]
+
+            one_field: str = None
+            for one_field in union_fields:
```
d1donlydfink, line 177: Now if there are fields to union, do those in a separate pass.
This handles the correct union-merging of "required" for sly_data_schema.properties.

neuro_san/internals/graph/filters/defaults_config_filter.py
```diff
@@ -80,38 +95,129 @@ def filter_config(self, basis_config: Dict[str, Any]) \
...
+                self.set_tool_value(tool, one_field_key, merged_field)
+
+    def set_tool_value(self, target: Dict[str, Any], key: str, value: Any):
+        """
+        Sets the value of a key in the tool dictionary
+
+        :param target: The dictionary to set the value in
+        :param key: The key to set the value for. This can be a multi-part key delimited by "."
+        :param value: The value to set
+        """
```
d1donlydfink, line 194: Recursive set_tool_value() method to handle deep.compound.keys.

tests/neuro_san/internals/graph/filters/test_defaults_config_filter.py
```diff
@@ -0,0 +1,291 @@
...
+        """
+        Can we construct?
+        """
+        my_filter: ConfigFilter = self.get_filter()
+        self.assertIsNotNone(my_filter)
+
+    def test_no_additions(self):
+        """
+        Tests basic non-operations of the ConfigFilter
+        """
```
d1donlydfink, line 52: New tests for the sly_data_schema cases we want to catch with the new behavior.

tests/neuro_san/internals/mappers/agent_mapper.py
```diff
@@ -23,6 +23,10 @@
 
 
 class TestMapper(unittest.TestCase):
+    """
+    Tests different file mappers
+    """
```
d1donlydfink, line 28: Lots of pylint fixes now that the `__init__.py` files are in the test directories and pylint actually knows to scan the files. :sigh:

tests/neuro_san/internals/run_context/langchain/llms/custom_llm_factory.py
```diff
@@ -63,3 +63,6 @@ def __init__(self):
             "test-openai": OpenAILlmPolicy
         }
         super().__init__(class_to_llm_policy_type)
+
+    def create_base_chat_model(self, config: Dict[str, str]) -> None:
+        _ = config
```
d1donlydfink, line 68: Pylint wanted this to be overridden.

tests/neuro_san/internals/run_context/openai/resources_cli.py
```diff
@@ -1,65 +0,0 @@
...
-# See the License for the specific language governing permissions and
-# limitations under the License.
-#
-# END COPYRIGHT
-from typing import Any
-from typing import List
-
-from asyncio import run
-
-from neuro_san.internals.run_context.openai.openai_client import OpenAIClient
```
d1donlydfink, line 22: This file was old from the original Babak days. Note the moldy import that hasn't existed for years.

tests/neuro_san/internals/validation/network/abstract_network_validator_test.py
```diff
@@ -105,3 +108,48 @@ def test_valid(self, hocon_file: str = "hello_world.hocon"):
         if len(errors) > 0:
             failure_message = errors[0]
         self.assertEqual(0, len(errors), failure_message)
+
+    def assertEqual(self, first: Any, second: Any, msg: str = None) -> None:
+        raise NotImplementedError
```
d1donlydfink, line 113: Pylint wanted these methods to be created.

tests/neuro_san/internals/validation/network/test_url_network_validator.py
```diff
@@ -37,7 +37,7 @@ def create_validator(self) -> DictionaryValidator:
         """
         return UrlNetworkValidator(external_agents=["/math_guy"])
 
-    def test_valid(self):
+    def test_valid(self, hocon_file: str = "hello_world.hocon"):
```
d1donlydfink, line 40: Pylint wanted consistent method signature with the superclass.

docs/agent_hocon_reference.md
```diff
@@ -396,6 +397,21 @@ which contains the following keys:
...
+channel when it is called.  The sly_data itself is generally considered to be private information
+that does not belong in the chat stream, for example: credential information.
+
+This is a "global" definition of sly_data_schema that will be merged with any tool-specific definition.
+While this is intended only for use with a front-man agent only, it can sometimes be useful to
+have partial definitions for sly_data_schema that pertain to all agent networks in the manifest
+defined globally with this key while network-specific definitions can be defined with the `sly_data_schema`
+key within the front-man tool.
+
+For more details, see the [sly_data_schema](#sly_data_schema-1) key below.
```
d1donlydfink, line 413: New docs for global sly_data_schema definition.

docs/agent_hocon_reference.md
```diff
@@ -530,6 +546,7 @@ definition, as sly_data itself is already visible to all other internal agents o
 Example networks that advertise their sly_data_schema:
 
 - [math_guy.hocon](../neuro_san/registries/math_guy.hocon)
+- [music_nerd_pro_sly_api_key.hocon](../neuro_san/registries/music_nerd_pro_sly_api_key.hocon)
```
d1donlydfink, line 550: Reference another example.

Noravee (approved):
Nice work — overall I like this. Being able to declare `sly_data_schema` once at the top level and have it merged down into `tool.function.sly_data_schema` is clean and declarative. It keeps the common BYOK boilerplate out of every front-man and makes the "what private inputs does this network need" contract obvious in one place.

One concern I have is the downstream-propagation side of BYOK as you mentioned in the PR description, e.g.:

```
"to_downstream": {
    "sly_data": ["agent_network_definition", "agent_network_name", ${?llm_config_key}]
}
```

Here we're mixing two unrelated concerns into the same list:

- **Domain data** the network genuinely needs to pass downstream (agent_network_definition, agent_network_name)
- **BYOK credential plumbing** (${?llm_config_key}), which is really infrastructure, not part of the network's functional contract

That coupling means every author who wants BYOK to work across external networks has to know to hand-thread the key into `to_downstream.sly_data`, and it's easy to get wrong or forget. It also muddies the meaning of that list — a reader can no longer tell "what data does this network actually pass" from "what credentials happen to need forwarding."

This PR nicely solves the declaration side; I think the propagation side deserves the same treatment eventually. A couple of directions I think might be worth considering (later / separate PR):

- Let fields declared in `sly_data_schema` (or a subset flagged as credentials) be propagated automatically, so authors don't manually restate keys in `to_downstream.sly_data`.
- Or give BYOK/credential keys their own channel separate from `to_downstream.sly_data`, so domain data and credential forwarding stay decoupled.

d1donlydfink:
I agree that it would be easy to forget to add these where they need to be for BYOK.
To that I would say that in general a deployment is very likely either going to be BYOK or not as a cost decision.
The neuro-san-studio repo ends up kinda special in that it has a dual role of being something that aspires to
be instructional, but also flexible enough to be deployable in several circumstances. There is tension in that role.
I think if someone were doing pure BYOK with multiple agent networks, they would have to have the key-transfer front-and-center in their mind the whole time.

@Noravee says:

> * Let fields declared in sly_data_schema (or a subset flagged as credentials) be propagated automatically, so authors don't manually restate keys in to_downstream.sly_data.
> * Or give BYOK/credential keys their own channel separate from to_downstream.sly_data, so domain data and credential forwarding stay decoupled.

These are both shrewd observations. 

I think for the first one, I would prefer erroring on the side of caution with private data.
That is: Do nothing with private data automatically. Always be explicit about what you want to do with it.
This stance will help us adhere to "security by default".  Flagging a subset would be the way to go for simplicity.
I think we can experiment with that in a later PR (as you also shrewdly suggest).
Minimizing an attack surface and maximizing convenience are both optimization problems of sorts.
Iterations are helpful

I think for the second one, having a second channel simply moves the dynamic routing of secrets to someplace else.
Sure, in what you are proposing, it wouldn't be sly_data, but someone still needs to decide "I need this private data to go there, I do not want this other private data going there".  Same problem, different information channel.
This is informational complexity that cannot be swept under a rug. Plus we'd be modifying the protocol and plumbing to make that happen.  Unclear if that's actually necessary, though I get the motivation.

andreidenissov-cog (approved):
Looks good, AFAICT.

PR #1074, opened 2026-06-19
NS-1022 Adjust GC timeout threshold for neuro-san AsyncioExecutorPool.

andreidenissov-cog:
Default GC timeout threshold for AsyncioExecutorPool proved to be too long (180 sec).
Adjust it to 30 sec and make it explicit in neuro-san code, so we know what's happening.
Note that in CLI client for direct connection we leave it at default - for single client session it will work OK.

### DIFF
--- a/neuro_san/service/utils/server_context.py
+++ b/neuro_san/service/utils/server_context.py
@@ -45,7 +45,7 @@ class ServerContext(ServerContextLite):
         Constructor.
         """
         self.server_status: ServerStatus = None
-        self.executor_pool = AsyncioExecutorPool(reuse_mode=True)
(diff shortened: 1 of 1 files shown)


d1donlydfink (approved):

neuro_san/service/utils/server_context.py
```diff
@@ -45,7 +45,7 @@ def __init__(self):
         Constructor.
         """
         self.server_status: ServerStatus = None
-        self.executor_pool = AsyncioExecutorPool(reuse_mode=True)
+        self.executor_pool = AsyncioExecutorPool(reuse_mode=True, idle_timeout_seconds=30)
```
d1donlydfink, line 48: Does this need to be configurable (better if not)

PR #1087, opened 2026-06-24
NS-1086 Do not use http/s when calling local /external agent networks

d1donlydfink:
Using an AsyncDirectAgentSession when calling local external agents will save us a lot of threads and maybe a lot of memory, thus having great potential to increase capacity of a single server.

It might seem the easy thing to do to "just" pass use_direct=True into ExternalSessionFactory, but 
there are 2 main stumbling blocks to this:
1. Adding network_storage_dict to ExternalSessionFactory constructor.  This gets you pretty far.
2. Just when Reservationist is being closed (shutting down its queue) is tricky

So we fix all that, and now use_direct=True is the default for "/external" agent references on the server.
This should be OK because if people really do want to go back out for an http connection, they can use an http(s)://<servername> specification for their external agents.

Initial memray flamegaphs show only 13MB per AND request, this is down from around 30MB per AND request before.  This is a good sign.

### DIFF
--- a/neuro_san/internals/chat/data_driven_chat_session.py
+++ b/neuro_san/internals/chat/data_driven_chat_session.py
@@ -33,7 +33,6 @@ from langchain_core.messages.base import BaseMessage
 
 from leaf_common.config.resolver_util import ResolverUtil
 
-from neuro_san.interfaces.reservationist import Reservationist
 from neuro_san.internals.chat.async_collating_queue import AsyncCollatingQueue
 from neuro_san.internals.chat.chat_history_message_processor import ChatHistoryMessageProcessor
 from neuro_san.internals.graph.registry.agent_network import AgentNetwork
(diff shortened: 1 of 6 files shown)


Noravee (dismissed):
lgtm

vince-leaf (changes requested):
This change is now use_direct=True is the default,
neuro_san/session/external_agent_session_factory.py, lines 133-134 has an existing bug.

The loop at lines 130-135 will actually execute in production. ExpiringAgentNetworkStorage.get_agent_network_provider() can return None, which would crash on the .get_agent_network() call inside the loop. Worth adding a None guard there too:

if agent_network_provider is not None and
agent_network_provider.get_agent_network() is not None:
break

ofrancon (dismissed):
lgtm

vince-leaf (approved):
LGTM.

Low priority.
get_networks_order(self, network_storage_dict) takes an explicit parameter even though the only caller passes self.network_storage_dict. This isn't a bug, but it's a minor design oddity — either make it a @staticmethod or remove the parameter and use self.network_storage_dict directly.

d1donlydfink:
Don't need to touch that now, Devince.

Issue #1029, opened 2026-06-05
Surface tool-call parse/validation failures on the journal, not just the server log


thilo11:
## Summary

When the agent loop retries after a recoverable error — most notably an output-parsing failure where the model's tool call can't be parsed — the reason is written only to the server logger, never to the journal / client message stream. From a client's activity feed this shows up as an unexplained stall or an uninformative agent event: you can see *that* a step happened but not *why* it is being retried.

This is common with models that occasionally emit a tool call wrapped in extra prose (so the tool-call JSON won't parse): the call is retried and usually recovers, but the activity stream gives no hint that anything went wrong.

## Where

`neuro_san/internals/run_context/langchain/core/run_context_runnable.py`, in the `run_it` retry loop. Each `except` branch logs the reason and retries silently, e.g.:

```python
except ValueError as value_error:
    ...
    self.logger.warning("retrying from ValueError")
    attempts = attempts - 1
    exception = value_error
```

with similar `self.logger.warning("retrying from ...")` for rate-limit, API, and `KeyError` cases. The class already holds `self.journal` (it writes the final `AIMessage` there), so the reason could be surfaced to clients from the same place.

## Proposal

When retrying from a recoverable error, also write a concise diagnostic `AgentMessage` to the journal carrying the underlying reason — the error class plus a short message — so clients see something like:

> Retrying: the model's output could not be parsed (`ValueError`)

Backtraces would stay in the server log only.

## Design consideration

This changes the client-visible event stream for every consumer, so it may be preferable to gate it behind a diagnostics/verbosity flag rather than emit it unconditionally. That trade-off (always-on vs. opt-in) is the main decision for maintainers and is why this is filed as an issue rather than a PR.

## Related

PR #1028 addresses a narrower slice of the same theme — the empty ``Invoking: `` `` label when a tool's serialized payload has no name. This issue is the broader follow-up: surfacing the *reason* behind failed/retried tool calls.

d1donlydfink:
A good idea.

Note to self: There is a consolidation that can be done between tool setup and middleware setup.

d1donlydfink:
Thinking about this a little bit: I am not inclined to use AgentMessages for this reporting, as those get put into the chat history,
bloating token counts going back and forth to LLMs.  These errors are retried, so the chat history would likely confuse the conversation. Probably best not to go there.

I am likely to put these in an AgentFramework message instead.
[...]

PR #1088, opened 2026-06-25
NS-1029 Surface retry-able tool errors as AgentFramework messages.

d1donlydfink:
Addresses #1029  using @thilo11 's original PR as a basis (https://github.com/cognizant-ai-lab/neuro-san/pull/1089)

The additions to the policy here are:
* We always use a SensitiveLogger to log the retry exception as a warning via the SensitiveLogger.  This guy checks the value of LEAF_LOG_SENSITIVE (which is on by default for developer convenience) before doing any logging so as to funnel SAST check concerns to a single API.
* We also always check the SensitiveLogger before sending out the journal message.  The idea here is that this is yet another place where we are sending potentially sensitive info/implementation details out of the server, this time to a client.

Worth noting that:
* We recommend that in production LEAF_LOG_SENSITIVE be turned off from its default of true (see Dockerfile comments here: https://github.com/cognizant-ai-lab/neuro-san/blob/main/neuro_san/deploy/Dockerfile#L349), but default is on for developer convenience
* Production clients shouldn't really care about the errors going on inside the server that are retried. These are implementation details.

A few other nits in here:
* An added comment in the Dockerfile with an info nugget about AGENT_MAX_CONCURRENT_REQUESTS server setting
* Move a block of code up in LangChainRunContext for better readability.

### DIFF
--- a/neuro_san/deploy/Dockerfile
+++ b/neuro_san/deploy/Dockerfile
@@ -212,6 +212,7 @@ ENV AGENT_FORWARDED_REQUEST_METADATA="request_id user_id"
 ENV AGENT_HTTP_PORT="8080"
 
 # Maximm number of requests that can be served at the same time
+# A value of 0 indicates unlimited requests are handled.
 ENV AGENT_MAX_CONCURRENT_REQUESTS="50"
 
 # Number of requests served before the server shuts down in an orderly fashion.
--- a/neuro_san/internals/run_context/langchain/core/langchain_run_context.py
+++ b/neuro_san/internals/run_context/langchain/core/langchain_run_context.py
@@ -174,6 +174,44 @@ class LangChainRunContext(RunContext):
         origination: Origination = self.invocation_context.get_origination()
         self.origin = origination.add_spec_name_to_origin(parent_origin, agent_name)
 
+    def update_from_chat_context(self, chat_context: Dict[str, Any]):
+        """
+        :param chat_context: A ChatContext dictionary that contains all the state necessary
+                to carry on a previous conversation, possibly from a different server.
+        """
+        self.chat_context = chat_context
+
+        if self.chat_context is None:
+            return
+
+        # See if our origin appears in the chat histories.
+        # If so, get ours from there.
+        empty: List[Any] = []
+        chat_histories: List[Dict[str, Any]] = self.chat_context.get("chat_histories", empty)
+        our_origin_str: str = Origination.get_full_name_from_origin(self.origin)
+        for one_chat_history in chat_histories:
+
+            # See if the origin matches our own
+            test_origin: List[Dict[str, Any]] = one_chat_history.get("origin", empty)
+            test_origin_str: str = Origination.get_full_name_from_origin(test_origin)
+            if test_origin_str != our_origin_str:
+                continue
+
+            one_messages: List[Dict[str, Any]] = one_chat_history.get("messages", empty)
+            if not one_messages:
+                # Empty list - Nothing to convert. Use default empty list.
+                break
+
+            converter = BaseMessageDictionaryConverter()
+            self.chat_history = []
+            for chat_message in one_messages:
+                base_message: BaseMessage = converter.from_dict(chat_message)
+                if base_message is not None:
+                    self.chat_history.append(base_message)
+
+            # Nothing left to search for
+            break
+
     async def create_resources(self, agent_name: str,
                                instructions: str,
                                assignments: str,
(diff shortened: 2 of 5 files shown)


neuro_san/internals/run_context/langchain/core/langchain_run_context.py
```diff
@@ -174,6 +174,44 @@ def update_from_parent_run_context(self, parent_run_context: RunContext):
         origination: Origination = self.invocation_context.get_origination()
         self.origin = origination.add_spec_name_to_origin(parent_origin, agent_name)
 
+    def update_from_chat_context(self, chat_context: Dict[str, Any]):
+        """
+        :param chat_context: A ChatContext dictionary that contains all the state necessary
+                to carry on a previous conversation, possibly from a different server.
+        """
```
d1donlydfink, line 181: Move this up from below so the file reads better.

neuro_san/deploy/Dockerfile
```diff
@@ -212,6 +212,7 @@ ENV AGENT_FORWARDED_REQUEST_METADATA="request_id user_id"
 ENV AGENT_HTTP_PORT="8080"
 
 # Maximm number of requests that can be served at the same time
+# A value of 0 indicates unlimited requests are handled.
 ENV AGENT_MAX_CONCURRENT_REQUESTS="50"
```
d1donlydfink, line 216: Add Dockerfile comment with new info nugget.

neuro_san/internals/run_context/langchain/core/run_context_runnable.py
```diff
@@ -208,9 +208,6 @@ async def invoke_agent_chain(self, inputs: Dict[str, Any], runnable_config: Dict
             try:
                 chain_result: Dict[str, Any] = await self.agent_chain.ainvoke(input=inputs, config=runnable_config)
             except RATE_LIMIT_ERROR_TYPES as rate_limit_error:
-                self.logger.warning("retrying from RateLimit error %s(%s)",
-                                    rate_limit_error.__class__.__name__,
-                                    str(rate_limit_error))
                 await self.journal_retry_reason(rate_limit_error, "the LLM provider rate-limited the request")
```
d1donlydfink, line 211: Move the logger warning into journal_retry_reason().
Here and a few other places below.

neuro_san/internals/run_context/langchain/tracing/neuro_san_runnable.py
```diff
@@ -86,6 +89,7 @@ def __init__(
         super().__init__(afunc=run_target.run_it, **kwargs)
 
         self.logger: Logger = getLogger(self.__class__.__name__)
+        self.sensitive_logger: SensitiveLogger = SensitiveLogger(self.logger)
```
d1donlydfink, line 92: Have a SensitiveLogger handy on the parent object.

neuro_san/internals/run_context/langchain/core/run_context_runnable.py
```diff
@@ -297,8 +288,15 @@ async def journal_retry_reason(self, error: Exception, reason: str):
...
-        await self.journal.write_message(AgentFrameworkMessage(content=text))
+        text: str = f"Retrying: {reason} ({error.__class__.__name__}) - {error}"
+
+        # Log the Exception as a warning, respecting server log sensitivity settings
+        self.sensitive_logger.warning(text)
+
+        # Also write the error message to the journal under the same
+        # LEAF_LOG_SENSITIVE env var setting as the SensitiveLogger uses.
+        if self.sensitive_logger.should_log():
+            await self.journal.write_message(AgentFrameworkMessage(content=text))
```
d1donlydfink, line 299: Only send the journal message to the client if the LEAF_LOG_SENSITIVE env var allows it.
By default this is on for developer convenience, but we recommend people turn it off in their deployments to prevent sensitive information leakage.  That is: Production servers should not be broadcasting errors to their production clients.

neuro_san/internals/run_context/langchain/core/run_context_runnable.py
```diff
@@ -297,8 +288,15 @@ async def journal_retry_reason(self, error: Exception, reason: str):
         :param error: The recoverable exception triggering the retry.
         :param reason: A concise, client-facing description of what went wrong.
         """
-        text: str = f"Retrying: {reason} ({error.__class__.__name__})"
-        await self.journal.write_message(AgentFrameworkMessage(content=text))
+        text: str = f"Retrying: {reason} ({error.__class__.__name__}) - {error}"
```
d1donlydfink, line 291: Use a common text string to log.  Add the error text to the message, not just the class.

neuro_san/internals/run_context/langchain/core/run_context_runnable.py
```diff
@@ -297,8 +288,15 @@ async def journal_retry_reason(self, error: Exception, reason: str):
...
+
+        # Log the Exception as a warning, respecting server log sensitivity settings
+        self.sensitive_logger.warning(text)
```
d1donlydfink, line 294: Use the SensitiveLogger to conditionally send exception information.
Using this class funnels down SAST scan false-positive complaints to a single method.

neuro_san/internals/run_context/langchain/tracing/neuro_san_runnable.py
```diff
@@ -63,6 +65,7 @@ class NeuroSanRunnable(RunnablePassthrough, RunTarget):
 
     # Default logger
     logger: Optional[Logger] = None
+    senstive_logger: Optional[SensitiveLogger] = None
```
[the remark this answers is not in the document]
d1donlydfink: Fixed.

neuro_san/internals/run_context/langchain/core/run_context_runnable.py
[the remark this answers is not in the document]
d1donlydfink: Interesting catch, but the sensitivity of the information is precisely why we check a SensitiveLogger for conditional journaling.

andreidenissov-cog (approved):
LGTM.

PR #1089, opened 2026-06-26
Surface recoverable-error retries on the journal as AgentFramework messages

### LINKED ISSUE #1029: Surface tool-call parse/validation failures on the journal, not just the server log

thilo11:
## Summary

Fixes #1029. When the agent loop retries after a recoverable error, the reason was written only to the server log — from a client's activity feed this showed up as an unexplained stall.

Each retry branch in `RunContextRunnable.invoke_agent_chain` (`run_context_runnable.py`) now also emits a concise diagnostic on the journal, e.g.:

> Retrying: the model's output could not be parsed (`ValueError`)

Covered branches: rate-limit, API error (regular-retry path only — the API-key short-circuit is unchanged), `KeyError`, and the non-parse `ValueError` path. Backtraces stay in the server log only.

## Design

Per @d1donlydfink's note on the issue, these are written as **`AgentFrameworkMessage`s** rather than `AgentMessage`s:

- `AgentFrameworkMessage` is excluded from chat history (`is_relevant_to_chat_history`), so retried/recovered errors don't bloat the tokens sent back to the LLM or confuse the conversation.
- It's written through the agent's own journal, so it carries that agent's **origin** and is never mistaken for "the answer" (the last origin-less `AgentFramework` message).
- It's the right architectural layer — these messages originate at the neuro-san level, not langchain.

Because `AgentFrameworkMessage`s don't pollute the chat, emission is unconditional (no verbosity flag needed) — this resolves the always-on vs. opt-in question raised in the issue.

## Tests

`tests/.../core/test_run_context_runnable.py`:
- `journal_retry_reason` writes a single `AgentFrameworkMessage` with the expected content.
- A retried `ValueError` emits a diagnostic per attempt, with the final `AIMessage` still last (journal stream order preserved).

### DIFF
--- a/neuro_san/internals/run_context/langchain/core/run_context_runnable.py
+++ b/neuro_san/internals/run_context/langchain/core/run_context_runnable.py
@@ -41,6 +41,7 @@ from leaf_common.config.resolver_util import ResolverUtil
 
 from neuro_san.internals.errors.error_detector import ErrorDetector
 from neuro_san.internals.journals.journal import Journal
+from neuro_san.internals.messages.agent_framework_message import AgentFrameworkMessage
 from neuro_san.internals.messages.origination import Origination
 from neuro_san.internals.run_context.interfaces.tool_caller import ToolCaller
 from neuro_san.internals.run_context.langchain.journaling.journaling_callback_handler import JournalingCallbackHandler
@@ -210,6 +211,7 @@ class RunContextRunnable(NeuroSanRunnable):
                 self.logger.warning("retrying from RateLimit error %s(%s)",
                                     rate_limit_error.__class__.__name__,
                                     str(rate_limit_error))
+                await self.journal_retry_reason(rate_limit_error, "the LLM provider rate-limited the request")
                 attempts = attempts - 1
                 exception = rate_limit_error
             except API_ERROR_TYPES as api_error:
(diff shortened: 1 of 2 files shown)


d1donlydfink:
@thilo11 : 1M-thanks again for the contribution!

I think I will combine the best of this PR with a different angle of attack in my own going on here:
https://github.com/cognizant-ai-lab/neuro-san/pull/1088

d1donlydfink (approved):
Will use yours as a basis tho.

PR #1104, opened 2026-07-07
NS-1103 Fix start logic for multi-instance service.

andreidenissov-cog:
This PR fixes problems with starting neuro-san service when scaling to several instances is requested.
Several Python objects - like threads and event loops - do not survive process fork,
and so must be created post-forking.

### DIFF
--- a/neuro_san/service/utils/server_context.py
+++ b/neuro_san/service/utils/server_context.py
@@ -17,6 +17,9 @@
 
 from typing import Any
 from typing import Dict
+from typing import Optional
+
+from threading import Lock
 
 from janus import Queue
 
@@ -45,7 +48,16 @@ class ServerContext(ServerContextLite):
         Constructor.
         """
         self.server_status: ServerStatus = None
-        self.executor_pool = AsyncioExecutorPool(reuse_mode=True, idle_timeout_seconds=30)
+        # NB: do NOT construct the AsyncioExecutorPool here. When Tornado is
+        # configured for multiple worker processes (AGENT_HTTP_SERVER_INSTANCES
+        # > 1), the server forks *after* this instance is created but before
+        # any request-path code runs. AsyncioExecutorPool starts a daemon GC
+        # thread in its constructor, and threads do not survive fork -- the
+        # child ends up with a _gc_thread reference to a thread that no longer
+        # exists, and stale executors are never reaped in the workers.
+        # Lazy-construct in get_executor_pool() so each worker builds its own.
+        self.executor_pool: Optional[AsyncioExecutorPool] = None
+        self._executor_pool_lock: Lock = Lock()
         self.queues: Queue[AsyncCollatingQueue] = Queue()
         self.mcp_server_context: McpServerContext = McpServerContext()
         self.server_port: int = AgentSessionConstants.DEFAULT_HTTP_PORT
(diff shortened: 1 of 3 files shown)


neuro_san/service/utils/server_context.py
```diff
@@ -72,8 +84,16 @@ def set_temp_storage_max_items(self, max_items: int):
...
+                 Constructed on first access so that each post-fork worker
+                 gets its own pool (with its own live GC thread), rather
+                 than inheriting a dead reference from the parent.
         """
+        if self.executor_pool is None:
+            with self._executor_pool_lock:
+                if self.executor_pool is None:
+                    self.executor_pool = AsyncioExecutorPool(reuse_mode=True,
+                                                             idle_timeout_seconds=30)
         return self.executor_pool
```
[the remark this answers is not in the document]
andreidenissov-cog: Not that important here.

neuro_san/service/utils/server_context.py
```diff
@@ -45,7 +48,16 @@ def __init__(self):
...
+        # NB: do NOT construct the AsyncioExecutorPool here. When Tornado is
+        # configured for multiple worker processes (AGENT_HTTP_SERVER_INSTANCES
+        # > 1), the server forks *after* this instance is created but before
+        # any request-path code runs. AsyncioExecutorPool starts a daemon GC
+        # thread in its constructor, and threads do not survive fork -- the
+        # child ends up with a _gc_thread reference to a thread that no longer
+        # exists, and stale executors are never reaped in the workers.
+        # Lazy-construct in get_executor_pool() so each worker builds its own.
+        self.executor_pool: Optional[AsyncioExecutorPool] = None
+        self._executor_pool_lock: Lock = Lock()
```
d1donlydfink, line 60: Cool! Great find!

neuro_san/service/watcher/event_initiator/periodic_event_initiator.py
```diff
@@ -55,13 +55,24 @@ def __init__(self, server_context: ServerContext):
...
-        self.executor = AsyncioExecutor()
+        # NB: do NOT construct the AsyncioExecutor here. When Tornado is
+        # configured for multiple worker processes (AGENT_HTTP_SERVER_INSTANCES
+        # > 1), the server forks *after* this instance is created but before
+        # run() is invoked. asyncio event loops -- and their kqueue/epoll
+        # selectors -- do not survive fork; a pre-fork loop yields
+        # "I/O operation on closed kqueue object" the moment a child tries to
+        # run it. Construct the executor in run() below so each worker gets
+        # its own fresh loop after fork.
+        self.executor: AsyncioExecutor | None = None
```
d1donlydfink, line 66: Makes sense.
Question: I would not expect there to be one of these EventInitiator guys per process.
Seems like we only need one.  Is there any way to be sure there is only one per server instance?  (Doesn't have to be this PR).

FWIW: We would have this singleton problem across multiple pods as well and that would needs its own separate mechanism.

neuro_san/service/watcher/temp_networks/temp_network_storage_updater.py
```diff
@@ -105,6 +97,15 @@ def start(self):
...
 
         self.logger.info("Starting TempNetworkStorageUpdater")
+
+        # We use an AsyncioExecutor and its event loop to process the collection of tasks
+        # this TempNetworkStorageUpdater instance is responsible for:
+        # 1. Checking for new queues to process coming in self.incoming "main" queue
+        # 2. Processing the individual queues that come in,
+        #    which involves processing the items that come in on those queues
+        #    and deploying reservations to the temp storage, which is done by self.reservationist.
+        self.executor = AsyncioExecutor()
```
d1donlydfink, line 107: I suppose it does make sense that each server process would need its own TempNetworkStorageUpdater so they can share a queue-of-queues (outer queue).

Will we need to be sure that those queue-of-queues are also lazily initialized?

d1donlydfink (approved):
Looks good for the question of AsyncioExecutors.

There are a few more ?s I have in the PR regarding similar lazy/optional initialization for server processes for:
* PeriodicEventInitiator (optional - only need 1 per server probably and we can solve later)
* TempNetworkStorageUpdater - each process will likely need its own queue-of-queues.  This will be more relevant sooner to MAUI w/ temp networks.

PR #1111, opened 2026-07-09
Use latest leaf-common==1.2.45

d1donlydfink:


requirements.txt
```diff
@@ -6,7 +6,7 @@
 # requirements-build.txt
 
 # In-house dependencies
-leaf-common>=1.2.44
+leaf-common>=1.2.45
```
d1donlydfink, line 9: Use latest leaf-common==1.2.45

PR #1115, opened 2026-07-09
Add visibility into Reservation deployment failures by logging

d1donlydfink:
Catching exceptions at this level will also no longer block the request, however that does not yet mean lack of success is communicated to the request. We want to see what kind of errors we are getting first.

### DIFF
--- a/neuro_san/service/watcher/temp_networks/temp_network_storage_updater.py
+++ b/neuro_san/service/watcher/temp_networks/temp_network_storage_updater.py
@@ -16,6 +16,7 @@
 # END COPYRIGHT
 from typing import Any
 from typing import Dict
+from typing import List
 
 from os import environ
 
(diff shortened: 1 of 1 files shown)


neuro_san/service/watcher/temp_networks/temp_network_storage_updater.py
```diff
@@ -206,9 +208,25 @@ async def process_one_queued_item(self, queued_item: Dict[str, Any]):
...
+            # Gather info about who will be effected by this failure
+            reservation_ids: List[str] = []
+            for reservation in deployment_dict.keys():
+                reservation_ids.append(reservation.get_reservation_id())
+
+            sensitive_logger = SensitiveLogger(self.logger)
+            sensitive_logger.exception("Exception while processing queued item for reservations: %s", reservation_ids)
 
         # Maybe notify the deployer.
+        # Note: even if deployment fails, we still notify (set the event) to avoid callers waiting indefinitely.
```
d1donlydfink, line 229: Make a note that even if there is a failure, we will (now) be un-blocking the request.
At some point we could communicate the failure to the request, but for now we really just want to understand what any failure might be.

neuro_san/service/watcher/temp_networks/temp_network_storage_updater.py
```diff
@@ -206,9 +208,25 @@ async def process_one_queued_item(self, queued_item: Dict[str, Any]):
...
+                self.logger.warning("Deployment dictionary was None")
+                deployment_dict = {}
+
+            # Gather info about who will be effected by this failure
+            reservation_ids: List[str] = []
+            for reservation in deployment_dict.keys():
+                reservation_ids.append(reservation.get_reservation_id())
+
+            sensitive_logger = SensitiveLogger(self.logger)
+            sensitive_logger.exception("Exception while processing queued item for reservations: %s", reservation_ids)
```
d1donlydfink, line 226: Log failures upon unknown exceptions when deploying Reservations so we know who is effected and by what kind of exception.

vince-leaf (approved):
Look Fine; not much risk.

PR #1125, opened 2026-07-11
Cache / Retry frozen s3 credentials for S3ReservationsWriter

d1donlydfink:
I learned that:
* We only need 1 Boto session for the lifetime of long-running stuff
* The session will cache authorization credentials
* It is possible for these cached credentials to expire depending on the AWS authorization setup

So in this PR we:
* Lazily get our own "frozen credentials" snapshot to re-use when creating a client.
* Cache these credentials for use with creating new clients.
  (Apparently creating a Client is a relatively lightweight operation. Creating a new Session is the heavyweight aspect)
* Invalidate the cached frozen credentials and retry when we detect these credentials have expired.

If this approach is successful in boosting performance we should replicate this across Reader and Expiration.

Tested locally on santa's workshop

This has potential to take upwards of 30s off some requests in the load testing.

### DIFF
--- a/neuro_san/service/watcher/temp_networks/s3_reservations_writer.py
+++ b/neuro_san/service/watcher/temp_networks/s3_reservations_writer.py
@@ -266,7 +266,7 @@ class S3ReservationsWriter:
             self.session: AioSession = get_session()
 
         credentials: Credentials = await self.session.get_credentials()
-        self.frozen_credentials = credentials.get_frozen_credentials()
+        self.frozen_credentials = await credentials.get_frozen_credentials()
 
         return self.frozen_credentials
 
--- a/neuro_san/service/watcher/temp_networks/s3_reservations_writer.py
+++ b/neuro_san/service/watcher/temp_networks/s3_reservations_writer.py
@@ -93,6 +93,7 @@ class S3ReservationsWriter:
         self.session: AioSession = None
 
         # Cached frozen credentials which can be invalidated should they expire
+        self.frozen_credentials: ReadOnlyCredentials = None
 
     async def add_reservations(self, reservations_dict: Dict[Reservation, Any],
                                source: str = None):
--- a/neuro_san/service/watcher/temp_networks/s3_reservations_writer.py
+++ b/neuro_san/service/watcher/temp_networks/s3_reservations_writer.py
@@ -274,9 +274,12 @@ class S3ReservationsWriter:
         We are assuming that we already have the async_s3_client_lock.
         """
 
+        # Use a local copy to avoid a race with the ClientError retry block
+        local_frozen: ReadOnlyCredentials = self.frozen_credentials
+
         # If we already have some credentials, use them
-        if self.frozen_credentials is not None:
-            return self.frozen_credentials
+        if local_frozen is not None:
+            return local_frozen
 
         # Create the session if needed
         # We should only need one for the lifetime of the object
@@ -287,7 +290,7 @@ class S3ReservationsWriter:
 
         # Avoid a small race condition with the ClientError retry block
         # by always returning a local copy
-        local_frozen: ReadOnlyCredentials = await credentials.get_frozen_credentials()
+        local_frozen = await credentials.get_frozen_credentials()
         self.frozen_credentials = local_frozen
 
         return local_frozen
--- a/neuro_san/service/watcher/temp_networks/s3_reservations_writer.py
+++ b/neuro_san/service/watcher/temp_networks/s3_reservations_writer.py
@@ -36,7 +36,9 @@ from aiobotocore.session import ClientCreatorContext
 
 from botocore.credentials import Credentials
 from botocore.credentials import ReadOnlyCredentials
-from botocore.exceptions import NoCredentialsError
+from botocore.exceptions import ClientError
+
+from leaf_common.parser.dictionary_extractor import DictionaryExtractor
 
 from neuro_san.interfaces.reservation import Reservation
 from neuro_san.internals.reservations.reservation_dictionary_converter import ReservationDictionaryConverter
@@ -161,7 +163,10 @@ class S3ReservationsWriter:
                 await self.add_reservations_with_new_client(reservations_dict, source, attempts)
                 success = True
 
-            except NoCredentialsError:
+            except ClientError as err:
+                extractor = DictionaryExtractor(err.response)
+                if "ExpiredToken" not in extractor.get("Error.Code", ""):
+                    raise
 
                 # Background: Certain IAM Instance Roles, ECS Task Roles or AWS SSO/IAM Identity Center
                 #             profiles have token-based credentials that may expire.
--- a/neuro_san/service/watcher/temp_networks/s3_reservations_writer.py
+++ b/neuro_san/service/watcher/temp_networks/s3_reservations_writer.py
@@ -266,7 +266,7 @@ class S3ReservationsWriter:
         # Create the session if needed
         # We should only need one for the lifetime of the object
         if self.session is None:
-            self.session: AioSession = get_session()
+            self.session = get_session()
 
         credentials: Credentials = await self.session.get_credentials()
         self.frozen_credentials = await credentials.get_frozen_credentials()
--- a/neuro_san/service/watcher/temp_networks/s3_reservations_writer.py
+++ b/neuro_san/service/watcher/temp_networks/s3_reservations_writer.py
@@ -145,6 +145,7 @@ class S3ReservationsWriter:
                                                                   source: str):
         """
         Retries adding the reservations when client credentials can expire.
+
         :param reservations_dict: A mapping of Reservation -> some deployable agent spec
         :param source: A string describing where the deployment was coming from
         """
@@ -156,10 +157,18 @@ class S3ReservationsWriter:
             try:
                 await self.add_reservations_with_new_client(reservations_dict, source, attempts)
                 success = True
+
             except NoCredentialsError:
+
+                # Background: Certain IAM Instance Roles, ECS Task Roles or AWS SSO/IAM Identity Center
+                #             profiles have token-based credentials that may expire.
+                #             See: https://docs.aws.amazon.com/boto3/latest/guide/configuration.html
+
                 # Reset the cached credentials as they are likely expired and try again.
                 self.frozen_credentials = None
-                self.logger.warning("%s (%d): S3 credentials seem to have expired. Retrying.", source, attempts)
+                self.logger.warning("%s (%d): S3 credentials seem to have expired. Retrying."
+                                    "If you believe you have non-expiring S3 credentials, be sure they are correct.",
+                                    source, attempts)
 
     async def add_reservations_with_new_client(self,
                                                reservations_dict: Dict[Reservation, Any],
--- a/neuro_san/service/watcher/temp_networks/s3_reservations_writer.py
+++ b/neuro_san/service/watcher/temp_networks/s3_reservations_writer.py
@@ -250,7 +250,7 @@ class S3ReservationsWriter:
                 self.async_s3_client_lock.release()
 
         finish_time: float = perf_counter()
-        self.logger.info("%s (%d): Lock acquisition in: %fs. Client creation after: %fs. "
+        self.logger.info("%s (%d): Lock acquisition in: %fs. Client context creation after: %fs. "
                          "Lock release after: %fs. Finish after: %fs",
                          self.name, attempt,
                          lock_aquired_time - start_time,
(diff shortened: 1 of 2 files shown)


neuro_san/service/watcher/temp_networks/s3_reservations_writer.py
```diff
@@ -81,10 +86,18 @@ def __init__(self, name: str = "S3ReservationsWriter", bucket_name: str = "", pr
 
         self.converter = ReservationDictionaryConverter()
 
+        # Boto Machinations
+        # We should be able to have a single Session for the lifetime of this object
+        self.session: AioSession = None
+
+        # Cached frozed credentials which can be invalidated should they expire
+        self.frozen_credentials: ReadOnlyCredentials = None
```
d1donlydfink, line 96: New members to keep around.

neuro_san/service/watcher/temp_networks/s3_reservations_writer.py
```diff
@@ -122,12 +135,56 @@ async def add_reservations(self, reservations_dict: Dict[Reservation, Any],
...
+
+    async def add_reservations_with_new_client_credential_retries(self,
+                                                                  reservations_dict: Dict[Reservation, Any],
+                                                                  source: str):
+        """
+        Retries adding the reservations when client credentials can expire.
+
+        :param reservations_dict: A mapping of Reservation -> some deployable agent spec
+        :param source: A string describing where the deployment was coming from
+        """
```
d1donlydfink, line 156: Introduce a new layer of retries for NoClientCredentials to catch the cases where token-based AWS auth expires.

neuro_san/service/watcher/temp_networks/s3_reservations_writer.py
```diff
@@ -138,20 +195,23 @@ async def add_reservations(self, reservations_dict: Dict[Reservation, Any],
 
         start_time: float = perf_counter()
         lock_aquired_time: float = 0.0
-        session_created_time: float = 0.0
+        client_created_time: float = 0.0
```
d1donlydfink, line 213: Rename timing checkpoint as we are now interested in how long it takes to create a client
which is born of a Session and should take much less time now.

neuro_san/service/watcher/temp_networks/s3_reservations_writer.py
```diff
@@ -122,12 +135,56 @@ async def add_reservations(self, reservations_dict: Dict[Reservation, Any],
...
+                                               source: str,
+                                               attempt: int = 1):
+        """
+        This method separates the machinactions of obtaining a proper S3 client
+        from add_all_reservations() which does all the actual work.
+
+        :param reservations_dict: A mapping of Reservation -> some deployable agent spec
+        :param source: A string describing where the deployment was coming from
+        :param attempt: Attempt number
+        """
```
d1donlydfink, line 202: The idea behind what is in this method is to contain all the S3 setup machinations.
It then calls add_all_reservations() which does the actual work.
This way we can mess with how the client gets set-up without interfering with the logic of the work that is done.

neuro_san/service/watcher/temp_networks/s3_reservations_writer.py
```diff
@@ -138,20 +195,23 @@ async def add_reservations(self, reservations_dict: Dict[Reservation, Any],
...
-            session_created_time = perf_counter()
+            # Get the current notion of frozen credentials.
+            frozen_creds: ReadOnlyCredentials = await self.get_frozen_credentials()
+            async_s3_client_creator_context = self.session.create_client(
+                "s3",
+                aws_access_key_id=frozen_creds.access_key,
+                aws_secret_access_key=frozen_creds.secret_key,
+                aws_session_token=frozen_creds.token
+            )
+            client_created_time = perf_counter()
```
d1donlydfink, line 229: Instead of creating a Session at this level, focus instead on getting the "frozen credentials" we want to pass into the client.  Call the new get_frozen_credentials() method to manage what we cache for this.

neuro_san/service/watcher/temp_networks/s3_reservations_writer.py
```diff
@@ -165,11 +225,7 @@ async def add_reservations(self, reservations_dict: Dict[Reservation, Any],
                 lock_released_time = perf_counter()
                 lock_released = True
 
-                # Process each reservation/agent spec pair individually
-                reservation: Reservation = None
-                agent_spec: Dict[str, Any] = None
-                for reservation, agent_spec in reservations_dict.items():
-                    await self.add_one_reservation(async_s3_client, reservation, agent_spec, source)
+                await self.add_all_reservations(async_s3_client, reservations_dict, source)
```
d1donlydfink, line 243: The above tiny block moves into the new "work" method add_all_reservations()

neuro_san/service/watcher/temp_networks/s3_reservations_writer.py
```diff
@@ -178,18 +234,64 @@ async def add_reservations(self, reservations_dict: Dict[Reservation, Any],
...
-
-        self.logger.info("Lock acquisition in: %fs. Session creation after: %fs. "
+        self.logger.info("%s (%d): Lock acquisition in: %fs. Client creation after: %fs. "
                          "Lock release after: %fs. Finish after: %fs",
+                         self.name, attempt,
                          lock_aquired_time - start_time,
-                         session_created_time - start_time,
+                         client_created_time - start_time,
                          lock_released_time - start_time,
                          finish_time - start_time)
```
d1donlydfink, line 258: Update log message.

neuro_san/service/watcher/temp_networks/s3_reservations_writer.py
```diff
@@ -178,18 +234,64 @@ async def add_reservations(self, reservations_dict: Dict[Reservation, Any],
...
+    def ensure_async_lock_exists(self):
+        """
+        Be sure we have an asyncio Lock to get our sessions.
+        """
+        # Note that none of this is async in and of itself.
+        if self.async_s3_client_lock is None:
+            with self.sync_lock:
+                # Be sure everyone has the same lock
+                if self.async_s3_client_lock is None:
+                    self.async_s3_client_lock = AsyncLock()
```
d1donlydfink, line 269: Logic for ensuring the async lock exists moves down here.

neuro_san/service/watcher/temp_networks/s3_reservations_writer.py
```diff
@@ -178,18 +234,64 @@ async def add_reservations(self, reservations_dict: Dict[Reservation, Any],
...
+
+    async def get_frozen_credentials(self) -> ReadOnlyCredentials:
+        """
+        Get the frozen credentials for the current session.
+        We are asssuming that we already have the async_s3_client_lock.
+        """
```
d1donlydfink, line 275: Cache management of frozen s3 credentials.

neuro_san/service/watcher/temp_networks/s3_reservations_writer.py
```diff
@@ -178,18 +234,64 @@ async def add_reservations(self, reservations_dict: Dict[Reservation, Any],
...
+
+    async def add_all_reservations(self, async_s3_client: AioBaseClient,
+                                   reservations_dict: Dict[Reservation, Dict[str, Any]],
+                                   source: str):
+        """
+        Add all reservations to S3 storage.
+        :param async_s3_client: An aiobotocore S3 client to use for the put_object call
+        :param reservations_dict: A mapping of Reservation -> some deployable agent spec
+        :param source: A string describing where the deployment was coming from
+        """
```
d1donlydfink, line 306: The work that is done with the client is now this add_all_reservations() method.

neuro_san/service/watcher/temp_networks/s3_reservations_writer.py
```diff
@@ -221,8 +323,6 @@ async def add_one_reservation(self, async_s3_client: AioBaseClient,
                                Body=json_body,
                                ContentType="application/json")
 
-        if source is None:
-            source = self.name
```
d1donlydfink, line 225: Checking for a None source moves to add_reservations()
No need for the optional args all over.

neuro_san/service/watcher/temp_networks/s3_reservations_writer.py
```diff
@@ -122,12 +132,56 @@ async def add_reservations(self, reservations_dict: Dict[Reservation, Any],
...
-                # Be sure everyone has the same lock
-                if self.async_s3_client_lock is None:
-                    self.async_s3_client_lock = AsyncLock()
+        if source is None:
+            source = self.name
+
+        # Aync lock has to be created in the thread that uses it.
+        self.ensure_async_lock_exists()
+
+        await self.add_reservations_with_new_client(reservations_dict, source)
```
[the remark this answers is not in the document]
d1donlydfink: Done.

neuro_san/service/watcher/temp_networks/s3_reservations_writer.py
```diff
@@ -122,12 +132,56 @@ async def add_reservations(self, reservations_dict: Dict[Reservation, Any],
...
+
+    async def add_reservations_with_new_client_credential_retries(self,
+                                                                  reservations_dict: Dict[Reservation, Any],
+                                                                  source: str):
+        """
```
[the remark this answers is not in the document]
d1donlydfink: Want to see if this approach works before investing.

neuro_san/service/watcher/temp_networks/s3_reservations_writer.py
```diff
@@ -122,12 +132,56 @@ async def add_reservations(self, reservations_dict: Dict[Reservation, Any],
...
+                # Background: Certain IAM Instance Roles, ECS Task Roles or AWS SSO/IAM Identity Center
+                #             profiles have token-based credentials that may expire.
+                #             See: https://docs.aws.amazon.com/boto3/latest/guide/configuration.html
+
+                # Reset the cached credentials as they are likely expired and try again.
+                self.frozen_credentials = None
+                self.logger.warning("%s (%d): S3 credentials seem to have expired. Retrying."
+                                    "If you believe you have non-expiring S3 credentials, be sure they are correct.",
+                                    source, attempts)
+
```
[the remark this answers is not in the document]
d1donlydfink: Addressed.

tests/neuro_san/service/watcher/temp_networks/s3_reservations_storage_test_base.py
```diff
@@ -97,6 +99,16 @@ def setUp(self):
...
 
+        # Patch AioSession at the import boundary in s3_reservations_storage
+        # so storage.start() receives our fake credentials.
+        credentials_patcher = patch(
+            "neuro_san.service.watcher.temp_networks.s3_reservations_writer.AioSession.get_credentials",
+            return_value=AioCredentials("bogus", "bogus")
+        )
+        credentials_patcher.start()
+        # Restores the real AioSession symbol after the test completes, regardless of pass/fail.
+        self.addCleanup(credentials_patcher.stop)
```
d1donlydfink, line 113: Patch Credentials in tests.

neuro_san/service/watcher/temp_networks/s3_reservations_writer.py
```diff
@@ -165,11 +240,7 @@ async def add_reservations(self, reservations_dict: Dict[Reservation, Any],
                 lock_released_time = perf_counter()
                 lock_released = True
```
d1donlydfink, line 241: FWIW: I suspect that we should be able to get rid of this unusual locking situation we have here because I think the "get the auth done" steps are now in the get_frozen_credentials() bit.  I didn't want to put real progress at risk if that was not the case though. So, perhaps after this is proven to be better we can then prove the locking strategy can be more sane.

neuro_san/service/watcher/temp_networks/s3_reservations_writer.py
```diff
@@ -138,20 +210,23 @@ async def add_reservations(self, reservations_dict: Dict[Reservation, Any],
...
         start_time: float = perf_counter()
         lock_aquired_time: float = 0.0
-        session_created_time: float = 0.0
+        client_created_time: float = 0.0
         lock_released_time: float = 0.0
         try:
             # Serialize creation of the ClientCreatorContext with the lock to avoid credential-chain races.
             await self.async_s3_client_lock.acquire()
             lock_aquired_time = perf_counter()
             acquired_lock = True
```
andreidenissov-cog, line 220: Do we still need self.async_s3_client_lock and all related machinations?

d1donlydfink: I believe so. This is what prevents the NoCredentialsError in the first place.
When the credentials are actually cached, the critical section becomes much smaller.

d1donlydfink:
Merging for progress. Feel free to continue to comment.

PR #1138, opened 2026-07-14
Neuro-san probes running on a separate thread/event loop.

andreidenissov-cog:
Problem

  Under load, neuro-san's Tornado main event loop can become saturated by request work (streaming chat, LLM I/O, etc.). Because /, /healthz, /readyz, and /livez were handled on that same loop, probe responses stalled along with everything else — leading to Kubernetes killing pods that were slow rather than dead ("death by health check").

  Solution

  Runs the health-probe endpoints on a separate port, in a dedicated OS thread with its own asyncio event loop. The main Tornado loop can be arbitrarily busy and probes still return quickly.

  New module — neuro_san/service/http/server/health_probe_server.py:
  - HealthProbeServer(Startable) — launches a daemon thread that creates its own asyncio loop and runs a minimal Tornado Application with only the four health routes.
  - Reuses the existing HealthCheckHandler verbatim, so probe response bodies are byte-for-byte identical to the main port.
  - Binds with SO_REUSEPORT so every Tornado worker binds the same probe port; the kernel dispatches accepted connections across workers — same model Tornado uses for the main port.
  - Instantiated pre-fork but start() runs post-fork (as a Startable), so each worker mints its own fresh loop — no kqueue/epoll fd inherited from parent.
  - Clean shutdown via contextlib.suppress(OSError) on socket teardown; daemon thread dies at process exit if stop() isn't called.
  
  http_server.py wiring:
  - New static _resolve_health_probe_port() reads AGENT_HEALTH_PROBE_PORT (default 8081, 0 disables).
  - Appends the HealthProbeServer to the startables list just before server.bind(), so it starts per-worker after server.start(N) fork.
  - Backward compatibility: main-port handlers for /, /healthz, /readyz, /livez remain registered — existing k8s configs pointing at the main port keep working.
  
  Dockerfile — new ENV AGENT_HEALTH_PROBE_PORT="8081" with commentary on the mechanism and how to disable.

  Recommended k8s probe wiring

  livenessProbe:
    httpGet:
      path: /livez
      port: 8081                # isolated loop → probes survive main-loop saturation
    failureThreshold: 3
    periodSeconds: 15
    timeoutSeconds: 3
  readinessProbe:
    httpGet:
      path: /readyz
      port: 8080                # main port → readiness fails when loop is saturated
                                        # (LB drops the pod, no restart)
    failureThreshold: 1
    periodSeconds: 5
    timeoutSeconds: 3

  Split gives the correct semantic for each concern: liveness reflects "process alive" from the isolated loop (survives transient main-loop stalls); readiness reflects "can I serve requests" from the main loop (correctly withdraws saturated pods from LB without restarting them).

  Trade-off worth flagging

  A /livez 200 on port 8081 no longer means "the main loop is healthy" — only "the process is alive." That's the right semantic for liveness (which triggers restart), but the failure mode "main loop wedged, probes still fine, requests still failing" is now hidden from the liveness probe by design. The readiness probe on the main port catches that case.

  Verification

  Smoke test against the isolated probe server showed all four routes responding on the dedicated loop with sub-10ms latency and correct status-code behavior (200 when ready, 503 when readiness signals not set).

### DIFF
--- a/neuro_san/deploy/Dockerfile
+++ b/neuro_san/deploy/Dockerfile
@@ -211,6 +211,20 @@ ENV AGENT_FORWARDED_REQUEST_METADATA="request_id user_id"
 # for it on your docker run command line.
 ENV AGENT_HTTP_PORT="8080"
 
+# Dedicated port for the isolated health-probe server. Serves /, /healthz,
+# /readyz, /livez on its own asyncio event loop in a dedicated thread, so
+# probes stay fast even when the main Tornado loop is saturated by request
+# work. Uses SO_REUSEPORT so every worker process binds the same port and
+# the kernel distributes accepted connections across them.
+#
+# Set to 0 to disable the isolated probe server (in that case probes must
+# hit AGENT_HTTP_PORT, which continues to register the same handlers for
+# backward compatibility).
+#
+# Point your Kubernetes livenessProbe.httpGet.port at this port for
+# best-case probe responsiveness under load.
+ENV AGENT_HEALTH_PROBE_PORT="8081"
+
 # Maximm number of requests that can be served at the same time
 # A value of 0 indicates unlimited requests are handled.
 ENV AGENT_MAX_CONCURRENT_REQUESTS="50"
--- a/neuro_san/service/http/config/http_server_config.py
+++ b/neuro_san/service/http/config/http_server_config.py
@@ -17,12 +17,17 @@
 """
 See class comment for details
 """
+# Env var: dedicated port for the isolated health-probe server. 0 disables it,
+# in which case k8s/LB probes must hit the main server port's handlers (which
+# remain registered for backward compatibility).
+ENV_HEALTH_PROBE_PORT: str = "AGENT_HEALTH_PROBE_PORT"
 
 DEFAULT_HTTP_CONNECTIONS_BACKLOG: int = 128
 DEFAULT_HTTP_IDLE_CONNECTIONS_TIMEOUT_SECONDS: int = 3600
 DEFAULT_HTTP_SERVER_INSTANCES: int = 1
 DEFAULT_HTTP_SERVER_MONITOR_INTERVAL_SECONDS: int = 0
 DEFAULT_KEEP_ALIVE_INTERVAL_SECONDS: int = 0
+DEFAULT_HEALTH_PROBE_PORT: int = 8081
 
 
 class HttpServerConfig:
@@ -35,5 +40,6 @@ class HttpServerConfig:
         self.http_idle_connection_timeout_seconds: int = DEFAULT_HTTP_IDLE_CONNECTIONS_TIMEOUT_SECONDS
         self.http_server_instances: int = DEFAULT_HTTP_SERVER_INSTANCES
         self.http_port: int = 80
+        self.http_probe_port: int = DEFAULT_HEALTH_PROBE_PORT
         self.http_server_monitor_interval_seconds: int = DEFAULT_HTTP_SERVER_MONITOR_INTERVAL_SECONDS
         self.stream_keep_alive_with_progress_interval_seconds: int = DEFAULT_KEEP_ALIVE_INTERVAL_SECONDS
--- a/neuro_san/service/http/handlers/health_check_handler.py
+++ b/neuro_san/service/http/handlers/health_check_handler.py
@@ -25,10 +25,6 @@ import http
 import os
 
 from http import HTTPStatus
-from threading import Lock
-
-from importlib.metadata import version as library_version
-from importlib.metadata import PackageNotFoundError
 
 from tornado.web import RequestHandler
 from neuro_san.service.http.logging.http_logger import HttpLogger
@@ -40,18 +36,14 @@ class HealthCheckHandler(RequestHandler):
     Handler class for API endpoint health check.
     """
 
-    # Current service versions are cached in a class variable to avoid repeated lookups.
-    # Note that we cannot do it in the handler initialization
-    # because the handler is created new for each request
-    service_versions: Dict[str, Any] | None = None
-    versions_lock: Lock = Lock()
-
     # pylint: disable=attribute-defined-outside-init
+    # pylint: disable=too-many-positional-arguments,too-many-arguments
     def initialize(self,
                    forwarded_request_metadata: List[str],
                    server_status: ServerStatus,
                    op: str,
-                   logging_config: Dict[str, Any]):
+                   logging_config: Dict[str, Any],
+                   versions: Dict[str, Any]):
         """
         This method is called by Tornado framework to allow
         injecting service-specific data into local handler context.
@@ -62,6 +54,10 @@ class HealthCheckHandler(RequestHandler):
                    "ready" for /readyz query
                    "live" for /livez query
         :param logging_config: logging configuration dictionary
+        :param versions: pre-computed library version dict returned in the
+                   "versions" field of a successful health response. Supplied
+                   from ServerStatus.get_versions() so per-request handlers
+                   do not have to re-resolve importlib metadata on every probe.
         """
         self.logger = HttpLogger(forwarded_request_metadata, logging_config)
         if op == "ready":
@@ -69,6 +65,7 @@ class HealthCheckHandler(RequestHandler):
         else:
             self.status = server_status.is_server_live()
         self.server_name = server_status.get_server_name()
+        self.versions = versions
 
         if os.environ.get("AGENT_ALLOW_CORS_HEADERS") is not None:
             self.set_header("Access-Control-Allow-Origin", "*")
@@ -82,14 +79,10 @@ class HealthCheckHandler(RequestHandler):
 
         try:
             if self.status:
-                if HealthCheckHandler.service_versions is None:
-                    with HealthCheckHandler.versions_lock:
-                        if HealthCheckHandler.service_versions is None:
-                            HealthCheckHandler.service_versions = HealthCheckHandler.determine_versions()
                 result_dict: Dict[str, Any] = {
                     "service": self.server_name,
                     "status": "ok",
-                    "versions": HealthCheckHandler.service_versions
+                    "versions": self.versions,
                 }
                 self.set_header("Content-Type", "application/json")
                 self.write(result_dict)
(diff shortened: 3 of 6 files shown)


neuro_san/deploy/Dockerfile
```diff
@@ -211,6 +211,20 @@ ENV AGENT_FORWARDED_REQUEST_METADATA="request_id user_id"
...
+# the kernel distributes accepted connections across them.
+#
+# Set to 0 to disable the isolated probe server (in that case probes must
+# hit AGENT_HTTP_PORT, which continues to register the same handlers for
+# backward compatibility).
+#
+# Point your Kubernetes livenessProbe.httpGet.port at this port for
+# best-case probe responsiveness under load.
+ENV AGENT_HEALTH_PROBE_PORT="8081"
+
```
andreidenissov-cog, line 227: New env variable for configuring probes server HTTP port.

neuro_san/service/http/config/http_server_config.py
```diff
@@ -35,5 +40,6 @@ def __init__(self):
         self.http_idle_connection_timeout_seconds: int = DEFAULT_HTTP_IDLE_CONNECTIONS_TIMEOUT_SECONDS
         self.http_server_instances: int = DEFAULT_HTTP_SERVER_INSTANCES
         self.http_port: int = 80
+        self.http_probe_port: int = DEFAULT_HEALTH_PROBE_PORT
         self.http_server_monitor_interval_seconds: int = DEFAULT_HTTP_SERVER_MONITOR_INTERVAL_SECONDS
```
andreidenissov-cog, line 44: One more system default - for separate probes serving mini-server.

neuro_san/service/http/server/health_probe_server.py
```diff
@@ -0,0 +1,202 @@
...
+            self.logger.debug("HealthProbeServer already started; skipping")
+            return
+
+        self._thread = threading.Thread(
+            target=self._run,
+            name=f"HealthProbeServer-{self.port}",
+            daemon=True,
+        )
+        self._thread.start()
+        self.logger.info("HealthProbeServer started on port %d", self.port)
```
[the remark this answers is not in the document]
andreidenissov-cog: Log message moved to the more appropriate place - when thread has really started functioning.

neuro_san/deploy/Dockerfile
d1donlydfink: This has to propagate upwards to neuro-san-studio Dockerfile when this is released.

neuro_san/service/http/handlers/health_check_handler.py
```diff
@@ -75,11 +82,14 @@ async def get(self):
 
         try:
             if self.status:
-                versions: Dict[str, Any] = self.determine_versions()
+                if HealthCheckHandler.service_versions is None:
+                    with HealthCheckHandler.versions_lock:
+                        if HealthCheckHandler.service_versions is None:
+                            HealthCheckHandler.service_versions = HealthCheckHandler.determine_versions()
```
d1donlydfink: Seems like we could do this without the lock if it was passed in as an arg
after doing it once within the server.

neuro_san/service/http/server/health_probe_server.py
```diff
@@ -0,0 +1,209 @@
...
+
+
+class HealthProbeServer(Startable):
+    """
+    Runs the /, /healthz, /readyz, /livez health-probe endpoints on a
+    dedicated port, in a dedicated OS thread with its own asyncio event
+    loop. This isolates probe responsiveness from the main Tornado loop
+    that serves streaming_chat and other request traffic -- so probes
+    keep returning quickly even when the main loop is saturated by
+    request work.
```
d1donlydfink, line 48: Does this defeat the purpose of the health checks?

neuro_san/service/http/config/http_server_config.py
```diff
@@ -17,12 +17,17 @@
...
+# in which case k8s/LB probes must hit the main server port's handlers (which
+# remain registered for backward compatibility).
+ENV_HEALTH_PROBE_PORT: str = "AGENT_HEALTH_PROBE_PORT"
 
 DEFAULT_HTTP_CONNECTIONS_BACKLOG: int = 128
 DEFAULT_HTTP_IDLE_CONNECTIONS_TIMEOUT_SECONDS: int = 3600
 DEFAULT_HTTP_SERVER_INSTANCES: int = 1
 DEFAULT_HTTP_SERVER_MONITOR_INTERVAL_SECONDS: int = 0
 DEFAULT_KEEP_ALIVE_INTERVAL_SECONDS: int = 0
+DEFAULT_HEALTH_PROBE_PORT: int = 8081
```
d1donlydfink, line 30: I can tell you from my own experience that dev ops people do not like opening a new port in their firewall specs.  This is probably a non-standard way to do this.

@nick-cognizant : Thoughts?

neuro_san/service/http/server/http_server.py
```diff
@@ -156,6 +159,20 @@ def start(self, startables: List[Startable]):
...
+        # a dedicated thread -- so probes stay fast even when the main
+        # Tornado loop is saturated by request work.
+        probe_port: int = self.resolve_health_probe_port()
+        if probe_port > 0:
+            startables.append(HealthProbeServer(
+                port=probe_port,
+                server_context=self.server_context,
+                forwarded_request_metadata=self.forwarded_request_metadata,
+                logging_config=self.logging_config,
+            ))
```
d1donlydfink, line 180: what happens if the probe port is the same as the http server itself?

d1donlydfink (commented):
A bunch of questions.
I think you should have @nick-cognizant be a reviewer as well.

neuro_san/service/http/server/health_probe_server.py
andreidenissov-cog: In what way?

neuro_san/service/http/server/http_server.py
andreidenissov-cog: That one is a good question.
Then we should log and drop starting this thread.
Or maybe make this terminal configuration error.

neuro_san/service/http/config/http_server_config.py
nick-cognizant: if this is just an internal port I am fine with it.

nick-cognizant (commented):
1. Liveness may be too permissive to be useful. With /livez on the separate loop and failureThreshold: 15, it basically only means "the process is running" and will almost never restart a pod — so a truly hung main loop won't self-heal. Suggestion: have /livez report unhealthy if the main loop hasn't made progress in N seconds (a heartbeat/last-tick check), so it still catches a real hang. If we keep it as pure process-alive, let's drop failureThreshold to something like 3–5 so it can actually act.
2. Readiness failureThreshold: 1 will flap. One 3s slow response drops the pod from the LB, so pods will bounce in and out under load. Suggestion: raise to failureThreshold: 3 (with periodSeconds: 5) so it takes sustained saturation, not a single blip, to pull a pod.

neuro_san/service/http/server/health_probe_server.py
d1donlydfink: See @nick-cognizant 's more erudite comments below.

d1donlydfink:
I agree with @nick-cognizant 's assessment that as-is, this PR isn't really reporting on health.
His comment about a "truly hung main loop" is especially astute.
The goal is not to report health as good at all costs, but to report on actual health.

My 2 cents:
* You don't need a separate thread, because you really want to measure the health of the server main loop
* ... and therefore you don't need a separate port

Instead:
* Do the version checks once when the server starts and cache the results (cuz it's real work to go looking in the pip database for library versions)
* Pass the cached result to any health check request that needs it - this way there's no real compute and/or swap to obscure (to the app) package databases ... and you won't need locks for this.
* Keep the health check requests as part of the regular single server instance. This will mean the health check itself is actually measuring what is supposed to.

Now, this will make the health check much more efficient and that is still very worthwhile, but it might not actually solve our problems with pods restarting.  That just means the root of the problems with pod restarting are elsewhere.

@nick-cognizant 's recent observations of the 8GB pods' memory being pretty saturated is a good sign that we're memory bound, which implies 2 directions to go at once:
1. We simply increase the memory on the pods in the deployment.  Easy - can help us meet immediate goals more quickly.
2. We dive into what makes this a memory-bound app at 8GB, which is a healthy thing to do at this point anyway.  You can begin to use the instrumentation brewing in this PR: https://github.com/cognizant-ai-lab/neuro-san/pull/1140   to plant object inspections at strategic times/locations within the code and see what's hogging memory. (See some theories in slack channel which implicate either AsyncCollatingQueue and/or Futures/AsyncioExecutor at this time) The place I started was in DataDrivenChatSession.close_of_work() - taking a look at the memory of the InvocationContext first.

d1donlydfink (changes requested):
See latest comments

andreidenissov-cog:
@nick-cognizant great comments - really helpful.
1. To clarify again the essence of the problem: our main service event loop is getting saturated pretty fast. @donn-leaf is observing problems starting from around 30 reqs in load test, I'm seeing higher numbers, but anyway. If we claim that our single instance of neuro-san is able to handle 200 concurrent requests (and it can) if we try to probe the main event loop during this run, the service will be neither live nor ready nor healthy. And pod will be killed. Death by health check.
2. That is why we do need a backdoor to the system - separate event loop to have a chance at getting the service status.
3.  Deployment configuration - failureThreshold, probe timeouts and such. Service itself knows nothing about it, so @nick-cognizant and @donn-leaf this is up to you - whatever you deem necessary.
4. Execution time of the probe itself is negligible, micro-optimizations or not. Problem is elsewhere - in event loop responsiveness.
5. So interesting question is what it means for our service to be live, ready and healthy.
6. I agree that maybe just running is not enough for service to be "live". But on the other hand, given our request latencies, detecting "truly hung" service is tricky. My suggestion to leave this improvement for next ticket/PR.
7. System "ready" - this we can probe on the main event loop, where probe timeouts will signal that service is too busy to accept new load. Again, tuning parameters is deployment config thing. @nick-cognizant what do you think?
8. System "healthy" - right now I'm reporting healthy is all requested service components have started and running. Smarter diagnosis? Of course, when it is figured out. We actually have an old ticket for that.

nick-cognizant:
> @nick-cognizant great comments - really helpful.
> 
> 1. To clarify again the essence of the problem: our main service event loop is getting saturated pretty fast. @donn-leaf is observing problems starting from around 30 reqs in load test, I'm seeing higher numbers, but anyway. If we claim that our single instance of neuro-san is able to handle 200 concurrent requests (and it can) if we try to probe the main event loop during this run, the service will be neither live nor ready nor healthy. And pod will be killed. Death by health check.
> 2. That is why we do need a backdoor to the system - separate event loop to have a chance at getting the service status.
> 3. Deployment configuration - failureThreshold, probe timeouts and such. Service itself knows nothing about it, so @nick-cognizant and @donn-leaf this is up to you - whatever you deem necessary.
> 4. Execution time of the probe itself is negligible, micro-optimizations or not. Problem is elsewhere - in event loop responsiveness.
> 5. So interesting question is what it means for our service to be live, ready and healthy.
> 6. I agree that maybe just running is not enough for service to be "live". But on the other hand, given our request latencies, detecting "truly hung" service is tricky. My suggestion to leave this improvement for next ticket/PR.
> 7. System "ready" - this we can probe on the main event loop, where probe timeouts will signal that service is too busy to accept new load. Again, tuning parameters is deployment config thing. @nick-cognizant what do you think?
> 8. System "healthy" - right now I'm reporting healthy is all requested service components have started and running. Smarter diagnosis? Of course, when it is figured out. We actually have an old ticket for that.

I agree with the split: readiness on the main loop, liveness heartbeat as a later ticket, healthy meaning components running.

One trap with readiness on the main loop: if the loop can't answer a probe at ~30 concurrent but one instance can really serve ~200, a tight timeout marks the pod not ready way before its real capacity. We'd drop pods from the LB at 30 and never reach 200, and probably flap them all out under a burst. So I'll set the timeout to fire on real overload, not brief busyness, knowing that makes readiness a rough signal.

That points to what's worth chasing next (not this PR): why does the loop choke at ~30? A healthy async loop should still answer a probe in milliseconds even with 200 requests in flight. If it can't, something CPU heavy or blocking is running on the loop and starving it, which is also why requests stall until the ALB times out. The separate probe loop is useful for visibility, but that blocking is the real ceiling.

neuro_san/service/http/server/http_server.py
d1donlydfink: Would it be worth having the regular server handle the health probes when the port == 0 or the same as the regular server port?  Seems like we could detect and act on that, and then we have something completely configurable.  That's actually kinda cool to have those options at our fingertips as we go forward trying to figure this thing out.

andreidenissov-cog: All probes are still handled on the "main" port as well, for fall-back option.
So it's a matter of configuration external to our service how probes are executed.

d1donlydfink (approved):

PR #1144, opened 2026-07-15
Add token cost for old OpenAI models

### LINKED ISSUE #1141: Add token price of older models in default llm info

Noravee:
This info used to come from `langchain-community`, but since the package has been deprecated, we put it directly in the default llm info. Note that newer models already have token costs.

Fix #1141


andreidenissov-cog (approved):
LGTM.

PR #1145, opened 2026-07-15
S3 Reservations: Robust s3 creds handling with new client for each invocation

d1donlydfink:
Factor our sync and asynchronous versions of a Aws[A]syncClientWorker, which handles the correct locking
and retries for getting AWS credentials from a single Session object.  These guys acknowledge that the credentials can change (as we now know some configurations are token based).

Refactor the Retriever, Writer and Expiration to all use the appropriate ClientWorker for their stuff.
Now, each time they are invoked via their main entry point, they get a new version of a client with correct credentials.
This will actually fix a latent undiscovered problem we had with the previous Reader code where each request would use the same client, meaning everyone would have to wait for each other.

Some other refactoring:
* S3RetryUtil -> S3Util
* Actual do_with_retry() methods move to their respective AwsClientWorker
* Have a single constant for the "reservations/" prefix
* Use DictionaryExtractors to better manage getting malformed data when doing deep dictionary dives. This should fix what @vince-leaf saw in the Expiration with "NoneType has no 'get()' attribute" messages

Tested santas workshop

### DIFF
--- a/neuro_san/service/watcher/temp_networks/aws_async_client_worker.py
+++ b/neuro_san/service/watcher/temp_networks/aws_async_client_worker.py
@@ -115,7 +115,7 @@ class AwsAsyncClientWorker:
         if last_err is not None:
             raise last_err
 
-        raise RuntimeError("S3 credential retries exhausted without capturing an error") from last_err
+        raise RuntimeError(f"{self.aws_service} credential retries exhausted without capturing an error") from last_err
 
     def ensure_async_lock_exists(self):
         """
--- a/neuro_san/service/watcher/temp_networks/aws_async_client_worker.py
+++ b/neuro_san/service/watcher/temp_networks/aws_async_client_worker.py
@@ -109,12 +109,7 @@ class AwsAsyncClientWorker:
                 self.frozen_credentials = None
                 if source is None:
                     source = self.name
-                self.logger.warning(
-                    "%s (%d): S3 credentials seem to have expired. Retrying. "
-                    "If you believe you have non-expiring S3 credentials, be sure they are correct.",
-                    source,
-                    attempt,
-                )
+                self.logger.warning("%s (%d): %s credentials seem to have expired. Retrying. If you believe you have non-expiring %s credentials, be sure they are correct.", source, attempt, self.aws_service, self.aws_service)
 
         # Exhausted retries
         if last_err is not None:
--- a/neuro_san/service/watcher/temp_networks/aws_async_client_worker.py
+++ b/neuro_san/service/watcher/temp_networks/aws_async_client_worker.py
@@ -109,7 +109,9 @@ class AwsAsyncClientWorker:
                 self.frozen_credentials = None
                 if source is None:
                     source = self.name
-                self.logger.warning("%s (%d): %s credentials seem to have expired. Retrying. If you believe you have non-expiring %s credentials, be sure they are correct.", source, attempt, self.aws_service, self.aws_service)
+                self.logger.warning("%s (%d): %s credentials seem to have expired. Retrying. "
+                                    "If you believe you have non-expiring %s credentials, be sure they are correct.",
+                                    source, attempt, self.aws_service, self.aws_service)
 
         # Exhausted retries
         if last_err is not None:
--- a/neuro_san/service/watcher/temp_networks/aws_async_client_worker.py
+++ b/neuro_san/service/watcher/temp_networks/aws_async_client_worker.py
@@ -44,24 +44,28 @@ from neuro_san.service.watcher.temp_networks.s3_retry_util import S3RetryUtil
 
 
 # pylint: disable=too-many-instance-attributes
-class S3AsyncClientWorker:
+class AwsAsyncClientWorker:
     """
-    AWS S3-based class that manages a particular async work_function (from functools.partial)
-    that has async_s3_client as an argument whose credentials are properly handled for
-    short-lived async S3 clients.
+    Class that manages a particular AWS boto async work_function (from functools.partial)
+    that has async_aws_client as an argument whose credentials are properly handled for
+    short-lived async boto clients.
     """
 
-    def __init__(self, name: str):
+    def __init__(self, name: str, aws_service: str = "s3"):
         """
-        Initialize S3 reservations storage.
+        Constructor
 
         :param name: Name for logging as to whose behalf this class is operating.
+        :param aws_service: Name of the AWS service to initialize the boto client
         """
         self.name: str = name
+        self.aws_service: str = aws_service
         self.logger: Logger = getLogger(self.__class__.__name__)
 
+        # Sync lock is only needed to properly create the async lock
+        # at the right time in the right thread.
         self.sync_lock: SyncLock = SyncLock()
-        self.async_s3_client_lock: AsyncLock = None
+        self.async_aws_client_lock: AsyncLock = None
 
         # Boto Machinations
         # We should be able to have a single Session for the lifetime of this object
@@ -123,11 +127,11 @@ class S3AsyncClientWorker:
         Be sure we have an asyncio Lock to get our sessions.
         """
         # Note that none of this is async in and of itself.
-        if self.async_s3_client_lock is None:
+        if self.async_aws_client_lock is None:
             with self.sync_lock:
                 # Be sure everyone has the same lock
-                if self.async_s3_client_lock is None:
-                    self.async_s3_client_lock = AsyncLock()
+                if self.async_aws_client_lock is None:
+                    self.async_aws_client_lock = AsyncLock()
 
     async def do_work_with_new_client(self, work_function: Awaitable, *, attempt: int = 1) -> Any:
         """
@@ -143,9 +147,9 @@ class S3AsyncClientWorker:
         retval: Any = None
 
         # Create an aiobotocore client for async operations.
-        async_s3_client: AioBaseClient = None
+        async_aws_client: AioBaseClient = None
 
-        async_s3_client_creator_context: ClientCreatorContext = None
+        async_aws_client_creator_context: ClientCreatorContext = None
         lock_released: bool = False
         acquired_lock: bool = False
 
@@ -155,14 +159,14 @@ class S3AsyncClientWorker:
         lock_released_time: float = 0.0
         try:
             # Serialize creation of the ClientCreatorContext with the lock to avoid credential-chain races.
-            await self.async_s3_client_lock.acquire()
+            await self.async_aws_client_lock.acquire()
             lock_aquired_time = perf_counter()
             acquired_lock = True
 
             # Get the current notion of frozen credentials.
             frozen_creds: ReadOnlyCredentials = await self.get_frozen_credentials()
-            async_s3_client_creator_context = self.session.create_client(
-                "s3",
+            async_aws_client_creator_context = self.session.create_client(
+                self.aws_service,
                 aws_access_key_id=frozen_creds.access_key,
                 aws_secret_access_key=frozen_creds.secret_key,
                 aws_session_token=frozen_creds.token,
@@ -173,21 +177,21 @@ class S3AsyncClientWorker:
             # but we want to be holding the lock while we create the client to avoid
             # credential-chain races like NoCredentialsError.
 
-            async with async_s3_client_creator_context as async_s3_client:
+            async with async_aws_client_creator_context as async_aws_client:
 
                 # Release the lock while we process, allowing other tasks to work on
-                # getting their own async_s3_client. (Not an async method)
-                self.async_s3_client_lock.release()
+                # getting their own async_aws_client. (Not an async method)
+                self.async_aws_client_lock.release()
                 lock_released_time = perf_counter()
                 lock_released = True
 
-                retval = await work_function(async_s3_client=async_s3_client)
+                retval = await work_function(async_aws_client=async_aws_client)
 
         finally:
             # Always release the lock if we successfully acquired it and have not already done so,
             # in case there was an error getting/entering the context manager.
             if acquired_lock and not lock_released:
-                self.async_s3_client_lock.release()
+                self.async_aws_client_lock.release()
 
         finish_time: float = perf_counter()
         self.logger.info("%s (%d): Lock acquisition in: %fs. Client context creation after: %fs. "
@@ -202,7 +206,7 @@ class S3AsyncClientWorker:
     async def get_frozen_credentials(self) -> ReadOnlyCredentials:
         """
         Get the frozen credentials for the current session.
-        We are assuming that we already have the async_s3_client_lock.
+        We are assuming that we already have the async_aws_client_lock.
         """
 
         # Use a local copy to avoid a race with the ClientError retry block
--- a/neuro_san/service/watcher/temp_networks/aws_async_client_worker.py
+++ b/neuro_san/service/watcher/temp_networks/aws_async_client_worker.py
@@ -40,7 +40,7 @@ from botocore.exceptions import ClientError
 from leaf_common.logging.sensitive_logger import SensitiveLogger
 from leaf_common.parsers.dictionary_extractor import DictionaryExtractor
 
-from neuro_san.service.watcher.temp_networks.s3_retry_util import S3RetryUtil
+from neuro_san.service.watcher.temp_networks.s3_util import S3Util
 
 
 # pylint: disable=too-many-instance-attributes
@@ -244,10 +244,10 @@ class AwsAsyncClientWorker:
             try:
                 return await fn()
             except ClientError as err:
-                if attempt >= max_attempts or not S3RetryUtil.is_retryable_client_error(err):
+                if attempt >= max_attempts or not S3Util.is_retryable_client_error(err):
                     raise
 
-                sleep = S3RetryUtil.exponential_backoff_with_jitter(base_sleep, attempt)
+                sleep = S3Util.exponential_backoff_with_jitter(base_sleep, attempt)
                 sensitive_logger.warning("%s: Retryable async ClientError (%s). attempt=%d", source, err, attempt)
                 await async_sleep(sleep)
                 attempt += 1
@@ -256,7 +256,7 @@ class AwsAsyncClientWorker:
                 if attempt >= max_attempts:
                     raise
 
-                sleep = S3RetryUtil.exponential_backoff_with_jitter(base_sleep, attempt)
+                sleep = S3Util.exponential_backoff_with_jitter(base_sleep, attempt)
                 sensitive_logger.warning("%s: Retryable async BotoCoreError (%s). attempt=%d", source, err, attempt)
                 await async_sleep(sleep)
                 attempt += 1
--- a/neuro_san/service/watcher/temp_networks/aws_sync_client_worker.py
+++ b/neuro_san/service/watcher/temp_networks/aws_sync_client_worker.py
@@ -99,12 +99,7 @@ class AwsSyncClientWorker:
                 self.frozen_credentials = None
                 if source is None:
                     source = self.name
-                self.logger.warning(
-                    "%s (%d): S3 credentials seem to have expired. Retrying. "
-                    "If you believe you have non-expiring S3 credentials, be sure they are correct.",
-                    source,
-                    attempt,
-                )
+                self.logger.warning("%s (%d): %s credentials seem to have expired. Retrying. If you believe you have non-expiring %s credentials, be sure they are correct.", source, attempt, self.aws_service, self.aws_service)
 
         # Exhausted retries
         if last_err is not None:
--- a/neuro_san/service/watcher/temp_networks/aws_sync_client_worker.py
+++ b/neuro_san/service/watcher/temp_networks/aws_sync_client_worker.py
@@ -99,7 +99,9 @@ class AwsSyncClientWorker:
                 self.frozen_credentials = None
                 if source is None:
                     source = self.name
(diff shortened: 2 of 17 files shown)


neuro_san/service/watcher/temp_networks/aws_async_client_worker.py
```diff
@@ -0,0 +1,269 @@
...
+from neuro_san.service.watcher.temp_networks.s3_util import S3Util
+
+
+# pylint: disable=too-many-instance-attributes
+class AwsAsyncClientWorker:
+    """
+    Class that manages a particular AWS boto async work_function (from functools.partial)
+    that has async_aws_client as an argument whose credentials are properly handled for
+    short-lived async boto clients.
+    """
```
d1donlydfink, line 52: Class that encapsulates correct credentials getting, credential failure retries, and locking around that for the asynchronous case.

neuro_san/service/watcher/temp_networks/aws_sync_client_worker.py
```diff
@@ -0,0 +1,246 @@
...
+from neuro_san.service.watcher.temp_networks.s3_util import S3Util
+
+
+# pylint: disable=too-many-instance-attributes
+class AwsSyncClientWorker:
+    """
+    Class that manages a particular AWS boto synchronous work_function (from functools.partial)
+    that has sync_aws_client as an argument whose credentials are properly handled for
+    short-lived synchronous boto clients.
+    """
```
d1donlydfink, line 48: Class that encapsulates correct credentials getting, credential failure retries, and locking around that for the synchronous case.

neuro_san/service/watcher/temp_networks/aws_async_client_worker.py
```diff
@@ -0,0 +1,269 @@
...
+        self.frozen_credentials: ReadOnlyCredentials = None
+
+    async def retry_with_new_client(self, work_function: Awaitable, *, source: str = None) -> Any:
+        """
+        Retries the async work_function when client credentials can expire.
+        :param work_function: The async work function to retry
+        :param *: extra keyword arguments for work_function
+        :param source: A string describing where the deployment was coming from
+        :return: What work_function returns
+        """
```
d1donlydfink, line 84: This kinda gets lifted from S3ReservatrionsWriter.
Callers can give a work_function as a partial() function to correctly wrap in the retries for credentials going bad.  Each call to work_function gets a new client.

neuro_san/service/watcher/temp_networks/aws_async_client_worker.py
```diff
@@ -0,0 +1,269 @@
...
+        # Exhausted retries
+        if last_err is not None:
+            raise last_err
+
+        raise RuntimeError("S3 credential retries exhausted without capturing an error") from last_err
+
+    def ensure_async_lock_exists(self):
+        """
+        Be sure we have an asyncio Lock to get our sessions.
+        """
```
d1donlydfink, line 125: This gets lifted from S3ReservatrionsWriter.

neuro_san/service/watcher/temp_networks/aws_async_client_worker.py
```diff
@@ -0,0 +1,269 @@
...
+    async def do_work_with_new_client(self, work_function: Awaitable, *, attempt: int = 1) -> Any:
+        """
+        This method separates the machinations of obtaining a proper S3 client
+        from add_all_reservations() which does all the actual work.
+
+        :param work_function: The async work function to retry
+        :param *: extra keyword arguments for work_function
+        :param attempt: Attempt number
+        :return: What work_function returns
+        """
```
d1donlydfink, line 142: This gets lifted from S3ReservatrionsWriter

neuro_san/service/watcher/temp_networks/aws_async_client_worker.py
```diff
@@ -0,0 +1,269 @@
...
+                         client_created_time - start_time,
+                         lock_released_time - start_time,
+                         finish_time - start_time)
+        return retval
+
+    async def get_frozen_credentials(self) -> ReadOnlyCredentials:
+        """
+        Get the frozen credentials for the current session.
+        We are assuming that we already have the async_aws_client_lock.
+        """
```
d1donlydfink, line 207: This gets lifted from S3ReservatrionsWriter

neuro_san/service/watcher/temp_networks/aws_async_client_worker.py
```diff
@@ -0,0 +1,269 @@
...
+        self.frozen_credentials = local_frozen
+
+        return local_frozen
+
+    @staticmethod
+    async def do_with_retries(source: str, fn, *, max_attempts: int = 8, base_sleep: float = 0.25):
+        """
+        Generic retry wrapper for boto3 calls.
+        boto3/botocore already retries, but this adds a bit of extra resilience and backoff for batch operations.
+        """
```
d1donlydfink, line 235: This gets lifted from S3RetryUtil, since this is the only place that needs it.

neuro_san/service/watcher/temp_networks/aws_sync_client_worker.py
```diff
@@ -0,0 +1,246 @@
...
+        self.frozen_credentials: ReadOnlyCredentials = None
+
+    def retry_with_new_client(self, work_function: Callable, *, source: str = None) -> Any:
+        """
+        Retries the work_function when client credentials can expire.
+        :param work_function: The work function to retry
+        :param *: extra keyword arguments for work_function
+        :param source: A string describing where the deployment was coming from
+        :return: What work_function returns
+        """
```
d1donlydfink, line 77: This gets lifted from async version and is translated to be a sync version.
Like its sibling class above, call with a partial() work_function to have it wrapped correctly for credential failure retries.

neuro_san/service/watcher/temp_networks/s3_reservations_expiration.py
```diff
@@ -41,10 +44,12 @@ class S3ReservationsExpiration:
     S3ReservationsStorage watcher loop.
     """
 
-    def __init__(self, name: str = "S3ReservationsExpiration", bucket_name: str = "", prefix: str = "reservations/"):
+    def __init__(self, name: str = "S3ReservationsExpiration", bucket_name: str = "",
+                 prefix: str = S3Util.DEFAULT_RESERVATIONS_PREFIX):
```
d1donlydfink, line 48: Use the constant for the default as a single source of truth.

neuro_san/service/watcher/temp_networks/s3_reservations_expiration.py
```diff
@@ -71,14 +76,26 @@ def expire_reservations(self):
...
+        expire_function = partial(self.expire_any_reservations)
+
+        client_worker: AwsSyncClientWorker = self.retriever.get_sync_client_worker()
+        client_worker.do_work_with_new_client(expire_function)
+
+    def expire_any_reservations(self, sync_aws_client: BaseClient = None):
+        """
+        Remove expired reservations from S3 storage.
+        :param sync_aws_client: S3 client
+        """
```
d1donlydfink, line 88: Same as guts before but now we get passed which sync_aws_client we are to use by the ClientWorker infra.  That guy gets passed around as needed.  The client lasts for the scope of this method.

neuro_san/service/watcher/temp_networks/s3_reservations_expiration.py
```diff
@@ -71,14 +76,26 @@ def expire_reservations(self):
...
         # Track how many reservations we expire for reporting
         expired_count: int = 0
         # Get current timestamp once for consistent expiration checking
         current_time: float = time()
 
-        for reservation_key in self.iter_reservation_keys():
+        for reservation_key in self.iter_reservation_keys(sync_aws_client):
+            if not reservation_key:
+                continue
```
d1donlydfink, line 96: Extra protection from None case from iterator.

neuro_san/service/watcher/temp_networks/s3_reservations_expiration.py
```diff
@@ -110,19 +129,20 @@ def iter_reservation_keys(self) -> Iterable[str]:
             if continuation_token:
                 kwargs["ContinuationToken"] = continuation_token
 
-            s3_client: BaseClient = self.retriever.get_client()
-            response = S3RetryUtil.do_with_retries(
-                self.name,
-                partial(s3_client.list_objects_v2, **kwargs)
-            )
+            list_objects_function = partial(sync_aws_client.list_objects_v2, **kwargs)
+            response = client_worker.do_with_retries(self.name, list_objects_function)
```
d1donlydfink, line 133: Use the ClientWorker to do the retries with the client given.

neuro_san/service/watcher/temp_networks/s3_reservations_expiration.py
```diff
@@ -110,19 +129,20 @@ def iter_reservation_keys(self) -> Iterable[str]:
...
+            if response is None:
+                response = {}
             for obj in response.get("Contents", []):
-                yield obj["Key"]
+                yield obj.get("Key")
```
d1donlydfink, line 137: Extra None protections to aid in fixing the issue Vince saw with what was probably malformed content.

neuro_san/service/watcher/temp_networks/s3_reservations_expiration.py
```diff
@@ -132,29 +152,36 @@ def expire_one_reservation(self, obj_key: str, current_time: float, source: str
...
         try:
             # Retrieve the reservation object from S3
             agent_spec: Dict[str, Any] = self.retriever.retrieve_object_with_retries(obj_key)
-            metadata: Dict[str, Any] = agent_spec.get("metadata")
-            reservation_data: Dict[str, Any] = metadata.get("reservation")
+            if agent_spec is None:
+                agent_spec = empty
+
+            extractor = DictionaryExtractor(agent_spec)
+            reservation_data: Dict[str, Any] = extractor.get("metadata.reservation", empty)
```
d1donlydfink, line 169: More None protections using a DictionaryExtractor to get to the meat.

neuro_san/service/watcher/temp_networks/s3_reservations_reader.py
```diff
@@ -71,16 +76,32 @@ def get_one_reservation(self, reservation_id: str) -> Tuple[Reservation, AgentNe
         """
         reservation: Reservation = None
         agent_network: AgentNetwork = None
+
         # Construct the S3 object key for this reservation ID
-        s3_obj_key: str = S3RetryUtil.get_obj_key_for_reservation(self.retriever.get_prefix(), reservation_id)
+        s3_obj_key: str = S3Util.get_obj_key_for_reservation(self.retriever.get_prefix(), reservation_id)
```
d1donlydfink, line 81: This get_obj_key_for_reservation() method moves to renamed S3Util.
(Better name for shell auto-complete)

neuro_san/service/watcher/temp_networks/s3_reservations_writer.py
```diff
@@ -82,18 +74,9 @@ def __init__(self, name: str = "S3ReservationsWriter", bucket_name: str = "", pr
...
-
         self.converter = ReservationDictionaryConverter()
 
-        # Boto Machinations
-        # We should be able to have a single Session for the lifetime of this object
-        self.session: AioSession = None
-
-        # Cached frozen credentials which can be invalidated should they expire
-        self.frozen_credentials: ReadOnlyCredentials = None
+        self.s3_async_client_worker = AwsAsyncClientWorker(self.name, "s3")
```
d1donlydfink, line 79: All the above members move to the AwsAsyncClientWorker.

neuro_san/service/watcher/temp_networks/s3_reservations_writer.py
```diff
@@ -140,183 +123,32 @@ async def add_reservations(self, reservations_dict: Dict[Reservation, Any],
...
-
-    async def add_reservations_with_new_client_credential_retries(self,
-                                                                  reservations_dict: Dict[Reservation, Any],
-                                                                  source: str):
-        """
-        Retries adding the reservations when client credentials can expire.
-
-        :param reservations_dict: A mapping of Reservation -> some deployable agent spec
-        :param source: A string describing where the deployment was coming from
-        """
```
d1donlydfink, line 156: This and all the rest moves to AwsAsyncClientWorker.

neuro_san/service/watcher/temp_networks/s3_reservations_writer.py
```diff
@@ -140,183 +123,32 @@ async def add_reservations(self, reservations_dict: Dict[Reservation, Any],
...
-        return local_frozen
+        work_function: Awaitable = partial(self.add_all_reservations, reservations_dict, source)
+        await self.s3_async_client_worker.retry_with_new_client(work_function)
 
-    async def add_all_reservations(self, async_s3_client: AioBaseClient,
+    async def add_all_reservations(self,
                                    reservations_dict: Dict[Reservation, Dict[str, Any]],
-                                   source: str):
+                                   source: str,
+                                   async_aws_client: AioBaseClient = None):
```
d1donlydfink, line 132: New convention has the client being passed in as a keyword arg.

neuro_san/service/watcher/temp_networks/s3_retry_util.py
```diff
@@ -1,163 +0,0 @@
...
-from botocore.exceptions import BotoCoreError
-from botocore.exceptions import ClientError
-
-from leaf_common.logging.sensitive_logger import SensitiveLogger
-
-
-class S3RetryUtil:
-    """
-    Utilities for retrying AWS S3 operations.
-    """
```
d1donlydfink, line 35: What's left from this file has moved to S3Util.

tests/neuro_san/service/watcher/temp_networks/fake_s3_client.py
```diff
@@ -79,3 +79,9 @@ def get_object(self, Bucket: str, Key: str):
                 "GetObject",
             )
         return {"Body": io.BytesIO(self.objects[Key])}
+
+    def close(self):
+        """
+        Close the in-memory bucket.
+        """
+        # Do nothing
```
d1donlydfink, line 87: Need this now because even the sync version closes the client before moving on.
This allows the tests to pass in the new regime.

tests/neuro_san/service/watcher/temp_networks/s3_reservations_storage_test_base.py
```diff
@@ -78,20 +79,33 @@ def setUp(self):
         # Patch boto3_client at the import boundary in s3_reservations_storage
         # so storage.start() receives our fake instead of a real boto3 client.
         boto3_patcher = patch(
-            "neuro_san.service.watcher.temp_networks.s3_reservations_retriever.boto3_client",
+            "neuro_san.service.watcher.temp_networks.aws_sync_client_worker.Session.create_client",
```
d1donlydfink, line 82: Patches have to move references.

tests/neuro_san/service/watcher/temp_networks/s3_reservations_storage_test_base.py
```diff
@@ -78,20 +79,33 @@ def setUp(self):
...
+        def _fake_get_credentials(*_args, **_kwargs):
+            return Credentials("bogus", "bogus")
+
+        credentials_patcher = patch(
+            "neuro_san.service.watcher.temp_networks.aws_sync_client_worker.Session.get_credentials",
+            new=_fake_get_credentials,
+        )
+        credentials_patcher.start()
+        # Restores the real AioSession symbol after the test completes, regardless of pass/fail.
+        self.addCleanup(credentials_patcher.stop)
```
d1donlydfink, line 101: Now we have to patch for synchronous get_credentials() as well.

neuro_san/service/watcher/temp_networks/s3_reservations_expiration.py
```diff
@@ -132,29 +152,36 @@ def expire_one_reservation(self, obj_key: str, current_time: float, source: str
...
+        empty: Dict[str, Any] = {}
         expired: bool = False
         try:
             # Retrieve the reservation object from S3
-            agent_spec: Dict[str, Any] = self.retriever.retrieve_object_with_retries(obj_key)
-            metadata: Dict[str, Any] = agent_spec.get("metadata")
-            reservation_data: Dict[str, Any] = metadata.get("reservation")
+agent_spec: Dict[str, Any] = self.retriever.retrieve_object_with_retries(obj_key=obj_key, source=source, sync_aws_client=sync_aws_client)
+            if agent_spec is None:
+                agent_spec = empty
```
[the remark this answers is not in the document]
d1donlydfink: You did that ya silly goose.

neuro_san/service/watcher/temp_networks/s3_reservations_retriever.py
```diff
@@ -113,18 +117,23 @@ def start(self):
...
-        get_function = partial(self.client.get_object, Bucket=self.bucket_name, Key=obj_key)
-        if source is None:
+        if obj_key is None:
+            raise ValueError(f"{self.name}: S3 object key must be provided")
+        if sync_aws_client is None:
+            raise ValueError(f"{self.name}: S3 client must be provided")
+        get_function = partial(sync_aws_client.get_object, Bucket=self.bucket_name, Key=obj_key)
             source = self.name
-        obj_response: Dict[str, Any] = S3RetryUtil.do_with_retries(source, get_function)
+        obj_response: Dict[str, Any] = self.s3_sync_client_worker.do_with_retries(source, get_function)
```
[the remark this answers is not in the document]
d1donlydfink: You did that ya silly goose.

neuro_san/service/watcher/temp_networks/s3_reservations_expiration.py
```diff
@@ -71,14 +76,26 @@ def expire_reservations(self):
...
         current_time: float = time()
 
-        for reservation_key in self.iter_reservation_keys():
+        for reservation_key in self.iter_reservation_keys(sync_aws_client):
+            if not reservation_key:
+                continue
             # Attempt to expire this reservation and increment counter if successful
-            if self.expire_one_reservation(reservation_key, current_time):
+            if self.expire_one_reservation(reservation_key, current_time, sync_aws_client):
                 expired_count += 1
```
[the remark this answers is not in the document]
d1donlydfink: Addressed.

neuro_san/service/watcher/temp_networks/s3_reservations_expiration.py
```diff
@@ -71,14 +76,26 @@ def expire_reservations(self):
         """
         self.logger.debug("%s: Starting expiration process for S3 reservations", self.name)
 
+        expire_function = partial(self.expire_any_reservations)
+
+        client_worker: AwsSyncClientWorker = self.retriever.get_sync_client_worker()
+        client_worker.retry_with_new_client(expire_function, source=self.name)
```
d1donlydfink, line 82: Use the AwsSyncClientWorker to correctly wrap the expire_any_reservations() method below with credential-oriented retries.

This could perhaps be done more elegantly maybe with a ContextManager and/or a better Futures interface.  Can iterate on that - at least this approach solves immediate problems and at least in the same way for sync and async

neuro_san/service/watcher/temp_networks/s3_reservations_reader.py
```diff
@@ -71,16 +76,31 @@ def get_one_reservation(self, reservation_id: str) -> Tuple[Reservation, AgentNe
...
+                agent_spec = empty
 
             # Reconstruct the Reservation object from stored dictionary
-            reservation_dict: Dict[str, Any] = metadata.get("reservation")
+            extractor = DictionaryExtractor(agent_spec)
+            reservation_dict: Dict[str, Any] = extractor.get("metadata.reservation", empty)
+
             reservation = self.converter.from_dict(reservation_dict)
+            if not reservation:
+                self.logger.error("%s: Failed to parse reservation payload for %s", self.name, reservation_id)
```
d1donlydfink, line 101: Log when we get malformed content on read (for request)

Noravee:
Post-merge follow-up: a deeper review of this refactor surfaced four findings, now tracked in two issues (split by the kind of change needed).

**Behavioral fixes — #1152, fix up in #1154:**
1. *Malformed-object handling*: the DictionaryExtractor defaults make an unparseable object read as "expired at epoch 0" (`current_time > 0` is always true), so the sweep silently **deletes** any object it can't parse — and the stored-null shape (`{"metadata": {"reservation": null}}`) still crashes with the same `'NoneType' object has no attribute 'get'` error this PR set out to fix, because the extractor returns a present-but-null value in preference to its default. #1154 centralizes the shape check in one helper and switches the sweep to skip-and-WARN (policy discussion welcome there).
2. *Mid-sweep credential expiry*: `expire_one_reservation()`'s broad `except ClientError` also swallows `ExpiredToken` from get/delete, so `retry_with_new_client()` never sees it — the sweep makes one doomed call per remaining key, expires nothing, and reports success. #1154 re-raises `ExpiredToken` (only) so the credential-refresh retry can actually fire.

**Design follow-up — #1153:**
3. *Frozen credentials*: passing explicit key/secret/token to `create_client()` makes botocore build a static `Credentials` object, which disables its native signing-time auto-refresh (verified against the installed botocore/aiobotocore). Recovery becomes reactive — one real failed `ExpiredToken` round trip per expiry cycle — and it's what makes the mid-sweep scenario above possible at all.
4. *Per-call clients on the read path*: every `get_one_reservation()` now builds a client under the worker lock and `close()`s it, discarding the connection pool — so each S3 GET pays a fresh TCP+TLS handshake. One note on the description's premise: botocore clients are thread-safe and serve concurrent requests through their pool, so the previous shared client didn't serialize requests. #1153 proposes one long-lived keyless client per worker, which addresses both points at once and comes with two ready acceptance tests.

None of this diminishes the refactor's direction — pulling the credential machinery into reusable workers is what made these fixes tractable to make in one place.

PR #1148, opened 2026-07-15
Optimize restore

### LINKED ISSUE #1147: Slow dev loop: saving one HOCON re-parses every served network
## Summary

With the manifest watcher enabled (default for `ns run` / `neuro-san-studio run`), saving **any single** `.hocon`/`.json` file re-restores **every** served network — all are re-parsed from scratch, not just the edited one. Since HOCON parsing dominates a restore, the wait after each save scales with the total number of served networks. On lower-powered machines this is **2+ minutes per save**.

## Reproduce

```bash
# this repo (33 networks)
python scripts/evidence_repeated_parsing.py

# neuro-san-studio checkout (85 networks)
STUDIO=/path/to/neuro-san-studio
python scripts/evidence_repeated_parsing.py --manifest "$STUDIO/registries/manifest.hocon" --chdir "$STUDIO"
```

Two consecutive restores of the studio manifest (mimics a watcher reload):

```
1st restore: 86 files parsed
2nd restore: 86 files parsed   <- all re-parsed again, nothing reused
```

Parse cost per restore: ~1.6s (33 networks, this repo) / ~12s (85 networks, studio) — dev-machine figures; slower elsewhere.

## Root cause

`RegistryStorageUpdater.update_storage()` (`neuro_san/service/watcher/registries/registry_storage_updater.py`) calls a full `RegistryManifestRestorer(...).restore()` on **any** change reported by the recursive tree observer (`polling_registry_observer.py`). Nothing reuses previously-parsed unchanged files between restores.

Scope: the watcher does **not** reparse on an idle server (it early-returns when nothing changed), and cold `ns run` startup is unaffected. The waste is per-**save**.

## Possible directions

1. Opt-in, include-aware parse cache so a re-restore only re-parses changed files (proposed in the accompanying PR; ~21× faster/save on studio once warm).
2. Incremental restore — re-parse only changed file(s) and splice into the in-memory set.
3. ~~Parallelize parsing~~ — investigated, a wash (CPU-bound, GIL-limited); noted so it isn't re-attempted.

No correctness bug — purely wasted repeated work.

kaushik-cognizant:
# Motivation

At present in the studio whenever a hocon is changed, the server reloads each enabled hocon present in the repository. This is not a problem for people with beefy machines or for people with only a couple of hocons, but on associate machines with not enough juice, it creates a very negative developer experience. For example, waiting 2 mins or more for the after saving a hocon. 
There could be many ways to fix this. This is just a proposal to start the discussion. 

I have added a cache which is enabled by `NEURO_SAN_PARSE_CACHE="1"`. This change is only for single user environments like `neuro-san-studio`. So far the tests have passed locally but this can have bigger ramifications that I might have overlooked.

## Description

Adds an **opt-in, in-process cache for parsed HOCON/JSON config files** to speed up the local development loop, where the same agent-network files are parsed over and over.

**The problem.** When a neuro-san server is running with the manifest watcher enabled (the default for `ns run` / `neuro-san-studio run`), `RegistryStorageUpdater.update_storage()` calls `RegistryManifestRestorer(...).restore()` — a **full re-restore of every served network** — whenever *any* single `.hocon`/`.json` file under the registry tree changes (the observer watches the tree recursively). HOCON parsing is the dominant cost of a restore, so in a dev loop, **saving one file re-parses all N served networks from scratch**, even though only the one file you touched actually changed.

To be precise about what this does and does not address:
- The watcher does **not** reparse on an idle server — it early-returns when no files changed. The waste is per-*save*, not per-*tick*.
- This cache does **not** speed up cold `ns run` startup (a fresh process starts with an empty cache). Its entire benefit is the **warm dev loop**: repeated saves within one long-lived server process.

**The proposal.** A class-level cache on `AbstractAsyncConfigRestorer.deserialize_file_contents()`:
- **Opt-in via `NEURO_SAN_PARSE_CACHE`** (values `1`/`true`/`yes`/`on`). Off by default, so production/deployed servers are completely unaffected — when disabled, neither the cache lookup nor the extra hashing runs.
- **Include-aware cache key.** The key folds in the content hash of every file a HOCON file transitively `include`s, so editing a shared include (e.g. `aaosa.hocon`, `llm_config.hocon`) correctly invalidates every dependent file's cache entry.
- **Returns/stores deep copies.** Callers (e.g. the filter chains) are not guaranteed to treat the parsed dict as immutable; deep-copying on both store and hit ensures an in-place mutation by one caller can never corrupt what a later cache hit sees.
- **Bounded with LRU eviction** (`_CACHE_MAX_ENTRIES = 256`). Each edit to a watched file produces a new cache key, so a long dev session needs eviction to avoid unbounded growth.

Fixes #1147 

## Impact

- **Local dev:** with `NEURO_SAN_PARSE_CACHE=1`, a watcher-triggered reload re-parses only the file(s) that actually changed instead of all N served networks (see evidence below). No source edits to agent networks required.
- **Production / default:** no behavioral or performance change — the cache is inert unless the env var is set.
- **Correctness:** guarded against the two failure modes a naive cache would introduce — stale entries when a shared include changes (handled by the include-aware hash) and shared-mutable-object corruption (handled by deep copy).

## Screenshots / Logs / Examples (optional)

Evidence collected with the throwaway scripts below (see note about removal). All numbers are from one dev machine — the absolute times scale with the machine (the associate machines this targets are much slower, hence the "2 min" in Motivation), so the **speedup ratio** is the portable figure, not the seconds.

### Measured speedups (both repos)

Measured against this repo's own manifest (`neuro_san/registries/manifest.hocon`, 33 served networks) and a `neuro-san-studio` checkout (`registries/manifest.hocon`, 85 served networks).

**Per-save cost — end-to-end `restore()` wall-clock** (`scripts/test_cache.py`). This is what a developer actually waits for on each file save (the watcher-triggered reload). "Without cache" is today's behavior:

| Repo | Served networks | Without cache (per save) | With cache (warm) | Speedup |
|---|---|---|---|---|
| `neuro-san` | 33 | ~0.85s | ~0.03s | **~28×** |
| `neuro-san-studio` | 85 | ~7.1s | ~0.34s | **~21×** |

**Parse cost in isolation — `deserialize_file_contents` cold vs warm** (`scripts/profile_manifest_restorer.py`, cProfile cumulative time). This isolates the HOCON parsing the cache actually eliminates (the wall-clock above also includes validation/filtering, which the cache does not touch, so its ratio is lower):

| Repo | Cold (1st restore) | Warm (2nd restore) | Speedup |
|---|---|---|---|
| `neuro-san` | 1.62s | 0.003s | **~540×** |
| `neuro-san-studio` | 12.35s | 0.20s | **~61×** |

> **Honest caveat:** the *first* restore with the cache enabled is slightly **slower** than without it (studio: ~8.2s vs ~7.1s) — it pays the include-aware hashing plus a deep-copy-on-store before any hit exists. The benefit is entirely on the *subsequent* (warm) restores, i.e. every save after the first. This is consistent with "helps the warm dev loop, not cold startup." The studio warm floor (~0.2s) is higher than neuro-san's (~0.003s) because studio's networks pull in shared includes (`aaosa.hocon`, etc.) that still get hashed on every restore, whereas most neuro-san sample networks have no includes.

**`scripts/evidence_repeated_parsing.py`** — instruments the actual parse call (`HoconSerializationFormat.to_object`, which a cache hit never reaches) and counts real parses vs cache hits per restore.

Part A — real studio manifest (85 served networks), two consecutive restores, mimicking a watcher-triggered reload:

```
[CACHE OFF]  1st restore: 86 parsed,  0 cached
             2nd restore: 86 parsed,  0 cached   <- every file re-parsed again = wasted work
[CACHE ON ]  1st restore: 86 parsed,  0 cached
             2nd restore:  0 parsed, 86 cached   <- repeat restore absorbed entirely
```

The same run against this repo's own manifest (33 served networks) shows the identical pattern: cache off re-parses all 34 files (33 networks + manifest) on every restore; cache on parses 34 once, then 0 on the repeat.

Part B — controlled 3-network registry, cache on, edit exactly one file, re-restore:

```
Initial restore: 4 parsed, 0 cached
Edited one file: net1.hocon
Re-restore:      1 parsed, 3 cached   -> only re-parsed ['net1.hocon'], not all 3
```

### Reproduce

Against **this repo** (uses its own `neuro_san/registries/manifest.hocon`, no extra setup):

```bash
# from the neuro-san repo root
python scripts/test_cache.py                       # wall-clock: cache off vs on
NEURO_SAN_PARSE_CACHE=1 python scripts/profile_manifest_restorer.py   # cold vs warm parse breakdown
python scripts/evidence_repeated_parsing.py        # per-restore parse vs cache-hit counts (Part A + B)
```

Against a **neuro-san-studio** checkout (point at its manifest; `--chdir` is required because studio's manifest uses project-root-relative `include`s). Replace `/path/to/neuro-san-studio` with your checkout:

```bash
STUDIO=/path/to/neuro-san-studio
python scripts/test_cache.py --manifest "$STUDIO/registries/manifest.hocon" --chdir "$STUDIO"
NEURO_SAN_PARSE_CACHE=1 python scripts/profile_manifest_restorer.py --manifest "$STUDIO/registries/manifest.hocon" --chdir "$STUDIO"
python scripts/evidence_repeated_parsing.py --manifest "$STUDIO/registries/manifest.hocon" --chdir "$STUDIO"
```

Notes:
- `test_cache.py` prints `Without cache - <1st restore> to <2nd restore>` then `With cache - <1st> to <2nd>`; the dev-loop speedup is *without-cache 2nd restore* ÷ *with-cache 2nd restore*.
- `profile_manifest_restorer.py` reports the cache breakdown only meaningfully with `NEURO_SAN_PARSE_CACHE=1`; without it, both restores re-parse and it says so.
- `evidence_repeated_parsing.py` needs no env var — it toggles the cache internally for Part A and always runs Part B.
- Run the unit tests with: `python -m pytest tests/neuro_san/internals/persistence/test_abstract_async_config_restorer.py`

**Scripts added (throwaway — safe to delete before/after merge):** each defaults to this repo's own manifest (runs bare, no setup) and accepts `--manifest PATH` / `--chdir DIR` to point at a `neuro-san-studio` checkout (the `--chdir` is needed there because studio's manifest uses project-root-relative `include`s).
- `scripts/evidence_repeated_parsing.py` — proves the repeated-parsing behavior and the cache's effect (above).
- `scripts/profile_manifest_restorer.py` — cProfile of a cold vs warm restore, filtered to the cache/restore call chain, with a cache-specific cold→warm breakdown.
- `scripts/test_cache.py` — quick wall-clock timing of restore with the cache off vs on.

> **Note:** the three files under `scripts/` were used only to investigate and demonstrate this change. They are not part of the shipped functionality and can be removed at any point.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Dependency upgrade
- [x] Refactoring / Improvement
- [ ] Documentation
- [ ] New agent / tool

## Checklist

- [x] Tested locally
- [x] Added/updated tests
- [ ] Updated relevant documentation
- [ ] Added dependencies to appropriate `requirements.txt` (tool-specific preferred; main only for core functionality)

---

**By submitting this pull request, I confirm that my contribution is made under the terms of the project's [Apache 2.0 License](../LICENSE.txt).**


neuro_san/internals/persistence/abstract_async_config_restorer.py
```diff
@@ -45,6 +72,23 @@ class AbstractAsyncConfigRestorer(Restorer, ConfigFilter):
...
+
+    @staticmethod
+    def is_cache_enabled() -> bool:
+        """
+        :return: True if the deserialization cache is opted in via the
+                _CACHE_ENV_VAR environment variable, False otherwise.
+                Evaluated per call so the setting can be toggled at runtime
+                and so production processes (which never set it) pay nothing.
+        """
+        return environ.get(_CACHE_ENV_VAR, "").strip().lower() in _CACHE_TRUTHY_VALUES
```
d1donlydfink, line 90: Part of the problem with this caching mechanism is that this class is used for many kinds of configuration files and so you don't really have control over just _what_ is being cached.

If we keep a caching mechanism (maybe, maybe not) it'd be best to have the caching mechanism separate-from and/or pluggable-into this individual restore class so the caching is optional for different kinds of files.    Caching takes up a bunch of extra memory. Per recent stats, any hocon files are ~100k when fully unfurled.

neuro_san/internals/persistence/abstract_async_config_restorer.py
```diff
@@ -194,8 +262,109 @@ def deserialize_file_contents(self, file_path: str, file_contents: bytes) -> Dic
             )
             raise ValueError(message) from exception
 
+        # Cache the parsed config for future use (only when opted in).
+        # Store a deepcopy so this caller (who gets the original `config` back below)
+        # can't mutate the cached entry out from under future cache hits, and evict the
+        # least-recently-used entry once we're over the cap so a long dev session with
+        # many edit/reload cycles doesn't grow this dict without bound.
+        if cache_enabled:
+            with self._cache_lock:
```
d1donlydfink, line 271: Given the caching is monolithic at this point, introducing this lock will kill performance when different kinds of files are read in.  For instance: I believe AND requests ultimately use this class to read in its own config files.

scripts/evidence_repeated_parsing.py
```diff
@@ -0,0 +1,246 @@
...
+          Shows cache-off pays the full parse cost on BOTH restores, while
+          cache-on pays it once and then parses 0 on the repeat.
+  Part B: a controlled 3-network temp registry, cache ON. Restore, then change
+          exactly ONE file's contents and restore again. Shows only the changed
+          file re-parses (the delta), not all N - the include-aware content hash
+          in action.
+
+Run:
+  python scripts/evidence_repeated_parsing.py
+"""
```
d1donlydfink, line 30: Anything in this folder should probably live under top-level tests directory.
These are not things that everyone is going to run all the time and so do not really deserver top-level status.

scripts/evidence_repeated_parsing.py
```diff
@@ -0,0 +1,246 @@
...
+        """Number of files that were actually (re-)parsed this restore."""
+        return sum(1 for _, kind in self.calls if kind == "miss")
+
+    @property
+    def hits(self):
+        """Number of files served from the cache this restore."""
+        return sum(1 for _, kind in self.calls if kind == "hit")
+
+
+def timed_restore(manifest_file):
```
d1donlydfink, line 90: OK this next one falls under the banner of "Yes you can do X in Python, but maybe you really shouldn't".

If you haven't noticed: _everything_ within neuro-san starts with a class, unlike these free-standing methods.  

Why? 
1) Enforcing this significantly reduces the hacker-mentality of using global variables to skirt around data plumbing issues.
2) We surprise ourselves as to what we end up reusing in part or in whole. Starting out with methods does not lead to OO-enabled design, even in CLI-related stuff.
3. Pylint enforces creation of class comments which make you think and explain about why you are grouping all this stuff together.  Later additions become less rando.

scripts/evidence_repeated_parsing.py
```diff
@@ -0,0 +1,246 @@
...
+
+
+def resolve_manifest(manifest_arg):
+    """:return: absolute manifest path - the arg if given, else the packaged default."""
+    if manifest_arg:
+        return os.path.abspath(manifest_arg)
+    return REGISTRIES_DIR.get_file_in_basis("manifest.hocon")
+
+
+def main():
```
d1donlydfink, line 237: Take a look at agent_cli for how we can begin to make CLI tools OO

scripts/profile_manifest_restorer.py
```diff
@@ -0,0 +1,236 @@
...
+    first_rows, _ = profile_one_restore(manifest_file)
+    second_rows, second_networks = profile_one_restore(manifest_file)
+
+    print_comparison_table(first_rows, second_rows)
+    print_cache_breakdown(first_rows, second_rows)
+
+    print("\n📊 SUMMARY (2nd call):")
+    print(f"   Total Storage Classes: {len(second_networks)}")
+    for storage_class, networks in second_networks.items():
+        print(f"   - {storage_class}: {len(networks)} networks")
```
d1donlydfink, line 232: More of the same from previous file.

scripts/test_cache.py
```diff
@@ -0,0 +1,76 @@
...
+    start_time = time.time()
+    restorer = RegistryManifestRestorer(manifest_file)
+    agent_networks: Dict[str, Dict[str, AgentNetwork]] = restorer.restore()
+    manifest_files: List[str] = restorer.get_manifest_files()
+    # Reference the results so they are not flagged as unused.
+    _ = (agent_networks, manifest_files)
+    return time.time() - start_time
+
+
+def parse_args():
```
d1donlydfink, line 35: If this really is a test, why is it a CLI tool to only do once/manually?
why isn't it a unit test derving from TestCase?

tests/neuro_san/internals/persistence/test_abstract_async_config_restorer.py
```diff
@@ -429,3 +429,82 @@ def test_async_restore_raises_value_error_on_malformed_hocon(self, tmp_path: Pat
         r: ConcreteRestorer = self.make_restorer()
         with pytest.raises(ValueError):
             self.run(r.async_restore(path))
+
+    # -----------------------------------------------------------------------
+    # deserialize_file_contents caching (NEURO_SAN_PARSE_CACHE)
+    # -----------------------------------------------------------------------
```
d1donlydfink, line 435: If the caching aspect were in a different class, we would be testing that in a separate clearly labeled test-case file as well. No text delineation needed, because it'd already be clear.

d1donlydfink (changes requested):
OK, first a summary critique of these changes as they are:

* Separate caching mechanism from config-restore mechanism.
* Limit scope of what is cached - locking at too broad a scope will inadvertently kill performance in other placess
* New files don't belong in scripts - they belong under test
* Use classes, not tons of free-standing methods. Not OO!
* Take a look at agent_cli for an example as to how to make a CLI tool OO
* test_cache should probably be a unit test, not a never-enforced manual CLI tool
* Separate unit test file for testing separated caching mechanism.
* Oh yeah - always make your tests pass before handing it off for review :)  Without that first , the review will just be invalidated with your next change and the reviewer has to look again later.  "Always be your own first reviewer"

---

OK, now, my higher-level reasons as to why this may not be the right approach (to the very-right problem!) at this time:

The approach taken with this PR as-is begins to solve for the re-read case, which effects neuro-san developers. It is great and good to attack this problem. However there is an aspect it does not cover which also needs addressing which is initial server start-up time.   

Here we trade memory to optimize the secondary load time, but maybe what we really should be doing is to also optimize the initial load time. Perhaps a better approach that would hit these two birds with a single stone is an effort to parallelize the reading of the manifest file.  Then startup wouldn't take 10s of seconds either, it would lift all boats and not cost any residual memory, placing the overall cost of processing on only _when_ the load/re-loads happen and not on the entire system.  Making this better could also reduce or at least make constant some cluster-level config that waits for a particular server pod to be ready to accept requests.

Now, that said there are perhaps 2 avenues for parallelization.
1) Read hocons themselves in parallel (top-level)
2) Parallelize validation of the hocon file (lower level)

I believe (without evidence) that (2) is where the real time is being taken because it is multiple passes across all the nodes of a single hocon file. We can't really control the number and interconnectedness of the network nodes - we want that to be supremely flexible.  Attacking (2) would also help the performance of AND, since it goes through multiple validation steps.

In other words:  O(# files x # validation steps x # nodes in network)

I would encourage you down this other path perhaps selfishly because in my own work I stop/restart the server all the time and it's annoying how long it takes to be able to send the first request. Help me in my quest to not be annoyed! :)

d1donlydfink:
See https://github.com/cognizant-ai-lab/neuro-san/pull/1177

PR #1154, opened 2026-07-21
Fix S3 expiration: malformed-object policy and ExpiredToken recovery

### LINKED ISSUE #1152: S3 reservations expiration: malformed-object handling deletes/crashes; mid-sweep ExpiredToken swallowed (follow-up to #1145)
## Context

Follow-up to #1145 (S3 creds handling refactor). A review of the merged changes found two behavioral issues in the expiration sweep plus a few smaller ones. All are reproduced by tests (currently red on `main`), and a candidate fix exists.

## 1. Unparseable objects under the `reservations/` prefix are silently deleted

`expire_one_reservation()` defaults a missing `metadata.reservation` to `{}` and a missing `expiration_time_in_seconds` to `0`, so `current_time > 0` is always true — any valid-JSON object that isn't reservation-shaped gets `delete_object`'d, logged only at DEBUG as reservation `<unknown>`.

Why it matters:
- "We couldn't parse it" and "it has expired" are different facts. The most realistic unparseable object isn't garbage — it's **schema drift during a rolling deploy** (an old server's sweep meeting objects written by a newer writer), and those get deleted while live.
- It also destroys the only evidence of whatever wrote the malformed object — the very thing the original production report ("NoneType has no 'get()' attribute" messages) could have been diagnosed from.

(Before #1145 the same input raised instead — object survived but the sweep died; see next point. Neither shipped behavior is the one we want.)

## 2. The stored-null variant still crashes the sweep (the original NoneType error)

`DictionaryExtractor.get()` only applies its default when a key is *missing*; a key present with a stored JSON `null` is returned as `None`. So an object with body `{"metadata": {"reservation": null}}` still raises `AttributeError: 'NoneType' object has no attribute 'get'` — the exact error #1145 set out to fix. One such object aborts the sweep at the same key every cycle, so nothing sorted after it ever expires until the object is removed by hand.

## 3. Mid-sweep `ExpiredToken` is swallowed, so the sweep silently no-ops

`expire_one_reservation()`'s broad `except ClientError` logs-and-continues for every non-`NoSuchKey` code — including `ExpiredToken` from `get_object`/`delete_object`. Because the sweep's client is built from frozen credentials (no auto-refresh; see the companion client-lifecycle issue), the only refresh mechanism is `retry_with_new_client()` at the top of the sweep — which never sees the error. Net effect when the token expires mid-sweep: one doomed S3 call per remaining key, **nothing expired**, and the sweep reports success. Expired reservations linger until a later sweep's `list_objects_v2` happens to fail outside the swallowing handler.

## Also in scope (small, one-liners)

- `session.get_credentials()` returns `None` when no credentials are configured → opaque `AttributeError: 'NoneType' object has no attribute 'get_frozen_credentials'` instead of the pre-#1145 friendly "AWS credentials not found" `ValueError` (both workers).
- The reader's new `if not reservation:` guard is dead code — `ReservationDictionaryConverter.from_dict()` always returns a truthy object — so a malformed spec reaches the request path and crashes `is_expired()` with `'>' not supported between 'float' and 'NoneType'`.
- `IsTruncated=True` without `NextContinuationToken` now loops the listing forever re-fetching page 1 (was a loud `KeyError` before #1145).
- `add_reservations()` drops `source=` when calling `retry_with_new_client()`, so credential-expiry warnings log the generic worker name instead of the deployment origin.
- `"ExpiredToken" not in extractor.get("Error.Code", "")` raises `TypeError` when the parsed error body stores `Code` as null (botocore's REST-XML parser can produce this for an empty `<Code/>`).

Noravee:
Fixes #1152

The expiration sweep treated any object it could not parse as expired (silently deleting it), still crashed on stored-null reservation blocks with the reported "'NoneType' object has no attribute 'get'" error, and swallowed mid-sweep ExpiredToken errors so a sweep could report success while expiring nothing.

**Changes:**
- Add S3Util.extract_reservation_data() as the single policy point for parsing reservation objects. The sweep now skips and WARNs on unparseable objects instead of deleting them, and the reader treats them as not-found instead of building a bogus Reservation whose None expiration crashed the request path.
- Re-raise ExpiredToken (only) from expire_one_reservation() so retry_with_new_client() can reset cached credentials and re-run the sweep. NoSuchKey and other ClientErrors keep their existing handling.
- Add S3Util.get_error_code()/is_expired_token_error() for safe Error.Code extraction: DictionaryExtractor returns a stored null in preference to its default, which could raise TypeError inside except handlers.
- Raise NoCredentialsError instead of an opaque AttributeError when the credential chain resolves nothing (both client workers).
- Guard against an infinite listing loop when IsTruncated=True arrives without a NextContinuationToken.
- Pass source= through to retry_with_new_client() in the writer so credential-expiry warnings identify the deployment origin.

**Tests:**
- test_expiration_malformed_object_policy.py: wrong-shape objects survive the sweep, stored-null objects no longer kill it, and a control case pins that well-formed expired objects are still deleted.
- test_expiration_credential_expiry_mid_sweep.py: mid-sweep ExpiredToken triggers a client rebuild and the sweep completes.
- FakeS3Client grows list_objects_v2/delete_object mirroring real S3 semantics (lexicographic key order, idempotent delete).
- `AND` with UNHCR network.

---

## Update 2026-07-21 — changes after review

**Malformed-object policy reworked** (addressing the change request: "All Reservations should have an expiration... detritus builds up"):
- Skip-forever is gone. New `handle_malformed_object()` applies **age-gated deletion**: unparseable objects are skipped with a WARNING while younger than `MALFORMED_OBJECT_GRACE_SECONDS` (48h = 2× `Reservationist.DEFAULT_LIFETIME`), then **deleted with a WARNING** once older than any reservation could live. Every object under the prefix now has a bounded life, including unparseable ones (effectively `LastModified + grace`); rolling-deploy safety and diagnosability are preserved during the grace window (full reasoning in the review thread).
- Age comes from one `head_object` call made only on the malformed path — normal sweeps pay nothing extra.

**"What happens if expiration_time is None?"** — answered in code: null/missing/bool/non-numeric all fail `extract_reservation_data()`'s isinstance check and take the malformed path, so the `current_time > expiration_time` comparison can never see a None. Comments added at the comparison and in the helper; pinned by `test_null_expiration_time_is_treated_as_malformed`.

**HEAD-404 race fixed** (Copilot catch on the age-gating commit): an object deleted by another process between `get_object` and the `head_object` age check now counts as already-expired instead of logging an error — HEAD responses have no body, so botocore reports bare `"404"` rather than `NoSuchKey`. Pinned by `test_object_deleted_mid_check_counts_as_expired`.

**Smaller items from the Copilot rounds:** `Optional[...]` annotations where None is returned (`extract_reservation_data()`, `get_one_reservation()`, call sites); `Callable` on all `partial()` work functions (they were annotated `Awaitable`, which a partial is not); `is_expired_token_error()` used in the expiration handler instead of a manual substring check; docstrings synced to the age-gated policy (helper, test module, test class).

**Tests:** `FakeS3Client` gains `head_object` with backdatable per-key `LastModified`; new tests pin over-age deletion, the null-expiration case, and the 404 race. Suite is now 20/20.

### DIFF
--- a/neuro_san/service/watcher/temp_networks/aws_async_client_worker.py
+++ b/neuro_san/service/watcher/temp_networks/aws_async_client_worker.py
@@ -36,9 +36,9 @@ from botocore.exceptions import BotoCoreError
 from botocore.credentials import Credentials
 from botocore.credentials import ReadOnlyCredentials
 from botocore.exceptions import ClientError
+from botocore.exceptions import NoCredentialsError
 
 from leaf_common.logging.sensitive_logger import SensitiveLogger
-from leaf_common.parsers.dictionary_extractor import DictionaryExtractor
 
 from neuro_san.service.watcher.temp_networks.s3_util import S3Util
 
@@ -95,8 +95,13 @@ class AwsAsyncClientWorker:
                 return retval
 
             except ClientError as err:
-                extractor = DictionaryExtractor(err.response)
-                if "ExpiredToken" not in extractor.get("Error.Code", ""):
+                # S3Util.is_expired_token_error() is used instead of raw
+                # DictionaryExtractor access: the extractor returns a stored
+                # None in preference to its default, and
+                # '"ExpiredToken" not in None' would raise TypeError inside
+                # this handler, masking the original ClientError
+                # (see S3Util.get_error_code for details).
+                if not S3Util.is_expired_token_error(err):
                     raise
 
                 last_err = err
@@ -219,6 +224,17 @@ class AwsAsyncClientWorker:
             self.session = get_session()
 
         credentials: Credentials = await self.session.get_credentials()
+        if credentials is None:
+            # aiobotocore's credential resolver (like botocore's) returns None -
+            # it does not raise - when nothing in the chain (env vars, config
+            # files, IMDS/ECS, SSO) yields credentials. Without this check the
+            # call below fails with an opaque "AttributeError: 'NoneType' object
+            # has no attribute 'get_frozen_credentials'". Raise the
+            # botocore-idiomatic error instead; a client using the default
+            # credential chain surfaces this same misconfiguration as a
+            # NoCredentialsError from the S3 call itself, so callers/operators
+            # already know what it means.
+            raise NoCredentialsError()
 
         # Avoid a small race condition with the ClientError retry block
         # by always returning a local copy
--- a/neuro_san/service/watcher/temp_networks/aws_async_client_worker.py
+++ b/neuro_san/service/watcher/temp_networks/aws_async_client_worker.py
@@ -16,7 +16,7 @@
 # END COPYRIGHT
 
 from typing import Any
-from typing import Awaitable
+from typing import Callable
 
 from asyncio import CancelledError
 from asyncio import Lock as AsyncLock
@@ -74,7 +74,7 @@ class AwsAsyncClientWorker:
         # Cached frozen credentials which can be invalidated should they expire
         self.frozen_credentials: ReadOnlyCredentials = None
 
-    async def retry_with_new_client(self, work_function: Awaitable, *, source: str = None) -> Any:
+    async def retry_with_new_client(self, work_function: Callable, *, source: str = None) -> Any:
         """
         Retries the async work_function when client credentials can expire.
         :param work_function: The async work function to retry
@@ -135,7 +135,7 @@ class AwsAsyncClientWorker:
                 if self.async_aws_client_lock is None:
                     self.async_aws_client_lock = AsyncLock()
 
-    async def do_work_with_new_client(self, work_function: Awaitable, *, attempt: int = 1) -> Any:
+    async def do_work_with_new_client(self, work_function: Callable, *, attempt: int = 1) -> Any:
         """
         This method separates the machinations of obtaining a proper S3 client
         from add_all_reservations() which does all the actual work.
--- a/neuro_san/service/watcher/temp_networks/aws_sync_client_worker.py
+++ b/neuro_san/service/watcher/temp_networks/aws_sync_client_worker.py
@@ -30,11 +30,11 @@ from botocore.credentials import Credentials
 from botocore.credentials import ReadOnlyCredentials
 from botocore.exceptions import BotoCoreError
 from botocore.exceptions import ClientError
+from botocore.exceptions import NoCredentialsError
 from botocore.session import get_session
 from botocore.session import Session
 
 from leaf_common.logging.sensitive_logger import SensitiveLogger
-from leaf_common.parsers.dictionary_extractor import DictionaryExtractor
 
 from neuro_san.service.watcher.temp_networks.s3_util import S3Util
 
@@ -85,8 +85,13 @@ class AwsSyncClientWorker:
                 return retval
 
             except ClientError as err:
-                extractor = DictionaryExtractor(err.response)
-                if "ExpiredToken" not in extractor.get("Error.Code", ""):
+                # S3Util.is_expired_token_error() is used instead of raw
+                # DictionaryExtractor access: the extractor returns a stored
+                # None in preference to its default, and
+                # '"ExpiredToken" not in None' would raise TypeError inside
+                # this handler, masking the original ClientError
+                # (see S3Util.get_error_code for details).
+                if not S3Util.is_expired_token_error(err):
                     raise
 
                 last_err = err
@@ -199,6 +204,16 @@ class AwsSyncClientWorker:
             self.session = get_session()
 
         credentials: Credentials = self.session.get_credentials()
+        if credentials is None:
+            # botocore's credential resolver returns None - it does not raise -
+            # when nothing in the chain (env vars, config files, IMDS/ECS, SSO)
+            # yields credentials. Without this check the call below fails with
+            # an opaque "AttributeError: 'NoneType' object has no attribute
+            # 'get_frozen_credentials'". Raise the botocore-idiomatic error
+            # instead; a client using the default credential chain surfaces
+            # this same misconfiguration as a NoCredentialsError from the S3
+            # call itself, so callers/operators already know what it means.
+            raise NoCredentialsError()
 
         # Avoid a small race condition with the ClientError retry block
         # by always returning a local copy
--- a/neuro_san/service/watcher/temp_networks/s3_reservations_expiration.py
+++ b/neuro_san/service/watcher/temp_networks/s3_reservations_expiration.py
@@ -29,8 +29,6 @@ from logging import Logger
 from botocore.client import BaseClient
 from botocore.exceptions import ClientError
 
-from leaf_common.parsers.dictionary_extractor import DictionaryExtractor
-
 from neuro_san.service.watcher.temp_networks.aws_sync_client_worker import AwsSyncClientWorker
 from neuro_san.service.watcher.temp_networks.s3_reservations_retriever import S3ReservationsRetriever
 from neuro_san.service.watcher.temp_networks.s3_util import S3Util
@@ -139,6 +137,17 @@ class S3ReservationsExpiration:
                 # This was the last page - exit loop
                 break
             continuation_token = response.get("NextContinuationToken")
+            if not continuation_token:
+                # S3's ListObjectsV2 contract pairs IsTruncated=True with a
+                # NextContinuationToken, so real S3 never lands here. But if a
+                # non-compliant S3-compatible endpoint (or a test fake) omits
+                # the token, looping again without it would re-request the
+                # first page forever - an infinite loop that also burns S3 API
+                # calls for every re-listed key. Keep the failure loud
+                # rather than looping silently.
+                raise RuntimeError(
+                    f"{self.name}: list_objects_v2 returned IsTruncated=True without a "
+                    "NextContinuationToken; aborting listing to avoid an infinite pagination loop")
 
     # pylint: disable=too-many-locals
     def expire_one_reservation(self, obj_key: str, current_time: float,
@@ -153,7 +162,6 @@ class S3ReservationsExpiration:
         if source is None:
             source = self.name
 
-        empty: Dict[str, Any] = {}
         expired: bool = False
         try:
             # Retrieve the reservation object from S3
@@ -162,16 +170,40 @@ class S3ReservationsExpiration:
                 source=source,
                 sync_aws_client=sync_aws_client
             )
-            if agent_spec is None:
-                agent_spec = empty
 
-            extractor = DictionaryExtractor(agent_spec)
-            reservation_data: Dict[str, Any] = extractor.get("metadata.reservation", empty)
+            reservation_data: Dict[str, Any] = S3Util.extract_reservation_data(agent_spec)
+            if reservation_data is None:
+                # Policy for objects under the prefix that do not parse as
+                # reservations: leave them alone and warn. Three behaviors are
+                # possible here, and each has been tried or considered:
+                #   1. Raise: one malformed object makes every sweep crash at
+                #      the same key ("'NoneType' object has no attribute
+                #      'get'"), so nothing sorted after it ever expires.
+                #   2. Treat as expired and delete (e.g. by defaulting
+                #      expiration_time to 0, where current_time > 0 is
+                #      always true): silently destroys any
+                #      object whose shape we merely fail to understand - e.g.
+                #      one written by a newer/older writer version - with only
+                #      a debug-level log naming reservation '<unknown>'.
+                #   3. Skip and warn (current): safe for both the data and the
+                #      sweep. The cost is that true garbage lingers and warns
+                #      on every sweep. That is deliberate: a human should
+                #      decide to delete data, not a parse failure.
+                self.logger.warning(
+                    "%s: Object %s under the reservations prefix does not parse as a "
+                    "reservation (missing/null metadata.reservation or non-numeric "
+                    "expiration_time_in_seconds). Leaving it in place. "
+                    "If this is expected garbage, remove it manually; if it recurs for "
+                    "objects the writer produced, check for schema drift.",
+                    self.name, obj_key)
+                return False
 
             client_worker: AwsSyncClientWorker = self.retriever.get_sync_client_worker()
 
-            # Compare current time against reservation's expiration timestamp
-            expiration_time: float = reservation_data.get("expiration_time_in_seconds", 0)
+            # Compare current time against reservation's expiration timestamp.
+            # extract_reservation_data() guarantees this key exists and is numeric,
+            # so no default is needed and the comparison below cannot TypeError.
+            expiration_time: float = reservation_data.get("expiration_time_in_seconds")
             if current_time > expiration_time:
                 # Reservation has expired - remove it from S3 storage
                 try:
@@ -183,9 +215,11 @@ class S3ReservationsExpiration:
                     self.logger.debug("%s: Deleted expired reservation %s from S3", self.name, reservation_id)
                     expired = True
                 except ClientError as delete_error:
-                    extractor = DictionaryExtractor(delete_error.response)
-                    # Handle case where another process already deleted the object
-                    if extractor.get("Error.Code") != "NoSuchKey":
+                    # Handle case where another process already deleted the object.
+                    # get_error_code() is used instead of raw response access because
+                    # it always returns a str, even when the parsed error body stores
+                    # Code as null (see S3Util.get_error_code for details).
+                    if S3Util.get_error_code(delete_error) != "NoSuchKey":
                         # Re-raise other delete errors
                         raise delete_error
 
@@ -195,11 +229,26 @@ class S3ReservationsExpiration:
             # Reservation is still active - no action needed
 
         except ClientError as exception:
+            error_code: str = S3Util.get_error_code(exception)
             # Handle case where another process already removed the object before we could read it
-            extractor = DictionaryExtractor(exception.response)
-            if extractor.get("Error.Code") == "NoSuchKey":
+            if error_code == "NoSuchKey":
                 self.logger.debug("%s: Reservation %s was already removed by another process", self.name, obj_key)
                 expired = True  # Object is gone, which is the desired outcome for expiration
+            elif "ExpiredToken" in error_code:
+                # Credential expiry must NOT be swallowed here. The client this sweep
+                # uses was created from frozen credentials (see AwsSyncClientWorker),
+                # which never auto-refresh, and the only code that can refresh them is
+                # retry_with_new_client() wrapping expire_any_reservations() at the top
+                # of the sweep - and it can only react to ClientErrors that actually
+                # reach it. If ExpiredToken were merely logged like the codes below,
+                # every remaining key in the sweep would make one doomed S3 call,
+                # nothing would be expired, and the sweep would still report success;
+                # credentials would only refresh when a later sweep's list_objects_v2
+                # call happened to fail outside this handler. Re-raising lets the
+                # wrapper reset the cached credentials and re-run the sweep with a
+                # working client. Restarting the sweep is safe because deletes are
+                # idempotent: NoSuchKey on the re-run is treated as success above.
+                raise
             else:
                 # Log other S3 errors but don't raise - allows expiration to continue
                 self.logger.error("%s: S3 error processing reservation object %s: %s",
--- a/neuro_san/service/watcher/temp_networks/s3_reservations_expiration.py
+++ b/neuro_san/service/watcher/temp_networks/s3_reservations_expiration.py
@@ -16,6 +16,7 @@
 # END COPYRIGHT
 
 from typing import Any
+from typing import Callable
 from typing import Dict
 from typing import Iterable
 from typing import Optional
@@ -30,6 +31,7 @@ from logging import Logger
 from botocore.client import BaseClient
 from botocore.exceptions import ClientError
 
+from neuro_san.interfaces.reservationist import Reservationist
 from neuro_san.service.watcher.temp_networks.aws_sync_client_worker import AwsSyncClientWorker
 from neuro_san.service.watcher.temp_networks.s3_reservations_retriever import S3ReservationsRetriever
 from neuro_san.service.watcher.temp_networks.s3_util import S3Util
@@ -42,6 +44,15 @@ class S3ReservationsExpiration:
     The main entry point to this guy is expire_reservations() which gets called as part of
     S3ReservationsStorage watcher loop.
     """
+
+    # Grace period before an object under the prefix that does not parse as a
+    # reservation is deleted (see handle_malformed_object). Every reservation
+    # has a bounded lifetime (clamped against a server max, which defaults to
+    # Reservationist.DEFAULT_LIFETIME), so an object that has existed for
+    # twice that cannot be a live reservation under ANY schema version.
+    # Deployments that raise the server max lifetime beyond the default
+    # should raise this accordingly.
+    MALFORMED_OBJECT_GRACE_SECONDS: float = 2 * Reservationist.DEFAULT_LIFETIME
 
     def __init__(self, name: str = "S3ReservationsExpiration", bucket_name: str = "",
                  prefix: str = S3Util.DEFAULT_RESERVATIONS_PREFIX):
@@ -75,7 +86,7 @@ class S3ReservationsExpiration:
         """
         self.logger.debug("%s: Starting expiration process for S3 reservations", self.name)
 
-        expire_function = partial(self.expire_any_reservations)
+        expire_function: Callable = partial(self.expire_any_reservations)
 
         client_worker: AwsSyncClientWorker = self.retriever.get_sync_client_worker()
         client_worker.retry_with_new_client(expire_function, source=self.name)
@@ -128,7 +139,7 @@ class S3ReservationsExpiration:
             if continuation_token:
                 kwargs["ContinuationToken"] = continuation_token
 
-            list_objects_function = partial(sync_aws_client.list_objects_v2, **kwargs)
+            list_objects_function: Callable = partial(sync_aws_client.list_objects_v2, **kwargs)
             response = client_worker.do_with_retries(self.name, list_objects_function)
             if response is None:
                 response = {}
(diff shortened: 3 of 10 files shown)


neuro_san/service/watcher/temp_networks/s3_util.py
```diff
@@ -58,6 +61,92 @@ def is_retryable_client_error(err: ClientError) -> bool:
...
+        """
+        :param err: The ClientError exception to evaluate
+        :return: True if the error indicates expired credentials.
+                 Substring match intentionally covers both "ExpiredToken" and
+                 "ExpiredTokenException" (different AWS services/paths use both).
+        """
+        return "ExpiredToken" in S3Util.get_error_code(err)
+
+    @staticmethod
+    def extract_reservation_data(agent_spec: Any) -> Dict[str, Any]:
```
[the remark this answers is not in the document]
Noravee: Good catch — fixed in 751c05d2: `extract_reservation_data()` now returns `Optional[Dict[str, Any]]`, matching its docstring's "or None if the object is not shaped like a reservation" contract.

neuro_san/service/watcher/temp_networks/s3_reservations_reader.py
```diff
@@ -84,23 +82,32 @@ def get_one_reservation(self, reservation_id: str) -> Tuple[Reservation, AgentNe
...
+            # an AgentReservation instance - truthy even when built from {} -
+            # so a post-construction "if not reservation:" can never fire, and
+            # the bogus reservation's expiration_time_in_seconds of None would
+            # later crash ExpiringAgentNetworkStorage.is_expired() on the
+            # request path with
+            # "'>' not supported between instances of 'float' and 'NoneType'".
+            # Validating the raw dict here is the one reliable place to detect
+            # malformed payloads and treat them as not-found.
+            reservation_dict: Dict[str, Any] = S3Util.extract_reservation_data(agent_spec)
+            if reservation_dict is None:
```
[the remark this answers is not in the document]
Noravee: Fixed in 751c05d2: `reservation_dict` is now `Optional[Dict[str, Any]]`, and `agent_spec` is annotated `Any` — a parsed S3 body can be any JSON type (retrieve returns `loads()` output), which is exactly the input `extract_reservation_data()` is designed to accept and validate.

neuro_san/service/watcher/temp_networks/s3_reservations_expiration.py
```diff
@@ -162,16 +170,40 @@ def expire_one_reservation(self, obj_key: str, current_time: float,
                 source=source,
                 sync_aws_client=sync_aws_client
             )
-            if agent_spec is None:
-                agent_spec = empty
 
-            extractor = DictionaryExtractor(agent_spec)
-            reservation_data: Dict[str, Any] = extractor.get("metadata.reservation", empty)
+            reservation_data: Dict[str, Any] = S3Util.extract_reservation_data(agent_spec)
+            if reservation_data is None:
```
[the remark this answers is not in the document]
Noravee: Fixed in 751c05d2: `reservation_data` is now `Optional[Dict[str, Any]]` (and `agent_spec` is `Any` here too, same reasoning as the reader — the parsed body can be any JSON type).

neuro_san/service/watcher/temp_networks/s3_reservations_expiration.py
```diff
@@ -195,11 +232,26 @@ def expire_one_reservation(self, obj_key: str, current_time: float,
...
 
         except ClientError as exception:
+            error_code: str = S3Util.get_error_code(exception)
             # Handle case where another process already removed the object before we could read it
-            extractor = DictionaryExtractor(exception.response)
-            if extractor.get("Error.Code") == "NoSuchKey":
+            if error_code == "NoSuchKey":
                 self.logger.debug("%s: Reservation %s was already removed by another process", self.name, obj_key)
                 expired = True  # Object is gone, which is the desired outcome for expiration
+            elif "ExpiredToken" in error_code:
```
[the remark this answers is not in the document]
Noravee: Agreed — fixed in f30b603f: the handler now uses `S3Util.is_expired_token_error(exception)`, so the ExpiredToken-matching policy lives in exactly one place.

neuro_san/service/watcher/temp_networks/s3_reservations_reader.py
```diff
@@ -84,22 +83,33 @@ def get_one_reservation(self, reservation_id: str) -> Tuple[Reservation, AgentNe
...
+            # later crash ExpiringAgentNetworkStorage.is_expired() on the
+            # request path with
+            # "'>' not supported between instances of 'float' and 'NoneType'".
+            # Validating the raw dict here is the one reliable place to detect
+            # malformed payloads and treat them as not-found.
+            reservation_dict: Optional[Dict[str, Any]] = S3Util.extract_reservation_data(agent_spec)
+            if reservation_dict is None:
+                # Log when we get malformed content on read (for request)
+                self.logger.error("%s: Failed to parse reservation payload for %s", self.name, reservation_id)
+                return None, None
```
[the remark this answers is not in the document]
Noravee: Fixed in f30b603f: the return type is now `Tuple[Optional[Reservation], Optional[AgentNetwork]]`, matching the docstring's "(None, None) otherwise" contract.

neuro_san/service/watcher/temp_networks/s3_reservations_writer.py
```diff
@@ -124,7 +124,11 @@ async def add_reservations(self, reservations_dict: Dict[Reservation, Any],
             source = self.name
 
         work_function: Awaitable = partial(self.add_all_reservations, reservations_dict, source)
```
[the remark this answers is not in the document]
Noravee: Fixed in f30b603f: `work_function` (and `put_function`, which had the same issue) are now annotated `Callable` — a `partial()` of an async function returns an awaitable when invoked, it isn't one itself. Also updated `AwsAsyncClientWorker.retry_with_new_client()`/`do_work_with_new_client()` signatures to `Callable` for the same reason, matching the sync worker's existing convention.

neuro_san/service/watcher/temp_networks/s3_util.py
```diff
@@ -58,6 +62,92 @@ def is_retryable_client_error(err: ClientError) -> bool:
             return True
         return False
 
+    @staticmethod
+    def get_error_code(err: ClientError) -> str:
```
Noravee, line 66: DictionaryExtractor returns a stored null in preference to its default, so raw extractor access inside an except handler can itself raise TypeError. This helper guarantees a str.

neuro_san/service/watcher/temp_networks/s3_util.py
```diff
@@ -58,6 +62,92 @@ def is_retryable_client_error(err: ClientError) -> bool:
...
+        :return: The error code as a string, or "" if it cannot be determined
+        """
+        extractor = DictionaryExtractor(err.response)
+        code: Any = extractor.get("Error.Code", "")
+        if not isinstance(code, str):
+            code = "" if code is None else str(code)
+        return code
+
+    @staticmethod
+    def is_expired_token_error(err: ClientError) -> bool:
```
Noravee, line 91: Single home for the ExpiredToken-matching policy (substring intentionally also covers ExpiredTokenException). Both workers and the expiration sweep use this.

neuro_san/service/watcher/temp_networks/s3_util.py
```diff
@@ -58,6 +62,92 @@ def is_retryable_client_error(err: ClientError) -> bool:
...
+        """
+        :param err: The ClientError exception to evaluate
+        :return: True if the error indicates expired credentials.
+                 Substring match intentionally covers both "ExpiredToken" and
+                 "ExpiredTokenException" (different AWS services/paths use both).
+        """
+        return "ExpiredToken" in S3Util.get_error_code(err)
+
+    @staticmethod
+    def extract_reservation_data(agent_spec: Any) -> Optional[Dict[str, Any]]:
```
Noravee, line 101: The heart of the change: one shared answer to "is this object shaped like a reservation?". Previously the reader and the expiration sweep each had their own (contradictory) handling of malformed content.

neuro_san/service/watcher/temp_networks/s3_reservations_expiration.py
```diff
@@ -153,25 +163,50 @@ def expire_one_reservation(self, obj_key: str, current_time: float,
...
                 source=source,
                 sync_aws_client=sync_aws_client
             )
-            if agent_spec is None:
-                agent_spec = empty
 
-            extractor = DictionaryExtractor(agent_spec)
-            reservation_data: Dict[str, Any] = extractor.get("metadata.reservation", empty)
+            reservation_data: Optional[Dict[str, Any]] = S3Util.extract_reservation_data(agent_spec)
+            if reservation_data is None:
```
Noravee, line 189: Policy change: unparseable objects are skipped with a WARNING, never deleted. Before, a missing expiration defaulted to 0 so current_time > 0 always deleted the object. The comment below lays out the three candidate policies - happy to discuss the choice.

neuro_san/service/watcher/temp_networks/s3_reservations_expiration.py
```diff
@@ -195,11 +232,26 @@ def expire_one_reservation(self, obj_key: str, current_time: float,
...
 
         except ClientError as exception:
+            error_code: str = S3Util.get_error_code(exception)
             # Handle case where another process already removed the object before we could read it
-            extractor = DictionaryExtractor(exception.response)
-            if extractor.get("Error.Code") == "NoSuchKey":
+            if error_code == "NoSuchKey":
                 self.logger.debug("%s: Reservation %s was already removed by another process", self.name, obj_key)
                 expired = True  # Object is gone, which is the desired outcome for expiration
+            elif S3Util.is_expired_token_error(exception):
```
Noravee, line 242: The one ClientError that must NOT be swallowed: ExpiredToken has to reach retry_with_new_client() so it can refresh credentials and re-run. Everything else keeps the existing log-and-continue behavior.

neuro_san/service/watcher/temp_networks/s3_reservations_expiration.py
```diff
@@ -139,6 +138,17 @@ def iter_reservation_keys(self, sync_aws_client: BaseClient) -> Iterable[str]:
...
             continuation_token = response.get("NextContinuationToken")
+            if not continuation_token:
+                # S3's ListObjectsV2 contract pairs IsTruncated=True with a
+                # NextContinuationToken, so real S3 never lands here. But if a
+                # non-compliant S3-compatible endpoint (or a test fake) omits
+                # the token, looping again without it would re-request the
+                # first page forever - an infinite loop that also burns S3 API
+                # calls for every re-listed key. Keep the failure loud
+                # rather than looping silently.
+                raise RuntimeError(
```
Noravee, line 160: Defensive: real S3 always pairs IsTruncated with a token, but without this guard a non-compliant endpoint/fake would silently refetch page 1 forever.

neuro_san/service/watcher/temp_networks/s3_reservations_reader.py
```diff
@@ -84,22 +83,33 @@ def get_one_reservation(self, reservation_id: str) -> Tuple[Reservation, AgentNe
...
+            # ReservationDictionaryConverter.from_dict() unconditionally returns
+            # an AgentReservation instance - truthy even when built from {} -
+            # so a post-construction "if not reservation:" can never fire, and
+            # the bogus reservation's expiration_time_in_seconds of None would
+            # later crash ExpiringAgentNetworkStorage.is_expired() on the
+            # request path with
+            # "'>' not supported between instances of 'float' and 'NoneType'".
+            # Validating the raw dict here is the one reliable place to detect
+            # malformed payloads and treat them as not-found.
+            reservation_dict: Optional[Dict[str, Any]] = S3Util.extract_reservation_data(agent_spec)
```
Noravee, line 107: Validation moved BEFORE object construction: from_dict() always returns a truthy object, so the previous post-construction check could never fire, and a malformed payload crashed the request path later with float-vs-None.

neuro_san/service/watcher/temp_networks/aws_sync_client_worker.py
```diff
@@ -199,6 +204,16 @@ def get_frozen_credentials(self) -> ReadOnlyCredentials:
...
+        if credentials is None:
+            # botocore's credential resolver returns None - it does not raise -
+            # when nothing in the chain (env vars, config files, IMDS/ECS, SSO)
+            # yields credentials. Without this check the call below fails with
+            # an opaque "AttributeError: 'NoneType' object has no attribute
+            # 'get_frozen_credentials'". Raise the botocore-idiomatic error
+            # instead; a client using the default credential chain surfaces
+            # this same misconfiguration as a NoCredentialsError from the S3
+            # call itself, so callers/operators already know what it means.
+            raise NoCredentialsError()
```
Noravee, line 216: botocore's resolver returns None rather than raising when no credentials exist; without this the failure was an opaque AttributeError. Same guard in the async worker.

neuro_san/service/watcher/temp_networks/s3_reservations_writer.py
```diff
@@ -123,8 +123,14 @@ async def add_reservations(self, reservations_dict: Dict[Reservation, Any],
...
-        work_function: Awaitable = partial(self.add_all_reservations, reservations_dict, source)
-        await self.s3_async_client_worker.retry_with_new_client(work_function)
+        # partial() of an async method is a callable that RETURNS an awaitable
+        # when invoked, not an awaitable itself - hence Callable, not Awaitable.
+        work_function: Callable = partial(self.add_all_reservations, reservations_dict, source)
+        # Pass source through so the worker's credential-expiry warnings identify
+        # which deployment triggered the write. Without it the worker falls back
+        # to its own name ("S3ReservationsWriter"), losing the log correlation
+        # that the pre-refactor retry loop provided.
+        await self.s3_async_client_worker.retry_with_new_client(work_function, source=source)
```
Noravee, line 133: Restores the deployment source in credential-expiry warnings (previously fell back to the generic worker name).

tests/neuro_san/service/watcher/temp_networks/fake_s3_client.py
```diff
@@ -80,6 +81,43 @@ def get_object(self, Bucket: str, Key: str):
             )
         return {"Body": io.BytesIO(self.objects[Key])}
 
+    def list_objects_v2(self, Bucket: str = None, Prefix: str = "",
```
Noravee, line 92: The fake grows the two methods the expiration sweep actually calls, mirroring real S3 semantics the tests rely on: lexicographic key order and idempotent delete.

tests/neuro_san/service/watcher/temp_networks/test_expiration_malformed_object_policy.py
```diff
@@ -0,0 +1,204 @@
...
+import json
+import time
+
+from typing import Any
+
+from tests.neuro_san.service.watcher.temp_networks.s3_reservations_storage_test_base \
+    import S3ReservationsStorageTestBase
+
+
+class TestExpirationMalformedObjectPolicy(S3ReservationsStorageTestBase):
```
Noravee, line 80: Pins the malformed-object policy. The null-poison test reproduces the exact "'NoneType' object has no attribute 'get'" error reported from production; the control test anchors that well-formed expired objects are still deleted.

tests/neuro_san/service/watcher/temp_networks/test_expiration_credential_expiry_mid_sweep.py
```diff
@@ -0,0 +1,180 @@
...
+
+from unittest.mock import patch
+
+from botocore.exceptions import ClientError
+
+from tests.neuro_san.service.watcher.temp_networks.s3_reservations_storage_test_base \
+    import S3ReservationsStorageTestBase
+
+
+class TestExpirationCredentialExpiryMidSweep(S3ReservationsStorageTestBase):
```
Noravee, line 54: Simulates the token expiring between the listing and the per-object get/delete calls, and asserts the sweep rebuilds its client and still completes - guards the ExpiredToken re-raise path.

Noravee (commented):
Short annotations on the load-bearing parts of this change, for easier navigation.

neuro_san/service/watcher/temp_networks/aws_async_client_worker.py
```diff
@@ -219,6 +224,17 @@ async def get_frozen_credentials(self) -> ReadOnlyCredentials:
...
+            # aiobotocore's credential resolver (like botocore's) returns None -
+            # it does not raise - when nothing in the chain (env vars, config
+            # files, IMDS/ECS, SSO) yields credentials. Without this check the
+            # call below fails with an opaque "AttributeError: 'NoneType' object
+            # has no attribute 'get_frozen_credentials'". Raise the
+            # botocore-idiomatic error instead; a client using the default
+            # credential chain surfaces this same misconfiguration as a
+            # NoCredentialsError from the S3 call itself, so callers/operators
+            # already know what it means.
+            raise NoCredentialsError()
```
d1donlydfink, line 237: This can actually be handy. Thanks!

neuro_san/service/watcher/temp_networks/s3_reservations_expiration.py
d1donlydfink: I actually do not like the idea of keeping around Reservation files without any expiration. 
_All_ Reservations should have an expiration. Ignoring this would just mean that detritus builds
up in the system over time and there would be more and more files to read and ignore and re-log about over every expiration pass.

neuro_san/service/watcher/temp_networks/s3_reservations_expiration.py
```diff
@@ -153,25 +163,50 @@ def expire_one_reservation(self, obj_key: str, current_time: float,
...
+                return False
 
             client_worker: AwsSyncClientWorker = self.retriever.get_sync_client_worker()
 
-            # Compare current time against reservation's expiration timestamp
-            expiration_time: float = reservation_data.get("expiration_time_in_seconds", 0)
+            # Compare current time against reservation's expiration timestamp.
+            # extract_reservation_data() guarantees this key exists and is numeric,
+            # so no default is needed and the comparison below cannot TypeError.
+            expiration_time: float = reservation_data.get("expiration_time_in_seconds")
```
d1donlydfink, line 207: What happens if expiration_time is None now?

neuro_san/service/watcher/temp_networks/s3_reservations_expiration.py
```diff
@@ -195,11 +232,26 @@ def expire_one_reservation(self, obj_key: str, current_time: float,
...
+                # of the sweep - and it can only react to ClientErrors that actually
+                # reach it. If ExpiredToken were merely logged like the codes below,
+                # every remaining key in the sweep would make one doomed S3 call,
+                # nothing would be expired, and the sweep would still report success;
+                # credentials would only refresh when a later sweep's list_objects_v2
+                # call happened to fail outside this handler. Re-raising lets the
+                # wrapper reset the cached credentials and re-run the sweep with a
+                # working client. Restarting the sweep is safe because deletes are
+                # idempotent: NoSuchKey on the re-run is treated as success above.
+                raise
```
d1donlydfink, line 256: Interesting catch.

d1donlydfink (changes requested):
Overall nice changes. Thanks for catching corners.

I actually do not like the idea of keeping around Reservation files without any expiration.
All Reservations should have an expiration. Ignoring this would just mean that detritus builds
up in the system over time and there would be more and more files to read and ignore and re-log about over every expiration pass.

neuro_san/service/watcher/temp_networks/s3_util.py
```diff
@@ -58,6 +62,96 @@ def is_retryable_client_error(err: ClientError) -> bool:
...
+          * S3ReservationsReader defaulted to {} and built a bogus Reservation
+            (id=None, expiration=None) whose expiration=None later crashed
+            request handling with "'>' not supported between float and NoneType".
+          * S3ReservationsExpiration defaulted expiration_time to 0, which made
+            "malformed" indistinguishable from "expired at epoch" and silently
+            DELETED the object.
+        Centralizing the shape check means both consumers agree on what a
+        well-formed reservation looks like; each caller decides what a None
+        return means for it (reader: treat as not-found; expiration: skip and
+        warn, but never delete).
```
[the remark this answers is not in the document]
Noravee: Fixed in 978b17dd: the docstring now describes the age-gated policy (skip-and-warn while young, delete once older than any reservation could live) and points at `S3ReservationsExpiration.handle_malformed_object()` as the policy's home — it had gone stale when the policy changed in response to review feedback.

neuro_san/service/watcher/temp_networks/s3_reservations_expiration.py
```diff
@@ -195,11 +230,26 @@ def expire_one_reservation(self, obj_key: str, current_time: float,
             # Reservation is still active - no action needed
 
         except ClientError as exception:
+            error_code: str = S3Util.get_error_code(exception)
             # Handle case where another process already removed the object before we could read it
-            extractor = DictionaryExtractor(exception.response)
-            if extractor.get("Error.Code") == "NoSuchKey":
+            if error_code == "NoSuchKey":
                 self.logger.debug("%s: Reservation %s was already removed by another process", self.name, obj_key)
                 expired = True  # Object is gone, which is the desired outcome for expiration
```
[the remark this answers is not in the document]
Noravee: Sharp catch — a real race. Fixed in 978b17dd: the handler now treats both `NoSuchKey` (GET errors, code carried in the response body) and bare `"404"` (HEAD errors, no body for botocore to parse) as "already removed by another process", counting it as a successful expiration. Pinned by the new `test_object_deleted_mid_check_counts_as_expired`.

tests/neuro_san/service/watcher/temp_networks/test_expiration_malformed_object_policy.py
```diff
@@ -0,0 +1,330 @@
...
+    """
+    Verifies that expire_reservations() survives objects that are not
+    shaped like reservations, keeps expiring well-formed reservations
+    around them, and never deletes an object it could not parse.
+    """
```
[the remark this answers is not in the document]
Noravee: Fixed in 613577bc: the class docstring now describes both halves of the age-gated policy (skip and WARN within the grace period, delete with a WARNING beyond it) — same doc drift as the `extract_reservation_data()` docstring from the previous round, now hopefully the last remnant of the pre-review skip-forever wording.

neuro_san/service/watcher/temp_networks/s3_reservations_expiration.py
Noravee: Agreed, and adopted — with one refinement. The policy is now **age-gated deletion** (see `handle_malformed_object()`): an unparseable object is skipped with a WARNING while younger than `MALFORMED_OBJECT_GRACE_SECONDS` (2× `Reservationist.DEFAULT_LIFETIME` = 48h), and **deleted with a WARNING** once older. So your requirement holds in full generality: *every* object under the prefix now has a bounded life — including the ones we can't parse, which effectively get `expiration = LastModified + grace`. Detritus can't accumulate, and the re-read/re-log cost you flagged is bounded to the grace window instead of forever.

Why the grace window rather than deleting on first sight:

1. **Rolling deploys.** The bucket is shared across servers, so during a deploy old and new code sweep the same objects. If a newer writer version changes the reservation schema, an older server's sweep can't parse the new server's **live** reservations — immediate deletion would destroy them mid-lease. The grace period is ≥ any possible reservation lifetime, so an old sweeper can never delete anything *any* schema version could still consider live. "Expired" is a fact about the data; "unparseable" is a fact about whichever code version happens to be reading — the grace window keeps the second from masquerading as the first.

2. **Diagnosability.** An unparseable object means something wrote unexpected data — a writer bug, other tooling, schema drift. Deleting on parse failure destroys the only evidence *and* the signal: the original `'NoneType' has no 'get()'` reports would have been undiagnosable if the offending objects had been quietly removed before anyone looked. The grace window plus WARNINGs (naming the key) give a human time to notice and inspect before cleanup happens.

3. **Cost is confined to the bad path.** The age check is one `head_object` call made only when parsing fails — normal sweeps pay nothing. And the eventual deletion now logs at WARNING with the key and age, versus the previous debug-level line naming reservation `<unknown>`.

If 48h feels too long (or too short for deployments that raise the max lifetime), it's a single constant with a comment pointing there.

neuro_san/service/watcher/temp_networks/s3_reservations_expiration.py
Noravee: It can't be None by that line — but the question deserved a better answer than a comment two lines up, so it now has three: (1) `extract_reservation_data()` returns `None` for the *whole payload* when `expiration_time_in_seconds` is missing, null, bool, or non-numeric (a stored JSON null is the tricky case: `DictionaryExtractor.get()` returns it in preference to any default — that exact shape was still producing the original NoneType crash), so those objects take the malformed path and never reach the comparison; (2) expanded comments at the comparison and in the helper spell this out; (3) `test_null_expiration_time_is_treated_as_malformed` pins it executably: an object with `"expiration_time_in_seconds": null` is skipped as malformed, doesn't crash the sweep, and doesn't block later expirations.

neuro_san/service/watcher/temp_networks/s3_reservations_expiration.py
```diff
@@ -44,6 +45,15 @@ class S3ReservationsExpiration:
...
     """
 
+    # Grace period before an object under the prefix that does not parse as a
+    # reservation is deleted (see handle_malformed_object). Every reservation
+    # has a bounded lifetime (clamped against a server max, which defaults to
+    # Reservationist.DEFAULT_LIFETIME), so an object that has existed for
+    # twice that cannot be a live reservation under ANY schema version.
+    # Deployments that raise the server max lifetime beyond the default
+    # should raise this accordingly.
+    MALFORMED_OBJECT_GRACE_SECONDS: float = 2 * Reservationist.DEFAULT_LIFETIME
```
Noravee, line 55: 2x the default max reservation lifetime: an object older than this cannot be live under ANY schema version, so deleting it is provably safe. Deployments that raise the server max lifetime should raise this too (per the comment).

neuro_san/service/watcher/temp_networks/s3_reservations_expiration.py
```diff
@@ -211,3 +265,74 @@ def expire_one_reservation(self, obj_key: str, current_time: float,
                               self.name, obj_key, str(exception))
 
         return expired
+
+    def handle_malformed_object(self, obj_key: str, current_time: float,
```
Noravee, line 269: New home of the malformed-object policy (age-gated deletion, per review feedback). The docstring records all four candidate policies and why this one won. The head_object age check runs only on this rare path - normal sweeps pay nothing extra.

neuro_san/service/watcher/temp_networks/s3_reservations_expiration.py
```diff
@@ -195,11 +230,30 @@ def expire_one_reservation(self, obj_key: str, current_time: float,
...
-            # Handle case where another process already removed the object before we could read it
-            extractor = DictionaryExtractor(exception.response)
-            if extractor.get("Error.Code") == "NoSuchKey":
+            error_code: str = S3Util.get_error_code(exception)
+            # Handle case where another process already removed the object
+            # before we could read it. Two codes mean "gone": GET errors carry
+            # NoSuchKey in the response body, while HEAD errors (from the
+            # malformed-path head_object age check) have no body for botocore
+            # to parse, so a missing key surfaces as the bare HTTP status "404".
+            if error_code in ("NoSuchKey", "404"):
```
Noravee, line 239: "404" joined NoSuchKey here because HEAD responses have no body for botocore to parse - a missing key on the malformed-path age check surfaces as the bare HTTP status. Covers the deleted-between-GET-and-HEAD race.

neuro_san/service/watcher/temp_networks/s3_reservations_expiration.py
```diff
@@ -153,39 +174,53 @@ def expire_one_reservation(self, obj_key: str, current_time: float,
...
-            expiration_time: float = reservation_data.get("expiration_time_in_seconds", 0)
+            # Compare current time against reservation's expiration timestamp.
+            # What if expiration_time_in_seconds is None (a stored JSON null)
+            # or missing? It cannot be, by this line: null, missing, bool and
+            # non-numeric values all fail extract_reservation_data()'s
+            # isinstance check, which returns None for the WHOLE payload, and
+            # that case exits above via handle_malformed_object(). Here the
+            # value is guaranteed to be an int or float, so no default is
+            # needed and this comparison cannot raise TypeError.
+            expiration_time: float = reservation_data.get("expiration_time_in_seconds")
```
Noravee, line 207: Direct answer to the review question "what happens if expiration_time is None?": it can't reach this line - null/missing/bool/non-numeric all fail extract_reservation_data()'s isinstance check and exit above via the malformed path. Pinned by test_null_expiration_time_is_treated_as_malformed.

tests/neuro_san/service/watcher/temp_networks/fake_s3_client.py
```diff
@@ -80,6 +89,67 @@ def get_object(self, Bucket: str, Key: str):
...
+        returns 204 success, NOT a NoSuchKey error. The fake mirrors that so
+        tests exercise the storage code against real S3 semantics rather than
+        a stricter fake-only contract.
+        """
+        _ = Bucket
+        self.objects.pop(Key, None)
+        self.last_modified.pop(Key, None)
+        return {}
+
+    def head_object(self, Bucket: str, Key: str) -> Dict[str, Any]:
```
Noravee, line 130: Backs the age gate: put_object stamps "now", direct inserts default to "now" (young), and tests backdate via self.last_modified to exercise over-age deletion. Mirrors real S3's bare-404 behavior on HEAD for missing keys.

Noravee (commented):
**Changes since the review** (summary, with inline notes on the new code below):

1. **Malformed-object policy reworked per @d1donlydfink's feedback**: skip-forever is gone; the sweep now applies age-gated deletion via a new `handle_malformed_object()` - skip + WARN while younger than `MALFORMED_OBJECT_GRACE_SECONDS` (48h = 2x default max lifetime), delete + WARN once older than any reservation could live. Detritus is bounded; rolling-deploy safety and diagnosability are preserved (detailed reasoning in the review thread).
2. **"What happens if expiration_time is None?" answered in code**: expanded comments at the comparison and in `extract_reservation_data()`, plus an executable answer (`test_null_expiration_time_is_treated_as_malformed`).
3. **HEAD-404 race fixed** (Copilot catch): an object deleted by another process between `get_object` and the new `head_object` age check now counts as already-expired instead of logging an error - HEAD responses carry bare `"404"`, not `NoSuchKey`. Pinned by `test_object_deleted_mid_check_counts_as_expired`.
4. **Docstrings synced to the new policy** (helper, test module, test class) and **type annotations tightened** from the Copilot rounds: `Optional[...]` where None is returned, `Callable` on all `partial()` work functions.
5. **Test scaffolding**: `FakeS3Client` gains `head_object` with backdatable per-key `LastModified`; new tests pin over-age deletion, the null-expiration case, and the 404 race. 20/20 passing.

tests/neuro_san/service/watcher/temp_networks/test_expiration_malformed_object_policy.py
```diff
@@ -0,0 +1,332 @@
...
+            f"Expected the unparseable object {wrong_shape_key} to be left in place "
+            f"while younger than the grace period; it was deleted, meaning the sweep "
+            f"treats 'could not parse' as 'expired' and silently destroys data.",
+        )
+        self.assertIn(
+            live_key, self.fake_s3.objects,
+            f"Expected the live reservation {live_key} to survive the sweep.",
+        )
+
+    def test_old_wrong_shape_object_is_deleted(self):
```
Noravee, line 180: Pins the review requirement: unparseable objects older than the grace period ARE deleted - skip-forever would fail here, so detritus buildup is now guarded by a test rather than policy intent.

tests/neuro_san/service/watcher/temp_networks/test_expiration_malformed_object_policy.py
```diff
@@ -0,0 +1,332 @@
...
+            racing_key, time.time(), sync_aws_client=self.fake_s3)
+
+        self.assertTrue(
+            expired,
+            "Expected a 404 from the head_object age check to be treated as "
+            "'already removed by another process' (a successful expiration "
+            "outcome), not logged as an S3 error.",
+        )
+
+    def test_null_expiration_time_is_treated_as_malformed(self):
```
Noravee, line 251: The executable answer to the expiration_time-is-None question: a stored JSON null takes the malformed path (DictionaryExtractor returns stored nulls over defaults), never the comparison.

neuro_san/service/watcher/temp_networks/s3_reservations_expiration.py
d1donlydfink: OK. Reasonable.

d1donlydfink (approved):

PR #1156, opened 2026-07-22
Explicitly disable Gemini Automatic Function Calling (AFC) (#1096)

### LINKED ISSUE #1096: Gemini Automatic Function Calling (AFC) enabled by default (which is not good)
1C asks: We are currently using Google Gemini models for one of our programs and have observed that Automatic Function Calling (AFC) is being enabled during agent execution. We suggest, the LLM configuration for Gemini models should include options to explicitly disable AFC, as the function-calling behavior is primarily governed by NeuroSan. 

References:
https://github.com/googleapis/python-genai/blob/main/README.md#disabling-automatic-function-calling
https://forum.langchain.com/t/how-to-disable-automatic-function-calling-with-chatgooglegenerativeai/3352

Noravee:
## TL;DR — this is a cosmetic fix

**Nothing about NeuroSan's behavior or performance changes.** Automatic Function Calling (AFC) is a google-genai SDK feature that never actually runs in NeuroSan. This PR only silences the misleading INFO banner the SDK prints:

```
AFC is enabled with max remote calls: 10.
```

## Why AFC never runs here

AFC only auto-executes tools passed to the SDK as *Python callables*. NeuroSan drives Gemini through LangChain, which binds tools as function declarations (schemas) — the SDK never has anything to execute, and function calling remains governed entirely by NeuroSan's own agent loop. The banner is printed before the SDK discovers it has nothing to do, on LLM calls made with no tools bound (e.g. by agents with no down-chain tools).

## What this PR does

- `GeminiLlmPolicy.create_llm()` now returns `llm.bind(automatic_function_calling=AutomaticFunctionCallingConfig(disable=True))` — the only way LangChain lets us pass the setting (`ChatGoogleGenerativeAI` has no constructor argument for it).
- `delete_resources()` unwraps the binding via its `bound` attribute so web client cleanup keeps working.
- The langchain-core floor is bumped to 1.4.0 — the first version whose `bind()` returns a chat-model-aware binding (with `bind_tools()`) — so the fix is active on every install the requirements permit.

## Graceful fallbacks (return the llm unchanged, i.e. today's behavior)

- Without the google-genai SDK (langchain-google-genai < 4.0 builds on google-ai-generativelanguage): there is no AFC, and no banner, to worry about.
- On langchain-core < 1.4 (e.g. overridden pins): `bind()` yields a generic `RunnableBinding` whose missing `bind_tools()` would break agent creation for agents with tools, so the cosmetic fix is skipped and the harmless banner stays.

## Note on tool-bound requests

Requests with tools bound do not inherit `bind()` kwargs, so they do not carry the explicit disable — but they are banner-free regardless: google-genai >= 1.48 skips its AFC path entirely for declaration-style tools, and our langchain-google-genai >= 4.1.2 floor requires google-genai >= 1.56.

Fixes #1096

### DIFF
--- a/neuro_san/internals/run_context/langchain/llms/gemini_llm_policy.py
+++ b/neuro_san/internals/run_context/langchain/llms/gemini_llm_policy.py
@@ -47,6 +47,7 @@ class GeminiLlmPolicy(LlmPolicy):
         ChatGoogleGenerativeAI = self.resolver.resolve_class_in_module("ChatGoogleGenerativeAI",
                                                                        module_name="langchain_google_genai.chat_models",
                                                                        install_if_missing="langchain-google-genai")
+
         llm = ChatGoogleGenerativeAI(
             model=model_name,
             google_api_key=self.get_value_or_env(config, "google_api_key",
@@ -83,7 +84,58 @@ class GeminiLlmPolicy(LlmPolicy):
             # global verbose value) so that the warning is never triggered.
             verbose=False,
         )
-        return llm
+        return self._disable_afc(llm)
+
+    def _disable_afc(self, llm: BaseLanguageModel) -> Any:
+        """
+        Explicitly disable the google-genai SDK's Automatic Function Calling (AFC)
+        on the given llm, if the environment allows it.
+
+        AFC is a client-side convenience feature of the google-genai SDK that can
+        only auto-execute tools passed to the SDK as Python callables.  LangChain
+        binds tools as function declarations (schemas), so AFC can never execute
+        anything here -- function calling is governed entirely by neuro-san's own
+        agent loop.  However, when AFC is not explicitly disabled, the SDK logs a
+        misleading "AFC is enabled with max remote calls: 10." INFO banner on
+        LLM calls made with no tools bound (e.g. by agents with no down-chain
+        tools).  Disabling AFC is behaviorally a no-op that silences that banner.
+        See https://github.com/cognizant-ai-lab/neuro-san/issues/1096
+
+        ChatGoogleGenerativeAI has no constructor argument for this -- the setting
+        is only honored as a per-request kwarg -- hence bind().
+
+        Note that requests with tools bound do not inherit bind() kwargs, so they
+        do not carry the explicit disable.  They are banner-free regardless:
+        google-genai >= 1.48 skips its AFC path entirely for declaration-style
+        tools, and langchain-google-genai >= 4.1.2 (our floor) requires
+        google-genai >= 1.56.
+
+        :param llm: The ChatGoogleGenerativeAI instance to disable AFC on.
+        :return: llm.bind(...) with AFC disabled, which behaves like the original
+                llm.  The original llm is returned unchanged when the environment
+                cannot support the bind:
+                * Without the google-genai SDK (langchain-google-genai < 4.0 builds
+                  on google-ai-generativelanguage), there is no AFC to disable.
+                * On langchain-core < 1.4, bind() returns a generic RunnableBinding
+                  with no bind_tools() method, which would break agent creation for
+                  agents with tools.  There the cosmetic banner is left alone.
+        """
+        # Use lazy loading to prevent installing the world.
+        # google-genai comes with langchain-google-genai >= 4.0, so intentionally
+        # no install_if_missing here: an older stack has no AFC to disable.
+        # pylint: disable=invalid-name
+        AutomaticFunctionCallingConfig = self.resolver.resolve_class_in_module(
+            "AutomaticFunctionCallingConfig",
+            module_name="google.genai.types",
+            raise_if_not_found=False)
+        if AutomaticFunctionCallingConfig is None:
+            return llm
+
+        bound: Any = llm.bind(automatic_function_calling=AutomaticFunctionCallingConfig(disable=True))
+        if not hasattr(bound, "bind_tools"):
+            return llm
+
+        return bound
 
     async def delete_resources(self):
         """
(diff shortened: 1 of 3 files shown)


neuro_san/internals/run_context/langchain/llms/gemini_llm_policy.py
```diff
@@ -83,7 +84,64 @@ def create_llm(self, config: Dict[str, Any], model_name: str, client: Any) -> Ba
             # global verbose value) so that the warning is never triggered.
             verbose=False,
         )
-        return llm
+        # Cosmetic only: silences google-genai's misleading "AFC is enabled" log banner.
+        # Behavior is identical with or without this.
+        return self._disable_afc(llm)
```
Noravee, line 89: The only functional change to model creation: the constructed llm is wrapped (or returned unchanged) by `_disable_afc()` below. All constructor args are untouched.

neuro_san/internals/run_context/langchain/llms/gemini_llm_policy.py
```diff
@@ -83,7 +84,64 @@ def create_llm(self, config: Dict[str, Any], model_name: str, client: Any) -> Ba
...
+                  on google-ai-generativelanguage), there is no AFC to disable.
+                * On langchain-core < 1.4, bind() returns a generic RunnableBinding
+                  with no bind_tools() method, which would break agent creation for
+                  agents with tools.  There the cosmetic banner is left alone.
+        """
+        # Use lazy loading to prevent installing the world.
+        # google-genai comes with langchain-google-genai >= 4.0, so intentionally
+        # no install_if_missing here: an older stack has no AFC to disable.
+        # pylint: disable=invalid-name
+        AutomaticFunctionCallingConfig = self.resolver.resolve_class_in_module(
```
Noravee, line 129: Same lazy-loading pattern as `ChatGoogleGenerativeAI` above, but deliberately **no** `install_if_missing`: on stacks with langchain-google-genai < 4.0, google-genai isn't used at all, so there is nothing to install or disable. `raise_if_not_found=False` returns `None` instead of raising, which keeps the fallback a plain None-check.

neuro_san/internals/run_context/langchain/llms/gemini_llm_policy.py
```diff
@@ -83,7 +84,64 @@ def create_llm(self, config: Dict[str, Any], model_name: str, client: Any) -> Ba
...
+            "AutomaticFunctionCallingConfig",
+            module_name="google.genai.types",
+            raise_if_not_found=False)
+        if AutomaticFunctionCallingConfig is None:
+            # No google-genai SDK, so no AFC (and no banner) to worry about.
+            return llm
+
+        # AFC can only be disabled per-request, so bind the setting onto every call.
+        bound: Any = llm.bind(automatic_function_calling=AutomaticFunctionCallingConfig(disable=True))
```
Noravee, line 138: `bind()` is the only way LangChain lets us pass this setting: there is no constructor argument, and unknown constructor kwargs get shunted into `model_kwargs` and silently dropped from requests (tested).

neuro_san/internals/run_context/langchain/llms/gemini_llm_policy.py
```diff
@@ -83,7 +84,64 @@ def create_llm(self, config: Dict[str, Any], model_name: str, client: Any) -> Ba
...
+        if not hasattr(bound, "bind_tools"):
```
Noravee, line 139: Guard verified against both worlds: langchain-core 1.3.0 (generic `RunnableBinding`, no `bind_tools` -> raw llm returned, banner stays) and 1.5.0 (`_ChatModelBinding` -> fix active). With the requirements floor bump the active path is the norm; this covers environments that override pins.

neuro_san/internals/run_context/langchain/llms/gemini_llm_policy.py
```diff
@@ -120,15 +178,19 @@ async def delete_resources(self):
         # https://reference.langchain.com/python/integrations/langchain_google_genai/ChatGoogleGenerativeAI/
         # #langchain_google_genai.ChatGoogleGenerativeAI.async_client
 
+        # The llm may be a bind() wrapper around the actual chat model
+        # (see _disable_afc), in which case the clients live on its "bound" attribute.
+        base_llm: Any = getattr(self.llm, "bound", self.llm)
```
Noravee, line 183: `bind()` wraps the chat model, so the google web clients live one level down on `.bound`. The `getattr` fallback keeps unwrapped cases (old stacks, guard-tripped) working - both shapes are covered by the delete_resources unit tests.

requirements.txt
```diff
@@ -28,7 +28,9 @@ ruamel.yaml>=0.18.6
 # Keep all these guys limited for now, as it can be the wild wild west out there
 # We don't want to be on a dependencies treadmill with timing not of our choosing.
 langchain>=1.2.0,<2.0
-langchain-core>=1.2.5,<2.0
+# Floor 1.4.0: bind() on a chat model must return a chat-model-aware binding
+# (with bind_tools) for the Gemini AFC-disable in gemini_llm_policy.py to activate.
+langchain-core>=1.4.0,<2.0
```
Noravee, line 33: 1.4.0 is where langchain-core's chat-model-aware `bind()` (with `bind_tools`) landed - confirmed against the released wheels. Fresh installs already resolve 1.5.0, so this floor mainly guarantees the fix activates in constrained environments.

tests/neuro_san/internals/run_context/langchain/llms/test_gemini_llm_policy.py
```diff
@@ -0,0 +1,135 @@
...
+from neuro_san.internals.run_context.langchain.llms.gemini_llm_policy import GeminiLlmPolicy
+
+try:
+    from google.genai.types import AutomaticFunctionCallingConfig
+    HAS_GOOGLE_GENAI = True
+except ImportError:
+    HAS_GOOGLE_GENAI = False
+
+
+class _ChatModelLikeBinding:
```
Noravee, line 32: The stubs emulate both binding shapes - modern chat-model-aware (has `bind_tools`) and legacy generic (does not) - so both branches of the guard are covered without needing two langchain-core versions installed.

Noravee (commented):
Notes on the important lines.

neuro_san/internals/run_context/langchain/llms/gemini_llm_policy.py
d1donlydfink: Bizzare.

d1donlydfink (approved):
If you say so

PR #1157, opened 2026-07-22
S3 reservations: long-lived keyless clients — proactive credential refresh and connection reuse on the read path (#1153)

### LINKED ISSUE #1153: S3 reservations: revisit client lifecycle — frozen credentials disable botocore auto-refresh; per-call clients on the read path (follow-up to #1145)
## Context

Follow-up to #1145, companion to #1152. This one is design-level rather than a bug list: the credential handling in `AwsSyncClientWorker`/`AwsAsyncClientWorker` re-implements — reactively — what botocore already provides natively, and the per-call client lifecycle has measurable request-path cost. Both trace to a single decision: passing **frozen credentials as explicit keys** to `create_client()`.

## Part A: frozen credentials disable botocore's native auto-refresh

Temporary AWS credentials (IAM Instance Role, ECS Task Role, SSO — the exact sources the workers' own comments name) are a triple `access_key/secret_key/session_token` with an expiry, held in botocore as `RefreshableCredentials`: a client created **without** explicit keys freezes them per request at signing time, refreshing proactively inside the expiry window. A long-lived keyless client never presents an expired token to S3.

Passing explicit `aws_access_key_id/secret/token` to `create_client()` makes botocore build a plain **static** `Credentials` object instead — no refresh machinery, token pinned forever. Verified against the installed botocore:

```
keyless client:       signer credentials IS the RefreshableCredentials object: True
stale token        -> get_frozen_credentials() auto-refreshed: True
explicit-keys client: signer credentials type: plain static Credentials, has _refresh: False
                      token pinned forever (even as the provider rotates)
```

Consequences of the frozen-keys design:
- Recovery from token expiry is **reactive**: every expiry cycle costs one real failed `ExpiredToken` round trip per worker (user-visible latency + error-log noise on the read path) before `retry_with_new_client()` re-freezes and retries.
- The frozen-credentials cache, the ExpiredToken retry loops, and the client-creation lock all exist to compensate for the disabled refresh.
- Mid-sweep expiry (#1152, part 3) becomes possible at all — with signing-time refresh it cannot happen.
- For credentials rotated *externally* (env/file rewritten by another process), `session.get_credentials()` caches its resolved object, so the 8-attempt retry loop re-freezes the same expired token every time, back-to-back with no backoff.

## Part B: per-call client create/close on the read hot path

Every `get_one_reservation()` — per local-cache miss, and per request for an unknown reservation id since negative lookups are never cached — now:
- acquires the worker-wide threading lock,
- builds a brand-new botocore client (~2 ms warm measured; ~170 ms cold), serialized under that lock, and
- `close()`s it in a `finally`, discarding the urllib3 connection pool — so **every S3 GET pays a fresh TCP+TLS handshake** and churns FDs/TIME_WAIT sockets under concurrent load.

Before #1145 the retriever created one client in `start()` and reused it (warm pool, no lock).

One premise worth correcting from the #1145 description ("each request would use the same client, meaning everyone would have to wait for each other"): botocore clients are thread-safe and serve concurrent requests through their connection pool — a shared client does not serialize requests. It's the new per-call `create_client` under a single lock that introduces a serialization point.

## Proposal

One long-lived client per worker, created **without** explicit keys, reused across calls:
- botocore refreshes role/SSO tokens at signing time — proactive, zero failed calls, no retry loop needed for expiry;
- the connection pool stays warm across reads;
- most of the frozen-credentials cache / ExpiredToken retry / lock machinery in both workers can be deleted (the transient-throttle `do_with_retries` backoff is still worth keeping);
- async writer: enter the aiobotocore client context once at `start()` and exit at `stop()` (or per batch, if lifecycle management is undesirable).

Relationship to #1152: independent and compatible. The `ExpiredToken` re-raise proposed there is a mitigation *within* the current design; if this issue's proposal lands, that path simply becomes dormant.

Noravee:
### What the current design does

On `main`, every S3 operation in both workers goes through the same sequence:

1. Resolve the session's credentials and **freeze** them into a raw key/secret/token snapshot, cached on the worker (`frozen_credentials`).
2. Build a **brand-new client** from those explicit strings via `create_client(aws_access_key_id=..., aws_secret_access_key=..., aws_session_token=...)`, serialized under a worker-wide lock.
3. Run the operation.
4. `close()` the client (sync worker), discarding its connection pool.

When a token expires, an 8-attempt retry loop catches `ExpiredToken`, clears the frozen snapshot, and re-freezes.

### Why that is a problem

**1. Explicit keys switch off botocore's own credential machinery.** Temporary credentials (IAM Instance Role, ECS Task Role, SSO — the exact sources the workers' own comments name) are held by botocore as `RefreshableCredentials`: an object that re-checks its expiry window at *every request signing* and refreshes itself **before** the token dies. Passing explicit key strings to `create_client()` makes botocore build a plain static `Credentials` object instead — no refresh machinery, token pinned forever. Verified against the installed botocore (from #1153):

```
keyless client:       signer credentials IS the RefreshableCredentials object: True
stale token        -> get_frozen_credentials() auto-refreshed: True
explicit-keys client: signer credentials type: plain static Credentials, has _refresh: False
                      token pinned forever (even as the provider rotates)
```

So the current design opts out of the mechanism AWS built for exactly this problem, and then re-implements it — reactively.

**2. Reactive recovery costs one real failed S3 call per expiry cycle.** Role/STS tokens expire on the order of an hour. Each cycle, some request has to actually *fail* with `ExpiredToken` (user-visible latency on the read path, error-log noise) before the retry loop re-freezes. Worse, on `main` the retry loop re-called `session.get_credentials()`, which botocore *caches* — so for externally rotated static credentials it re-froze the same stale token 8 times back-to-back and could never recover at all. The mid-sweep expiry scenario (#1152, part 3) exists *only* because of this design; with signing-time refresh it cannot happen for token-based credentials.

**3. Per-call clients put construction plus TCP+TLS on the request path.** `get_one_reservation()` runs per reservation-cache miss and per request naming an unknown agent id (negative lookups are never cached). Every such read paid: the worker-wide lock, ~2 ms of client construction (measured warm; ~170 ms cold), and — because `close()` discards the urllib3 pool — a fresh TCP+TLS handshake for the actual GET, churning FDs/TIME_WAIT sockets under concurrent load.

**4. The compensating machinery IS the complexity.** The frozen-credentials cache, the 8-attempt no-backoff retry loop, and the create-under-lock dance exist only to patch around points 1–3. One premise from the #1145 discussion is also worth correcting: botocore clients are thread-safe and serve concurrent requests through their connection pool, so a shared client does **not** make callers "wait for each other" — it was the per-call `create_client` under a single lock that introduced a serialization point.

### Real-world evidence from prod (added 2026-08-07)

cognizant-ai-lab/neuro-san-studio#1310 ("AND is sporadically broken") is this design failing in production. The Datadog error:

```
S3ReservationsReader: S3 error processing reservation object tech_news_aggregation_network-… during sync:
An error occurred (InvalidToken) when calling the GetObject operation: The provided token is malformed or otherwise invalid.
```

The user-visible symptom — a network is created but "not found" when clicked — is the reader's swallow path: the credential error is logged and the reservation is reported as not-found, so the frontend gets a network id the backend then can't serve.

How the current design produces and *sustains* this: prod runs on EKS with IRSA (role-based, hourly-refreshable credentials), and the worker freezes a key/secret/token snapshot and builds every client from it. `InvalidToken` — S3 rejecting the token as malformed/mismatched, the signature of a snapshot captured at an unlucky moment — does not match the `ExpiredToken`-only reset gate, so **the bad snapshot is never invalidated**: every cache-miss read keeps failing until a pod restart or a coincidental `ExpiredToken` clears the cache. Under this PR there is no snapshot and no cached triple to go stale — each request signs through the session's `RefreshableCredentials`, produced atomically at signing time — so both the creation of that state and, more importantly, its persistence are gone. And should such a state still arise from a bad source, the credential-rejection gate now matches `InvalidToken` too (see Hardening below), so the worker actively rebuilds and recovers instead of merely not causing the problem.

### Why keyless + long-lived is the right replacement

- **It is the botocore-intended usage.** A client created *without* explicit keys holds the session's credential object and freezes it per request at signing time; refreshable sources refresh proactively. Zero failed calls on token rotation, and the entire reactive apparatus becomes unnecessary rather than patched.
- **The connection pool stays warm across reads** and steady-state client acquisition is a single unlocked attribute read — no lock, no construction, no handshake on the hot path.
- **The new failure modes are handled explicitly, not discovered later.** A long-lived *cached* client introduces hazards per-call clients never had (e.g. caching a client built while the credential chain was momentarily empty). Those are guarded and tested — see the Hardening section below.
- **What deliberately does *not* change:** the write path stays per-batch (aiobotocore clients are loop-bound context managers and writes are not hot); `retry_with_new_client()` keeps its name and signature so no caller changes; and #1154's `ExpiredToken` re-raise remains as the trigger for the now-dormant fallback.

**References — where the auto-refresh actually lives.** (Links pin botocore **1.43.0**, the version in our env. Note boto3's credential machinery *is* botocore's: `boto3.Session.get_credentials()` simply returns `self._session.get_credentials()` from its wrapped `botocore.session.Session`, so the boto3 guide below is the official prose for the botocore behavior our workers call directly.)

- [`RefreshableCredentials`](https://github.com/boto/botocore/blob/1.43.0/botocore/credentials.py#L388) — *"Holds the credentials needed to authenticate requests. In addition, it knows how to refresh itself."* This is what the credential chain resolves to for role/SSO/IRSA sources.
- [`RefreshableCredentials.get_frozen_credentials()`](https://github.com/boto/botocore/blob/1.43.0/botocore/credentials.py#L663) — its accessors *"always check and refresh credentials if needed before returning"*; the freeze exists precisely to avoid signing with a torn key/token triple mid-refresh (pod point 2 below).
- [`RequestSigner.sign()`](https://github.com/boto/botocore/blob/1.43.0/botocore/signers.py#L312) — calls `credentials.get_frozen_credentials()` on **every request**, so the refresh check runs *before* each request is signed. That chain — sign → freeze → refresh-if-needed — is the zero-failed-calls mechanism.
- [`Session.create_client()`](https://github.com/boto/botocore/blob/1.43.0/botocore/session.py#L962) — the counterpoint: passing explicit `aws_access_key_id`/`aws_secret_access_key` makes botocore construct plain static `Credentials` (no refresh machinery) instead of the session's refreshable object. The previous design's explicit-keys call took exactly this branch — this line is what disabled auto-refresh.
- Official AWS prose: the [boto3 credentials guide](https://docs.aws.amazon.com/boto3/latest/guide/credentials.html) (botocore has no credentials guide of its own; AWS documents the shared chain here). Assume-role provider: *"It will handle in-memory caching as well as refreshing credentials as needed."* Web-identity provider — what EKS IRSA uses, i.e. prod: *"Boto3 will automatically make the corresponding `AssumeRoleWithWebIdentity` calls to AWS STS on your behalf. It will handle in-memory caching as well as refreshing credentials, as needed."*

### Why this holds up on the pod (where the old races lived)

Credential handling on a pod is a different world from a local machine — and that is precisely where this design is *more* at home, not less. Locally, credentials are static file/env values resolved with a local read. On a pod (EC2 instance role, ECS task role, EKS IRSA) they are **temporary tokens fetched over the network** — IMDS, the ECS credential endpoint, or STS — with latency, timeouts, rate limits, and hourly expiry. The classic pod-only race conditions all trace to one of those, and each is addressed structurally here:

1. **Concurrent credential-chain resolution.** botocore `Session` is not thread-safe during resolution; several threads resolving at once against a slow IMDS is how you get the intermittent `NoCredentialsError` races the old code's comments describe guarding against with a lock. The old design *entered* that guarded path on every S3 call; the new design resolves the chain under the lock **once per process** (plus rare resets). Fewer entries into the racy region beats guarding it per call.

2. **Torn credential reads during refresh.** The historical botocore bug — a signer reading `access_key`/`secret_key`/`token` as three separate property reads while a refresh landed in between, yielding a mismatched triple and auth failures — is plausibly where the "freeze the credentials" approach came from, and freezing was the era-appropriate workaround. botocore has since fixed it upstream: signing takes an **atomic `get_frozen_credentials()` snapshot under the refresh lock**. This design leans on that internally-locked `RefreshableCredentials` machinery — the single most exercised concurrency path in botocore, since it is what every boto3 client on EKS/ECS runs through constantly.

3. **IMDS/STS throttling storms.** Repeated resolutions in a tight window get rate-limited, and throttled IMDS produces exactly the kind of intermittent failure that never reproduces locally. The new design performs ~one resolution per worker at startup plus one refresh per expiry window (three workers total — trivial load), and the only path that resolves repeatedly — the fallback retry — has jittered backoff, 4 attempts instead of 8, and compare-and-swap resets so N concurrent failures cannot fan out into N chain re-resolutions.

Two more points worth making explicit:

- **The pod is where the old design was at its worst.** Role tokens expire hourly, so the frozen snapshot *guaranteed* one real `ExpiredToken` failure per worker per hour, forever. With signing-time refresh that cycle does not exist.
- **This is not a novel pattern.** A long-lived keyless client on role credentials is the default way boto3 is deployed on EKS/ECS everywhere; the frozen-explicit-keys pattern is the unusual one.

Honest caveats: once an hour the first request inside the refresh window triggers the IMDS/STS call (in botocore's advisory window this happens alongside still-valid credentials, so it is normally invisible); and a pod that cannot reach IMDS at all (the classic `http_put_response_hop_limit=1` misconfiguration) fails under either design — but the new one fails with an explicit `NoCredentialsError` from the creation-time guard rather than caching a broken client. Since the old races were found empirically, the plan is to validate empirically too: run the high-concurrency load test on a pod against this branch before merging.

---

## Summary

Follow-up to the #1145 review, closing #1153 (concerns 2 and 3: frozen credentials disable botocore's native auto-refresh; per-call clients on the read hot path). Companion to #1152 / #1154, which fixed the behavioral bugs — this PR fixes the client-lifecycle design underneath them.

Both problems traced to one decision: passing **frozen credentials as explicit keys** to `create_client()` and building/closing a client around every call.

### Credential handling: delegate refresh to botocore (concern 2, Part A)

Clients are now created **without** explicit keys. A keyless client holds the session's credential *object* and freezes it per request at signing time; for token-based sources (IAM Instance Role, ECS Task Role, SSO) that object is `RefreshableCredentials`, which checks its expiry window on every request and refreshes itself *before* signing. A long-lived keyless client therefore never presents an expired token to S3 — refresh is proactive and costs zero failed calls.

Deleted along with the old design: the `frozen_credentials` cache and `get_frozen_credentials()` in both workers, and the manual `NoCredentialsError` guard (a keyless client raises it natively at signing time).

### Client lifecycle (concern 3, Part B)

- **`AwsSyncClientWorker` (read path + expiration sweep):** one long-lived client per worker, created lazily under the lock, then used lock-free — botocore clients are thread-safe and serve concurrent requests through their connection pool. The per-read lock/create/close cycle is gone and the pool stays warm across reads, so request-path S3 GETs no longer pay client construction plus a fresh TCP+TLS handshake each.
- **`AwsAsyncClientWorker` (writer):** still one client per deployment batch — aiobotocore clients are async context managers whose lifetime must be managed inside the event loop that uses them, the writer has no reachable stop() hook to exit a long-lived context at shutdown, and the write path runs per batch, not per request. What changed is the keyless creation; the session is also now reused across batches.

`retry_with_new_client()` keeps its name and signature — **no caller changed** — but is now a dormant fallback: it rebuilds session + client only when S3 rejects the credentials a request was signed with (externally rotated static session tokens, which botocore resolves once per Session and never re-reads on its own). The `ExpiredToken` re-raise that landed in #1154 is exactly what feeds this fallback.

### Hardening (second commit)

Before opening this PR I ran an adversarial multi-angle review of the first commit; it surfaced real gaps, all fixed in the second commit:

- **Credential-chain guard in `get_client()`** — `create_client()` pins whatever the chain resolves *at that instant*, including `None` (e.g. a credentials file mid-rewrite during rotation). A client pinned with `None` raises `NoCredentialsError` on every request, so caching it would have wedged the worker until process restart. `get_client()` now raises without caching anything; the credential retry treats it as a credential rejection (see the gate bullet below), and `Session.get_credentials()` re-resolves whenever its previous resolution came back None, so the worker self-heals once credentials reappear.
- **`NoCredentialsError` fails fast in `do_with_retries()`** (both workers) — it is a `BotoCoreError` subclass, and the generic transient-error branch would otherwise sleep through ~16–48 s of exponential backoff on an error that cannot heal within *this* loop; re-raising immediately routes it to `retry_with_new_client()`'s credential retry, which discards the session and can.
- **Connection pool sized for concurrency** — the long-lived client is created with `Config(max_pool_connections=MAX_POOL_CONNECTIONS)` (50) instead of botocore's default 10. One pool now serves *all* concurrent readers, and with urllib3's `block=False` every request beyond the default cap would silently open a fresh TCP+TLS connection and discard it on release — quietly reintroducing the per-call handshake cost under exactly the concurrent-read load this PR targets.
- **Retry policy: 4 attempts with jittered backoff** (`CREDENTIAL_RETRY_MAX_ATTEMPTS` / `CREDENTIAL_RETRY_BASE_SLEEP_SECONDS`) instead of 8 back-to-back — each attempt now re-resolves the full credential chain (possible IMDS/ECS/SSO network calls), so zero-backoff retries would hammer those endpoints and still fail before an in-progress rotation lands.
- **Compare-and-swap resets** — `reset_client(failed_client=...)` skips the reset when a concurrent thread already rebuilt a fresh client, so N threads failing on the *old* client during a rotation no longer serially discard each other's rebuilds; the async worker's session gets the same CAS treatment (`do_work_with_new_client()` now receives the session as a parameter so the retry handler knows exactly which session an attempt used).
- **Honest recovery contract in the docs** — what self-heals (session tokens rotated in a credentials file, surfacing as `ExpiredToken`) vs. what deliberately requires a restart: env-var "rotation" (a process's environment cannot be changed from outside) and rotated access-key *pairs* (surface as `InvalidAccessKeyId`/`SignatureDoesNotMatch`; widening the gate to match those would make genuine signing bugs retry instead of surfacing). Also documented: HEAD error responses carry no body, so an expired-token `head_object` failure on the (rare) malformed-object path surfaces as bare `"400"` and recovery happens on the next key's GET.
- **Credential-rejection gate widened to `InvalidToken`** (added 2026-08-07, motivated by cognizant-ai-lab/neuro-san-studio#1310) — `S3Util.is_credential_rejection_error()` now treats `InvalidToken` ("malformed or otherwise invalid") like `ExpiredToken`: discard session + client, re-resolve, retry. Safe to widen *because* clients are keyless: our code never assembles key/secret/token triples itself, so `InvalidToken` can only mean the resolved credential state is bad, and one bounded rebuild cycle (4 attempts, jittered backoff) is the right reflex. `InvalidAccessKeyId`/`SignatureDoesNotMatch` remain deliberately unmatched. The retry warning now names the rejection code it is recovering from.
- **Gate widened again to local resolution failures and `TokenRefreshRequired`** (added 2026-08-07, follow-up from an adversarial review of this branch) — `NoCredentialsError`/`PartialCredentialsError` are `BotoCoreError`s, *not* `ClientError`s, so both workers' `except ClientError` retry gates let the mid-rewrite phase of a rotation (chain resolves empty or half-written) escape with retry budget unused: an unhandled exception on the read request path, and a silently lost deployment batch on the write path (verified with live botocore/aiobotocore repros). Both retry loops now catch them alongside `ClientError`, and `S3Util.is_credential_rejection_error()` classifies them — plus S3's third temporary-token rejection code, `TokenRefreshRequired` — as credential rejections. The sync loop also pre-binds its client local before the `try` so the handler can never see an unbound/stale reference, and skips the reset when `get_client()` itself raised (nothing was cached).

## Tests

Two new acceptance tests, both originally written **red against the old design** and green here (verified by stashing the production changes and re-running — they fail on `main`'s design):

- `test_reader_client_reuse.py` — N reads must construct exactly **one** S3 client.
- `test_proactive_credential_refresh.py` — a token rotation between two reads must cause **zero** S3 calls with the stale token, plus a mechanism check that `create_client` receives no explicit credential arguments (the thing that would disable refresh).

Test infrastructure:
- The base class now sets fake `AWS_*` env credentials as a **hermeticity barrier**: the workers build real botocore/aiobotocore sessions (only `create_client` is patched), and the sync worker now consults the credential chain, so resolution must be instant and deterministic rather than walking the developer/CI machine's real chain (config files, SSO, IMDS on EC2).
- Shared scaffolding moved to the base (`_put_live_reservation` via `S3Util.get_obj_key_for_reservation`, the `_fresh_reader_client` repatch-and-reset context manager, `make_expired_token_error`) instead of being duplicated per test file.
- The mid-sweep credential-expiry test from #1154 passes with unchanged assertions; its narration was updated to describe the new design. Dead `close()` methods on the fakes were removed (nothing on the sync path calls `close()` anymore).

Also added with the `InvalidToken` widening: `test_invalid_token_recovery.py` — a reader-path recovery test reproducing the #1310 failure mode (verified red against the `ExpiredToken`-only gate), plus gate-boundary tests pinning which codes trigger re-resolution and which must surface.

Added with the second widening: `test_empty_chain_recovery.py` — a reader-path test pinning the ride-out of an empty credential chain (`NoCredentialsError` → backoff → re-resolve → read succeeds), plus gate-boundary tests for `TokenRefreshRequired`, the local resolution failures, and generic `BotoCoreError`s that must surface. All verified red against the `ClientError`-only gate. The test base also gained a hermeticity fix: `AWS_PROFILE`/`AWS_DEFAULT_PROFILE` are popped and the config-file env vars point at `os.devnull` — previously a bogus host profile made botocore raise `ProfileNotFound` during config load, before any credential provider ran, failing every test in setUp (reproduced end-to-end).

Writer-path counterparts (`test_writer_credential_recovery.py`, prompted by a Copilot review note): `add_reservations` batches driven through an `InvalidToken` rejection and a local `NoCredentialsError` at signing time, verifying the retry discards the shared `AioSession`, builds a second one, and lands the batch instead of losing it. The `NoCredentialsError` case verified red against a `ClientError`-only gate.

Suite: **30 passed** · pylint **10.00/10** · flake8 clean (repo config, max-line-length=120).

## Relationship to #1152 / #1154 / #1215

Independent and compatible. The `ExpiredToken` re-raise from #1154 is the trigger for this PR's fallback path; the malformed-object age-gating policy is untouched.

**#1215 (merged 2026-08-07)** is the minimal recovery-only half of this PR, cherry-picked onto the old design to stop the #1310 prod 404s immediately: the same `S3Util.is_credential_rejection_error()` gate (same name, same codes), wired into the old frozen-snapshot reset. This branch supersedes it — the merge resolved every overlap to this branch's version with zero net content change — and adds what #1215 deliberately left out: **prevention** (keyless clients, zero failed calls on rotation, no per-call client construction) plus the local `NoCredentialsError`/`PartialCredentialsError` handling that only matters under the keyless design.

## Update 2026-08-07 — merged main's `temp_networks` restructure

`main` reorganized `temp_networks/` into `s3/`, `azure/` (new Azure Blob support, #1193), `common/`, and `updater/` subpackages, which conflicted with this branch. Merged in 049e9da9:

- The redesign now lives at `neuro_san/service/watcher/temp_networks/s3/` (tests under `tests/.../temp_networks/s3/`); git rename detection carried every change across, verified piece by piece. **Nothing in the design changed** — everything above still describes the code.
- Conflict resolutions (both in tests) preserved this PR's approach: the test base keeps the fake-`AWS_*`-env hermeticity barrier *instead of* the `get_credentials` patchers main's copy still had, and the mid-sweep test keeps the shared `make_expired_token_error` helper.
- main's orthogonal move-time changes were adopted untouched: botocore/aiobotocore as lazily-resolved soft dependencies (`ResolverUtil`) in expiration/writer, and check-interval env parsing via `ExternalStorageUtil`.
- The two acceptance tests moved into `tests/.../temp_networks/s3/` with imports and the `_fresh_reader_client` patch target updated.
- The reviewer annotations posted earlier now show as "outdated" (they anchor to the pre-restructure paths); they have been re-posted at the new paths.
- Correction to the Tests note above: the repo's flake8 limit is 120, not 119 (main's aligned `# noqa` import lines are legal under it).
- Same day, after diagnosing cognizant-ai-lab/neuro-san-studio#1310: widened the credential-rejection gate to `InvalidToken` (commit 6cee6f16) with a recovery test verified red against the previous gate — see the evidence section and the Hardening bullet.
- Same day, after an adversarial multi-agent review of this branch: widened the gate again to `NoCredentialsError`/`PartialCredentialsError` and `TokenRefreshRequired`, fixed a latent unbound-local in the sync retry handler, and hardened test hermeticity against host `AWS_PROFILE` settings (commit 4827a17f) — see the Hardening and Tests sections.
- Same day: the recovery-only half of this design shipped to main as **#1215** (quick fix for the #1310 prod 404s) and was merged back into this branch (7759333d) with zero net content change — see the Relationship section. With recovery live in prod, what remains unique to this PR is **prevention**: no failed call per rotation, no per-call client construction on the read path.

Closes #1153


tests/neuro_san/service/watcher/temp_networks/test_reader_client_reuse.py
```diff
@@ -0,0 +1,99 @@
...
+  * acquired a worker-wide threading lock,
+  * built a brand-new botocore client (~2ms warm, measured - endpoint
+    resolution plus a fresh urllib3 pool), serialized under that lock, and
+  * close()d the client in a finally block, discarding the connection
+    pool - so every S3 GET paid a fresh TCP+TLS handshake and churned
+    file descriptors / TIME_WAIT sockets under concurrent load.
+
+Per-call clients are not required for credential correctness either: the
+long-lived client AwsSyncClientWorker now keeps is created WITHOUT
+explicit keys, so each request is signed through the session's
```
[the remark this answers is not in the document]
Noravee: Fixed in fe4a414d — reworded to "AwsSyncClientWorker's long-lived client is created WITHOUT explicit keys". The original was a garden-path sentence ("the long-lived client [that] AwsSyncClientWorker now keeps is created…") — grammatical, but it made readers backtrack, which defeats the point of the docstring.

tests/neuro_san/service/watcher/temp_networks/test_proactive_credential_refresh.py
```diff
@@ -0,0 +1,183 @@
...
+    stale STS/role token.
+    """
+
+    # pylint: disable=too-few-public-methods
+
+    def __init__(self, fake_s3: FakeS3Client, bound_token: str,
+                 provider_state: Dict[str, str],
+                 failed_calls: List[Tuple[str, str]]):
+        self._fake_s3: FakeS3Client = fake_s3
+        self._bound_token: str = bound_token
```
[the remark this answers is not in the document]
Noravee: Fixed in fe4a414d — `bound_token` is now `Optional[str]` in the signature and on the attribute, with a comment noting that None means "created keylessly". Agreed this one mattered: None isn't an edge case here, it's the value the whole test is about.

tests/neuro_san/service/watcher/temp_networks/test_proactive_credential_refresh.py
```diff
@@ -0,0 +1,183 @@
...
+    def get_object(self, Bucket: str, Key: str) -> Dict[str, Any]:
+        """
+        Reject the request if the token it presents is no longer the
+        provider's current token; otherwise delegate to the fake.
+        """
+        presented_token: str = self._bound_token
+        if presented_token is None:
+            # Keyless client: the request is signed with the token the
+            # provider holds RIGHT NOW, resolved at request time.
+            presented_token = self._provider_state["current_token"]
```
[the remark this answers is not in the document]
Noravee: Fixed in fe4a414d — `presented_token` is now `Optional[str]`, matching the keyless-client path where it starts as None before being resolved from the provider.

neuro_san/service/watcher/temp_networks/aws_sync_client_worker.py
```diff
@@ -96,140 +166,115 @@ def retry_with_new_client(self, work_function: Callable, *, source: str = None)
...
+                # arguments here: passing them would pin a static snapshot of
+                # the credentials into the client and disable auto-refresh.
+                self.sync_aws_client = self.session.create_client(
+                    self.aws_service,
+                    config=Config(max_pool_connections=self.MAX_POOL_CONNECTIONS))
+                self.logger.info("%s: Created long-lived %s client", self.name, self.aws_service)
+
+            return self.sync_aws_client
+
+    def reset_client(self, failed_client: BaseClient = None):
```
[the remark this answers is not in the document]
Noravee: Fixed in fe4a414d, using `Optional[BaseClient]` rather than `BaseClient | None` — this codebase imports from `typing` throughout, so `Optional` keeps the style consistent (same call as PR #1154's `Optional[Dict[str, Any]]`). Worth noting the surrounding code has many pre-existing `param: BaseClient = None` signatures; I only changed the one this PR introduces rather than sweeping the file.

neuro_san/service/watcher/temp_networks/s3/aws_sync_client_worker.py
```diff
@@ -58,30 +102,57 @@ def __init__(self, name: str, aws_service: str = "s3"):
...
+        match them would make genuine signing bugs retry instead of
+        surfacing). Both of those require a process restart.
+
+        NOTE: mirrored in AwsAsyncClientWorker.retry_with_new_client() -
+        keep the retry policies in sync when editing.
+
         :param work_function: The work function to retry
         :param *: extra keyword arguments for work_function
         :param source: A string describing where the deployment was coming from
         :return: What work_function returns
```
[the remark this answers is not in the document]
Noravee: Fixed in d8a359b2 — good catch. The bare `*` here is the keyword-only marker, not `*args`/`**kwargs`, so the line documented a parameter callers can't actually pass (it was inherited from the pre-#1153 docstrings). Removed it here, in the async twin you flagged, and in the same inherited line in the async `do_work_with_new_client` — that was the last `:param *:` in the repo.

neuro_san/service/watcher/temp_networks/s3/aws_async_client_worker.py
```diff
@@ -68,15 +99,33 @@ def __init__(self, name: str, aws_service: str = "s3"):
...
+        rotated access KEY PAIRS (S3 rejects those with InvalidAccessKeyId
+        or SignatureDoesNotMatch, not ExpiredToken). Both require a
+        process restart.
+
+        NOTE: mirrored in AwsSyncClientWorker.retry_with_new_client() -
+        keep the retry policies in sync when editing.
+
         :param work_function: The async work function to retry
         :param *: extra keyword arguments for work_function
         :param source: A string describing where the deployment was coming from
```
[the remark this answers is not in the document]
Noravee: Fixed in d8a359b2, along with the same inherited line in `do_work_with_new_client` below (not flagged, but identical) — no `:param *:` lines remain anywhere in the repo.

Noravee (commented):
Summary:

* **One long-lived keyless client per sync worker** - botocore refreshes role/SSO tokens at signing time; the pool stays warm; the read hot path acquires the client with a single unlocked attribute read.
* **`retry_with_new_client()` keeps its API** but is now a dormant fallback (rebuild session + client) for externally rotated session tokens only.
* **The async writer stays per-batch** (aiobotocore clients are loop-bound context managers and the write path is not hot) but creates keylessly.
* **Guard rails** below cover the failure modes a long-lived cached client introduces that per-call clients never had to worry about.

neuro_san/service/watcher/temp_networks/s3/aws_sync_client_worker.py
```diff
@@ -96,140 +166,115 @@ def retry_with_new_client(self, work_function: Callable, *, source: str = None)
...
+                # including None (an empty chain, e.g. a credentials file
+                # mid-rewrite during rotation). A client pinned with None
+                # raises NoCredentialsError - a BotoCoreError the
+                # ExpiredToken retry above can never see - on every request,
+                # so caching it would wedge this worker until restart.
+                # Raising here instead leaves the cache empty, and
+                # Session.get_credentials() re-resolves whenever its previous
+                # resolution came back None, so the worker self-heals on the
+                # next call once credentials are available again.
+                if self.session.get_credentials() is None:
```
Noravee: **The most important guard in the PR.** `create_client()` pins whatever the chain resolves *at this instant* - including `None` (credentials file mid-rewrite). A cached client pinned with `None` raises `NoCredentialsError` (a `BotoCoreError`) on every request, which the `except ClientError` retry can never see - so without this check, one unlucky rotation window would wedge the worker until process restart. Raising here leaves the cache empty, and `Session.get_credentials()` re-resolves whenever its previous result was None, so the worker self-heals. Verified against installed botocore before landing.

neuro_san/service/watcher/temp_networks/s3/aws_sync_client_worker.py
```diff
@@ -96,140 +166,115 @@ def retry_with_new_client(self, work_function: Callable, *, source: str = None)
...
+                    raise NoCredentialsError()
+
+                # No aws_access_key_id/aws_secret_access_key/aws_session_token
+                # arguments here: passing them would pin a static snapshot of
+                # the credentials into the client and disable auto-refresh.
+                self.sync_aws_client = self.session.create_client(
+                    self.aws_service,
+                    config=Config(max_pool_connections=self.MAX_POOL_CONNECTIONS))
```
Noravee: Without this, the shared client would keep botocore's default `max_pool_connections=10`, and with urllib3's `block=False` every request beyond 10 concurrent reads would silently open a fresh TCP+TLS connection and discard it on release - quietly reintroducing the per-call handshake cost under exactly the concurrent-read load this PR targets. Never mattered before because each per-call client had its own pool.

neuro_san/service/watcher/temp_networks/s3/aws_sync_client_worker.py
```diff
@@ -96,140 +166,115 @@ def retry_with_new_client(self, work_function: Callable, *, source: str = None)
...
-            raise NoCredentialsError()
-
-        # Avoid a small race condition with the ClientError retry block
-        # by always returning a local copy
-        local_frozen = credentials.get_frozen_credentials()
-        self.frozen_credentials = local_frozen
-
-        return local_frozen
+        with self.sync_aws_client_lock:
+            if failed_client is not None and self.sync_aws_client is not failed_client:
```
Noravee: Compare-and-swap: during a rotation, N threads that all failed on the *old* client would otherwise serially discard the fresh session+client the first thread's retry already rebuilt - N redundant credential-chain re-resolutions under the lock. Passing `failed_client=None` (as the tests' administrative reset does) keeps the unconditional behavior.

neuro_san/service/watcher/temp_networks/s3/aws_sync_client_worker.py
```diff
@@ -240,19 +285,26 @@ def do_with_retries(source: str, fn, *, max_attempts: int = 8, base_sleep: float
                     raise
 
                 sleep = S3Util.exponential_backoff_with_jitter(base_sleep, attempt)
-                sensitive_logger.warning("%s: Retryable sync ClientError (%s). attempt=%d", source, err, attempt)
+                _SENSITIVE_LOGGER.warning("%s: Retryable sync ClientError (%s). attempt=%d", source, err, attempt)
                 sync_sleep(sleep)
                 attempt += 1
+            except NoCredentialsError:
```
Noravee: Order matters: `NoCredentialsError` is a `BotoCoreError` subclass, so without this catch the generic branch below would sleep through ~16-48s of exponential backoff on an error that cannot heal mid-loop (the calling client's credential state is fixed), stalling every cache-miss request. The old design failed fast here; this preserves that.

neuro_san/service/watcher/temp_networks/s3/aws_async_client_worker.py
```diff
@@ -106,23 +157,45 @@ async def retry_with_new_client(self, work_function: Callable, *, source: str =
...
-                #             See: https://docs.aws.amazon.com/boto3/latest/guide/configuration.html
-
-                # Reset the cached credentials as they are likely expired and try again.
-                self.frozen_credentials = None
+                # Discard the session so the next attempt re-resolves the
+                # credential chain from scratch - but only if it is still
+                # the session this attempt failed with. A concurrent batch's
+                # retry may already have rebuilt a fresh session; discarding
+                # that would just force a redundant chain re-resolution.
+                if self.session is session:
```
Noravee: Same compare-and-swap idea as the sync worker's `reset_client(failed_client=...)`: deployment batches genuinely run concurrently on the writer's loop (one queue_processor task per queue), so an unconditional `self.session = None` here could discard a fresh session a peer's retry just rebuilt. `do_work_with_new_client()` takes the session as a parameter precisely so this handler knows which session the failed attempt actually used.

neuro_san/service/watcher/temp_networks/s3/s3_reservations_expiration.py
```diff
@@ -244,19 +248,28 @@ def expire_one_reservation(self, obj_key: str, current_time: float,
...
+                # expire_any_reservations() at the top of the sweep - and it can only
+                # react to ClientErrors that actually reach it. If ExpiredToken were
+                # merely logged like the codes below, every remaining key in the
+                # sweep would make one doomed S3 call, nothing would be expired, and
+                # the sweep would still report success. Re-raising lets the wrapper
+                # rebuild the session + client and re-run the sweep with working
+                # credentials. Restarting the sweep is safe because deletes are
                 # idempotent: NoSuchKey on the re-run is treated as success above.
+                #
+                # Caveat: this match only works for errors with a parseable body
```
Noravee: Honest limitation, documented rather than papered over: HEAD error responses have no body, so an expired-token `head_object` failure (only possible on the rare malformed-object path) surfaces as bare `"400"`, misses this re-raise, and gets logged by the else below. Recovery happens one key later via the next GET's parseable ExpiredToken. This behavior pre-exists on main; only the comments changed here. Widening the match to bare "400" would conflate genuine bad-request errors with credential expiry.

tests/neuro_san/service/watcher/temp_networks/s3/test_reader_client_reuse.py
```diff
@@ -0,0 +1,99 @@
...
+                # round-trip the object, or the client count means nothing.
+                self.assertIsNotNone(
+                    reservation,
+                    f"Expected read of {reservation_id} to succeed; reads must work "
+                    f"for the client-count assertion below to be meaningful.",
+                )
+                self.assertIsNotNone(agent_network)
+
+        self.assertEqual(
+            1, create_client_mock.call_count,
```
Noravee: The pinned invariant of concern 3: N reads, exactly one client construction. Originally written red against the per-call design (5 constructions). `_fresh_reader_client()` discards the client built in setUp so the construction being counted is provably caused by these reads - without that reset, `call_count` would be 0 and prove nothing.

tests/neuro_san/service/watcher/temp_networks/s3/test_proactive_credential_refresh.py
```diff
@@ -0,0 +1,185 @@
...
+            f"snapshot that bypassed the provider and paid a real ExpiredToken round "
+            f"trip before recovering - reactive refresh where botocore's "
+            f"keyless-client design gives proactive refresh for free.",
+        )
+
+        # Mechanism check: the proactive behavior above only holds because the
+        # worker creates its client WITHOUT explicit keys - passing any of
+        # these arguments makes botocore pin a static credential snapshot
+        # into the client and disables at-signing-time refresh.
+        explicit_key_args = {"aws_access_key_id", "aws_secret_access_key", "aws_session_token"}
```
Noravee: Mechanism check backing the behavioral assertion above: proactive refresh only works because the client is created with NO explicit credential arguments - passing any of these makes botocore pin a static snapshot and disables at-signing-time refresh. If someone reverts to explicit-key creation, this fails even if the behavioral half of the test is somehow fooled.

Noravee (commented):
*(Superseded — annotations refreshed in a consolidated review on 2026-08-07 after the credential-gate widening; see the latest review below.)*

neuro_san/service/watcher/temp_networks/s3/s3_util.py
```diff
@@ -97,6 +99,60 @@ def is_expired_token_error(err: ClientError) -> bool:
...
+                 SignatureDoesNotMatch. Those indicate rotated long-lived
+                 key pairs or genuine signing bugs - retrying would mask
+                 real misconfiguration (see the workers' retry docstrings).
+        """
+        if isinstance(err, (NoCredentialsError, PartialCredentialsError)):
+            return True
+        if not isinstance(err, ClientError):
+            return False
+        return S3Util.is_expired_token_error(err) \
+            or S3Util.get_error_code(err) in ("InvalidToken", "TokenRefreshRequired")
```
Noravee, line 154: **The whole credential policy, one predicate.** Both workers' retry loops and the expiration sweep's re-raise all classify through here, so gate changes are a single edit. Matches the three S3 temporary-token rejection codes plus the two local (BotoCoreError) resolution failures; deliberately excludes `InvalidAccessKeyId`/`SignatureDoesNotMatch` — those mean rotated key *pairs* or signing bugs, which re-resolution cannot fix and retrying would hide. The docstring records the #1310 evidence for `InvalidToken`.

neuro_san/service/watcher/temp_networks/s3/aws_sync_client_worker.py
```diff
@@ -58,178 +103,218 @@ def __init__(self, name: str, aws_service: str = "s3"):
...
-                         lock_aquired_time - start_time,
-                         client_created_time - start_time,
-                         lock_released_time - start_time,
-                         finish_time - start_time)
-        return retval
-
-    def get_frozen_credentials(self) -> ReadOnlyCredentials:
+        # Unlocked fast path: after first creation, this read is the whole
+        # cost of client acquisition on the read hot path.
+        local_client: BaseClient = self.sync_aws_client
```
Noravee, line 248: **The read hot path.** After first creation, client acquisition is this one unlocked attribute read — no lock, no credential resolution, no client construction. The previous design did all three per call, serialized under a worker-wide lock, plus a fresh TCP+TLS handshake per S3 request.

neuro_san/service/watcher/temp_networks/s3/aws_sync_client_worker.py
```diff
@@ -58,178 +103,218 @@ def __init__(self, name: str, aws_service: str = "s3"):
...
+                # mid-rewrite during rotation). A client pinned with None
+                # raises NoCredentialsError on every request, so caching it
+                # would wedge this worker until restart. Raising here instead
+                # leaves the cache empty; retry_with_new_client() treats it
+                # as a credential rejection (backoff + retry), and
+                # Session.get_credentials() re-resolves whenever its previous
+                # resolution came back None, so the worker self-heals as soon
+                # as credentials are available again.
+                if self.session.get_credentials() is None:
+                    raise NoCredentialsError()
```
Noravee, line 273: **The most important guard in the PR.** `create_client()` pins whatever the chain resolves *right now* — including `None` (credentials file mid-rewrite). A client pinned with `None` raises `NoCredentialsError` on every request forever, so caching it would wedge the worker until restart (verified empirically against botocore). Raising here leaves the cache empty; the retry loop above treats it as a credential rejection, and botocore re-resolves a `None` result on the next call, so the worker heals the moment credentials reappear.

neuro_san/service/watcher/temp_networks/s3/aws_sync_client_worker.py
```diff
@@ -58,178 +103,218 @@ def __init__(self, name: str, aws_service: str = "s3"):
...
-                # S3Util.is_expired_token_error() is used instead of raw
-                # DictionaryExtractor access: the extractor returns a stored
-                # None in preference to its default, and
-                # '"ExpiredToken" not in None' would raise TypeError inside
+            # NoCredentialsError/PartialCredentialsError are BotoCoreErrors,
+            # not ClientErrors - get_client()'s empty-chain guard raises the
+            # former - so an "except ClientError" alone would let the
+            # mid-rewrite phase of the very rotation window this loop exists
+            # to ride out escape with retry budget unused.
+            except (ClientError, NoCredentialsError, PartialCredentialsError) as err:
```
Noravee, line 186: **Why this catches more than `ClientError`.** `NoCredentialsError`/`PartialCredentialsError` are `BotoCoreError`s — the guard in `get_client()` below raises the former — so a `ClientError`-only gate would let the mid-rewrite phase of the exact rotation window this loop exists for escape to the caller with retry budget unused (a 500 on the request path). `sync_aws_client` is pre-bound to `None` above the `try` so this handler can never see an unbound local, and `None` doubles as "nothing was built, nothing to reset".

neuro_san/service/watcher/temp_networks/s3/aws_sync_client_worker.py
```diff
@@ -58,178 +103,218 @@ def __init__(self, name: str, aws_service: str = "s3"):
...
+                # as credentials are available again.
+                if self.session.get_credentials() is None:
+                    raise NoCredentialsError()
+
+                # No aws_access_key_id/aws_secret_access_key/aws_session_token
+                # arguments here: passing them would pin a static snapshot of
+                # the credentials into the client and disable auto-refresh.
+                self.sync_aws_client = self.session.create_client(
+                    self.aws_service,
+                    config=Config(max_pool_connections=self.MAX_POOL_CONNECTIONS))
```
Noravee, line 280: Without this, the shared client would keep botocore's default `max_pool_connections=10`; with urllib3's `block=False`, every request beyond the pool cap silently opens a fresh TCP+TLS connection and discards it on release — quietly reintroducing the per-call handshake cost under concurrent load. 50 sits comfortably above the service's achievable read concurrency (`asyncio.to_thread`'s default executor caps at min(32, cpu+4) threads).

neuro_san/service/watcher/temp_networks/s3/aws_sync_client_worker.py
```diff
@@ -58,178 +103,218 @@ def __init__(self, name: str, aws_service: str = "s3"):
...
-            raise NoCredentialsError()
-
-        # Avoid a small race condition with the ClientError retry block
-        # by always returning a local copy
-        local_frozen = credentials.get_frozen_credentials()
-        self.frozen_credentials = local_frozen
-
-        return local_frozen
+        with self.sync_aws_client_lock:
+            if failed_client is not None and self.sync_aws_client is not failed_client:
```
Noravee, line 307: **Compare-and-swap reset.** During a rotation, N threads that all failed on the *old* client would otherwise serially discard the fresh session+client the first thread's retry already rebuilt — N redundant credential-chain re-resolutions (possibly IMDS/SSO network calls) on a pod. Passing the failed client makes the reset a no-op unless the cache still holds exactly that client.

neuro_san/service/watcher/temp_networks/s3/aws_sync_client_worker.py
```diff
@@ -240,19 +325,29 @@ def do_with_retries(source: str, fn, *, max_attempts: int = 8, base_sleep: float
                     raise
 
                 sleep = S3Util.exponential_backoff_with_jitter(base_sleep, attempt)
-                sensitive_logger.warning("%s: Retryable sync ClientError (%s). attempt=%d", source, err, attempt)
+                _SENSITIVE_LOGGER.warning("%s: Retryable sync ClientError (%s). attempt=%d", source, err, attempt)
                 sync_sleep(sleep)
                 attempt += 1
+            except (NoCredentialsError, PartialCredentialsError):
```
Noravee, line 331: **Ordering matters twice here.** Both local credential errors are `BotoCoreError` subclasses, so this carve-out must precede the generic `BotoCoreError` branch or they'd be retried 8 times (~16–48s of backoff) against a client whose credential state cannot change. Fast-failing routes them out to `retry_with_new_client()` — the only loop that can heal them, because it discards the session. (`PartialCredentialsError` was added after a Copilot review note: it is a *sibling* of `NoCredentialsError`, not a subclass, so the original single-exception carve-out missed it.)

neuro_san/service/watcher/temp_networks/s3/aws_async_client_worker.py
```diff
@@ -68,61 +100,125 @@ def __init__(self, name: str, aws_service: str = "s3"):
...
-                # DictionaryExtractor access: the extractor returns a stored
-                # None in preference to its default, and
-                # '"ExpiredToken" not in None' would raise TypeError inside
+            # NoCredentialsError/PartialCredentialsError are BotoCoreErrors,
+            # not ClientErrors - a keyless client whose chain resolved empty
+            # raises the former at signing time (do_with_retries re-raises
+            # it immediately) - so an "except ClientError" alone would let
+            # the mid-rewrite phase of the very rotation window this loop
+            # exists to ride out escape with retry budget unused.
+            except (ClientError, NoCredentialsError, PartialCredentialsError) as err:
```
Noravee, line 164: **The write-path mirror.** Same union catch as the sync worker — here the stakes are a silently lost deployment batch: `NoCredentialsError` escaping this loop propagates to the updater, which logs and drops the item with no requeue, and other pods then report the just-created network not-found. The session CAS below (`if self.session is session`) is the async analog of the sync worker's `reset_client(failed_client=...)`: only the session this attempt actually failed with gets discarded.

neuro_san/service/watcher/temp_networks/s3/s3_reservations_expiration.py
```diff
@@ -243,20 +247,33 @@ def expire_one_reservation(self, obj_key: str, current_time: float,
...
+                # expire_any_reservations() at the top of the sweep - and it can only
+                # react to ClientErrors that actually reach it. If these errors were
+                # merely logged like the codes below, every remaining key in the
+                # sweep would make one doomed S3 call, nothing would be expired, and
+                # the sweep would still report success. Re-raising lets the wrapper
+                # rebuild the session + client and re-run the sweep with working
+                # credentials. Restarting the sweep is safe because deletes are
                 # idempotent: NoSuchKey on the re-run is treated as success above.
+                #
+                # Caveat: this match only works for errors with a parseable body
```
Noravee, line 271: **Honest limitation, documented rather than papered over.** HEAD error responses carry no body, so an expired-token rejection of `head_object` surfaces as the bare status `"400"`, misses the gate, and is logged by the else branch; the sweep recovers on the next GET/LIST, whose parseable body does reach this re-raise. Widening the gate to bare 400s was considered and rejected: a genuine bodyless 400 would then reset-and-rerun the whole sweep instead of skipping one key.

Noravee (commented):
**TL;DR — guided tour of the final design** *(consolidates and replaces the earlier annotation rounds; full rationale, prod evidence, and change log live in the PR description)*

* **One long-lived keyless client per sync worker.** Keyless means botocore keeps its credential OBJECT, and role/SSO/IRSA tokens refresh at signing time — token rotation costs **zero failed calls**, and the read hot path is one unlocked attribute read instead of per-call client construction under a lock (issue #1153, concerns 2+3).
* **Async writer:** keyless per-batch clients from a shared, lazily-created `AioSession` (aiobotocore clients are loop-bound context managers; the write path is per deployment batch, not per request).
* **One credential-rejection policy** (`S3Util.is_credential_rejection_error`): `ExpiredToken`/`ExpiredTokenException`, `InvalidToken` (the prod failure in cognizant-ai-lab/neuro-san-studio#1310), `TokenRefreshRequired`, plus the *local* resolution failures `NoCredentialsError`/`PartialCredentialsError` (BotoCoreErrors — a ClientError-only gate never sees them). `InvalidAccessKeyId`/`SignatureDoesNotMatch` deliberately surface: retrying would mask real misconfiguration.
* **Recovery = discard session + client, re-resolve the chain, retry**: 4 attempts, jittered backoff, compare-and-swap resets so concurrent threads don't destroy each other's rebuilds.
* **Tests: 30 passed** — every recovery behavior is pinned by a test that was first verified RED against the design (or gate) it replaces; suite is hermetic against host AWS config. pylint 10.00/10, flake8 clean.

The annotations below walk the load-bearing lines in reading order.

tests/neuro_san/service/watcher/temp_networks/s3/s3_reservations_storage_test_base.py
```diff
@@ -57,14 +60,37 @@ class S3ReservationsStorageTestBase(IsolatedAsyncioTestCase):
...
+            {
+                "AGENT_RESERVATIONS_EXTERNAL_STORAGE_CHECK_PERIOD_SECONDS": "0",
+                # Hermeticity barrier: the workers build REAL botocore /
+                # aiobotocore sessions (only create_client is patched below),
+                # and the sync worker consults the session's credential chain
+                # before creating its client. These fake env credentials make
+                # that resolution instant and deterministic instead of walking
+                # the developer/CI machine's real chain (config files, SSO,
+                # IMDS on EC2).
+                "AWS_ACCESS_KEY_ID": "testing",
```
Noravee, line 83: **Hermeticity barrier — a new requirement of this PR.** The workers now build REAL botocore/aiobotocore sessions (only `create_client` is patched), and the sync worker consults the credential chain before each client build, so resolution must be instant and deterministic. Fake env credentials satisfy the chain; the config-file redirects plus the `AWS_PROFILE` pops below close the remaining host leak — a bogus profile on a dev/CI machine raised `ProfileNotFound` during botocore's config load, before any provider ran, failing every test in setUp (reproduced end-to-end, then fixed).

tests/neuro_san/service/watcher/temp_networks/s3/test_reader_client_reuse.py
```diff
@@ -0,0 +1,99 @@
...
+                # Guard against a vacuous pass: every read must actually
+                # round-trip the object, or the client count means nothing.
+                self.assertIsNotNone(
+                    reservation,
+                    f"Expected read of {reservation_id} to succeed; reads must work "
+                    f"for the client-count assertion below to be meaningful.",
+                )
+                self.assertIsNotNone(agent_network)
+
+        self.assertEqual(
```
Noravee, line 91: **The two headline behaviors, pinned.** This test: N reads, exactly one client construction (concern 3). Its sibling `test_proactive_credential_refresh.py` pins concern 2: a token rotation between reads causes zero failed calls, with a mechanism check that `create_client` receives no explicit key/secret/token kwargs — the thing that would silently disable at-signing-time refresh. Both were written RED against the previous design.

tests/neuro_san/service/watcher/temp_networks/s3/test_writer_credential_recovery.py
```diff
@@ -0,0 +1,155 @@
...
+
+class TestWriterCredentialRecovery(S3ReservationsStorageTestBase):
+    """
+    Verifies that the async writer's credential retry discards the
+    shared session and completes the batch through a re-resolved chain,
+    for both a ClientError rejection (InvalidToken) and a local
+    resolution failure (NoCredentialsError).
+    """
+
+    async def _run_batch_with_bad_first_session(self, make_error) -> Dict[str, int]:
```
Noravee, line 62: **The recovery-test family.** Reader path: `test_invalid_token_recovery.py` (the #1310 failure mode) and `test_empty_chain_recovery.py` (chain resolves `None` mid-rewrite). Writer path (this file): a batch that fails against a bad first session must discard it, build a second one, and land the write — asserted via the session-creation counter, for both an `InvalidToken` rejection and a local `NoCredentialsError`. Each was verified red against the gate it hardens.

vince-leaf (dismissed):
Non-blocking, for a separate issue: even with this fix, S3ReservationsReader.get_one_reservation() still catches the ClientError and returns (None, None), which ExpiringAgentNetworkStorage.get_agent_network_provider() cannot distinguish from "no such network" — so any future credential/backend failure still reaches the user as a 404 that looks like data loss (exactly how #1310 presented). Surfacing it as a 503 instead would make the next occurrence obvious. Not for this PR. 


LGTM.

d1donlydfink (approved):
Nice shoring up. You are seeing things I never did.
Still let's keep an eye on this after deploying for the first time.
The S3 writes in particular are on a load-test performance path.

andreidenissov-cog:
Verified this implementation with load-150 test against 1 and 5 neuro-san instances (Boombox bask-end).
Runs OK.

andreidenissov-cog (approved):
Great refactoring.

PR #1159, opened 2026-07-22
Profiling locks

d1donlydfink:
* Add classes to profile acquisition and release of sync and asynchronous locks.
* Standardize how we import Lock and asyncio.Lock so they are easier to find by grepping.

These ProfilingLocks are not actually hooked up to anything permanently.
I have used them and observed that we are generally not waiting for locks at all in neuro_san.internals.
I can say that AND's GetMcpTool does _hold_ its request-scoped lock for > 8 seconds (in 20 request run), but it does not take anything to actually acquire.

Note that there are locks deeper down in the system I have not yet tried to replace.

### DIFF
--- a/neuro_san/internals/network_providers/agent_network_storage.py
+++ b/neuro_san/internals/network_providers/agent_network_storage.py
@@ -19,7 +19,7 @@ from typing import Dict
 from typing import List
 
 import logging
-import threading
+from threading import Lock
 
 from neuro_san.internals.graph.registry.agent_network import AgentNetwork
 from neuro_san.internals.interfaces.agent_network_provider import AgentNetworkProvider
@@ -39,7 +39,7 @@ class AgentNetworkStorage(AgentStorageSource):
     def __init__(self):
         self.agents_table: Dict[str, AgentNetwork] = {}
         self.logger = logging.getLogger(self.__class__.__name__)
-        self.lock = threading.Lock()
+        self.lock = Lock()
         self.listeners: List[AgentStateListener] = []
 
     def add_listener(self, listener: AgentStateListener):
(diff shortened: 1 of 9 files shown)


neuro_san/internals/network_providers/agent_network_storage.py
```diff
@@ -19,7 +19,7 @@
 from typing import List
 
 import logging
-import threading
+from threading import Lock
```
d1donlydfink, line 22: Be consistent about how we import sync Lock and async Lock.
This is so we can easily find them by grepping.

neuro_san/internals/utils/async_profiling_lock.py
```diff
@@ -0,0 +1,113 @@
...
+from logging import getLogger
+from logging import Logger
+from os import environ
+from time import perf_counter
+
+
+class AsyncProfilingLock:
+    """
+    Class that adds some profiling information to an asynchronous lock.
+    """
```
d1donlydfink, line 33: Profiling for async locks.

neuro_san/internals/utils/profiling_lock.py
```diff
@@ -0,0 +1,113 @@
...
+from logging import Logger
+from os import environ
+from threading import Lock
+from time import perf_counter
+
+
+class ProfilingLock:
+    """
+    Class that adds some profiling information to a synchronous lock.
+    """
```
d1donlydfink, line 33: Profiling for synchronous locks.

Noravee (approved):
LGTM.

andreidenissov-cog (approved):
LGTM.

PR #1162, opened 2026-07-23
fix: resolve manifest HOCON includes relative to the manifest file's directory

rstrickland-ohpen:
## Summary

- **`neuro_san/internals/graph/persistence/raw_manifest_restorer.py`**: Override `deserialize_file_contents()` to pass `basedir=<manifest directory>` when parsing manifest HOCON files. Previously, `parse_string()` resolved `include` paths relative to the process CWD, so sub-manifests (e.g. `include "experimental/manifest.hocon"` inside `registries/manifest.hocon`) silently failed to load when the server was started from a user project directory rather than the package root.

With this fix, `include` paths in manifest files are resolved relative to the manifest file itself — matching the intuitive expectation and making multi-directory deployments work correctly. Agent HOCON files are unaffected; they continue to resolve includes from the CWD as before (which is intentional — it allows user project configs to override package defaults).

Depends on the companion fix in `leaf-common` (adds `basedir` parameter to `HoconSerializationFormat.to_object()`).

## Test plan

- [ ] Start the server from a directory other than the package root with `AGENT_MANIFEST_FILE` pointing to a manifest that includes sub-manifests (e.g. `include "experimental/manifest.hocon"`); verify agents from the sub-manifest are loaded without "Cannot include file" warnings
- [ ] Verify agent HOCON files that use `include "config/llm_config.hocon"` still resolve correctly from the user project CWD

### DIFF
--- a/neuro_san/internals/graph/persistence/raw_manifest_restorer.py
+++ b/neuro_san/internals/graph/persistence/raw_manifest_restorer.py
@@ -15,6 +15,12 @@
 #
 # END COPYRIGHT
 
+import os
+from io import BytesIO
+from typing import Any, Dict
+
+from leaf_common.serialization.format.hocon_serialization_format import HoconSerializationFormat
+
 from neuro_san.internals.persistence.abstract_async_config_restorer import AbstractAsyncConfigRestorer
 
 
@@ -30,3 +36,18 @@ class RawManifestRestorer(AbstractAsyncConfigRestorer):
         Constructor
         """
         super().__init__(file_purpose="agent network manifest", env_var="AGENT_MANIFEST_FILE", must_exist=False)
+
+    def deserialize_file_contents(self, file_path: str, file_contents: bytes) -> Dict[str, Any]:
+        """
+        Override to resolve HOCON include paths relative to the manifest file's own directory.
+        This ensures that sub-manifest includes (e.g. include "experimental/manifest.hocon")
+        work regardless of the server's working directory.
+        """
+        if not file_path.endswith(".hocon"):
+            return super().deserialize_file_contents(file_path, file_contents)
+
+        basedir = os.path.dirname(os.path.abspath(file_path))
+        bytes_file = BytesIO(file_contents)
+        serialization = HoconSerializationFormat()
+        config = serialization.to_object(bytes_file, basedir=basedir)
+        return self.filter_config(config, file_path)
--- a/neuro_san/internals/graph/persistence/raw_manifest_restorer.py
+++ b/neuro_san/internals/graph/persistence/raw_manifest_restorer.py
@@ -15,11 +15,10 @@
 #
 # END COPYRIGHT
 
-import os
-from io import BytesIO
+import json
 from typing import Any, Dict
 
-from leaf_common.serialization.format.hocon_serialization_format import HoconSerializationFormat
+from pyhocon import ConfigFactory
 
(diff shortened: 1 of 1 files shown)


neuro_san/internals/graph/persistence/raw_manifest_restorer.py
```diff
@@ -30,3 +36,18 @@ def __init__(self):
...
+        This ensures that sub-manifest includes (e.g. include "experimental/manifest.hocon")
+        work regardless of the server's working directory.
+        """
+        if not file_path.endswith(".hocon"):
+            return super().deserialize_file_contents(file_path, file_contents)
+
+        basedir = os.path.dirname(os.path.abspath(file_path))
+        bytes_file = BytesIO(file_contents)
+        serialization = HoconSerializationFormat()
+        config = serialization.to_object(bytes_file, basedir=basedir)
```
d1donlydfink: This will perhaps break backwards compatibility with existing installations.
How to mitigate against this?

rstrickland-ohpen:
@d1donlydfink Good catch. The original approach (passing `basedir` to `parse_string()`) would have been a breaking change for any existing manifest that used CWD-relative include paths (e.g. `include "registries/experimental/manifest.hocon"` written for a server started from the project root).

I've updated the branch to use `ConfigFactory.parse_file()` instead. This is fully backwards-compatible: `parse_file()` has always resolved includes relative to the file being parsed (that's its native behaviour), so no existing include paths need to change. It also eliminates the dependency on a `leaf-common` change.

d1donlydfink:
@donn-leaf suggests closing this PR then re-opening it to get tests to run.
Going to try that.

neuro_san/internals/graph/persistence/raw_manifest_restorer.py
```diff
@@ -30,3 +35,22 @@ def __init__(self):
...
+        Agent HOCON files (not manifests) are unaffected — they continue to go through
+        the base class which uses parse_string() and CWD-relative resolution.
+        """
+        if not file_path.endswith(".hocon"):
+            return super().deserialize_file_contents(file_path, file_contents)
+
+        config = ConfigFactory.parse_file(file_path)
+        if config is not None:
+            config = json.loads(json.dumps(config))
+        return self.filter_config(config, file_path)
```
d1donlydfink, line 56: This still isn't working for me.

Here's what I'd like you to do:

1) Try running a neuro-san server from the top of the neuro-san-studio without this change in the mix.
```
neuro-san-studio % export AGENT_MANIFEST_FILE=<your-path>/neuro-san-studio/registries/manifest.hocon

neuro-san-studio % python -m neuro_san.service.main_loop.server_main_loop
```
Then in another shell, use the agent_cli tool to list the networks that are visible to count them:
```
neuro-san-studio % python -m neuro_san.client.agent_cli --http --list | grep agent_name | wc -l
52
```
You'll see about 52 agents.

2) Now try adding your change into the mix and do the same thing.
The agent count goes down to 2.

This is not what I would call backwards compatible.


It's worth noting that the same exercise within the neuro-san repo yields the same answer for both: 32.
But I use neuro-san-studio as a test case because it's more complicated, and that is the other repo that we have to keep working for everyone.

d1donlydfink (changes requested):
See comments about agent count from neuro-san-studio with and without the change

d1donlydfink:
Closing due to inactivity

PR #1163, opened 2026-07-23
No max_connection limits on httpx connection pools for OpenAI

d1donlydfink:
This is not really a limit on smaller loads, but can be a limit on larger loads.
For AND, we have an average of about 18 calls per request.

### DIFF
--- a/neuro_san/internals/run_context/langchain/llms/openai_llm_policy.py
+++ b/neuro_san/internals/run_context/langchain/llms/openai_llm_policy.py
@@ -19,6 +19,7 @@ from typing import Dict
 
 from contextlib import suppress
 from httpx import AsyncClient
+from httpx import Limits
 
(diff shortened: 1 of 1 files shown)


neuro_san/internals/run_context/langchain/llms/openai_llm_policy.py
```diff
@@ -36,6 +37,9 @@ class OpenAILlmPolicy(LlmPolicy):
     the llm reference, because of our create_client() implementation, we do not.
     """
 
+    # None means no limit
+    LIMITS: Limits = Limits(max_connections=None, max_keepalive_connections=None)
```
d1donlydfink, line 41: Have no limits on connection pools.

Noravee (approved):
LGTM.

I wonder if we need to concern about resource exhaustion with this.

PR #1167, opened 2026-07-24
Reuse the service's loaded toolbox factory for connectivity reporting

### LINKED ISSUE #1166: Connectivity requests rebuild and re-read toolbox info hocon files on every call
## Problem

Every `connectivity()` request constructs a fresh `ConnectivityReporter`, which builds a brand-new `ToolboxFactory` from the agent network config and `load()`s it — re-reading and re-parsing the bundled `toolbox_info.hocon` (plus any user `toolbox_info_file` / `AGENT_TOOLBOX_INFO_FILE` overlay) from disk on every single request:

- `neuro_san/session/direct_agent_session.py` and `neuro_san/session/async_direct_agent_session.py` construct `ConnectivityReporter(self.agent_network)` bare, so the reporter falls into its self-build fallback each time.
- Meanwhile `AgentService` / `AsyncAgentService` already build and `load()` a toolbox factory **once** at service construction and use it for chat — but their `connectivity()` paths create the session with `invocation_context=None`, so that warm factory never reaches the reporter.
- On the async HTTP path, this per-request file read + hocon parse runs synchronously **inside the event loop** (`AsyncDirectAgentSession.connectivity()` calls the reporter directly from `async def`), blocking other requests.

Under load testing this shows up as unnecessary per-request file descriptor and latency pressure on the Connectivity endpoint, for data that is loaded once and immutable at runtime in practice.

## What should happen

- The `ConnectivityReporter` should be able to reuse an already-loaded toolbox factory instead of always building its own (with the existing self-build as fallback).
- The services' cached factory should be threaded through `DirectAgentSession` / `AsyncDirectAgentSession` to the reporter; library sessions should reuse the factory already carried by their `SessionInvocationContext`.
- `ToolboxFactory.load()` should be a cheap no-op after the first load, and safe to call on a factory shared across request threads.

## Expected result

Zero toolbox-info file I/O per connectivity request on server and library paths; the only load happens at service/session-factory startup.

Noravee:
Fixes #1166

## Problem

Every `connectivity()` request built a fresh `ToolboxFactory` and re-read the toolbox info hocon files from disk, even though `AgentService` / `AsyncAgentService` already build and load a factory once at startup for chat. On the async HTTP path the per-request file read + parse ran synchronously inside the event loop. See #1166 for the full analysis.

## Changes

**Wiring (zero file I/O per connectivity request)**
- `DirectAgentSession` and `AsyncDirectAgentSession` accept an optional, keyword-only `toolbox_factory`. It is resolved once in the constructor: explicit param → `invocation_context.get_toolbox_factory()` → `None` (in which case `ConnectivityReporter` keeps its existing self-build fallback).
- `AgentService.connectivity()` and `AsyncAgentService.connectivity()` pass their already-loaded `self.toolbox_factory` into the session. Library sessions created by `DirectAgentSessionFactory` reuse the factory their invocation context already carries.
- `ConnectivityReporter` accepts the injected factory and logs when it has to self-build one, so any future call site that forgets the parameter is observable instead of silently regressing to per-request reads.

**Thread-safety / hardening**
- `ToolboxFactory.load()` is now load-once and thread-safe: double-checked `Lock` (same idiom as `LangChainMcpAdapter`), and the assembled infos are published in a single assignment so lock-free readers never observe a partially-overlaid dict.
- `AsyncAgentService.reload_factories()` loads factories into locals before assigning the fields request paths read, so a request can never see an unloaded factory.

**Tests**
- New `tests/neuro_san/session/` suite covers the actual wiring: injected-factory and invocation-context-fallback paths in both session twins (sync and async).
- New reporter test proves the injected factory's data is consulted (a distinctive `display_as` value appears in the report) rather than merely stored.
- `ToolboxFactory` test asserts a second `load()` performs no additional file reads.
- Connectivity reporter tests are hermetic against a set `AGENT_TOOLBOX_INFO_FILE` env var (including the pre-existing `test_hello_world`, which had the same exposure through the reporter's self-build path).

## Behavior notes

- Connectivity now uses the same construction-time factory as chat. After an agent-network hot-reload, the async provider rebuilds the service (and factory) via `reset_service()`, so reports stay consistent; in-flight requests during the swap window may briefly report with the prior factory, matching existing chat semantics.
- Sessions created by `ExternalAgentSessionFactory` carry the calling network's invocation context; nothing calls `connectivity()` on those sessions today, but if cross-network connectivity reporting is ever added, the factory-vs-network pairing there deserves a look.

## Testing

- `pytest tests/neuro_san/internals tests/neuro_san/session tests/neuro_san/service tests/neuro_san/client tests/neuro_san/middleware` — 369 passed.
- Affected tests also pass with a hostile `AGENT_TOOLBOX_INFO_FILE=/nonexistent/path.hocon` exported, verifying hermeticity.

### DIFF
--- a/neuro_san/internals/chat/connectivity_reporter.py
+++ b/neuro_san/internals/chat/connectivity_reporter.py
@@ -60,6 +60,12 @@ class ConnectivityReporter:
         Constructor
 
         :param inspector: The AgentNetworkInspector to use.
+        :param toolbox_factory: An optional ContextTypeToolboxFactory to consult
+                        for display_as information about toolbox tools.
+                        This should have been built from the same config as the
+                        inspector's agent network and may be passed pre-loaded
+                        so toolbox info files are not re-read per report.
+                        If None, one is created from the inspector's config.
         """
 
         self.inspector: AgentNetworkInspector = inspector
--- a/neuro_san/internals/chat/connectivity_reporter.py
+++ b/neuro_san/internals/chat/connectivity_reporter.py
@@ -19,6 +19,8 @@ from typing import Dict
 from typing import List
 from typing import Set
 
+import logging
+
 from leaf_common.parsers.dictionary_extractor import DictionaryExtractor
 
 from neuro_san.internals.interfaces.context_type_toolbox_factory import ContextTypeToolboxFactory
@@ -72,6 +74,9 @@ class ConnectivityReporter:
         self.toolbox_factory: ContextTypeToolboxFactory = toolbox_factory
 
         if self.inspector is not None and self.toolbox_factory is None:
+            logger = logging.getLogger(self.__class__.__name__)
+            logger.info("No toolbox_factory provided. Building one from the agent network config. "
+                        "Pass a pre-loaded factory to avoid re-reading toolbox info files per report.")
             config: Dict[str, Any] = self.inspector.get_config()
             self.toolbox_factory = MasterToolboxFactory.create_toolbox_factory(config)
 
--- a/neuro_san/internals/chat/connectivity_reporter.py
+++ b/neuro_san/internals/chat/connectivity_reporter.py
@@ -57,6 +57,10 @@ class ConnectivityReporter:
         Maybe someday.
     """
 
+    # Set once the self-build fallback below has been logged, so per-request
+    # reporter construction cannot flood the logs.
+    self_build_logged: bool = False
+
     def __init__(self, inspector: AgentNetworkInspector, toolbox_factory: ContextTypeToolboxFactory = None):
         """
         Constructor
@@ -74,9 +78,12 @@ class ConnectivityReporter:
         self.toolbox_factory: ContextTypeToolboxFactory = toolbox_factory
 
         if self.inspector is not None and self.toolbox_factory is None:
-            logger = logging.getLogger(self.__class__.__name__)
-            logger.info("No toolbox_factory provided. Building one from the agent network config. "
-                        "Pass a pre-loaded factory to avoid re-reading toolbox info files per report.")
+            if not ConnectivityReporter.self_build_logged:
+                ConnectivityReporter.self_build_logged = True
+                logger = logging.getLogger(self.__class__.__name__)
+                logger.info("No toolbox_factory provided. Building one from the agent network config. "
+                            "Pass a pre-loaded factory to avoid re-reading toolbox info files per report. "
+                            "(Logged once per process.)")
             config: Dict[str, Any] = self.inspector.get_config()
             self.toolbox_factory = MasterToolboxFactory.create_toolbox_factory(config)
 
--- a/neuro_san/internals/run_context/langchain/toolbox/toolbox_factory.py
+++ b/neuro_san/internals/run_context/langchain/toolbox/toolbox_factory.py
@@ -120,6 +120,9 @@ class ToolboxFactory(ContextTypeToolboxFactory):
         if self.loaded:
             return
 
+        # Double-checked locking: the test above keeps the per-report load()
+        # call cheap once loaded; the re-test under the lock makes the first
+        # load exclusive when a not-yet-loaded factory is shared by threads.
         with self.load_lock:
             if self.loaded:
                 return
--- a/neuro_san/internals/run_context/langchain/toolbox/toolbox_factory.py
+++ b/neuro_san/internals/run_context/langchain/toolbox/toolbox_factory.py
@@ -109,6 +109,10 @@ class ToolboxFactory(ContextTypeToolboxFactory):
     def load(self):
         """
         Loads the base tool information from hocon files.
+
+        Only the first call does any work.  Subsequent calls on the same
+        instance are no-ops, so a shared factory can be load()-ed cheaply
+        on every use.
         """
         if self.loaded:
             return
--- a/neuro_san/internals/run_context/langchain/toolbox/toolbox_factory.py
+++ b/neuro_san/internals/run_context/langchain/toolbox/toolbox_factory.py
@@ -27,6 +27,8 @@ from typing import Union
 import logging
 import os
 
+from threading import Lock
+
 from langchain_core.tools.base import BaseTool
 from langchain_core.tools.base import BaseToolkit
 from pydantic import BaseModel
@@ -85,6 +87,7 @@ class ToolboxFactory(ContextTypeToolboxFactory):
         self.toolbox_infos: Dict[str, Any] = {}
         self.overlayer = DictionaryOverlay()
         self.loaded: bool = False
+        self.load_lock: Lock = Lock()
 
         # Get user toolbox info file path with the following priority:
         # 1. "toolbox_info_file" from agent network hocon
@@ -112,20 +115,27 @@ class ToolboxFactory(ContextTypeToolboxFactory):
 
         Only the first call does any work.  Subsequent calls on the same
         instance are no-ops, so a shared factory can be load()-ed cheaply
-        on every use.
+        on every use, even from multiple threads.
         """
         if self.loaded:
             return
 
-        restorer = ToolboxInfoRestorer()
-        self.toolbox_infos = restorer.restore()
+        with self.load_lock:
+            if self.loaded:
+                return
 
-        # Mix in user-specified toolbox info, if available.
-        if self.toolbox_info_file:
-            extra_toolbox_infos: Dict[str, Any] = restorer.restore(file_reference=self.toolbox_info_file)
-            self.toolbox_infos = self.overlayer.overlay(self.toolbox_infos, extra_toolbox_infos)
+            restorer = ToolboxInfoRestorer()
+            toolbox_infos: Dict[str, Any] = restorer.restore()
 
-        self.loaded = True
+            # Mix in user-specified toolbox info, if available.
+            if self.toolbox_info_file:
+                extra_toolbox_infos: Dict[str, Any] = restorer.restore(file_reference=self.toolbox_info_file)
+                toolbox_infos = self.overlayer.overlay(toolbox_infos, extra_toolbox_infos)
+
+            # Publish the fully-assembled infos in a single assignment so that
+            # lock-free readers never observe a partially-overlaid dictionary.
+            self.toolbox_infos = toolbox_infos
+            self.loaded = True
 
     def create_tool_from_toolbox(
             self,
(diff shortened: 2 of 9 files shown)


neuro_san/session/direct_agent_session.py
```diff
@@ -48,7 +49,9 @@ def __init__(self,
                  invocation_context: SessionInvocationContext,
```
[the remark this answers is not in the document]
Noravee: Good catch — annotated as `Optional[SessionInvocationContext]` in a3070b9c. Kept the parameter required (no default): chat methods genuinely need a real context, so passing `None` should remain a deliberate choice for connectivity()/function()-only sessions (which is how the services have always called it). Also documented the None-allowance in the docstring.

neuro_san/session/async_direct_agent_session.py
```diff
@@ -47,7 +48,9 @@ def __init__(self,
                  agent_network: AgentNetwork,
                  invocation_context: SessionInvocationContext,
```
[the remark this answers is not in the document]
Noravee: Same fix as the sync twin — annotated `Optional[SessionInvocationContext]` in a3070b9c, kept required with the None-allowance documented in the docstring.

neuro_san/internals/chat/connectivity_reporter.py
```diff
@@ -60,12 +62,21 @@ def __init__(self, inspector: AgentNetworkInspector, toolbox_factory: ContextTyp
...
 
         self.inspector: AgentNetworkInspector = inspector
         self.toolbox_factory: ContextTypeToolboxFactory = toolbox_factory
 
         if self.inspector is not None and self.toolbox_factory is None:
+            logger = logging.getLogger(self.__class__.__name__)
+            logger.info("No toolbox_factory provided. Building one from the agent network config. "
+                        "Pass a pre-loaded factory to avoid re-reading toolbox info files per report.")
             config: Dict[str, Any] = self.inspector.get_config()
             self.toolbox_factory = MasterToolboxFactory.create_toolbox_factory(config)
```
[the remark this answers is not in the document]
Noravee: Agreed on the flood risk. Changed in a3070b9c to log once per process via a class-level flag, keeping INFO level so the unwired-call-site signal still shows up in production logs (DEBUG would usually be filtered out, which defeats the purpose of the log). The message now notes it is logged once per process.

neuro_san/session/direct_agent_session.py
```diff
@@ -72,6 +88,13 @@ def __init__(self,
         self.agent_network: AgentNetwork = agent_network
         self.request_id: str = None
         self.umbrella_timeout: Timeout = umbrella_timeout
+        # Resolve the toolbox factory once at construction: an invocation
+        # context's factory is fixed when the context is built, and resolving
+        # here keeps connectivity() working even after close() sets
+        # self.invocation_context to None.
+        self.toolbox_factory: ContextTypeToolboxFactory = toolbox_factory
```
Noravee, line 95: **Core wiring (sync).** Resolution order: explicit param → invocation context's factory → None (in which case ConnectivityReporter self-builds from the network config, preserving pre-PR behavior for callers that pass nothing). Resolved at construction because a context's factory is fixed when the context is built, and this keeps connectivity() working after close() nulls invocation_context. The async twin mirrors this block exactly.

neuro_san/service/generic/agent_service.py
```diff
@@ -206,10 +206,14 @@ def connectivity(self, request_dict: Dict[str, Any],
...
         agent_network: AgentNetwork = self.agent_network_provider.get_agent_network()
+        # Pass the toolbox factory that was created and loaded once at service
+        # construction, so connectivity reporting does not re-read toolbox
+        # info files on every request.
         session = DirectAgentSession(agent_network=agent_network,
                                      invocation_context=None,
                                      metadata=metadata,
-                                     security_cfg=self.security_cfg)
+                                     security_cfg=self.security_cfg,
+                                     toolbox_factory=self.toolbox_factory)
```
Noravee, line 216: This line is the sync-path fix for #1166: the factory built and loaded once in `AgentService.__init__` now reaches ConnectivityReporter, so connectivity() stops re-reading toolbox hocon files per request. Services are per-network, so no factory is ever shared across different networks' configs.

neuro_san/service/generic/async_agent_service.py
```diff
@@ -211,12 +215,16 @@ async def connectivity(self, request_dict: Dict[str, Any],
...
+        # construction, so connectivity reporting does not re-read toolbox
+        # info files on every request.
         session: AsyncDirectAgentSession =\
             AsyncDirectAgentSession(
                 agent_network=agent_network,
                 invocation_context=None,
                 metadata=metadata,
-                security_cfg=self.security_cfg)
+                security_cfg=self.security_cfg,
+                toolbox_factory=self.toolbox_factory)
```
Noravee, line 227: Async-path fix for #1166. Besides skipping the per-request re-read, this removes blocking file I/O from the event loop: `report_network_connectivity()`'s `load()` call is now a no-op on the pre-loaded factory.

neuro_san/service/generic/async_agent_service.py
```diff
@@ -110,12 +110,16 @@ def reload_factories(self):
...
 
-        # Load once.
-        self.llm_factory.load()
-        self.toolbox_factory.load()
+        # Load once, before publishing to the fields that request paths read,
+        # so no request can ever see an unloaded factory.
+        llm_factory.load()
+        toolbox_factory.load()
+
+        self.llm_factory: ContextTypeLlmFactory = llm_factory
```
Noravee, line 121: Defensive ordering with no behavior change today: `reload_factories()` is only called from `__init__` (before the provider publishes the service), but its docstring advertises runtime calls and `connectivity()` now reads `self.toolbox_factory` per request. Loading into locals and assigning last means a future runtime caller can never expose an unloaded factory to in-flight requests.

neuro_san/internals/run_context/langchain/toolbox/toolbox_factory.py
```diff
@@ -109,19 +112,33 @@ def __init__(self, config: Optional[Dict[str, Any]] = None):
...
-        # Mix in user-specified toolbox info, if available.
-        if self.toolbox_info_file:
-            extra_toolbox_infos: Dict[str, Any] = restorer.restore(file_reference=self.toolbox_info_file)
-            self.toolbox_infos = self.overlayer.overlay(self.toolbox_infos, extra_toolbox_infos)
-
-        self.loaded = True
+        # Double-checked locking: the test above keeps the per-report load()
+        # call cheap once loaded; the re-test under the lock makes the first
+        # load exclusive when a not-yet-loaded factory is shared by threads.
+        with self.load_lock:
```
Noravee, line 126: **This is the one real semantic change in the PR** (the rest is plumbing): `load()` is now load-once and thread-safe. Double-checked lock (same idiom as `LangChainMcpAdapter`), and the assembled infos are published in a single assignment so lock-free readers never observe a partially-overlaid dict. Trade-off to be aware of: a factory instance no longer picks up on-disk toolbox-file edits during its lifetime — fresh data arrives whenever a factory is rebuilt, which covers every in-tree refresh path (service reset on hot-reload, new session from the session factory, bare reporter construction).

neuro_san/internals/chat/connectivity_reporter.py
```diff
@@ -55,17 +57,33 @@ class ConnectivityReporter:
...
+                        inspector's agent network and may be passed pre-loaded
+                        so toolbox info files are not re-read per report.
+                        If None, one is created from the inspector's config.
         """
 
         self.inspector: AgentNetworkInspector = inspector
         self.toolbox_factory: ContextTypeToolboxFactory = toolbox_factory
 
         if self.inspector is not None and self.toolbox_factory is None:
+            if not ConnectivityReporter.self_build_logged:
```
Noravee, line 81: Observability guard: if a future call site forgets to pass a factory, this logs (once per process, at INFO) instead of silently regressing to per-request file reads — a regression that would otherwise only resurface as latency/FD pressure under load.

tests/neuro_san/session/test_direct_agent_session_connectivity.py
```diff
@@ -0,0 +1,120 @@
...
+
+from neuro_san import REGISTRIES_DIR
+from neuro_san.internals.graph.persistence.agent_network_restorer import AgentNetworkRestorer
+from neuro_san.internals.graph.registry.agent_network import AgentNetwork
+from neuro_san.internals.run_context.langchain.toolbox.toolbox_factory import ToolboxFactory
+from neuro_san.session.async_direct_agent_session import AsyncDirectAgentSession
+from neuro_san.session.direct_agent_session import DirectAgentSession
+
+
+class TestDirectAgentSessionConnectivity(TestCase):
```
Noravee, line 32: This suite pins the wiring the PR adds — injected-factory and invocation-context-fallback paths in both session twins. Without it, reverting the session/service wiring would leave the whole test suite green while the per-request reads quietly return.

tests/neuro_san/internals/chat/test_connectivity_reporter.py
```diff
@@ -77,6 +85,50 @@ def test_hello_world(self):
...
+        # Keep the test hermetic: no user toolbox info file from the environment.
+        toolbox_factory.toolbox_info_file = None
+        reporter = ConnectivityReporter(agent_network, toolbox_factory)
+        self.assertIs(reporter.toolbox_factory, toolbox_factory)
+
+        messages: List[Dict[str, Any]] = reporter.report_network_connectivity()
+        self.assertEqual(len(messages), 2)
+        self.assertTrue(toolbox_factory.loaded)
+
+    def test_injected_toolbox_factory_data_is_consulted(self):
```
Noravee, line 105: This test proves the injected factory's *data* is consulted (the distinctive display_as shows up in the report), not merely stored and loaded. hello_world.hocon has no toolbox agents, so the sibling test alone cannot catch a regression where the reporter silently consults a self-built factory instead.

Noravee (commented):
Inline comments below mark the load-bearing hunks and the one real semantic change. Known non-goals (details in the PR description): the external-session factory pairing (nothing calls connectivity() on those sessions today) and the pre-existing shared tool_info mutation, which is tracked separately.

neuro_san/service/generic/agent_service.py
d1donlydfink: This is a nice find.

d1donlydfink (approved):

PR #1169, opened 2026-07-26
Only do structure parsing on AGENT_FRAMEWORK messages that could be the answer

d1donlydfink:
Not doing structure parsing on every message saves us a bit of time.

### DIFF
--- a/neuro_san/message_processing/structure_message_processor.py
+++ b/neuro_san/message_processing/structure_message_processor.py
@@ -63,6 +63,10 @@ class StructureMessageProcessor(MessageProcessor):
         :param chat_message_dict: The ChatMessage dictionary to process.
         :param message_type: The ChatMessageType of the chat_message_dictionary to process.
         """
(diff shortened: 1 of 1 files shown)


neuro_san/message_processing/structure_message_processor.py
```diff
@@ -63,6 +63,10 @@ def process_message(self, chat_message_dict: Dict[str, Any], message_type: ChatM
         :param chat_message_dict: The ChatMessage dictionary to process.
         :param message_type: The ChatMessageType of the chat_message_dictionary to process.
         """
+        if message_type != ChatMessageType.AGENT_FRAMEWORK:
+            # No formats to look for
+            return
```
d1donlydfink, line 68: Here: We return early when it's not an AGENT_FRAMEWORK message.
We do the structure parsing work below only when it's an AGENT_FRAMEWORK message.

Noravee (approved):
lgtm

PR #1170, opened 2026-07-26
Get rid of the QueueFilter task

d1donlydfink:
Filtering is now done before anything gets thrown onto the hopper in MessageJournal.
And now we don't need 2 queues.
The MessageJournal queue itself now contains client-ready message dicts instead of need-to-be-processed BaseMessages.

Also: Use the latest leaf-common==1.2.49 for eager task factory.

### DIFF
--- a/neuro_san/internals/chat/queue_filter.py
+++ b/neuro_san/internals/chat/queue_filter.py
@@ -38,17 +38,13 @@ class QueueFilter:
     We want the work done in filter_queue() to be done off the main server thread.
     """
 
-    def __init__(self, invocation_context: InvocationContext,
-                 chat_filter: Dict[str, Any],
-                 agent_network: AgentNetwork):
+    def __init__(self, chat_filter: Dict[str, Any], agent_network: AgentNetwork):
         """
         Constructor
 
-        :param invocation_context: An instance of InvocationContext
         :param chat_filter: A dictionary form of chat.ChatFilter
         :param agent_network: An instance of AgentNetwork
         """
-        self.invocation_context: InvocationContext = invocation_context
         self.message_filter: MessageFilter = MessageFilterFactory.create_message_filter(chat_filter)
         self.message_processor: MessageProcessor = self.create_outgoing_message_processor(agent_network)
 
@@ -74,10 +70,10 @@ class QueueFilter:
         message_processor = StructureMessageProcessor(structure_formats)
         return message_processor
 
-    def apply_to_journal(self):
+    def apply_to_journal(self, invocation_context: InvocationContext):
         """
         Apply the filter and processor to the message journal
         """
-        message_journal: MessageJournal = self.invocation_context.get_journal()
+        message_journal: MessageJournal = invocation_context.get_journal()
         message_journal.set_message_filter(self.message_filter)
         message_journal.set_message_processor(self.message_processor)
--- a/neuro_san/internals/chat/queue_filter.py
+++ b/neuro_san/internals/chat/queue_filter.py
@@ -31,11 +31,9 @@ from neuro_san.message_processing.structure_message_processor import StructureMe
 
 class QueueFilter:
     """
-    Filters messages from the input queue (the one from the MessageJournal)
-    to be sure they are suitable to be sent to the client on the output queue.
-
-    The main entrypoint to this class is filter_queue().
-    We want the work done in filter_queue() to be done off the main server thread.
+    Prepares MessageFilter and MessageProcessors for use by the MessageJournal
+    to be sure the output prodduced by the MessageJournal is suitable given the parameters
+    of the request.
     """
 
     def __init__(self, chat_filter: Dict[str, Any], agent_network: AgentNetwork):
@@ -72,7 +70,7 @@ class QueueFilter:
 
     def apply_to_journal(self, invocation_context: InvocationContext):
         """
-        Apply the filter and processor to the message journal
+        Apply the filter and processor to the MessageJournal
         """
         message_journal: MessageJournal = invocation_context.get_journal()
         message_journal.set_message_filter(self.message_filter)
--- a/neuro_san/internals/chat/queue_filter.py
+++ b/neuro_san/internals/chat/queue_filter.py
@@ -23,7 +23,6 @@ from typing import Union
 from neuro_san.internals.filters.message_filter import MessageFilter
 from neuro_san.internals.filters.message_filter_factory import MessageFilterFactory
 from neuro_san.internals.graph.registry.agent_network import AgentNetwork
-from neuro_san.internals.interfaces.invocation_context import InvocationContext
 from neuro_san.internals.journals.message_journal import MessageJournal
 from neuro_san.message_processing.message_processor import MessageProcessor
 from neuro_san.message_processing.structure_message_processor import StructureMessageProcessor
@@ -68,10 +67,9 @@ class QueueFilter:
         message_processor = StructureMessageProcessor(structure_formats)
         return message_processor
 
-    def apply_to_journal(self, invocation_context: InvocationContext):
+    def apply_to_journal(self, message_journal: MessageJournal):
         """
         Apply the filter and processor to the MessageJournal
         """
-        message_journal: MessageJournal = invocation_context.get_journal()
         message_journal.set_message_filter(self.message_filter)
         message_journal.set_message_processor(self.message_processor)
--- a/neuro_san/internals/chat/queue_filter.py
+++ b/neuro_san/internals/chat/queue_filter.py
@@ -20,14 +20,11 @@ from typing import Dict
 from typing import List
 from typing import Union
 
-from copy import copy
-
-from neuro_san.internals.chat.async_collating_queue import AsyncCollatingQueue
 from neuro_san.internals.filters.message_filter import MessageFilter
 from neuro_san.internals.filters.message_filter_factory import MessageFilterFactory
 from neuro_san.internals.graph.registry.agent_network import AgentNetwork
 from neuro_san.internals.interfaces.invocation_context import InvocationContext
-from neuro_san.internals.messages.chat_message_type import ChatMessageType
+from neuro_san.internals.journals.message_journal import MessageJournal
 from neuro_san.message_processing.message_processor import MessageProcessor
 from neuro_san.message_processing.structure_message_processor import StructureMessageProcessor
 
@@ -42,21 +39,16 @@ class QueueFilter:
     """
 
     def __init__(self, invocation_context: InvocationContext,
-                 template_response_dict: Dict[str, Any],
                  chat_filter: Dict[str, Any],
                  agent_network: AgentNetwork):
         """
         Constructor
 
         :param invocation_context: An instance of InvocationContext
-        :param template_response_dict: A dictionary form of chat.ChatResponse
         :param chat_filter: A dictionary form of chat.ChatFilter
         :param agent_network: An instance of AgentNetwork
         """
-        self.input_queue: AsyncCollatingQueue = invocation_context.get_queue()
-        self.output_queue: AsyncCollatingQueue = invocation_context.get_filtered_queue()
-        self.template_response_dict: Dict[str, Any] = template_response_dict
-
+        self.invocation_context: InvocationContext = invocation_context
         self.message_filter: MessageFilter = MessageFilterFactory.create_message_filter(chat_filter)
         self.message_processor: MessageProcessor = self.create_outgoing_message_processor(agent_network)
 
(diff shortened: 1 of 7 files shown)


neuro_san/internals/chat/queue_filter.py
```diff
@@ -41,22 +38,13 @@ class QueueFilter:
...
         """
         Constructor
 
-        :param invocation_context: An instance of InvocationContext
-        :param template_response_dict: A dictionary form of chat.ChatResponse
         :param chat_filter: A dictionary form of chat.ChatFilter
         :param agent_network: An instance of AgentNetwork
         """
-        self.input_queue: AsyncCollatingQueue = invocation_context.get_queue()
-        self.output_queue: AsyncCollatingQueue = invocation_context.get_filtered_queue()
```
d1donlydfink, line 57: No need for input and output queues any more.

neuro_san/internals/chat/queue_filter.py
```diff
@@ -41,22 +38,13 @@ class QueueFilter:
...
-        self.template_response_dict: Dict[str, Any] = template_response_dict
```
d1donlydfink, line 58: This guy ended up being kinda bogus as an empty dict anyway. No need to store that.

neuro_san/internals/chat/queue_filter.py
```diff
@@ -82,42 +70,10 @@ def create_outgoing_message_processor(self, agent_network: AgentNetwork) -> Mess
         message_processor = StructureMessageProcessor(structure_formats)
         return message_processor
 
-    async def filter_queue(self):
-        """
-        Main task entry point for DirectAgentSession.
-        Filter the messages coming off the input queue so that they can be sent to the client
-        via the output queue.
-        We'd rather this work be done off the main server thread.
-        """
```
d1donlydfink, line 91: We don't need this as a task entry point any more cuz we don't need 2 queues any more.

neuro_san/internals/chat/queue_filter.py
```diff
@@ -82,42 +70,10 @@ def create_outgoing_message_processor(self, agent_network: AgentNetwork) -> Mess
...
-            message_type: ChatMessageType = message.get("type")
-            # Can modify message
-            await self.message_processor.async_process_message(message, message_type)
-
-        response_dict: Dict[str, Any] = copy(self.template_response_dict)
-        response_dict["response"] = message
-        return response_dict
+        message_journal: MessageJournal = invocation_context.get_journal()
+        message_journal.set_message_filter(self.message_filter)
+        message_journal.set_message_processor(self.message_processor)
```
d1donlydfink, line 75: Send the MessageFilter and the MessageProcessor to the MessageJournal.

neuro_san/internals/interfaces/invocation_context.py
```diff
@@ -89,13 +89,6 @@ def get_queue(self) -> AsyncCollatingQueue:
         """
         raise NotImplementedError
 
-    def get_filtered_queue(self) -> AsyncCollatingQueue:
-        """
-        :return: The AsyncCollatingQueue instance via which messages are streamed to the
-                AgentSession mechanics
-        """
-        raise NotImplementedError
```
d1donlydfink, line 97: This InvocationContext method no longer needed.

neuro_san/internals/journals/message_journal.py
```diff
@@ -39,6 +43,21 @@ def __init__(self, hopper: AsyncHopper):
...
+        """
+        Sets the message filter for this journal.
+        """
+        self.message_filter = message_filter
+
+    def set_message_processor(self, message_processor: MessageProcessor):
+        """
+        Sets the message processor for this journal.
+        """
+        self.message_processor = message_processor
```
d1donlydfink, line 59: Allow for the MessageFilter and MessageProcessor to be set later.
These differ on a per-request basis.

neuro_san/internals/journals/message_journal.py
```diff
@@ -53,11 +72,23 @@ async def write_message(self, message: BaseMessage, origin: List[Dict[str, Any]]
...
+        if not self.message_filter.allow_message(message_dict, message_type):
+            return
+
+        if self.message_processor is not None:
+            # Can modify message_dict
+            await self.message_processor.async_process_message(message_dict, message_type)
+
+        outgoing_dict: Dict[str, Any] = message_dict
+        if self.issue_response_dicts:
+            outgoing_dict = {"response": message_dict}
```
d1donlydfink: This is the meat of what was happening in the QueueFilter.filter_queue() task.
Now we are simply doing the work before we put anything into the single hopper of an event queue.
The MessageJournal queue now contains client-ready message dictionaries, thus still keeping a bunch of the conversion/filtering work off the main server thread.

neuro_san/session/async_direct_agent_session.py
```diff
@@ -185,15 +185,13 @@ async def streaming_chat(self, request_dict: Dict[str, Any]) -> AsyncGenerator[D
         _ = task
 
         # Task for late-stage conversions for any and all messages
-        queue_filter = QueueFilter(self.invocation_context, template_response_dict, chat_filter, self.agent_network)
-        task: Task = asyncio_executor.submit(self.request_id, queue_filter.filter_queue)
-        # Ignore the future. Live in the now.
-        _ = task
+        queue_filter = QueueFilter(chat_filter, self.agent_network)
+        queue_filter.apply_to_journal(self.invocation_context)
```
d1donlydfink: We don't need the extra task any more, but we do need the QueueFilter to apply to the MessageJournal.

neuro_san/session/async_direct_agent_session.py
```diff
@@ -185,15 +185,13 @@ async def streaming_chat(self, request_dict: Dict[str, Any]) -> AsyncGenerator[D
...
 
         # The generator below will asynchronously block waiting for
         # chat.ChatMessage dictionaries to come back asynchronously from the submit()
         # above until there are no more from the input.
-        queue_generator = self.invocation_context.get_filtered_queue()
+        queue_generator = self.invocation_context.get_queue()
```
d1donlydfink, line 217: Back to only one queue.

neuro_san/session/direct_agent_session.py
```diff
@@ -198,7 +196,7 @@ def streaming_chat(self, request_dict: Dict[str, Any]) -> Generator[Dict[str, An
             #    interrupted by caller-side "close" method.
             # 3. And we suppress all exceptions while deleting resources to keep things quieter.
             message: Dict[str, Any] = None
-            for message in generator.synchronously_iterate(self.invocation_context.get_filtered_queue()):
+            for message in generator.synchronously_iterate(self.invocation_context.get_queue()):
```
d1donlydfink, line 222: Same as async case above.

neuro_san/session/session_invocation_context.py
```diff
@@ -171,13 +170,6 @@ def get_queue(self) -> AsyncCollatingQueue:
         """
         return self.queue
 
-    def get_filtered_queue(self) -> AsyncCollatingQueue:
-        """
-        :return: The AsyncCollatingQueue instance via which messages are streamed to the
-                AgentSession mechanics
-        """
-        return self.filtered_queue
```
d1donlydfink, line 179: We no longer need 2 queues.

requirements.txt
```diff
@@ -6,7 +6,7 @@
 # requirements-build.txt
 
 # In-house dependencies
-leaf-common>=1.2.47
+leaf-common>=1.2.49
```
d1donlydfink, line 9: Use the latest leaf-common with eager task factories in AsyncioExecutor.
This saves us several seconds on large load-tests.

neuro_san/session/async_direct_agent_session.py
```diff
@@ -208,15 +208,13 @@ async def streaming_chat(self, request_dict: Dict[str, Any]) -> AsyncGenerator[D
         _ = task
 
         # Task for late-stage conversions for any and all messages
-        queue_filter = QueueFilter(self.invocation_context, template_response_dict, chat_filter, self.agent_network)
-        task: Task = asyncio_executor.submit(self.request_id, queue_filter.filter_queue)
-        # Ignore the future. Live in the now.
-        _ = task
+        queue_filter = QueueFilter(chat_filter, self.agent_network)
+        queue_filter.apply_to_journal(self.invocation_context)
```
Noravee: Now that filtering happens at write time instead of consume time, this ordering is a race: the chat task submitted a few lines above starts on the AsyncioExecutor's event-loop thread and can write to the journal before `apply_to_journal()` installs the filter/processor. Anything written in that window escapes with the default `MaximalMessageFilter` and no `StructureMessageProcessor`.

In the old design, submit-then-filter was safe because unfiltered messages just sat in the raw queue until the filter task drained them. And early writes do exist: the "Event acknowledged" message for `invocation == "event"` in `run_it()`, and the `set_up()` ValueError message. The eager task factory from leaf-common 1.2.49 (this same PR) makes the task start even sooner, widening the window. On session reuse, `reset()` keeps the old journal, so in the window messages would be filtered by the *previous* request's filter.

Fix is free: move these two lines above the `asyncio_executor.submit(...)` call — nothing about QueueFilter construction depends on the task.

neuro_san/session/direct_agent_session.py
```diff
@@ -198,10 +198,8 @@ def streaming_chat(self, request_dict: Dict[str, Any]) -> Generator[Dict[str, An
         _ = task
 
         # Task for late-stage conversions for any and all messages
-        queue_filter = QueueFilter(self.invocation_context, template_response_dict, chat_filter, self.agent_network)
-        task: Task = asyncio_executor.submit(self.request_id, queue_filter.filter_queue)
-        # Ignore the future. Live in the now.
-        _ = task
+        queue_filter = QueueFilter(chat_filter, self.agent_network)
+        queue_filter.apply_to_journal(self.invocation_context)
```
Noravee: Same race as the async version: `apply_to_journal()` should happen before the `submit()` above, otherwise early journal writes from the chat task can slip through unfiltered/unprocessed.

neuro_san/internals/journals/message_journal.py
```diff
@@ -53,11 +72,23 @@ async def write_message(self, message: BaseMessage, origin: List[Dict[str, Any]]
         """
         converter = BaseMessageDictionaryConverter(origin=origin)
         message_dict: Dict[str, Any] = converter.to_dict(message)
+        message_type: ChatMessageType = message_dict.get("type")
+
+        if not self.message_filter.allow_message(message_dict, message_type):
```
Noravee, line 76: Error-containment change worth a deliberate decision: the old `filter_queue()` wrapped this work in a try/finally that guaranteed `put_final_item()`, so a filter/processor exception ended the client stream cleanly. Now the same exception propagates out of `write_message()` into agent execution — and since `DataDrivenChatSession.chat()` only catches `PATIENCE_ERRORS`, it can kill the chat task before `finalize_request()` ever runs. `finalize_request()` is now the *only* producer of the final item, so the client's `async for` on the queue would hang forever.

Suggest a try/except around the filter + processor calls here that logs and lets the message through (or drops it), so a processing bug can't take down the whole request.

neuro_san/internals/journals/message_journal.py
```diff
@@ -39,6 +43,21 @@ def __init__(self, hopper: AsyncHopper):
                        any message will be put().
         """
         self.hopper: AsyncHopper = hopper
+        self.message_filter: MessageFilter = MaximalMessageFilter()
+        self.message_processor: MessageProcessor = None
+        self.issue_response_dicts: bool = True
```
Noravee: `issue_response_dicts` is always True — there's no setter and no False path anywhere. Either drop it and wrap unconditionally, or add the setter the flag implies. As-is it suggests a configuration that doesn't exist.

neuro_san/internals/chat/queue_filter.py
```diff
@@ -41,22 +38,13 @@ class QueueFilter:
     We want the work done in filter_queue() to be done off the main server thread.
```
Noravee: Stale docstring: it still describes `filter_queue()`, the input/output queues, and the off-main-thread task — none of which exist anymore. The class name is arguably a misnomer now too (it configures the journal rather than filtering a queue), but at minimum the docstring should catch up with the new role.

neuro_san/internals/chat/queue_filter.py
```diff
@@ -82,42 +70,10 @@ def create_outgoing_message_processor(self, agent_network: AgentNetwork) -> Mess
...
-        # We expect the message to be a dictionary form of chat.ChatMessage
-        if self.message_processor is not None:
-            message_type: ChatMessageType = message.get("type")
-            # Can modify message
-            await self.message_processor.async_process_message(message, message_type)
-
-        response_dict: Dict[str, Any] = copy(self.template_response_dict)
-        response_dict["response"] = message
-        return response_dict
+        message_journal: MessageJournal = invocation_context.get_journal()
```
Noravee: This assumes `get_journal()` returns a `MessageJournal`, but the setters aren't on the `Journal` interface — a custom `InvocationContext` from library code would hit an AttributeError here. Consider hoisting no-op `set_message_filter()`/`set_message_processor()` into `Journal` to make the contract explicit.

Relatedly: removing `get_filtered_queue()` from the public `InvocationContext` interface is a breaking change for outside implementers — probably worth a release-note line.

Noravee (commented):
Nice simplification — one less task and queue per request, and rejected messages never get queued at all. I traced through the behavioral equivalences and they hold up: the `{"response": ...}` wrapping matches the old always-empty `template_response_dict`; calling `allow_message()` with the raw dict `type` is equivalent to the old `allow()` path since `from_response_type()` is a pass-through for the `ChatMessageType` enums that `to_dict()` stores; filtering still sits downstream of the tracing/chat-history interceptors so traces and `chat_context` still see everything; and external-agent sub-sessions are unaffected since they request a MAXIMAL filter.

Two behavioral concerns inline — the setup-ordering race especially — plus a few minor notes. Also: `MessageJournal` is now trivially unit-testable (fake hopper, assert filter/processor/wrap behavior); a small test there would lock in the semantics this refactor carefully preserves.

neuro_san/session/async_direct_agent_session.py
d1donlydfink: That's an excellent point. I will go there.

d1donlydfink: Addressed. Good find.

neuro_san/session/direct_agent_session.py
d1donlydfink: Addressed. Good find.

neuro_san/internals/journals/message_journal.py
d1donlydfink: Removed issue_response_dicts boolean.

neuro_san/internals/chat/queue_filter.py
d1donlydfink: Refreshed the class comment.

neuro_san/internals/chat/queue_filter.py
d1donlydfink: get_filtered_queue() is not worth a release-note. It was short lived and we do not expect people to be implementing their own InvocationContexts, as there is no API for roll-your-own here.

d1donlydfink: It's now up to the caller to pass in a MessageJournal

neuro_san/internals/journals/message_journal.py
d1donlydfink: The try/finally block in the old filter_queue() was an end-of-task operation.
The responsibility for calling put_final_item() on the queue resides where it always has, in DataDrivenChatRequest.finalize_request()

The MessageFilter and MessageProcessor implementations have been exception-free for long enough and are Simple enough to not really have to worry about the exceptions from within those subsystems.
The finalize_request() method itself gets invoked at a number of places within DataDrivenChat to properly close out the MessageJournal queue.

In short: I'm not worried. We would have much bigger very in-your-face problems if the subsystems mentioned were not functioning.

Noravee (approved):
lgtm.

PR #1171, opened 2026-07-26
Http traffic tracer for neuro-san LLM connections.

andreidenissov-cog:
Summary

  Adds an opt-in, structured-logging tracer for outbound LLM traffic at the httpx layer. Emits one JSON event per HTTP request lifecycle phase, tagged with a per-user-request id that propagates through asyncio and executor bridges automatically. Purpose: understand what
  neuro-san is actually doing when it talks to LLM providers — how many calls, in what order, for how long, with what retries, per user request — for offline analysis via grep / jq / pandas.

  Zero code changes required to enable per-provider. httpx.AsyncClient is monkey-patched once at server startup, so every LLM SDK (openai, anthropic, google-genai partial, nvidia, ollama, and any other httpx-based provider) is captured uniformly with no adapters. Bedrock
  (aiobotocore) is a known gap; documented as future work.

  What the tracer captures

  Every LLM HTTP call generates a triple of structured JSON events, plus optional per-chunk events:

  - http_llm_out — request sent. Includes attempt_id, provider (host-inferred), method, URL, request body size, user_req_id (from ContextVar).
  - http_llm_in — response headers received. Same attempt_id, HTTP status, elapsed-ms since out (~TTFT for streaming), server_req_id from provider headers (openai-request-id / x-request-id / x-amzn-requestid).
  - http_llm_end — response body fully consumed or closed. Same attempt_id, total elapsed-ms, reason (stream_complete / stream_generator_exit / stream_error:<ExcType> / closed_without_iteration), stream bytes.
  - http_llm_chunk — (optional, high volume) one event per received body chunk with byte count and since_out_ms.

  Optional body capture (both request and response, capped at 64 KB per body) via a second env var — off by default because prompts/completions are large and often sensitive.

  Correlation across the lifecycle

  Two IDs make the log analyzable:

  - attempt_id — unique per httpx request. Groups the 3-4 events belonging to one HTTP roundtrip. Retries fire as fresh httpx requests, so each retry gets its own attempt_id.
  - user_req_id — per user-facing chat request. Set via contextvars.ContextVar at the top of StreamingChatHandler.post(); asyncio and run_coroutine_threadsafe propagate context automatically, so every LLM call anywhere in the call chain (agent → sub-agent → tool → LLM)
  inherits the same id.

  grep 'user_req_id":"<id>"' log.jsonl | jq . reconstructs the full LLM traffic lifecycle of one user request in time order, with retries clearly identified as multiple attempt_ids under the same user.

  Files touched

  - neuro_san/service/utils/http_llm_tracer.py (new, 382 lines) — HttpxLlmTracer class. install() monkey-patches httpx; set_user_request_id() sets the ContextVar; internal event hooks emit structured events; response aiter_raw and aclose are wrapped to emit correct
  stream_complete / closed_without_iteration events.
  - neuro_san/service/http/server/http_server.py — _maybe_install_http_llm_tracer() invoked before IOLoop.start(), gated on AGENT_HTTP_LLM_TRACE.
  - neuro_san/service/http/handlers/streaming_chat_handler.py — set a short UUID as user_req_id at the top of post().
  - neuro_san/deploy/Dockerfile — documented three new env vars.

  Env vars

  - AGENT_HTTP_LLM_TRACE (default "false") — master toggle. When enabled, installs the tracer on server startup.
  - AGENT_HTTP_LLM_TRACE_INCLUDE_BODIES (default "false") — capture request/response bodies verbatim (capped at 64 KB each). Prompts and completions are large and often sensitive; off by default.
  - AGENT_HTTP_LLM_TRACE_CHUNKS (default "false") — emit one event per received body chunk. Massive log volume (dozens of chunks per streaming LLM call); enable only when investigating inter-chunk cadence.

  Log routing

  Events go to a dedicated logger, neuro_san.diagnostics.http_llm_trace. Recommended production wiring in logging.hocon:

  handlers {
      http_llm_trace_file {
          class = "logging.handlers.RotatingFileHandler"
          filename = "/var/log/neuro-san/http_llm_trace.jsonl"
          formatter = "raw_message"
          maxBytes = 104857600
          backupCount = 5
      }
  }
  loggers {
      "neuro_san.diagnostics.http_llm_trace" {
          handlers = ["http_llm_trace_file"]
          level = "INFO"
          propagate = false     # keeps events out of the main log
      }
  }

  Without propagate = false, events would echo to any root-level handler as well.

  Design decisions worth flagging

  1. Monkey-patch httpx.AsyncClient.__init__ rather than injecting hooks per-SDK. Catches every provider without touching provider-specific plumbing. Trade-off: touches httpx internals, and any new provider using a non-httpx transport (Bedrock/aiobotocore is the current known
  example) is invisible until a separate hook is added.
  2. Response iterator wrapping was tricky. httpx internally calls response.aclose() from within its own body-exhaustion path — before our wrapped generator's post-loop code runs. First implementation emitted reason=aclose and lost the real stream_complete. Fixed by tracking
  iter_started on response.extensions and having aclose emit an end event only when iteration never began. Details in the module docstring.
  3. set_user_request_id uses ContextVar rather than passing an ID through the call chain. ContextVar propagates through every await and every asyncio.create_task / run_coroutine_threadsafe boundary automatically. Setting it once at the top of post() is enough for every
  downstream LLM call to inherit the ID. No plumbing through the deep call stack.
  4. Best-effort emission — _emit() catches all exceptions and swallows them. If a bug in the tracer produces malformed events, the tracer stops emitting silently rather than crashing httpx requests. Diagnostic tools must never break the traffic they observe.
  5. Overhead is negligible — two dict allocations and one logger.info per request event. Cheap enough to leave on in production if the log volume and routing are acceptable.

  Explicit non-goals

  - Not a replacement for LangChain / OpenTelemetry / Langfuse. No spans, no trace-id, no cross-service correlation. Purpose-built for local diagnostic runs.
  - Not per-token latency — captures headers-received time (TTFT) and body-complete time. Per-token cadence needs AGENT_HTTP_LLM_TRACE_CHUNKS.
  - Not Bedrock. Documented as future work; needs a separate aiobotocore hook.

  Verification / usage

  Confirmed end-to-end via a smoke test (local Tornado echo server, streaming response, monkey-patched client): the tracer produces the expected sequence:

  http_llm_out    user_req_id=user_ABC  attempt_id=7a3590  t=0        POST .../v1/x
  http_llm_in     user_req_id=user_ABC  attempt_id=7a3590  t=7.8ms    status=200
  http_llm_chunk  user_req_id=user_ABC  attempt_id=7a3590  t=7.9ms    bytes=8
  http_llm_chunk  user_req_id=user_ABC  attempt_id=7a3590  t=8.1ms    bytes=8
  http_llm_chunk  user_req_id=user_ABC  attempt_id=7a3590  t=8.1ms    bytes=8
  http_llm_end    user_req_id=user_ABC  attempt_id=7a3590  t=8.3ms    reason=stream_complete  bytes=24

  Real-world usage in this branch has driven the load-analysis work in burst_analysis_25_50_100.md (companion doc in neuro-san-studio) — confirmed zero HTTP 429s across 25/50/100-burst runs, ruling out provider rate limits as the wall-time doubling cause, and revealed LiteLLM
  proxy behavior that would have been invisible without HTTP-layer visibility.

### DIFF
--- a/neuro_san/service/http/server/http_server.py
+++ b/neuro_san/service/http/server/http_server.py
@@ -59,6 +59,7 @@ from neuro_san.service.interfaces.agent_server import AgentServer
 from neuro_san.service.interfaces.event_loop_logger import EventLoopLogger
 from neuro_san.internals.interfaces.startable import Startable
 from neuro_san.service.mcp.handlers.mcp_root_handler import McpRootHandler
+from neuro_san.service.utils.http_llm_tracer import HttpxLlmTracer
 from neuro_san.service.utils.loop_timeline_tracer import LoopTimelineTracer
 from neuro_san.service.utils.server_context import ServerContext
 from neuro_san.service.utils.server_status import ServerStatus
@@ -229,6 +230,12 @@ class HttpServer(AgentStateListener):
         # runs synchronously on the same thread.
         self._maybe_start_loop_timeline_tracer()
 
+        # Optionally install the httpx-level LLM traffic tracer. Must run
+        # BEFORE any LLM SDK constructs its httpx.AsyncClient, which happens
+        # lazily on first LLM call, so installing here (before IOLoop.start)
+        # is safely ahead of any traffic.
+        self._maybe_install_http_llm_tracer()
+
         tornado.ioloop.IOLoop.current().start()
         self.logger.info({}, "Http server stopped.")
 
@@ -318,6 +325,49 @@ class HttpServer(AgentStateListener):
 
         # Start LAST so the tracer captures all callbacks executed after the loop starts running.
         self._loop_timeline_tracer.start()
+
+    def _maybe_install_http_llm_tracer(self) -> None:
+        """
+        Install the httpx-level LLM traffic tracer if AGENT_HTTP_LLM_TRACE
+        is truthy. Emits structured JSON events (http_llm_out /
+        http_llm_in / http_llm_end / optional http_llm_chunk) to the
+        dedicated logger "neuro_san.diagnostics.http_llm_trace" so
+        post-run analysis can reconstruct the LLM traffic lifecycle of
+        any user request via the user_req_id field.
+
+        Environment variables:
+          AGENT_HTTP_LLM_TRACE (bool, default false)
+              Enables the tracer. Off by default.
+          AGENT_HTTP_LLM_TRACE_INCLUDE_BODIES (bool, default false)
+              When true, request and response bodies are logged verbatim
+              (capped at 64 KB per body). Prompts + completions are large
+              and sensitive; off by default.
+          AGENT_HTTP_LLM_TRACE_CHUNKS (bool, default false)
+              When true, emits one http_llm_chunk event per received
+              body chunk. Massive log volume; enable only when
+              investigating inter-chunk cadence.
+
+        The tracer overhead is negligible (a few log writes per LLM
+        call). Route the "neuro_san.diagnostics.http_llm_trace" logger
+        to a separate file in logging.hocon for clean offline analysis.
+        """
+        if not ConfigUtil.get_bool(os.environ, "AGENT_HTTP_LLM_TRACE"):
+            return
+
+        include_bodies: bool = ConfigUtil.get_bool(
+            os.environ, "AGENT_HTTP_LLM_TRACE_INCLUDE_BODIES")
+        include_chunks: bool = ConfigUtil.get_bool(
+            os.environ, "AGENT_HTTP_LLM_TRACE_CHUNKS")
+
+        HttpxLlmTracer.install(
+            include_bodies=include_bodies,
+            include_chunks=include_chunks,
+        )
+        self.logger.info(
+            {},
+            "HttpxLlmTracer installed (include_bodies=%s include_chunks=%s)",
+            include_bodies, include_chunks,
+        )
 
     def resolve_health_probe_port(self) -> int:
         """
--- a/neuro_san/service/utils/http_llm_tracer.py
+++ b/neuro_san/service/utils/http_llm_tracer.py
@@ -1,0 +1,382 @@
+
+# Copyright © 2023-2026 Cognizant Technology Solutions Corp, www.cognizant.com.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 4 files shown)


neuro_san/service/http/server/http_server.py
```diff
@@ -229,6 +230,12 @@ def start(self, startables: List[Startable]):
         # runs synchronously on the same thread.
         self._maybe_start_loop_timeline_tracer()
 
+        # Optionally install the httpx-level LLM traffic tracer. Must run
+        # BEFORE any LLM SDK constructs its httpx.AsyncClient, which happens
+        # lazily on first LLM call, so installing here (before IOLoop.start)
+        # is safely ahead of any traffic.
+        self._maybe_install_http_llm_tracer()
```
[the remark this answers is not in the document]
andreidenissov-cog: httpx clients we are interested in are constructed only on server-LLM transactions,
which happen after server is started. Not relevant here.

neuro_san/service/utils/http_llm_tracer.py
```diff
@@ -0,0 +1,407 @@
...
+
+        async def wrapped_aclose() -> None:
+            # aclose emits end only if the iterator never started (i.e.
+            # response closed without reading the body). If the iterator
+            # started, its own completion / exception path emits end --
+            # aclose here would otherwise race with httpx's own aclose
+            # firing mid-iteration and swallow the real stream_complete.
+            if not response.extensions.get("_llm_trace_iter_started"):
+                emit_end("closed_without_iteration", 0)
+            await original_aclose()
```
d1donlydfink: Can we make all these internal methods less internal?
That is, why can't these internal methods be static methods on this class (for instance)?

d1donlydfink (changes requested):
Very very cool.
Can we make all the internal methods not so internal?

neuro_san/service/utils/http_llm_tracer.py
andreidenissov-cog: Did that.

d1donlydfink (approved):
Thanks. Now at least theoretically those methods are individually testable.
Better than not testable at all.

PR #1176, opened 2026-07-28
Remove some dependencies from requirements.txt

d1donlydfink:
Remove the following dependencies from requirements.txt:
* langchain-classic
* langchain-community
* boto3
* validators

This came from a "I bet we don't use these any more ..." inquiry.

Worth noting that the only things that tie us to plain-old "langchain" library (not langchain_core) are Middleware concepts and a single use of langchain.agents.factory.create_agent

### DIFF
--- a/neuro_san/internals/run_context/langchain/core/run_context_runnable.py
+++ b/neuro_san/internals/run_context/langchain/core/run_context_runnable.py
@@ -166,6 +166,7 @@ class RunContextRunnable(NeuroSanRunnable):
             #       LangSmith, Langfuse, HoneyHive, or Arize Phoenix.
             # See the docs for Observability plugins in neuro-san-studio.
             # This was our only tie to langchain-classic, so make it optional in this case.
+            # pylint: disable=invalid-name
             LoggingCallbackHandler: Type[BaseCallbackHandler] = ResolverUtil.create_type(
                 "langchain_classic.callbacks.tracers.logging.LoggingCallbackHandler",
                 install_if_missing="langchain-classic"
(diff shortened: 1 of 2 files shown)


neuro_san/internals/run_context/langchain/core/run_context_runnable.py
```diff
@@ -162,6 +161,16 @@ async def run_it(self, inputs: Input) -> Output:
...
+            # If you find yourself curious about this setting, what you probably
+            # are really looking for is an Observability setup like:
+            #       LangSmith, Langfuse, HoneyHive, or Arize Phoenix.
+            # See the docs for Observability plugins in neuro-san-studio.
+            # This was our only tie to langchain-classic, so make it optional in this case.
+            # pylint: disable=invalid-name
+            LoggingCallbackHandler: Type[BaseCallbackHandler] = ResolverUtil.create_type(
+                "langchain_classic.callbacks.tracers.logging.LoggingCallbackHandler",
+                install_if_missing="langchain-classic"
+            )
```
d1donlydfink, line 173: This was the only place we were using langchain_classic, and its use if out-of-date.
Use ResolverUtil with a lead to the langchain-classic library and add a comment as to
what people are really looking for when they think they want this extra logging stuff.

requirements.txt
```diff
@@ -31,8 +30,6 @@ langchain>=1.2.0,<2.0
 # Floor 1.4.0: bind() on a chat model must return a chat-model-aware binding
 # (with bind_tools) for the Gemini AFC-disable in gemini_llm_policy.py to activate.
 langchain-core>=1.4.0,<2.0
-langchain-classic>=1.0.0,<2.0
-langchain-community>=0.4,<1.0
```
d1donlydfink, line 35: Remove a bunch of libs we don't actually use any more.

andreidenissov-cog (approved):
LGTM.

PR #1177, opened 2026-07-28
Refactor RegistryManifestRestorer to parallelize synchronous reads making server startup and nsflow iteration faster

### LINKED ISSUE #1147: Slow dev loop: saving one HOCON re-parses every served network

d1donlydfink:
It had been bugging me how long server startup was taking, and apparently it had been bugging @kaushik-cognizant how long watcher re-reads were taking when using nsflow.

* Refactor RegistryManifestRestorer so that read_one_manifest() can be parallelized.
* Use ThreadPoolExecutor for the synchronous parallelization by default, however if the number of agent networks is high, use a ProcessPoolExecutor instead because even though it's heavyweight, it's faster cuz no GIL and no filesystem lock contention.
* Still make the async path reflect the changes in the sync path, but don't parallelize that just yet, cuz it's probably pointless
* Consolidate some policy.
* Make some methods lower down the stack static so ProcessPoolExectuor won't have state problems when attempting to pickle "self" 

Note that the notion of "high number of agent networks" is currently an arbitrary 20 to use the ProcessPoolExecutor
Can iterate on that number as better data comes in.

Also, when we should revisit the ProcessPoolExeuctor approach when we get to Python 3.14T - the free-threading version which has no GIL.  At that point the lighter weight ThreadPoolExecutor should prevail.

This helps speed up server startup and watcher restore of neuro-san-studio from ~20 secs to ~3 secs using fork.
However, that is not as robust across OSes and python versions and can lead to deadlock. So with the more correct spawn, we get ~8 secs.

@Noravee : Can we get rid of async path in RegistryManifestRestorer now that AND doesn't use it any more?

Fixes #1147

### DIFF
--- a/neuro_san/internals/graph/persistence/registry_manifest_restorer.py
+++ b/neuro_san/internals/graph/persistence/registry_manifest_restorer.py
@@ -19,6 +19,7 @@ from typing import Dict
 from typing import List
 from typing import Sequence
 from typing import Tuple
+from typing import Type
 from typing import Union
 
 import os
@@ -176,7 +177,7 @@ class RegistryManifestRestorer(Restorer):
 
         # Determine how many threads to use and which style of executor to use
         max_workers: int = len(one_manifest)
-        pool_executor: Executor = ThreadPoolExecutor
+        pool_executor: Type[Executor] = ThreadPoolExecutor
 
         # By default use a ThreadPoolExecutor, but because of the GIL, in cases of large manifests,
         # a ProcessPoolExecutor ends up being more heavyweight, yet still more efficient because of
--- a/neuro_san/internals/graph/persistence/registry_manifest_restorer.py
+++ b/neuro_san/internals/graph/persistence/registry_manifest_restorer.py
@@ -18,6 +18,7 @@ from typing import Any
 from typing import Dict
 from typing import List
 from typing import Sequence
+from typing import Tuple
 from typing import Union
 
 import os
@@ -169,24 +170,55 @@ class RegistryManifestRestorer(Restorer):
 
         # At this point only hocon files we are going to serve up are in the one_manifest.
         for manifest_key, manifest_dict in one_manifest.items():
-
-            usable_network: bool = isinstance(manifest_dict, dict) and manifest_dict.get("serve", False)
-
-            # We'll need to use an agent mapper to get to this agent definition file.
-            agent_filepath: str = self.agent_mapper.agent_name_to_filepath(manifest_key)
-            network_name: str = self.agent_mapper.filepath_to_agent_network_name(agent_filepath)
-            validator = ManifestNetworkValidator(
-                external_network_names,
-                network_name=network_name)
             agent_network: AgentNetwork = None
-            if usable_network:
-                agent_network = self.restore_one_agent_network(manifest_dir, agent_filepath, manifest_key)
+            network_name: str = None
+            agent_network, network_name = self.read_one_agent_network(manifest_dir, manifest_file,
+                                                                      manifest_key, manifest_dict,
+                                                                      external_network_names)
+            if agent_network is not None:
 
-            self.process_one_agent_network(agent_network, usable_network, agent_filepath,
-                                           manifest_file, manifest_key, manifest_dict,
-                                           validator, agent_networks)
+                # Figure out where we want to put the network per the network's manifest dictionary
+                storage: str = StorageClass.PUBLIC
+                if not manifest_dict.get(StorageClass.PUBLIC):
+                    storage = StorageClass.PROTECTED
+                if manifest_dict.get("periodic", False):
+                    self.periodic_configs[network_name] = manifest_dict["periodic"]
+
+                agent_networks[storage][network_name] = agent_network
 
         return agent_networks
+
+    # pylint: disable=too-many-arguments,too-many-positional-arguments
+    def read_one_agent_network(
+                self,
+                manifest_dir: str,
+                manifest_file: str,
+                manifest_key: str,
+                manifest_dict: Dict[str, Any],
+                external_network_names: List[str]
+            ) -> Tuple[AgentNetwork, str]:
+        """
+        :param manifest_dir: The directory of the manifest file.
+        :param manifest_file: The file reference to use when restoring.
+        :param manifest_key: The key in the manifest file.
+        :param manifest_dict: The dictionary of the manifest file.
+        :param external_network_names: The list of external network names
+        :return: a nested map of storage type -> (mapping of name -> agent networks)
+        """
+        usable_network: bool = isinstance(manifest_dict, dict) and manifest_dict.get("serve", False)
+
+        # We'll need to use an agent mapper to get to this agent definition file.
+        agent_filepath: str = self.agent_mapper.agent_name_to_filepath(manifest_key)
+        network_name: str = self.agent_mapper.filepath_to_agent_network_name(agent_filepath)
+        validator = ManifestNetworkValidator(external_network_names, network_name=network_name)
+
+        agent_network: AgentNetwork = None
+        if usable_network:
+            agent_network = self.restore_one_agent_network(manifest_dir, agent_filepath, manifest_key)
+
+        self.process_one_agent_network(agent_network, usable_network, agent_filepath,
+                                       manifest_file, manifest_key, manifest_dict, validator)
+        return agent_network, network_name
 
     # pylint: disable=too-many-locals
     async def async_restore_one_manifest(self, manifest_file: str) -> Dict[str, Dict[str, AgentNetwork]]:
@@ -213,30 +245,60 @@ class RegistryManifestRestorer(Restorer):
 
         # At this point only hocon files we are going to serve up are in the one_manifest.
         for manifest_key, manifest_dict in one_manifest.items():
-
-            usable_network: bool = isinstance(manifest_dict, dict) and manifest_dict.get("serve", False)
-
-            # We'll need to use an agent mapper to get to this agent definition file.
-            agent_filepath: str = self.agent_mapper.agent_name_to_filepath(manifest_key)
-            network_name: str = self.agent_mapper.filepath_to_agent_network_name(agent_filepath)
-            validator = ManifestNetworkValidator(
-                external_network_names,
-                network_name=network_name)
             agent_network: AgentNetwork = None
-            if usable_network:
-                agent_network = await self.async_restore_one_agent_network(manifest_dir, agent_filepath, manifest_key)
+            network_name: str = None
+            agent_network, network_name = await self.async_read_one_agent_network(manifest_dir, manifest_file,
+                                                                                  manifest_key, manifest_dict,
+                                                                                  external_network_names)
+            if agent_network is not None:
 
-            self.process_one_agent_network(agent_network, usable_network, agent_filepath,
-                                           manifest_file, manifest_key, manifest_dict,
-                                           validator, agent_networks)
+                # Figure out where we want to put the network per the network's manifest dictionary
+                storage: str = StorageClass.PUBLIC
+                if not manifest_dict.get(StorageClass.PUBLIC):
+                    storage = StorageClass.PROTECTED
+                if manifest_dict.get("periodic", False):
+                    self.periodic_configs[network_name] = manifest_dict["periodic"]
+
+                agent_networks[storage][network_name] = agent_network
 
         return agent_networks
 
     # pylint: disable=too-many-arguments,too-many-positional-arguments
+    async def async_read_one_agent_network(
+                self,
+                manifest_dir: str,
+                manifest_file: str,
+                manifest_key: str,
+                manifest_dict: Dict[str, Any],
+                external_network_names: List[str]
+            ) -> Tuple[AgentNetwork, str]:
+        """
+        :param manifest_dir: The directory of the manifest file.
+        :param manifest_file: The file reference to use when restoring.
+        :param manifest_key: The key in the manifest file.
+        :param manifest_dict: The dictionary of the manifest file.
+        :param external_network_names: The list of external network names
+        :return: a nested map of storage type -> (mapping of name -> agent networks)
+        """
+        usable_network: bool = isinstance(manifest_dict, dict) and manifest_dict.get("serve", False)
+
+        # We'll need to use an agent mapper to get to this agent definition file.
+        agent_filepath: str = self.agent_mapper.agent_name_to_filepath(manifest_key)
+        network_name: str = self.agent_mapper.filepath_to_agent_network_name(agent_filepath)
+        validator = ManifestNetworkValidator(external_network_names, network_name=network_name)
+        agent_network: AgentNetwork = None
+        if usable_network:
+            agent_network = await self.async_restore_one_agent_network(manifest_dir, agent_filepath, manifest_key)
+
+        agent_network = self.process_one_agent_network(agent_network, usable_network, agent_filepath,
+                                                       manifest_file, manifest_key, manifest_dict, validator)
+
+        return agent_network, network_name
+
+    # pylint: disable=too-many-arguments,too-many-positional-arguments
     def process_one_agent_network(self, agent_network: AgentNetwork, usable_network: bool, agent_filepath: str,
                                   manifest_file: str, manifest_key: str, manifest_dict: Dict[str, Any],
-                                  validator: ManifestNetworkValidator,
-                                  agent_networks: Dict[str, Dict[str, AgentNetwork]]):
+                                  validator: ManifestNetworkValidator) -> AgentNetwork:
         """
         :param agent_network: The agent network to process.
         :param usable_network: Is this agent network usable?
@@ -245,7 +307,6 @@ class RegistryManifestRestorer(Restorer):
         :param manifest_key: The manifest key.
         :param manifest_dict: The manifest dictionary.
         :param validator: The validator.
-        :param agent_networks: The accumulated agent networks
         """
         network_name: str = self.agent_mapper.filepath_to_agent_network_name(agent_filepath)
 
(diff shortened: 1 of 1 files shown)


neuro_san/internals/graph/persistence/registry_manifest_restorer.py
```diff
@@ -145,6 +152,7 @@ def restore_one_manifest(self, manifest_file: str) -> Dict[str, Dict[str, AgentN
         :param manifest_file: The file reference to use when restoring.
         :return: a nested map of storage type -> (mapping of name -> agent networks)
         """
+        start: float = perf_counter()
```
d1donlydfink, line 155: Time the synchronous restore_one_manifest()

neuro_san/internals/graph/persistence/registry_manifest_restorer.py
```diff
@@ -167,26 +175,92 @@ def restore_one_manifest(self, manifest_file: str) -> Dict[str, Dict[str, AgentN
 
         external_network_names: List[str] = self.find_external_network_names(one_manifest)
 
-        # At this point only hocon files we are going to serve up are in the one_manifest.
-        for manifest_key, manifest_dict in one_manifest.items():
```
d1donlydfink, line 171: The sequential for-loop above is replaced by a parallelized Executor.map() call below to a new read_one_agent_network() that is the basis of iteration.

neuro_san/internals/graph/persistence/registry_manifest_restorer.py
```diff
@@ -167,26 +175,92 @@ def restore_one_manifest(self, manifest_file: str) -> Dict[str, Dict[str, AgentN
...
+        # Determine how many threads to use and which style of executor to use
+        max_workers: int = len(one_manifest)
+        pool_executor: Type[Executor] = ThreadPoolExecutor
+
+        # By default use a ThreadPoolExecutor, but because of the GIL, in cases of large manifests,
+        # a ProcessPoolExecutor ends up being more heavyweight, yet still more efficient because of
+        # the GIL.  When we get to free-threading in Python 3.14T, this can be changed back to ThreadPoolExecutor.
+        process_threshold: int = 20     # Somewhat arbitrary
+        if max_workers > process_threshold:
+            pool_executor = ProcessPoolExecutor
```
d1donlydfink: Here is where we decide between ThreadPoolExecutor and ProcessPoolExecutor.
Currently with neuro-san-studio ProcessPoolExeucutor is faster.

(@kaushik-cognizant : You were right about ProcessPoolExecutor!)

limit of 20 is somewhat arbitrary.  Can refine that as we have better data.

neuro_san/internals/graph/persistence/registry_manifest_restorer.py
```diff
@@ -167,26 +175,92 @@ def restore_one_manifest(self, manifest_file: str) -> Dict[str, Dict[str, AgentN
...
+            results: List[Tuple[AgentNetwork, str, Dict[str, Any]]] = executor.map(read_one, manifest_keys,
+                                                                                   manifest_dicts)
+            result: Tuple[AgentNetwork, str, Dict[str, Any]] = None
+            for result in results:
+                self.maybe_store_agent_network(result[0], result[1], result[2], agent_networks)
+
+        stop: float = perf_counter()
+        duration: float = stop - start
+        if duration > 5.0:
+            self.logger.warning("Manifest file %s restored in %.2f seconds.", manifest_file, duration)
```
d1donlydfink, line 222: At the end, spit out if the manifest file restore took more than 5 seconds.

Interestingly enough, the logging isn't quite set up correctly the first time around for infos to be spit out on timing all the time. That's its own puzzle.

neuro_san/internals/graph/persistence/registry_manifest_restorer.py
```diff
@@ -167,26 +175,92 @@ def restore_one_manifest(self, manifest_file: str) -> Dict[str, Dict[str, AgentN
...
+                external_network_names: List[str] = None
+            ) -> Tuple[AgentNetwork, str, Dict[str, Any]]:
+        """
+        :param manifest_dir: The directory of the manifest file.
+        :param manifest_file: The file reference to use when restoring.
+        :param manifest_key: The key in the manifest file.
+        :param manifest_dict: The dictionary of the manifest file.
+        :param external_network_names: The list of external network names
+        :return: A tuple of (agent_network, network_name, manifest_dict)
+        """
```
d1donlydfink, line 244: This new method encapsulates the work done in a single iteration of the above loop so it can be parallelized.  The return value is now a Tuple which aids in the processing of results from parallelization.

Note that from here down the call stack methods are now static so that ProcessPoolExecutor doesn't have argument-pickle problems across processes with "self". (I hate pickle, but I'll take the Easy now)

neuro_san/internals/graph/persistence/registry_manifest_restorer.py
```diff
@@ -167,26 +175,91 @@ def restore_one_manifest(self, manifest_file: str) -> Dict[str, Dict[str, AgentN
...
-        return agent_networks
+    def maybe_store_agent_network(self, agent_network: AgentNetwork, network_name: str, manifest_dict: Dict[str, Any],
+                                  agent_networks: Dict[str, Dict[str, AgentNetwork]]):
+        """
+        :param agent_network: The agent network to store
+        :param network_name: The name of the network
+        :param manifest_dict: The manifest dictionary
+        :param agent_networks: a nested map of storage type -> (mapping of name -> agent networks)
+                                potentially modified
+        """
```
d1donlydfink, line 285: Consolidated policy method for use in both sync and asynchronous cases which determines whether/where to store the agent network.

neuro_san/internals/graph/persistence/registry_manifest_restorer.py
```diff
@@ -213,30 +288,55 @@ async def async_restore_one_manifest(self, manifest_file: str) -> Dict[str, Dict
...
+            agent_network: AgentNetwork = None
+            network_name: str = None
+            agent_network, network_name, manifest_dict = await self.async_read_one_agent_network(
+                manifest_key,
+                manifest_dict,
+                manifest_file,
+                manifest_dir,
+                external_network_names
+            )
+            self.maybe_store_agent_network(agent_network, network_name, manifest_dict, agent_networks)
```
d1donlydfink, line 333: Refactor the asynchronous side similarly to keep sanity.
Not parallelizing here with gather() for 2 reasons:
1) it'd likely be pointless in terms of performance
2) it's not the main thrust of optimization.

Can either optimize this or eliminate this later as we better understand the use cases for asynchronous.

neuro_san/internals/graph/persistence/registry_manifest_restorer.py
```diff
@@ -213,30 +288,55 @@ async def async_restore_one_manifest(self, manifest_file: str) -> Dict[str, Dict
...
+        """
+        usable_network: bool = isinstance(manifest_dict, dict) and manifest_dict.get("serve", False)
+
+        # We'll need to use an agent mapper to get to this agent definition file.
+        agent_filepath: str = self.agent_mapper.agent_name_to_filepath(manifest_key)
+        network_name: str = self.agent_mapper.filepath_to_agent_network_name(agent_filepath)
+        validator = ManifestNetworkValidator(external_network_names, network_name=network_name)
+        agent_network: AgentNetwork = None
+        if usable_network:
+            agent_network = await self.async_restore_one_agent_network(manifest_dir, agent_filepath, manifest_key)
```
d1donlydfink: ... all for this one `await`   Maybe someday Python will make this simpler <wistful stare into the distance>

neuro_san/internals/graph/persistence/registry_manifest_restorer.py
```diff
@@ -213,30 +288,55 @@ async def async_restore_one_manifest(self, manifest_file: str) -> Dict[str, Dict
...
 
-            self.process_one_agent_network(agent_network, usable_network, agent_filepath,
-                                           manifest_file, manifest_key, manifest_dict,
-                                           validator, agent_networks)
+        agent_network = self.process_one_agent_network(agent_network, usable_network, agent_filepath,
+                                                       manifest_file, manifest_key, manifest_dict, validator)
 
-        return agent_networks
+        return agent_network, network_name, manifest_dict
```
d1donlydfink, line 372: Same return value to match the synchronous version for sanity.

neuro_san/internals/graph/persistence/registry_manifest_restorer.py
```diff
@@ -257,24 +356,17 @@ def process_one_agent_network(self, agent_network: AgentNetwork, usable_network:
...
 
-        # Figure out where we want to put the network per the network's manifest dictionary
-        storage: str = StorageClass.PUBLIC
-        if not manifest_dict.get(StorageClass.PUBLIC):
-            storage = StorageClass.PROTECTED
-        if manifest_dict.get("periodic", False):
-            self.periodic_configs[network_name] = manifest_dict["periodic"]
-
-        agent_networks[storage][network_name] = agent_network
+        return agent_network
```
d1donlydfink, line 418: No longer store in the agent_networks mapping in this method.
Do that outside this process() method and in the new maybe_store...() method above.

neuro_san/internals/graph/persistence/registry_manifest_restorer.py
```diff
@@ -213,30 +302,68 @@ async def async_restore_one_manifest(self, manifest_file: str) -> Dict[str, Dict
...
+        :param external_network_names: The list of external network names
+        :return: A tuple of (agent_network, network_name, manifest_dict)
+        """
+        usable_network: bool = isinstance(manifest_dict, dict) and manifest_dict.get("serve", False)
 
-            self.process_one_agent_network(agent_network, usable_network, agent_filepath,
-                                           manifest_file, manifest_key, manifest_dict,
-                                           validator, agent_networks)
+        # We'll need to use an agent mapper to get to this agent definition file.
+        agent_filepath: str = agent_mapper.agent_name_to_filepath(manifest_key)
```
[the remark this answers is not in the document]
d1donlydfink: Addressed.

deepsaia (dismissed):
Lgtm.

Noravee:
@d1donlydfink 

Yes — from the AND/studio side you can delete the async path in `RegistryManifestRestorer`. `GetSubnetwork` parses only the manifest via `RawManifestRestorer.async_restore()` (the `AbstractAsyncConfigRestorer` method), not `RegistryManifestRestorer` — so please keep the base-class async support on the raw restorer. Moreover, AND is about to depend on it even less: the fix for #1267 will move that read to the sync `RawManifestRestorer.restore()` in `asyncio.to_thread` behind a process-wide cache (same pattern as #1262).

neuro_san/internals/graph/persistence/registry_manifest_restorer.py
```diff
@@ -167,26 +176,107 @@ def restore_one_manifest(self, manifest_file: str) -> Dict[str, Dict[str, AgentN
...
+        Determine whether and where to store the agent network
+
+        :param agent_network: The agent network to store
+        :param network_name: The name of the network
+        :param manifest_dict: The manifest dictionary
+        :param agent_networks: a nested map of storage type -> (mapping of name -> agent networks)
+                                potentially modified
+        """
+        if agent_network is None:
+            return
```
Noravee, line 288: One behavior change in `maybe_store_agent_network()` I think is unintended: the early `return` when `agent_network is None` breaks the multi-manifest "unserve" overlay mechanism.

The old loop stored `None` in `agent_networks[storage][network_name]` for `serve: false` entries. `ManifestFilterChain` runs `ServedManifestConfigFilter` with `entry_for_skipped=True`. `restore_from_files()` then overlays manifests in order and only strips the `None`s at the very end with its own `ServedManifestConfigFilter` pass.

Concretely: with `AGENT_MANIFEST_FILE=a.hocon:b.hocon`, where `a.hocon` serves network X and `b.hocon` lists X with `serve: false` — before this PR, X ended up unserved (b's `None` overlaid a's network, then got filtered out). With this PR, the `None` is never stored, so X stays served. Single-manifest setups are unaffected. As a side effect, the final `ServedManifestConfigFilter` pass in `restore_from_files()` is now dead code.

Related (same root cause): the old code also registered `periodic_configs[network_name]` for those unserved entries; now that only happens for stored networks. That one might actually be an improvement — periodic on an unserved network is dubious — but flagging it so it's a deliberate choice rather than a silent change.

Suggestion: keep the early return only for restore/validation failures (which were never stored before either), and store `None` when the entry is explicitly unserved:

```python
if agent_network is None:
    if isinstance(manifest_dict, dict) and manifest_dict.get("serve", False):
        # Restore/validation failure — never stored, same as before
        return
    # Explicitly unserved: fall through and store None so a later
    # manifest can override an earlier one, per ServedManifestConfigFilter.
```

Letting the rest of the method run in the unserved case also restores the old `periodic_configs` behavior exactly, if that's what we want.

neuro_san/internals/graph/persistence/registry_manifest_restorer.py
Noravee: One concern with the `ProcessPoolExecutor` branch: no `mp_context` is passed, so it uses the platform-default multiprocessing start method. On macOS that's `spawn` (which is presumably what got tested), but on Linux with Python 3.10–3.13 the default is **`fork`** — and this code gets called from the `StorageWatcher` daemon thread inside a running, multi-threaded server process. Forking a multi-threaded process can deadlock the child on locks held by other threads at fork time (logging locks, event-loop internals, etc.); Python 3.12+ even emits a `DeprecationWarning` for exactly this:

```
DeprecationWarning: This process (pid=...) is multi-threaded, use of fork() may lead to deadlocks in the child.
```

So on a Linux deployment with >20 networks, every watcher re-read would fork the live server. It may work most of the time, but it's the kind of thing that hangs a pod once in a while and might be difficult to debug.

References:
- Default start methods and the fork-with-threads warning: https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods (note: "Changed in version 3.14: the default start method changed from fork to forkserver" on non-macOS POSIX — so 3.14 fixes this by default, but we support >=3.10)
- `os.fork()` DeprecationWarning in threaded processes (3.12+): https://docs.python.org/3/library/os.html#os.fork
- CPython ≤3.13 default selection (`darwin` → spawn, other POSIX → fork): https://github.com/python/cpython/blob/3.13/Lib/multiprocessing/context.py#L319-L328

Suggested fix — pin the context explicitly so behavior matches what was measured on macOS and is safe everywhere:

```python
executor_factory = partial(ThreadPoolExecutor, max_workers=max_workers)
if len(one_manifest) > process_threshold:
    executor_factory = partial(ProcessPoolExecutor, max_workers=max_workers,
                               mp_context=multiprocessing.get_context("spawn"))

with executor_factory() as executor:
```

d1donlydfink: Taking your suggestion, but it slows down start times on Linux.

neuro_san/internals/graph/persistence/registry_manifest_restorer.py
```diff
@@ -167,26 +176,107 @@ def restore_one_manifest(self, manifest_file: str) -> Dict[str, Dict[str, AgentN
...
+                                    manifest_dir=manifest_dir,
+                                    external_network_names=external_network_names,
+                                    agent_mapper=self.agent_mapper)
+            manifest_keys: List[str] = list(one_manifest.keys())
+            manifest_dicts: List[Dict[str, Any]] = list(one_manifest.values())
+
+            results: List[Tuple[AgentNetwork, str, Dict[str, Any]]] = executor.map(read_one, manifest_keys,
+                                                                                   manifest_dicts)
+            result: Tuple[AgentNetwork, str, Dict[str, Any]] = None
+            for result in results:
```
Noravee: One robustness gap with the `executor.map()` approach: exceptions from a worker re-raise in the parent when the results iterator is consumed ([docs](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor.map)), and the process pool adds failure modes that can't be caught inside the worker at all — [`BrokenProcessPool`](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.process.BrokenProcessPool) if a worker gets OOM-killed, pickling errors, etc. Since the watcher call chain (`RegistryStorageUpdater.update_storage()` → `StorageWatcher._run()`) has no try/except anywhere, one poisoned entry or crashed worker would silently kill the watcher thread — manifest hot-reload stops until the server restarts.

A try/except inside `read_one_agent_network()` (so one bad entry is skipped like the old per-entry policy) plus one around the executor block (so pool-level failures can't escape to the watcher) would cover it.

Minor related caveat, fine to leave as-is: in `ProcessPoolExecutor` mode the workers don't inherit the server's logging config (spawned children start fresh), so their `error()` output shows up as bare unformatted stderr and `info()` lines are dropped. Errors are still visible in container logs, just not formatted — worth knowing when debugging a >20-network manifest.

neuro_san/internals/graph/persistence/registry_manifest_restorer.py
d1donlydfink: This is a really good catch.

d1donlydfink: Addressed.

neuro_san/internals/graph/persistence/registry_manifest_restorer.py
d1donlydfink: Addressed exceptions in 2 places.

Noravee (approved):
LGTM.

PR #1180, opened 2026-07-29
Add AGENT_MANIFEST_CONCURRENCY_CONTEXT env var and ulimit hints to run.sh

d1donlydfink:
Left the previous PR on this subject thinking that some deployments expecting manifest reloads could have adverse memory consequences from the default of ProcessExecutorPool:"spawn".  So I decided to change how this is done with this PR to avoid backwards compatibility train-wrecks.

* Add a new AGENT_MANIFEST_CONCURRENCY_CONTEXT to have control over just how concurrent manifest reads are done.  The default is "thread", the safest, but also slowest, because GIL.
* Other possibilities are "spawn" and "thread".
* Add docs to Dockerfile
* Add ulimits hint to deploy/run.sh

### DIFF
--- a/neuro_san/deploy/Dockerfile
+++ b/neuro_san/deploy/Dockerfile
@@ -365,6 +365,8 @@ ENV AGENT_TOOL_EXECUTION_LOG_THRESHOLD_SECONDS=0
 #               but has potential to lead to deadlocks over repeated manifest reads.
 #               It can still be a reasonable option for server environments where the manifest
 #               is known never to change at all.
+# See https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods
+# for more details on "spawn" and "fork" concurrency models.
 ENV AGENT_MANIFEST_CONCURRENCY_CONTEXT="thread"
 
 # Optional URL describing how the server is to be referenced by the outside world.
--- a/neuro_san/deploy/Dockerfile
+++ b/neuro_san/deploy/Dockerfile
@@ -294,7 +294,7 @@ ENV AGENT_MANIFEST_UPDATE_PERIOD_SECONDS="0"
 # "fork":       uses a multi-process based concurrency model. It is actually the fastest option,
 #               but has potential to lead to deadlocks over repeated manifest reads.
 #               It can still be a reasonable option for server environments where the manifest
-#               is known never to change at all.
+#               is known never to change at all (AGENT_MANIFEST_UPDATE_PERIOD_SECONDS="0").
 # See https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods
 # for more details on "spawn" and "fork" concurrency models.
 ENV AGENT_MANIFEST_CONCURRENCY_CONTEXT="thread"
--- a/neuro_san/deploy/Dockerfile
+++ b/neuro_san/deploy/Dockerfile
@@ -281,6 +281,24 @@ ENV AGENT_REQUEST_LIMIT="-1"
 # if value is not specified or <= 0, no such dynamic updates will be executed.
 ENV AGENT_MANIFEST_UPDATE_PERIOD_SECONDS="0"
 
+# Describes the concurrency model used for the manifest file updates.
+# Valid values are "thread" (the default) and "spawn" and "fork".
+#
+# "thread":     uses the default threading concurrency model.  It is best for servers who
+#               know their manifest content will be changing over the course of their lifetime.
+#               It is slowest of all options, but has the least amount of overall memory overhead
+#               and is lightweight and generally the safest and problem-free.
+# "spawn":      uses a multi-process based concurrency model.  It is the safest option (avoiding deadlocks)
+#               for environments where the manifest is changing, but some speed is still desired
+#               (like development servers).
+# "fork":       uses a multi-process based concurrency model. It is actually the fastest option,
+#               but has potential to lead to deadlocks over repeated manifest reads.
+#               It can still be a reasonable option for server environments where the manifest
+#               is known never to change at all.
+# See https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods
+# for more details on "spawn" and "fork" concurrency models.
+ENV AGENT_MANIFEST_CONCURRENCY_CONTEXT="thread"
+
 # By default, the HTTP service reports the neuro-san library pip version in its health-check response.
 # It is possible to add other libraries to those results by listing them within this env var
 # below and separating them with spaces, like this: "langchain openai".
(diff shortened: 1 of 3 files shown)


neuro_san/deploy/Dockerfile
```diff
@@ -351,6 +351,24 @@ ENV AGENT_STREAM_KEEP_ALIVE_WITH_PROGRESS_INTERVAL_SECONDS=0
...
+# "spawn":      uses a multi-process based concurrency model.  It is the safest option (avoiding deadlocks)
+#               for environments where the manifest is changing, but some speed is still desired
+#               (like development servers).
+# "fork":       uses a multi-process based concurrency model. It is actually the fastest option,
+#               but has potential to lead to deadlocks over repeated manifest reads.
+#               It can still be a reasonable option for server environments where the manifest
+#               is known never to change at all.
+# See https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods
+# for more details on "spawn" and "fork" concurrency models.
+ENV AGENT_MANIFEST_CONCURRENCY_CONTEXT="thread"
```
d1donlydfink: Docs for new env var.

neuro_san/deploy/run.sh
```diff
@@ -56,11 +56,17 @@ function run() {
...
+    # command line.  We don't want the ceiling of fds to interfere with how many
+    # requests we can serve in the container.
+    FILE_DESCRIPTOR_MAX=100000
+
     # Run the docker container in interactive mode
     #   Mount the 1st command line arg as the place where input files come from
     #   Slurp in the rest as environment variables, all of which are optional.
 
     docker_cmd="docker run --rm -it \
+        --ulimit nofile=${FILE_DESCRIPTOR_MAX}:${FILE_DESCRIPTOR_MAX} \
```
d1donlydfink, line 69: Add hints about how to do max file descriptors in run.sh

neuro_san/internals/graph/persistence/registry_manifest_restorer.py
```diff
@@ -156,18 +156,31 @@ def restore_one_manifest(self, manifest_file: str) -> Dict[str, Dict[str, AgentN
...
+            # This is still a reasonable option for server environments where the manifest
+            # is known never to change at all.
             executor_factory = partial(ProcessPoolExecutor, max_workers=max_workers,
-                                       mp_context=get_context("spawn"))
+                                       mp_context=get_context(concurrency_context))
+        else:
+            # The default of "thread" is best for servers who know their manifest content will be
+            # changing over the course of their lifetime. It is slowest of all options, but has
+            # the least amount of overall memory overhead and is lightweight and safe and problem-free.
+            executor_factory = partial(ThreadPoolExecutor, max_workers=max_workers)
```
d1donlydfink: Remove the arbitrary-ness of when the Thread vs Process comes into play and opt for a safer and
more configurable path.

Default is now actually "thread" (slowest), a change from the previous PR's "spawn".

The idea is that within studio run.py, we can reset the developer default to "spawn" as it was before.

andreidenissov-cog (dismissed):
Looks good.

neuro_san/internals/graph/persistence/registry_manifest_restorer.py
```diff
@@ -156,18 +156,31 @@ def restore_one_manifest(self, manifest_file: str) -> Dict[str, Dict[str, AgentN
...
-            # While "spawn" below is more correct for more OSes and more Python versions,
-            # it is not necessarily the fastest. "fork" can be faster, but could lead to deadlocks.
-            # See https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods
+
+        # The default of "thread" is the least heavyweight, but not necessarily the fastest
+        # given the context of how/how often the manifest is read.
+        concurrency_context: str = os.environ.get("AGENT_MANIFEST_CONCURRENCY_CONTEXT", "thread").lower()
+
+        executor_factory = None
+        if concurrency_context in ("spawn", "fork"):
```
Noravee: Since anything that isn't exactly `spawn`/`fork` silently falls through to `thread`, a typo here (or `forkserver`, or a trailing space — there's `.lower()` but no `.strip()`) just makes startup mysteriously slow with nothing in the logs. And since the plan is for studio's `run.py` to set this to get its speed back, the silent-fallback case is exactly the one someone will eventually hit.

A related edge: `get_context("fork")` raises `ValueError` on Windows, and it's evaluated when the `partial` is constructed — *outside* the `try` that starts below — so a Windows box configured with `fork` would have the exception escape `restore_one_manifest()`, and on the watcher path that kills the watcher thread (the same failure class the try/except from #1177 was added for).

Suggest: `.strip().lower()`, validate against an explicit whitelist, warn on unrecognized values, and log the effective executor choice once at info — that last one also helps when diagnosing pod behavior.

neuro_san/internals/graph/persistence/registry_manifest_restorer.py
```diff
@@ -156,18 +156,31 @@ def restore_one_manifest(self, manifest_file: str) -> Dict[str, Dict[str, AgentN
...
+            #
+            # "spawn" is the safest option (avoiding deadlocks) for environments where the manifest is changing,
+            # but some speed is still desired (like development servers).
+            #
+            # "fork" is actually the fastest option, but has potential to lead to deadlocks.
+            # This is still a reasonable option for server environments where the manifest
+            # is known never to change at all.
             executor_factory = partial(ProcessPoolExecutor, max_workers=max_workers,
-                                       mp_context=get_context("spawn"))
+                                       mp_context=get_context(concurrency_context))
```
Noravee: Worth considering `forkserver` as a fourth option — it may be the best fit for the memory concern that motivated this PR: workers fork from a clean, single-threaded helper process, so it avoids the threaded-fork deadlock risk (safe on watcher re-reads, unlike `fork`) while being much cheaper per re-read than `spawn` (no full interpreter start + re-import per worker each time). It's what Python 3.14 made the default on Linux for exactly this reason.

Cost is one more string in the whitelist plus a Dockerfile paragraph. It's not available on Windows (`get_context("forkserver")` raises `ValueError` there), which the validation suggested above would also handle.

https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods

neuro_san/internals/graph/persistence/registry_manifest_restorer.py
d1donlydfink: All very good points. Addressed, try and all.

neuro_san/internals/graph/persistence/registry_manifest_restorer.py
d1donlydfink: Added fork server.  I left it out because I couldn't get my head immediately around it from the explanation in the docs, but I get it now.

d1donlydfink: Empirically forkserver is a little faster than spawn, but slower than fork

Noravee (approved):
LGTM.

PR #1184, opened 2026-07-31
Fix non-deterministic registry parse failures by serializing HOCON parsing

### LINKED ISSUE #1183: Non-deterministic registry parse failures at startup: pyhocon/pyparsing global-state race in parallel manifest reads (since #1177)
## Symptom

Since #1177 (merged 2026-07-28), loading registries is non-deterministic. Running e.g.

```
python -m neuro_san.client.agent_cli --agent music_nerd_pro_multi_agents
```

*sometimes* fails to load a random subset of registry files, with errors like:

```
Parse error in registry item music_nerd_pro_sly.hocon. Skipping. - Expected end of text, found '\n'  (at char 5295), (line:118, col:2)
manifest registry music_nerd_pro_sly.hocon not found in .../neuro_san/registries/manifest.hocon
```

or several files at once reporting:

```
manifest registry gist.hocon has validation errors. Skipping. Errors: [
    "Nothing to validate."
]
```

The same files parse fine on the next run. Nothing is wrong with the `.hocon` files themselves.

## Root cause

`pyhocon.ConfigParser.parse` is not thread-safe, and #1177 made `RegistryManifestRestorer.restore_one_manifest` parse all registry files concurrently in a `ThreadPoolExecutor` by default (`AGENT_MANIFEST_CONCURRENCY_CONTEXT=thread`).

On **every** parse call, pyhocon rebuilds its entire pyparsing grammar inside a context manager that mutates **process-global** state ([config_parser.py](https://github.com/chimpler/pyhocon/blob/master/pyhocon/config_parser.py), `set_default_white_spaces`):

```python
default = ParserElement.DEFAULT_WHITE_CHARS   # saves " \n\t\r" for the first thread
ParserElement.setDefaultWhitespaceChars(' \t')
yield                                          # grammar build + parse happen here
ParserElement.setDefaultWhitespaceChars(default)
```

pyparsing elements capture `DEFAULT_WHITE_CHARS` at construction time. So when the first-wave thread finishes its parse and restores the pristine `" \n\t\r"`, any other thread that is mid-grammar-build at that instant constructs the rest of its grammar with `\n` treated as skippable whitespace. HOCON is newline-sensitive, so a partially-poisoned grammar misparses. (pyparsing's maintainer has acknowledged that class-level state is not thread-safe: pyparsing/pyparsing#89.)

Before #1177 the manifest loop was sequential, so the global mutation was harmless. The race window is ~1 ms of grammar construction per parse, which is why failures are intermittent and cluster at startup (only threads that saved the pristine default can poison others).

## Reproduction / evidence

Plain concurrent parsing rarely hits the window, so the race was proven by deterministically flipping `DEFAULT_WHITE_CHARS` at each of the 221 `ParserElement` construction points during pyhocon's grammar build (simulating the racing restore), then flipping back after a window of w constructions (simulating a third thread re-entering parse). Sweeping all poison windows over `neuro_san/registries/music_nerd_pro_sly.hocon` (1768 combinations):

| Outcome | Count |
|---|---|
| ok | 1634 |
| `ConfigWrongTypeException: Token ... must be of type str` | 70 |
| **`ParseException: Expected end of text, found '\n'  (at char 5295), (line:118, col:2)`** | 27 |
| **Empty config → `"Nothing to validate."`** | 22 |
| `ParseSyntaxException: Expected '}', found '\n'` | 15 |

The `ParseException` reproduces the observed error **character-for-character** (same char/line/col). The empty-config case reproduces the `"Nothing to validate."` symptom; one poisoning event hits every thread that is mid-build at that moment, which is why several registries fail together in a single run.

<details>
<summary>Simulation script</summary>

```python
"""Simulate a racing thread restoring DEFAULT_WHITE_CHARS mid-grammar-build."""
from pyparsing import ParserElement
from pyhocon import ConfigFactory

orig_init = ParserElement.__init__
state = {"count": 0, "on": -1, "off": -1}

def patched_init(self, *a, **kw):
    state["count"] += 1
    if state["count"] == state["on"]:
        ParserElement.DEFAULT_WHITE_CHARS = " \n\t\r"   # racing restore
    elif state["count"] == state["off"]:
        ParserElement.DEFAULT_WHITE_CHARS = " \t"       # third thread re-enters parse
    orig_init(self, *a, **kw)

ParserElement.__init__ = patched_init

# e.g. on=1, off=22 yields the exact reported error on music_nerd_pro_sly.hocon
state["on"], state["off"] = 1, 22
state["count"] = 0
ConfigFactory.parse_file("neuro_san/registries/music_nerd_pro_sly.hocon")
```

</details>

## Workaround

```
export AGENT_MANIFEST_CONCURRENCY_CONTEXT=spawn
```

The `ProcessPoolExecutor` path gives each worker its own interpreter, so there is no shared pyparsing state.

## Environment

- neuro-san @ `218b730e` (main)
- Python 3.13, macOS (Darwin 24.6.0), 12 CPUs → 12 worker threads over 32 registry files
- pyhocon 0.3.61, pyparsing 3.2.1

Noravee:
## Problem

Since #1177 parallelized manifest reads, registry loading intermittently fails on perfectly valid files:

```
Parse error in registry item music_nerd_pro_sly.hocon. Skipping. - Expected end of text, found '\n'  (at char 5295), (line:118, col:2)
manifest registry gist.hocon has validation errors. Skipping. Errors: [ "Nothing to validate." ]
```

Root cause: pyhocon's `ConfigParser.parse()` rebuilds its pyparsing grammar on every call while temporarily overriding the **process-global** `ParserElement.DEFAULT_WHITE_CHARS`. pyparsing elements capture that global at construction time, so when one thread restores the default while another thread is mid-grammar-build, the second thread's grammar is partially poisoned (newlines become skippable whitespace). Reproduced deterministically by flipping the global at controlled construction indices — the exact reported error (char 5295, line 118, col 2) matches character-for-character. See #1183 for the full analysis and pyparsing/pyparsing#89 for upstream acknowledgment that this state is not thread-safe.

## Fix

* **`HoconParseLock`** (new): a context manager over a single process-wide `threading.Lock` that all in-process HOCON parsing must hold. JSON parsing shares no pyparsing state and does not funnel through it. No throughput is lost: pyhocon parsing is pure-Python and CPU-bound, so the GIL already prevented real parse parallelism between threads, and `ProcessPoolExecutor` workers each have their own copy of the lock.
* **Fork safety**: `os.register_at_fork` hooks acquire the lock before any fork, release it in the parent, and `_at_fork_reinit()` it in the child — the same treatment stdlib `logging` gives its module lock. Reinit (not release) matters: on CPython ≤ 3.12 platforms without POSIX semaphores (e.g. macOS), a waiter blocked in `acquire()` can hold the lock's internal pthread mutex at the fork instant, and a child-side `release()` would hang on that inherited mutex (bpo-40089).
* **Lock taken at all three in-repo parse sites**: `AbstractAsyncConfigRestorer.deserialize_file_contents` (covers registries, manifests, llm/toolbox/MCP configs), `LoggingConfigRestorer`, and `DataDrivenAgentTestDriver.parse_hocon_test_case`.
* **Drive-by fix** in `LoggingConfigRestorer`: the "Unsupported logging config file type" `ValueError` was attached to an unreachable branch, so unsupported extensions silently returned `{}`; it now raises at the source.
* **Regression tests**: mutual exclusion (16 parses / 8 threads, max-in-flight must be 1), lock blocking with a started-handshake, JSON bypass, and fork-during-parse child usability (deadline-bounded reaping, `waitstatus_to_exitcode`, deterministic Event-based ordering assertion). A scoped `conftest.py` exempts these LLM-free unit tests from the project-wide `OPENAI_API_KEY` skip so they actually run on secret-less CI.

## Verification

* Threaded soak: 30 consecutive full-manifest restores with a concurrent HOCON-parsing churn thread — 33/33 networks every round, zero parse errors (previously intermittent failures).
* All three `AGENT_MANIFEST_CONCURRENCY_CONTEXT` modes (`thread`, `spawn`, `fork`) restore identical results under churn; fork mode stressed 10/10 runs.
* Child-deadlock scenario for the fork hooks reproduced on Python 3.12/macOS with `release()` and verified fixed with `_at_fork_reinit()`.
* flake8 clean, pylint 10.00/10, full persistence/graph/service suites pass (123 tests).

## Known trade-offs (deliberate, documented in the class docstring)

* pyhocon resolves `include` directives inside the parse, so include I/O happens under the lock. Local file includes are unaffected (sub-millisecond); `include url(...)` is fetched by pyhocon with **no timeout** and must not be used in served registries.
* The lock is applied per call site rather than at a structural chokepoint — see the follow-up plan below.

## Next step: move the fix upstream to leaf-common

Tracked in **cognizant-ai-lab/leaf-common#168**.

The structural home for this lock is leaf-common: `HoconSerializationFormat.to_object` contains the only `ConfigFactory.parse_string` call in the entire library, and every HOCON path (including `EasyHoconPersistence`) funnels through it. Acquiring the lock there makes the invariant unbypassable for every consumer, instead of the per-call-site convention this PR uses (which the review behind this PR already caught being violated once in-repo).

This PR still locks per call site so the registry flakiness is fixed now, without waiting on a leaf-common release. Once leaf-common#168 ships (lock inside `to_object` + the `register_at_fork`/`_at_fork_reinit` fork hooks + ported tests), a follow-up PR here bumps the leaf-common minimum, deletes `hocon_parse_lock.py`, and removes the three `with HoconParseLock():` blocks — as a coordinated swap, since a leaf-common lock plus these call-site locks on the same non-reentrant `Lock` pattern would double-acquire. Full migration plan and design details are in the issue.

## Downstream impact (neuro-san-studio designer tools)

The Agent Network Designer's `get_subnetwork` and `get_toolbox` coded tools parse HOCON exclusively through `RegistryManifestRestorer`/`RawManifestRestorer` and `ToolboxInfoRestorer` — all subclasses of `AbstractAsyncConfigRestorer`, so both are covered by the lock taken there.

* **Correctness**: those tools were exposed to this race. Their process-wide caches (neuro-san-studio#1282) make parsing rare but don't close the window — the cache's once-gate is per event loop, so two sessions on different threads can still trigger the same cold-load parse concurrently, and `get_subnetwork` parses many registry files in one load. Before this PR one of those parses could sporadically fail on a valid file (the cache's publish-only-on-success contract keeps the bad result out of the cache, but that call degrades). This PR removes that failure mode.
* **Performance**: no regression. The warm path (the common case with the caches) never parses and never touches the lock; cold-start parses were already effectively serial under the GIL.

The two changes are complementary: the caches make parsing rare, this lock makes the rare parse safe.

Fixes #1183

### DIFF
--- a/neuro_san/internals/persistence/abstract_async_config_restorer.py
+++ b/neuro_san/internals/persistence/abstract_async_config_restorer.py
@@ -15,8 +15,10 @@
 #
 # END COPYRIGHT
 from typing import Any
+from typing import ContextManager
 from typing import Dict
 
+from contextlib import nullcontext
 from io import BytesIO
 from json.decoder import JSONDecodeError
 from logging import getLogger
@@ -33,6 +35,8 @@ from leaf_common.persistence.interface.restorer import Restorer
 from leaf_common.serialization.format.hocon_serialization_format import HoconSerializationFormat
 from leaf_common.serialization.format.json_serialization_format import JsonSerializationFormat
 from leaf_common.serialization.interface.serialization_format import SerializationFormat
+
+from neuro_san.internals.persistence.hocon_parse_lock import HoconParseLock
 
 
 class AbstractAsyncConfigRestorer(Restorer, ConfigFilter):
@@ -155,10 +159,14 @@ class AbstractAsyncConfigRestorer(Restorer, ConfigFilter):
 
         # Determine the serialization format
         serialization: SerializationFormat = None
+        parse_lock: ContextManager = nullcontext()
         if file_path.endswith(".hocon"):
             # Worth noting that includes within hocon files can only be done synchronously
             # The core pyhocon library does not support async includes.
             serialization = HoconSerializationFormat()
+            # pyhocon parsing mutates process-global pyparsing state and is not
+            # thread-safe. See HoconParseLock.
+            parse_lock = HoconParseLock()
         elif file_path.endswith(".json"):
             serialization = JsonSerializationFormat()
         else:
@@ -166,7 +174,8 @@ class AbstractAsyncConfigRestorer(Restorer, ConfigFilter):
 
         # Read the contents
         try:
-            config = serialization.to_object(bytes_file)
+            with parse_lock:
+                config = serialization.to_object(bytes_file)
         except (ParseException, ParseSyntaxException, JSONDecodeError, ConfigException) as exception:
             # Re-raise as ValueError, not pyparsing.ParseException, for two reasons:
             #
--- a/neuro_san/internals/persistence/hocon_parse_lock.py
+++ b/neuro_san/internals/persistence/hocon_parse_lock.py
@@ -48,10 +48,23 @@ class HoconParseLock:
     already prevents real parse parallelism between threads.
     ProcessPoolExecutor-based reads are unaffected because each worker process
     has its own copy of the lock.
+
+    Known trade-off: pyhocon resolves "include" directives (file or URL reads)
+    inside the parse, so that I/O happens while the lock is held. Ordinary
+    local-file includes (include "registries/foo.hocon") are fine: their reads
+    are sub-millisecond and their nested parse happens inside pyhocon, below
+    this lock, so there is no reentrancy. The hazard is unbounded I/O: pyhocon
+    fetches "include url(...)" with no timeout, and a file include on a hung
+    network mount blocks the same way, stalling every other HOCON parse (and
+    every fork, per the hooks below) instead of just its own thread. Do not
+    use "include url(...)" in served registries.
     """
 
     # One lock for the whole process, shared by all instances,
     # because the pyparsing state it guards is process-global.
+    # Never rebind this attribute: the register_at_fork hooks below capture
+    # bound methods of this exact object and cannot be re-registered, so a
+    # replacement lock would serialize parses without fork protection.
     lock: ClassVar[Lock] = Lock()
 
     if hasattr(os, "register_at_fork"):
@@ -59,16 +72,22 @@ class HoconParseLock:
         # Without this, an os.fork() happening while some thread holds the lock
         # (e.g. AGENT_MANIFEST_CONCURRENCY_CONTEXT="fork" while a watcher thread
         # is mid-parse) would give the child a lock that can never be released,
-        # deadlocking the child's first HOCON parse. Holding the lock across the
-        # fork means children always start with it released - the same treatment
-        # the standard library logging module gives its module lock.
-        # Windows has no fork, hence the hasattr guard.
-        # The bare "lock" reference is required: the HoconParseLock name is not
-        # bound until the class body finishes executing.
+        # deadlocking the child's first HOCON parse. This mirrors what the
+        # standard library logging module does for its module lock: acquire
+        # before the fork, release in the parent, and _at_fork_reinit() in the
+        # child. The child must reinit rather than release: on CPython <= 3.12
+        # platforms without POSIX semaphores (e.g. macOS), a waiter blocked in
+        # acquire() can hold the lock's internal pthread mutex at the fork
+        # instant, and a child-side release() would hang on that inherited
+        # mutex whose owner is dead (see bpo-40089). Windows has no fork,
+        # hence the hasattr guard.
+        # The bare "lock" references are required: the HoconParseLock name is
+        # not bound until the class body finishes executing.
         os.register_at_fork(
             before=lock.acquire,
             after_in_parent=lock.release,
-            after_in_child=lock.release,
+            # pylint: disable=no-member,protected-access
+            after_in_child=lock._at_fork_reinit,
         )
 
     def __enter__(self) -> "HoconParseLock":
--- a/neuro_san/internals/persistence/hocon_parse_lock.py
+++ b/neuro_san/internals/persistence/hocon_parse_lock.py
@@ -1,0 +1,92 @@
+
+# Copyright © 2023-2026 Cognizant Technology Solutions Corp, www.cognizant.com.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# END COPYRIGHT
+from typing import Any
+from typing import ClassVar
+
+import os
+
+from threading import Lock
+
+
+class HoconParseLock:
+    """
+    Context manager serializing all in-process pyhocon parsing.
+
+    pyhocon's ConfigParser.parse() rebuilds its pyparsing grammar on every call
+    while temporarily overriding the process-global
+    pyparsing.ParserElement.DEFAULT_WHITE_CHARS (see set_default_white_spaces()
+    in pyhocon's config_parser.py). pyparsing elements capture that global at
+    construction time, so two HOCON parses running concurrently in the same
+    process can poison each other's grammar mid-build. Symptoms are
+    non-deterministic parse failures on perfectly valid files: spurious
+    ParseExceptions ("Expected end of text, found '\\n'"), empty parse results
+    (surfacing as "Nothing to validate." validation errors), or
+    ConfigWrongTypeExceptions.
+    See https://github.com/cognizant-ai-lab/neuro-san/issues/1183 and
+    https://github.com/pyparsing/pyparsing/issues/89 for details.
+
+    Every code path that parses HOCON must do so under "with HoconParseLock():".
+    JSON/YAML parsing shares no such global state and should not funnel through
+    this lock.
+
+    Serializing HOCON parsing does not sacrifice the speedup of parallel
+    manifest reads: pyhocon parsing is pure-Python and CPU-bound, so the GIL
+    already prevents real parse parallelism between threads.
+    ProcessPoolExecutor-based reads are unaffected because each worker process
+    has its own copy of the lock.
+    """
+
+    # One lock for the whole process, shared by all instances,
+    # because the pyparsing state it guards is process-global.
+    lock: ClassVar[Lock] = Lock()
+
+    if hasattr(os, "register_at_fork"):
+        # Runs once, when the class body is executed at import time.
+        # Without this, an os.fork() happening while some thread holds the lock
+        # (e.g. AGENT_MANIFEST_CONCURRENCY_CONTEXT="fork" while a watcher thread
+        # is mid-parse) would give the child a lock that can never be released,
+        # deadlocking the child's first HOCON parse. Holding the lock across the
+        # fork means children always start with it released - the same treatment
+        # the standard library logging module gives its module lock.
+        # Windows has no fork, hence the hasattr guard.
+        # The bare "lock" reference is required: the HoconParseLock name is not
+        # bound until the class body finishes executing.
+        os.register_at_fork(
+            before=lock.acquire,
+            after_in_parent=lock.release,
+            after_in_child=lock.release,
+        )
+
+    def __enter__(self) -> "HoconParseLock":
+        """
+        Acquire the process-wide parse lock.
+        :return: this instance
+        """
+        self.lock.acquire()
+        return self
+
+    def __exit__(self, exc_type: Any, exc_value: Any, traceback: Any) -> bool:
+        """
+        Release the process-wide parse lock.
+
+        :param exc_type: The type of any exception raised inside the with-block
+        :param exc_value: The exception instance, if any
+        :param traceback: The traceback, if any
+        :return: False, so any exception from the with-block propagates
+        """
+        self.lock.release()
+        return False
--- a/neuro_san/service/http/logging/logging_config_restorer.py
+++ b/neuro_san/service/http/logging/logging_config_restorer.py
@@ -26,6 +26,7 @@ from leaf_common.persistence.easy.easy_json_persistence import EasyJsonPersisten
 from leaf_common.persistence.easy.easy_yaml_persistence import EasyYamlPersistence
 
 from neuro_san import DEPLOY_DIR
+from neuro_san.internals.persistence.hocon_parse_lock import HoconParseLock
 
 
 class LoggingConfigRestorer(Restorer):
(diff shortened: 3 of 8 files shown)


tests/neuro_san/internals/persistence/test_hocon_parse_lock.py
```diff
@@ -0,0 +1,218 @@
...
+                time.sleep(hold_seconds)
+                # Set strictly before the lock release that lets the
+                # before-fork hook's acquire (and hence the fork) proceed.
+                released.set()
+
+        holder = threading.Thread(target=hold_lock, daemon=True)
+        holder.start()
+        # Make sure the holder actually has the lock before forking.
+        while not HoconParseLock.lock.locked():
+            time.sleep(0.001)
```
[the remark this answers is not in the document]
Noravee: Good catch — fixed in 6f834e2b. The poll loop now has a 5s monotonic deadline that fails the test via pytest.fail instead of spinning forever, matching the deadline-bounded reap_child() approach used later in the same test.

neuro_san/internals/persistence/hocon_parse_lock.py
```diff
@@ -0,0 +1,111 @@
...
+    every fork, per the hooks below) instead of just its own thread. Do not
+    use "include url(...)" in served registries.
+    """
+
+    # One lock for the whole process, shared by all instances,
+    # because the pyparsing state it guards is process-global.
+    # Never rebind this attribute: the register_at_fork hooks below capture
+    # bound methods of this exact object and cannot be re-registered, so a
+    # replacement lock would serialize parses without fork protection.
+    lock: ClassVar[Lock] = Lock()
```
Noravee, line 68: One `Lock` for the whole process, shared by all instances, because the pyparsing state it guards is process-global. The "never rebind" warning is load-bearing: the fork hooks below capture bound methods of this exact object and can never be re-registered, so a replacement lock would still serialize parses but silently lose fork protection.

neuro_san/internals/persistence/hocon_parse_lock.py
```diff
@@ -0,0 +1,111 @@
...
+        # instant, and a child-side release() would hang on that inherited
+        # mutex whose owner is dead (see bpo-40089). Windows has no fork,
+        # hence the hasattr guard.
+        # The bare "lock" references are required: the HoconParseLock name is
+        # not bound until the class body finishes executing.
+        os.register_at_fork(
+            before=lock.acquire,
+            after_in_parent=lock.release,
+            # pylint: disable=no-member,protected-access
+            after_in_child=lock._at_fork_reinit,
```
Noravee, line 90: The subtlest line in the PR. The child hook must be `_at_fork_reinit`, **not** `release`: on CPython <= 3.12 platforms without POSIX semaphores (macOS), a waiter blocked in `acquire()` can hold the lock's *internal* pthread mutex at the fork instant; the child inherits that mutex locked with its owner dead, so a child-side `release()` hangs inside `os.fork()` before reaching any user code (bpo-40089). Reproduced both ways on Python 3.12/macOS: deadlock with `release`, clean exit with `_at_fork_reinit`. This mirrors stdlib `logging`'s treatment of its module lock.

neuro_san/internals/persistence/abstract_async_config_restorer.py
```diff
@@ -155,18 +159,23 @@ def deserialize_file_contents(self, file_path: str, file_contents: bytes) -> Dic
...
         # Determine the serialization format
         serialization: SerializationFormat = None
+        parse_lock: ContextManager = nullcontext()
         if file_path.endswith(".hocon"):
             # Worth noting that includes within hocon files can only be done synchronously
             # The core pyhocon library does not support async includes.
             serialization = HoconSerializationFormat()
+            # pyhocon parsing mutates process-global pyparsing state and is not
+            # thread-safe. See HoconParseLock.
+            parse_lock = HoconParseLock()
```
Noravee, line 169: This is the main chokepoint in neuro-san: every restorer subclass (registries, manifests, llm-info, toolbox, MCP configs) funnels through `deserialize_file_contents`, so this one site covers the parallel manifest reads from #1177 — including the watcher's re-parse storms. JSON deliberately bypasses the lock via `nullcontext` (json parsing shares no pyparsing state). Lock scope is just `to_object`: measured >99% of that time is `ConfigFactory.parse_string` itself.

Noravee (commented):
**TL;DR**

- **Bug:** pyhocon rebuilds its pyparsing grammar on every parse while mutating a **process-global** (`ParserElement.DEFAULT_WHITE_CHARS`). Once #1177 parallelized manifest reads, concurrent parses could poison each other's grammar mid-build → the random `Expected end of text, found '\n'` / `"Nothing to validate."` failures on valid registries (#1183, reproduced character-for-character).
- **Fix:** one process-wide `HoconParseLock` held around every in-repo pyhocon parse (3 call sites), plus `os.register_at_fork` hooks so `AGENT_MANIFEST_CONCURRENCY_CONTEXT=fork` cannot deadlock a child forked mid-parse.
- **Cost:** none in practice — the GIL already serialized pure-Python parsing (>99% of the locked region is `ConfigFactory.parse_string` itself); process-pool modes are unaffected because each worker has its own lock.
- **Verified:** 30× threaded full-manifest soak with concurrent-parse churn = zero errors; `thread`/`spawn`/`fork` produce identical results; fork stressed 10/10; the mutual-exclusion regression test fails against the unfixed code.
- **Next:** move the lock upstream to leaf-common's `HoconSerializationFormat.to_object` chokepoint — tracked in cognizant-ai-lab/leaf-common#168; then this repo's lock module and the three `with` blocks get deleted.

neuro_san/service/http/logging/logging_config_restorer.py
```diff
@@ -58,13 +59,18 @@ def restore(self, file_reference: str = None) -> Dict[str, Any]:
...
-            if use_file_reference.endswith(".json"):
-                logging_config = EasyJsonPersistence().restore(file_reference=use_file_reference)
-            if use_file_reference.endswith(".yaml") or use_file_reference.endswith(".yml"):
-                logging_config = EasyYamlPersistence().restore(file_reference=use_file_reference)
+        elif use_file_reference.endswith(".json"):
+            logging_config = EasyJsonPersistence().restore(file_reference=use_file_reference)
+        elif use_file_reference.endswith(".yaml") or use_file_reference.endswith(".yml"):
+            logging_config = EasyYamlPersistence().restore(file_reference=use_file_reference)
         else:
             raise ValueError(f"Unsupported logging config file type: {use_file_reference}")
```
Noravee, line 75: Drive-by fix in the touched function: previously this `raise` sat on the `use_file_reference is None` else-branch, which is unreachable (the constructor fallback is never None) — so an unsupported extension silently returned `{}` and the server later crashed deep in leaf_common's `dictConfig` with a misattributed error. Now `if/elif/else` makes the intended error fire at the source. Behavior change: unsupported extensions fail fast instead of silently running with empty logging config.

neuro_san/test/driver/data_driven_agent_test_driver.py
```diff
@@ -254,7 +255,11 @@ def parse_hocon_test_case(self, hocon_file: str) -> Dict[str, Any]:
         if self.fixtures is not None:
             test_path = self.fixtures.get_file_in_basis(hocon_file)
         hocon = EasyHoconPersistence(must_exist=True)
-        test_case: Dict[str, Any] = hocon.restore(file_reference=test_path)
+        # pyhocon parsing mutates process-global pyparsing state and is not
+        # thread-safe. See HoconParseLock. Sessions driven by this class can
+        # leave background threads parsing agent hocons concurrently.
+        with HoconParseLock():
```
Noravee, line 261: Third parse site — found by review after the first version of this PR shipped with only two. `neuro_san.test` is packaged in the wheel (only `tests*` is excluded), and sessions driven by this class can leave background threads parsing agent hocons concurrently with the next test case's parse. The fact that the "every parse must hold the lock" convention was violated within the same PR that introduced it is the core argument for moving the lock into leaf-common's `to_object` (leaf-common#168).

tests/neuro_san/internals/persistence/test_hocon_parse_lock.py
```diff
@@ -0,0 +1,224 @@
...
+            time.sleep(0.01)
+
+        # Child overran the deadline: kill and reap so nothing leaks.
+        # waitstatus_to_exitcode() will report -SIGKILL, failing the caller's
+        # assertion with a recognizable value instead of hanging the run.
+        os.kill(pid, signal.SIGKILL)
+        _, status = os.waitpid(pid, 0)
+        return status
+
+    def test_concurrent_hocon_parses_are_mutually_exclusive(self) -> None:
```
Noravee, line 114: The core regression test: 16 parses across 8 threads, each `to_object` sleeps 50ms with the GIL released, so any two unserialized parses would overlap with near-certainty — `max_in_flight == 1` is deterministic proof of mutual exclusion, independent of machine load. Verified to FAIL against the unfixed code. Note it patches the name binding in the restorer module, so it pins this call site specifically; chokepoint-level coverage moves to leaf-common with #168.

tests/neuro_san/internals/persistence/test_hocon_parse_lock.py
```diff
@@ -0,0 +1,224 @@
...
+            thread = threading.Thread(
+                target=lambda: (self.deserialize("valid.json"), done.set()), daemon=True)
+            thread.start()
+            # The JSON parse completes even though the HOCON lock is held.
+            assert done.wait(timeout=5.0)
+        thread.join(timeout=5.0)
+        assert not thread.is_alive()
+
+    @pytest.mark.skipif(not hasattr(os, "fork"), reason="os.fork not available on this platform")
+    def test_fork_while_lock_held_leaves_child_usable(self) -> None:
```
Noravee, line 165: The fork test is deliberately paranoid, each guard closes a reviewed hole: (1) `released` Event asserts the before-fork hook waited out the holder — no wall-clock arithmetic, so no scheduler flakes; (2) `reap_child` polls with a 30s deadline + SIGKILL fallback so a hung child fails the test instead of hanging the whole pytest run; (3) `waitstatus_to_exitcode` (not bare `WEXITSTATUS`) so a signal-killed child reads as -N instead of a false-passing 0; (4) the child prints its traceback before `os._exit(3)` so CI failures aren't opaque exit codes.

tests/neuro_san/internals/persistence/conftest.py
```diff
@@ -0,0 +1,44 @@
...
+local fixture files and do not call any LLM provider, so the
+project-wide skip condition does not apply. This conftest overrides
+that fixture with a no-op for tests in this subtree only.
+"""
+import pytest
+
+
+# pylint: disable=unused-argument
+@pytest.fixture(autouse=True)
+def configure_llm_provider_keys(request, monkeypatch):
```
Noravee, line 38: Without this override, the project-wide autouse fixture silently SKIPS every test in this directory whenever `OPENAI_API_KEY` is unset — meaning the regression tests for this fix would not run on secret-less CI or contributor machines. These are pure threading/parsing unit tests with no LLM calls. Same pattern as `tests/neuro_san/service/watcher/temp_networks/conftest.py`.

d1donlydfink (approved):
Thanks for getting to the bottom of this.

Question: Would this lock be better situation at the leaf-common level inside
HoconSerializationFormat.to_object()?

Noravee:
@d1donlydfink Yes. Please see https://github.com/cognizant-ai-lab/leaf-common/issues/168 as well as the "Next step: move the fix upstream to leaf-common" section in the description.

PR #1187, opened 2026-07-31
Fix request-level token accounting double-counting; report main-network and request totals separately

### LINKED ISSUE #1186: Request-level token accounting double-counts usage for networks calling external agent networks
## Summary

Request-level token accounting is inflated (~2x) for agent networks that call other agent networks. Both the "Request reporting" block in `server.log` and the final client-visible accounting message (the one carrying the `models` key) over-report token usage and call counts. Per-agent (per-origin) accounting messages are correct.

## Observed behavior

Running `agent_network_designer` (which calls external networks via direct sessions on the same server):

- Reported request total: **1,192,157 tokens / 199 calls**
- Actual usage (summed from provider-side records): **569,464 tokens / 76 calls**

A simpler reproduction with `math_guy_passthrough` shows the same pattern: reported 2,004 tokens / 6 calls vs. actual 1,347 / 4 — the front man's subtree total gets re-added on top of the external network's own contribution.

## Root cause

Three interacting pieces:

1. Every agent's `LlmTokenCallbackHandler` is registered via `register_configure_hook(llm_token_callback_var, inheritable=True)`. Since commit `2646d20c` ("Get parenting correct", v0.6.55) introduced `merge_configs(ensure_config(), runnable_config)` for Langfuse trace parenting, ancestor handlers are *merged* (not replaced) into every descendant run's callbacks, so each handler's tallies are **subtree-cumulative**, not per-agent-exclusive. The older comments in `langchain_token_counter.py` claiming counts are "limited to the calling agent only" no longer hold.
2. `LangChainTokenCounter.report()` merges each agent's `models_token_dict` into `request_reporting["token_accounting"]` at **every** agent completion, so an LLM call observed at nesting depth N is added N times.
3. `SessionInvocationContext.safe_shallow_copy()` shares the same `request_reporting` dict with external networks invoked via direct sessions on the same server, so their front men merge into the same accumulator that the parent front man later re-merges.

Networks reached over HTTP are not affected (separate process/state). Agent-level per-origin accounting is subtree-cumulative by design and remains correct.

## Expected behavior

- The request-level accounting should count each LLM call exactly once.
- `token_accounting` should honor its documented caveat: main agent network usage only, external agents excluded.
- A new `total_token_accounting` entry should carry the true request total (including same-server externals invoked via direct sessions) along with the per-model breakdown, so the server log and the client-visible accounting agree.

Noravee:
Fixes #1186

## Problem

Request-level token accounting (the "Request reporting" block in `server.log` and the final client-visible accounting message carrying the `models` key) is inflated for agent networks that call other agent networks. Verified on `agent_network_designer`: reported **1,192,157 tokens / 199 calls** vs. an actual **569,464 / 76**. Per-agent (per-origin) accounting messages are correct and unchanged.

## Root cause

Since commit `2646d20c` (v0.6.55) introduced `merge_configs(...)` for Langfuse trace parenting, every ancestor's `LlmTokenCallbackHandler` (registered as an inheritable callback) hears all descendant LLM calls, making each handler's tallies **subtree-cumulative**. `LangChainTokenCounter.report()` then merges each agent's `models_token_dict` into the shared `request_reporting` at **every** agent completion, so a call at nesting depth N is counted N times. External networks invoked via direct sessions on the same server share the same `request_reporting` dict (via `safe_shallow_copy()`), compounding the inflation.

## Fix

- **Exclusive attribution in `LlmTokenCallbackHandler`**: per-model tallies (`models_token_dict`) are only updated when the handler is the nearest enclosing agent scope (`llm_token_callback_var.get() is self`). Scalar tallies stay subtree-cumulative so per-agent accounting messages are untouched. Merging every agent's exclusive contribution now sums each LLM call exactly once.
- **`token_accounting` = main network only**: restored to its historically documented semantics ("usage from external agents is not included"), reported without a `models` key.
- **New `total_token_accounting` = request total**: includes external agents invoked via direct sessions on this server, carries the per-model `models` breakdown, and is written last. The main network's front man streams both dicts verbatim as its two accounting messages, so the server log and the client-side thinking file now agree.
- The client-side `TokenAccountingMessageProcessor` keeps the last qualifying front-man message, so client state still ends up with the request total (with `models`) — no client-side change needed.

### Hardening included

- Accounting messages always carry `total_tokens` (zero-seeded stats), so `TokenAccountingMessageFilter` never drops them when a provider returns no `usage_metadata`.
- Cloned (external) reporters preserve the front man's `time_taken_in_seconds` and seed zeroed accounting if they report first.
- Front-man detection is structural (`len(origin) == 1` and not cloned) instead of checking for `"."` in the origin string, so dotted agent names work.
- `SessionInvocationContext.reset()` clears `request_reporting` so totals are per-exchange in library mode.
- A caveat is appended when agents that did not complete leave tokens unattributed.

### Format note for log scrapers

`token_accounting.models` moved to `total_token_accounting.models`, and `token_accounting`'s numbers shrink to main-network-only.

## Testing

- New integration test builds a passthrough topology (front man + internal agent + cloned external network) with a fake usage-emitting chat model and asserts main = 30/2, total = 45/3, key order pinned.
- New unit tests for exclusive model attribution, front-man gating, cloned-reporter behavior, late-external latency preservation, unattributed-token caveat, and `safe_shallow_copy()` sharing semantics.
- Full suite: 348 passed / 2 skipped; pylint 10.00/10; flake8 clean.

### DIFF
--- a/neuro_san/internals/interfaces/invocation_context.py
+++ b/neuro_san/internals/interfaces/invocation_context.py
@@ -119,6 +119,19 @@ class InvocationContext(LingeringResource):
         """
         raise NotImplementedError
 
+    def is_cloned(self) -> bool:
+        """
+        :return: True if this instance is a clone of a request's original
+                InvocationContext, created to invoke an external agent network
+                on the same server via a direct session.
+                False for the original InvocationContext of a request.
+
+        Note: Not raising NotImplementedError here on purpose.
+        Never being cloned is a reasonable default for implementations
+        that do not deal in same-server external agents.
+        """
+        return False
+
     def get_llm_factory(self) -> ContextTypeLlmFactory:
         """
         :return: The ContextTypeLlmFactory instance for the session
--- a/neuro_san/internals/interfaces/invocation_context.py
+++ b/neuro_san/internals/interfaces/invocation_context.py
@@ -125,12 +125,8 @@ class InvocationContext(LingeringResource):
                 InvocationContext, created to invoke an external agent network
                 on the same server via a direct session.
                 False for the original InvocationContext of a request.
-
-        Note: Not raising NotImplementedError here on purpose.
-        Never being cloned is a reasonable default for implementations
-        that do not deal in same-server external agents.
         """
-        return False
+        raise NotImplementedError
 
     def get_llm_factory(self) -> ContextTypeLlmFactory:
         """
--- a/neuro_san/internals/run_context/langchain/token_counting/get_llm_token_callback.py
+++ b/neuro_san/internals/run_context/langchain/token_counting/get_llm_token_callback.py
@@ -22,10 +22,9 @@ from typing import Dict
 
 from neuro_san.internals.run_context.langchain.token_counting.llm_token_callback_handler \
     import LlmTokenCallbackHandler
-# Re-exported for backwards compatibility: the ContextVar (and its langchain
-# configure-hook registration) lives with the handler class so the handler can
-# consult it to attribute events to their owning agent.
-# pylint: disable=unused-import
+# The ContextVar (and its langchain configure-hook registration) lives with the
+# handler class so the handler can consult it to attribute events to their
+# owning agent.
 from neuro_san.internals.run_context.langchain.token_counting.llm_token_callback_handler \
     import llm_token_callback_var
 
--- a/neuro_san/internals/run_context/langchain/token_counting/get_llm_token_callback.py
+++ b/neuro_san/internals/run_context/langchain/token_counting/get_llm_token_callback.py
@@ -17,6 +17,7 @@
 
 from collections.abc import Generator
 from contextlib import contextmanager
+from contextvars import Token
 from typing import Any
 from typing import Dict
 
@@ -48,11 +49,18 @@ def get_llm_token_callback(llm_infos: Dict[str, Any]) -> Generator[LlmTokenCallb
     # Create a new callback handler instance for tracking token usage
     cb = LlmTokenCallbackHandler(llm_infos)
 
-    # Set the context variable to the newly created callback handler
-    llm_token_callback_var.set(cb)
+    # Set the context variable to the newly created callback handler,
+    # keeping the token so the previous value can be restored on exit.
+    token: Token = llm_token_callback_var.set(cb)
 
-    # Yield the callback handler to the context block
-    yield cb
-
-    # Reset the context variable to None when the context exits
-    llm_token_callback_var.set(None)
+    try:
+        # Yield the callback handler to the context block
+        yield cb
+    finally:
+        # Restore the context variable to the value it had on entry - even if
+        # the block raised.  In practice each agent scope runs its work in its
+        # own copied Context (see LangChainTokenCounter.count_tokens()), so no
+        # other scope ever observes this mutation, but restoring (rather than
+        # clobbering to None) keeps this context manager correct on its own,
+        # independent of how callers arrange their contexts.
+        llm_token_callback_var.reset(token)
--- a/neuro_san/internals/run_context/langchain/token_counting/get_llm_token_callback.py
+++ b/neuro_san/internals/run_context/langchain/token_counting/get_llm_token_callback.py
@@ -17,21 +17,17 @@
 
 from collections.abc import Generator
 from contextlib import contextmanager
-from contextvars import ContextVar
 from typing import Any
 from typing import Dict
-from typing import Optional
-
-from langchain_core.tracers.context import register_configure_hook
 
 from neuro_san.internals.run_context.langchain.token_counting.llm_token_callback_handler \
     import LlmTokenCallbackHandler
-
-
-llm_token_callback_var: ContextVar[Optional[LlmTokenCallbackHandler]] = (
-        ContextVar("llm_token_callback", default=None)
-    )
-register_configure_hook(llm_token_callback_var, inheritable=True)
+# Re-exported for backwards compatibility: the ContextVar (and its langchain
+# configure-hook registration) lives with the handler class so the handler can
+# consult it to attribute events to their owning agent.
+# pylint: disable=unused-import
+from neuro_san.internals.run_context.langchain.token_counting.llm_token_callback_handler \
+    import llm_token_callback_var
 
 
 @contextmanager
--- a/neuro_san/internals/run_context/langchain/token_counting/langchain_token_counter.py
+++ b/neuro_san/internals/run_context/langchain/token_counting/langchain_token_counter.py
@@ -41,13 +41,41 @@ from neuro_san.internals.journals.journal import Journal
 from neuro_san.internals.messages.agent_message import AgentMessage
 from neuro_san.internals.messages.origination import Origination
 from neuro_san.internals.run_context.langchain.token_counting.get_llm_token_callback import get_llm_token_callback
-from neuro_san.internals.run_context.langchain.token_counting.get_llm_token_callback import llm_token_callback_var
+from neuro_san.internals.run_context.langchain.token_counting.llm_token_callback_handler import llm_token_callback_var
 
 # Keep a ContextVar for the origin info.  We do this because the
 # langchain callbacks this stuff is based on also uses ContextVars
 # and we want to be sure these are in sync.
 # See: https://docs.python.org/3/library/contextvars.html
 ORIGIN_INFO: ContextVar[str] = ContextVar('origin_info', default=None)
+
+# Baseline for aggregated token stats.  Aggregations start from these zeros so
+# request-level accounting always carries the standard keys even when no LLM
+# usage was recorded - clients key off "total_tokens" being present
+# (see TokenAccountingMessageFilter).
+ZERO_TOKEN_STATS: Dict[str, Any] = {
+    "total_tokens": 0,
+    "prompt_tokens": 0,
+    "completion_tokens": 0,
+    "successful_requests": 0,
+    "empty_responses": 0,
+    "total_cost": 0.0,
+}
+
+# Caveats shared by every token accounting message.
+COMMON_CAVEATS: List[str] = [
+    "Token counts are approximate and estimated using tiktoken.",
+    "time_taken_in_seconds includes overhead from Langchain and Neuro SAN"
+]
+
+# Coverage statements for the two request-level accountings.
+MAIN_NETWORK_CAVEAT: str = \
+    "Token usage from agents of the main agent network only. " \
+    "Usage from external agents is not included."
+TOTAL_CAVEAT: str = \
+    "Request total: includes token usage from external agents invoked via direct " \
+    "sessions on this server. Usage from external agents reached over the network " \
+    "is not included."
 
 
 class LangChainTokenCounter:
@@ -129,6 +157,12 @@ class LangChainTokenCounter:
         #   with each call counted exactly once, no matter how deeply agents nest.
         # * External agents on *remote* servers handle their own requests in a separate
         #   context and are not seen by this handler at all.
+        # Note on cancellation: if this task is cancelled outright (e.g. client
+        # disconnect), CancelledError propagates and report() below never runs -
+        # by design, since journaling from a cancelled task would fail anyway.
+        # The timeout path does report.  Descendant scopes cancelled mid-run are
+        # accounted for by the "unattributed tokens" caveat the front man adds
+        # in report().
         timed_out: bool = False
         with get_llm_token_callback(llm_infos) as callback:
             # Create a new context for different ContextVar values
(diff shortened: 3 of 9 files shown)


neuro_san/internals/run_context/langchain/token_counting/llm_token_callback_handler.py
```diff
@@ -102,6 +128,16 @@ def __repr__(self) -> str:
...
 
+    def _is_own_call(self) -> bool:
+        """
+        :return: True if the event being handled belongs to an LLM call made by this
+                handler's own agent.  False for calls made by downstream agents, whose
+                events this handler also receives because handlers are inheritable
+                callbacks.  At event time, llm_token_callback_var holds the handler
+                of the nearest enclosing agent scope - the call's owner.
+        """
+        return llm_token_callback_var.get() is self
```
Noravee, line 147: The heart of the fix. Handlers are registered as *inheritable* callbacks (`register_configure_hook(..., inheritable=True)` at the top of this module), so one handler instance receives events for its own agent's LLM calls **and** for every downstream agent's calls. At event time, `llm_token_callback_var` holds the handler of the nearest enclosing agent scope — the call's owner — so `is self` is True for exactly one handler per LLM call. This is what lets `report()` merge per-agent contributions without double-counting.

neuro_san/internals/run_context/langchain/token_counting/llm_token_callback_handler.py
```diff
@@ -175,23 +235,24 @@ async def on_llm_end(self, response: LLMResult, **kwargs: Any):
...
-                    completion_tokens
-                self.models_token_dict[self.provider_class][model_name]["successful_requests"] += 1
-                self.models_token_dict[self.provider_class][model_name]["empty_responses"] += \
-                    int(is_empty_response)
-                self.models_token_dict[self.provider_class][model_name]["total_cost"] += total_cost
-                self.models_token_dict[self.provider_class][model_name]["time_taken_in_seconds"] += \
-                    time_taken_in_seconds
-
-                # Update per-agent stats
+                if is_own_call:
```
Noravee, line 246: Dual bookkeeping, on purpose: the scalar tallies above (`total_tokens`, `successful_requests`, ...) update *unconditionally*, staying subtree-cumulative — they feed the per-agent accounting messages, which are intentionally unchanged (a front man's per-agent message still covers its whole network). Only this per-model dict is exclusive to the agent's own calls.

neuro_san/internals/run_context/langchain/token_counting/langchain_token_counter.py
```diff
@@ -186,45 +224,110 @@ async def report(self, callback: AsyncCallbackHandler, time_taken_in_seconds: fl
...
+        # The front man of the main network streams these two dictionaries verbatim
+        # as its two accounting messages, so the server log and the client-visible
+        # messages read the same.
         request_reporting: Dict[str, Any] = self.invocation_context.get_request_reporting()
-        token_accounting: Dict[str, Any] = request_reporting.get("token_accounting", {})
+        total_accounting: Dict[str, Any] = request_reporting.get("total_token_accounting", {})
         models_token_dict: Dict[str, Any] = \
-            self.merge_dicts(token_accounting.get("models", {}), callback.models_token_dict)
-        network_token_dict: Dict[str, Any] = self.sum_all_tokens(models_token_dict, time_taken_in_seconds)
+            self.merge_dicts(total_accounting.get("models", {}), callback.models_token_dict)
```
Noravee, line 252: Because each agent's `models_token_dict` now contains only its *own* calls, merging every completing agent's dict into the shared `request_reporting` yields each LLM call exactly once. Before this PR these dicts were subtree-cumulative, so a call at nesting depth N was added N times — the source of the ~2x inflation.

neuro_san/internals/run_context/langchain/token_counting/langchain_token_counter.py
```diff
@@ -186,45 +224,110 @@ async def report(self, callback: AsyncCallbackHandler, time_taken_in_seconds: fl
...
-            "time_taken_in_seconds includes overhead from Langchain and Neuro SAN"
-        ]
-        request_reporting["token_accounting"] = \
+        network_token_dict["caveats"] = [TOTAL_CAVEAT] + COMMON_CAVEATS
+
+        # Maintain the accounting covering only the main (top-level) agent network.
+        # Agents of external networks invoked via direct sessions run on cloned
+        # InvocationContexts, so they contribute to the totals but not to this one.
+        main_accounting: Dict[str, Any] = request_reporting.get("token_accounting", {})
+        if not self.invocation_context.is_cloned():
```
Noravee, line 270: Same-server external networks (invoked via direct sessions) run on **cloned** invocation contexts, so this branch skips them: they contribute to `total_token_accounting` above, but not to `token_accounting`. That restores this key's historically documented semantics ("usage from external agents is not included"), which the old implementation never actually delivered.

neuro_san/internals/run_context/langchain/token_counting/langchain_token_counter.py
```diff
@@ -186,45 +224,110 @@ async def report(self, callback: AsyncCallbackHandler, time_taken_in_seconds: fl
...
-        # Token counting results are collected in the callback.
-        # Create a token counting dictionary for each agent
-        agent_token_dict: Dict[str, Any] = self._generate_agent_token_dict(callback, time_taken_in_seconds)
+        # Figure out whether we are the front man of the main network:
+        # * Only front men sit at the root of the origin path (single-element origin).
+        # * Front men of same-server external networks invoked via direct sessions
+        #   also have single-element origins, but they run on a cloned InvocationContext.
+        is_main_front_man: bool = \
+            self.origin is not None and len(self.origin) == 1 and \
+            not self.invocation_context.is_cloned()
```
Noravee, line 299: Structural front-man detection: only front men sit at the root of the origin path (single-element origin). Front men of *external* networks also have single-element origins, but they run on cloned contexts — `is_cloned()` tells them apart. This replaces a `"." not in origin` string check, which would have broken on dotted agent names.

neuro_san/internals/run_context/langchain/token_counting/langchain_token_counter.py
```diff
@@ -186,45 +224,110 @@ async def report(self, callback: AsyncCallbackHandler, time_taken_in_seconds: fl
...
+
+        if is_main_front_man:
+            # This front man's scalar totals cover everything its handler heard
+            # (its whole subtree).  Anything above what completed scopes merged
+            # into the total (e.g. LLM calls of agents cancelled mid-run) is
+            # unattributed - say so rather than silently under-reporting.
+            total_token_dict: Dict[str, Any] = request_reporting["total_token_accounting"]
+            unattributed_tokens: int = callback.total_tokens - total_token_dict["total_tokens"]
```
Noravee, line 307: Safety net for cancelled work: the front man's subtree scalars are ground truth for everything its handler heard. If agents were cancelled mid-run (e.g. a timeout), their scopes never reported, so their calls are missing from the merged total — this caveat surfaces the gap instead of silently under-reporting.

neuro_san/internals/run_context/langchain/token_counting/langchain_token_counter.py
```diff
@@ -186,45 +224,110 @@ async def report(self, callback: AsyncCallbackHandler, time_taken_in_seconds: fl
...
+                # The front man of the main network reports the two request-level
+                # accountings as two complementary messages - main network first,
+                # request total second - each stating its coverage in its caveats.
+                # External front men do not do this: their tokens are already merged into
+                # the shared request_reporting above, and a request-level message from them
+                # mid-request would report a partial (and potentially sibling-polluted) total.
+                await self.journal.write_message(
+                    AgentMessage(structure=request_reporting["token_accounting"]))
+                await self.journal.write_message(
+                    AgentMessage(structure=request_reporting["total_token_accounting"]))
```
Noravee, line 325: These two dicts are streamed **verbatim** — the same structures the service logs from `request_reporting` — so the server log and the client-visible thinking file are now identical. Client-side, `TokenAccountingMessageProcessor` keeps the *last* qualifying front-man message, so client state still ends up with the request total (with `models`): no client changes needed.

Noravee (commented):
## TL;DR

**The bug:** every agent's token handler is an *inheritable* langchain callback, so an ancestor's handler also hears all descendant LLM calls (subtree-cumulative tallies). `report()` merged each completing agent's per-model tallies into the shared `request_reporting`, so a call at nesting depth N was counted N times (~2x on real networks calling external networks).

**The fix, in one line:** per-model tallies (`models_token_dict`) are now updated only by the handler of the agent that *owns* the call (`llm_token_callback_var.get() is self`), so merging every agent's contribution at completion counts each LLM call exactly once. Scalar tallies stay subtree-cumulative on purpose — per-agent accounting messages are unchanged.

**Reporting format:** `request_reporting` now holds two entries, printed main-first / total-last:
- `token_accounting` — main network only (no `models` key), finally honoring its long-documented caveat that external agent usage is not included
- `total_token_accounting` — request total including same-server externals invoked via direct sessions, with the per-model `models` breakdown

The main network's front man streams both dicts verbatim as its two accounting messages, so the server log and the client-visible messages read the same. The client-side processor keeps the last qualifying message, so client state still ends up with the total — no client changes needed.

neuro_san/session/session_invocation_context.py
```diff
@@ -263,6 +272,11 @@ def reset(self):
         # in subsequent interactions with the same network.
         self.origination.reset()
 
+        # Token accounting is per-exchange ("Request total"), so clear it for the
+        # next exchange rather than accumulating across turns.  Clear in place:
+        # the dictionary instance is shared by reference with any clones.
+        self.request_reporting.clear()
```
Noravee, line 278: Totals are per-exchange: without this, library-mode multi-turn conversations (reusing one `DirectAgentSession`) would accumulate token counts across turns. Cleared *in place* because clones share this exact dict instance by reference.

neuro_san/session/session_invocation_context.py
```diff
@@ -288,10 +302,18 @@ def safe_shallow_copy(self, invocation: str = None) -> SessionInvocationContext:
...
+        #   invoked on this server contributes to the request-wide totals, with
+        #   each LLM call counted exactly once (see LangChainTokenCounter.report()).
+        # * origination - so tool instantiation indices stay consistent across
+        #   the whole request.
+
         # Mark the invocation context as cloned
         # This tells us that it is not the original/root and will prevent
         # mis-happenings like returning the executor to the pool too early.
-        invocation_context.is_cloned = True
+        invocation_context.cloned = True
```
Noravee, line 321: This marker is what `LangChainTokenCounter.report()` uses to tell main-network agents apart from same-server external ones. (The attribute was renamed `is_cloned` -> `cloned` to make room for the `is_cloned()` interface accessor.)

neuro_san/internals/interfaces/invocation_context.py
```diff
@@ -119,6 +119,19 @@ def get_request_reporting(self) -> Dict[str, Any]:
...
+        :return: True if this instance is a clone of a request's original
+                InvocationContext, created to invoke an external agent network
+                on the same server via a direct session.
+                False for the original InvocationContext of a request.
+
+        Note: Not raising NotImplementedError here on purpose.
+        Never being cloned is a reasonable default for implementations
+        that do not deal in same-server external agents.
+        """
+        return False
```
d1donlydfink: Best to raise NotImplementedError here like all the other methods in this interface.

neuro_san/internals/run_context/langchain/token_counting/llm_token_callback_handler.py
d1donlydfink: Hmm.Tricky.
Worth putting more of this PR comment into the code comments?

d1donlydfink (changes requested):
* InvocationContext method should raise NotImplementedError just like everything else in there.   SessionInvocationContext is really the sole implementation and will take care of the meat.
* While your commenting is usually very thorough, I felt some of the PR comments near the "crux of the change" should actually be in the code
* Missing? SessionInvocationContext has a method called safe_shallow_copy(). This is where is_cloned gets turned to True when an external agent on the same server is called.  Should there be any resetting/reframing of request_reporting in there?
It's quite possible that what you have is fine, I just find this is a case that is often overlooked.

Question:  Were we generally undercounting tokens/cost before? or overcounting?

neuro_san/internals/run_context/langchain/token_counting/get_llm_token_callback.py
```diff
@@ -17,21 +17,16 @@
...
-
-llm_token_callback_var: ContextVar[Optional[LlmTokenCallbackHandler]] = (
-        ContextVar("llm_token_callback", default=None)
-    )
-register_configure_hook(llm_token_callback_var, inheritable=True)
+# The ContextVar (and its langchain configure-hook registration) lives with the
+# handler class so the handler can consult it to attribute events to their
+# owning agent.
+from neuro_san.internals.run_context.langchain.token_counting.llm_token_callback_handler \
+    import llm_token_callback_var
```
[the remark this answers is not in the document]
Noravee: Fixed in d6178892 — `get_llm_token_callback()` now uses the canonical `token = var.set(cb)` / `var.reset(token)` pattern inside `try/finally`, so the previous value is restored (rather than clobbered to `None`) even when the block raises.

Worth noting the failure mode described here could not actually occur in the current architecture: each agent scope runs its work inside its own copied `Context` (`LangChainTokenCounter.count_tokens()` does `copy_context()` and runs the chain as a task within it, and asyncio tasks/langchain's `coro_with_context` copy contexts as well), so an inner scope's exit mutation was never visible to the outer scope's context. The integration test (`test_token_accounting_integration.py`) exercises exactly this nesting with correct attribution. That said, the reset-token pattern makes the context manager correct on its own, independent of how callers arrange their contexts — a strictly better shape, so adopted.

neuro_san/internals/interfaces/invocation_context.py
Noravee: Done — it now raises NotImplementedError like the rest of the interface.

neuro_san/internals/run_context/langchain/token_counting/llm_token_callback_handler.py
Noravee: Done — folded this explanation into the code: the module comment above llm_token_callback_var and the _is_own_call() docstring now spell out the exactly-once ownership.

Noravee:
Thanks for the review!:

- `is_cloned()` now raises `NotImplementedError` like the rest of the interface.
- Moved the exactly-once ownership explanation from the PR comments into the code (module comment above `llm_token_callback_var` and the `_is_own_call()` docstring).
- Good catch on `safe_shallow_copy()` — the sharing of `request_reporting` there is deliberate (a fresh dict would orphan the external network's usage from the request totals), and the staleness case is handled by `reset()`, which this PR changed to clear the shared dict in place per exchange. Added a comment in `safe_shallow_copy()` saying exactly that.

To your question: we were **overcounting** — roughly 2x for networks that call external networks (a call at nesting depth N was merged N times), and cost was inflated the same way. Per-agent accounting was unaffected.

d1donlydfink (approved):

PR #1189, opened 2026-07-31
Fix sensitive logging situation in HttpLlmTracer

d1donlydfink:
New issue found in SAST scan.

### DIFF
--- a/neuro_san/service/utils/http_llm_tracer.py
+++ b/neuro_san/service/utils/http_llm_tracer.py
@@ -33,6 +33,8 @@ import time
 import uuid
 import httpx
 
+from leaf_common.logging.sensitive_logger import SensitiveLogger
+
 
 class HttpxLlmTracer:
     """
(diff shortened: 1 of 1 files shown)


neuro_san/service/utils/http_llm_tracer.py
```diff
@@ -188,6 +191,7 @@ def install(cls,
         cls._include_bodies = include_bodies
         cls._include_chunks = include_chunks
         cls._logger = logging.getLogger("neuro_san.diagnostics.http_llm_trace")
+        cls._sensitive_logger = SensitiveLogger(cls._logger)
```
d1donlydfink, line 194: Have a class-wide SensitiveLogger that wraps the class logger.

neuro_san/service/utils/http_llm_tracer.py
```diff
@@ -457,7 +461,7 @@ def _emit(cls, event: str, **fields: Any) -> None:
                 "user_req_id": cls._user_req_id_var.get(),
                 **fields,
             }
-            cls._logger.info(json.dumps(payload, default=str))
+            cls._sensitive_logger.info(json.dumps(payload, default=str))
```
d1donlydfink, line 464: This is the line that was flagged in the SAST scan.
It will continue to be flagged in the SAST scan, but at least we have it behind our SensitiveLogger policy, which more easily relegates it to a false-positive upon sighting it in future SAST scans.

andreidenissov-cog (approved):
Good catch.

PR #1190, opened 2026-07-31
NS-1188 VEnv creation for Python3.14t

andreidenissov-cog:
This PR adds a bash script creating a neuro-san virtual environment based on Python3.14t - free threading.
Also added is a GIL state reporting on service startup, so we know what we are running.

Tested by creating said virtual environment and running neuro-san server under it.


neuro_san/service/main_loop/server_main_loop.py
```diff
@@ -294,6 +295,10 @@ def main_loop(self):
             storage: AgentNetworkStorage = network_storage_dict.get(storage_class)
             storage.setup_agent_networks(self.agent_networks.get(storage_class))
 
+        # Report the process's GIL / free-threading state so logs make it
+        # unambiguous whether the server is actually running free-threaded.
+        GilStateReporter.report()
+
```
andreidenissov-cog: Add GIL state reporting.

build_scripts/make_venv_py314t.sh
```diff
@@ -0,0 +1,271 @@
+#!/bin/bash
+
+# Copyright © 2023-2026 Cognizant Technology Solutions Corp, www.cognizant.com.
+#
```
andreidenissov-cog, line 4: VEnv creation script - reasonably well documented in its source.

neuro_san/service/utils/gil_state_reporter.py
```diff
@@ -0,0 +1,114 @@
...
+        elif state["gil_enabled"] is True:
+            logger.warning(
+                "GIL state: free-threaded build but GIL is ENABLED - free-threading NOT in effect "
+                "(a loaded native extension re-enabled it, or PYTHON_GIL=1). %s. PYTHON_GIL=%s",
+                runtime, env_str)
+        else:
+            logger.info(
+                "GIL state: free-threaded build; runtime GIL state not queryable on this interpreter. "
+                "%s. PYTHON_GIL=%s",
+                runtime, env_str)
```
d1donlydfink: Please watch for SAST scans that concern these new logger lines on the next release.

build_scripts/make_venv_py314t.sh
```diff
@@ -0,0 +1,277 @@
...
+#   AUTO_INSTALL_RUST=1      install a Rust toolchain (rustup) automatically if missing
+
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(cd "${SCRIPT_DIR}/.." && pwd)"
+
+PYTHON_VERSION="${PYTHON_VERSION:-3.14t}"
+LEAF_COMMON_DIR="${LEAF_COMMON_DIR:-${REPO_ROOT}/../leaf-common}"
+LEAF_SERVER_COMMON_DIR="${LEAF_SERVER_COMMON_DIR:-${REPO_ROOT}/../leaf-server-common}"
```
d1donlydfink: What if you installed leaf-server-common and leaf-common as specified to an alternate venv
and took the "source" from inside that venv?    Then you wouldn't have to do wonky pinning of version numbers in here and expect code to be installed at particular places (cuz that is what pip does!)

build_scripts/make_venv_py314t.sh
```diff
@@ -0,0 +1,277 @@
...
+
+# Create a local virtual environment for running/developing the neuro-san
+# service on FREE-THREADED CPython 3.14t (no-GIL). This script mirrors the
+# choices made in the py314t deployment container build (where applicable):
+#
+#   * The free-threaded interpreter is provisioned with `uv` (there is no
+#     official python:3.14t image or, usually, a system 3.14t interpreter).
+#   * The in-house libraries leaf-common and leaf-server-common are built and
+#     installed FROM LOCAL SOURCE (never PyPI); their two lines are stripped
+#     from requirements.txt so the resolver does not fetch them from the index.
```
d1donlydfink: It's not immediately clear why it's necessary to build leaf-common and leaf-server-common from local source.  Can you please explain up here?

build_scripts/make_venv_py314t.sh
```diff
@@ -0,0 +1,277 @@
...
+        [ -f "${d}/pyproject.toml" ] || die "local dependency source not found at '${d}' (no pyproject.toml).
+   Set LEAF_COMMON_DIR / LEAF_SERVER_COMMON_DIR to point at the local repos."
+    done
+
+    ensure_uv
+    ensure_build_toolchain
+
+    local leaf_common_version leaf_server_common_version
+    leaf_common_version="${LEAF_COMMON_VERSION:-$(derive_version "${LEAF_COMMON_DIR}" "1.2.43")}"
+    leaf_server_common_version="${LEAF_SERVER_COMMON_VERSION:-$(derive_version "${LEAF_SERVER_COMMON_DIR}" "0.1.17")}"
```
d1donlydfink: Simply looking for a way to get these pinned versions out of your script and have their truth come from a single source like requirements.txt

build_scripts/make_venv_py314t.sh
andreidenissov-cog: OK all this "get local leaf-*.py sources and build them together" nonsense is removed.
Now we just take them as normal dependencies from PyPi.
Problems if any will not be in that area.

build_scripts/make_venv_py314t.sh
d1donlydfink: It's possible we might want to move this to neuro_san/deploy because we might want to export it.
(Stuff in build_scripts never gets exported as part of the library)

neuro_san/service/utils/gil_state_reporter.py
```diff
@@ -0,0 +1,114 @@
...
+        gil_query = getattr(sys, "_is_gil_enabled", None)
+        # pylint: disable=not-callable
+        gil_enabled: Optional[bool] = gil_query() if callable(gil_query) else None
+        return {
+            "implementation": platform.python_implementation(),
+            "python_version": platform.python_version(),
+            "free_threaded_build": bool(int(sysconfig.get_config_var("Py_GIL_DISABLED") or 0)),
+            "gil_enabled": gil_enabled,
+            "python_gil_env": os.environ.get("PYTHON_GIL"),
+        }
```
d1donlydfink, line 70: I think this python version information is already spat out in entrypoint.sh
Does this reporting really belong in there?

neuro_san/service/utils/gil_state_reporter.py
```diff
@@ -0,0 +1,114 @@
...
+                 "gil_enabled" is None when the interpreter is too old to query
+                 (pre-3.13), where the GIL is always enabled.
+        """
+        gil_query = getattr(sys, "_is_gil_enabled", None)
+        # pylint: disable=not-callable
+        gil_enabled: Optional[bool] = gil_query() if callable(gil_query) else None
+        return {
+            "implementation": platform.python_implementation(),
+            "python_version": platform.python_version(),
+            "free_threaded_build": bool(int(sysconfig.get_config_var("Py_GIL_DISABLED") or 0)),
```
d1donlydfink, line 67: Wow, is it really called "Py_GIL_DISABLED" (all those caps and a small 'y') ?

Haven't seen a more inconsistent config key name in a long time ... and it's new.

d1donlydfink (dismissed):
OK for now.
I can see:
* We are likely to want to eventually move make_venv_py314t.sh out of build_scripts and into deploy, because we do not ship anything in build_scripts (nor do we ever use it in Dockerfiles)
* It doesn't look like we will percolate the GIL state down anywhere else.  What you have is OK, but it's really for a bunch of logging. While I can see that for now it's nice to have assurances from Python about what the app thinks its GIL state is,  when confidence is gained, strongly consider moving the GIL reporting out of the server and into entrypoint.sh

d1donlydfink (approved):

PR #1192, opened 2026-08-02
Register Langfuse handler lazily, gated on LANGFUSE_ENABLED

### LINKED ISSUE #1191: Langfuse CallbackHandler registers at import time regardless of LANGFUSE_ENABLED - traces export whenever langfuse is installed and keys are present
## Summary

`LangfuseTracingContext` registers its `langfuse.langchain.CallbackHandler` with LangChain's global `register_configure_hook` **at import time**, gated only on whether the `langfuse` package is importable — not on `LANGFUSE_ENABLED`. As a result:

1. In any process where `langfuse` is installed and `LANGFUSE_SECRET_KEY`/`LANGFUSE_PUBLIC_KEY` are present in the environment, **trace data (prompts/completions) is exported to the Langfuse host even when `LANGFUSE_ENABLED` is `false` or unset**. The flag the docs present as the on/off switch does not actually stop export.
2. In processes with `langfuse` installed but no keys (e.g. any dev machine once a downstream repo adds `langfuse` to its requirements), every process start logs `Authentication error: Langfuse client initialized without public_key. Client will be disabled. ...`, and the disabled handler is still dispatched on **every** LangChain invocation (verified: its `on_chain_start` fires on a bare `RunnableLambda.invoke()`), plus a noticeable cold-import cost (langfuse + OpenTelemetry stack, ~200 extra modules).
3. Because LangChain dedupes configure-hook handlers **by object identity only**, any deployment that registers its own Langfuse handler produces a second handler and **every observation in every trace appears exactly twice**, with doubled token usage/cost reporting. This actually happened in neuro-san-studio (see cognizant-ai-lab/neuro-san-studio#1292; symptom also reported as cognizant-ai-lab/neuro-san-studio#1256, where the doubled span trees were interpreted as the agent executing twice).

## Root cause

In `neuro_san/internals/run_context/langchain/tracing/langfuse_tracing_context.py` (0.6.88), the class body ends with:

```python
# Do the register() once at class load time.
HANDLER_CONTEXT_VAR: ContextVar = register()
```

`register()` resolves `langfuse.langchain.CallbackHandler` via `ResolverUtil.create_type(..., raise_if_not_found=False)`, instantiates it (which constructs the Langfuse client from the `LANGFUSE_*` env vars, spawning its OTel export threads when keys are present), and calls `register_configure_hook(context_var, inheritable=True)`. There is no `LANGFUSE_ENABLED` check anywhere in this path.

`LANGFUSE_ENABLED` is read in exactly one place — `langchain_tracing_context_factory.py` (`getenv("LANGFUSE_ENABLED", "false").lower() == "true"`) — and it only selects which `TracingContext` class wraps a request (root AGENT span, `sessionId`/`userId` metadata, per-request flush). It never gates the global handler registration. The factory module imports `LangfuseTracingContext` unconditionally at module top and is reached via the core chat path (`data_driven_chat_session` → `master_tracing_context_factory`), so the registration side effect happens in every process that touches a chat session.

Because registration is a process-lifetime global performed at import, changing `LANGFUSE_ENABLED` on a running server has no effect on it, and the effective on/off switch is "is the `langfuse` package installed?" — which is surprising and not what the env var promises.

## Reproduction

Export with the flag off:

```bash
pip install neuro-san==0.6.88 langfuse==4.6.1
export LANGFUSE_SECRET_KEY=sk-lf-... LANGFUSE_PUBLIC_KEY=pk-lf-... LANGFUSE_HOST=https://us.cloud.langfuse.com
export LANGFUSE_ENABLED=false
# start any neuro-san server / run any agent session, send one request
```

The request produces a trace in Langfuse (a "bare" span tree — no root AGENT span or session metadata, since the factory-side context is off). Minimal in-process check:

```python
import neuro_san.internals.chat.data_driven_chat_session  # triggers class-load registration
from langchain_core.runnables import RunnableLambda
RunnableLambda(lambda x: x + 1).invoke(1)  # handler fires; with keys set, spans export
```

Warning-noise variant: unset all `LANGFUSE_*` vars, import the same module — the SDK's `Authentication error: ... Client will be disabled.` warning is logged at import, and the disabled handler is still attached to every subsequent LangChain run.

## Suggested fix

Four parts, all in `langfuse_tracing_context.py`:

**1. Register lazily instead of at class load.** Move the `register()` call into a lock-guarded classmethod invoked from `__init__`:

```python
class LangfuseTracingContext(LangChainTracingContext):

    HANDLER_CONTEXT_VAR: ContextVar = None
    _REGISTER_LOCK = threading.Lock()

    @classmethod
    def _ensure_registered(cls) -> ContextVar:
        with cls._REGISTER_LOCK:
            if cls.HANDLER_CONTEXT_VAR is not None:
                return cls.HANDLER_CONTEXT_VAR

            # Immunity guard: if another component in this process already
            # registered a langfuse handler hook, adopt its ContextVar rather
            # than registering a second handler (the duplicate-trace bug).
            existing = _find_configure_hook_by_name("langfuse_handler")
            if existing is not None:
                cls.HANDLER_CONTEXT_VAR = existing
                return existing

            handler_type = ResolverUtil.create_type(
                "langfuse.langchain.CallbackHandler", raise_if_not_found=False)
            handler = handler_type() if handler_type is not None else None
            # default= (not .set()) so the handler is visible in every
            # thread/context, not only descendants of the registering thread.
            var = ContextVar("langfuse_handler", default=handler)
            if handler is not None:
                register_configure_hook(var, inheritable=True)
            cls.HANDLER_CONTEXT_VAR = var
            return var

    def __init__(self, ...):
        handler_var = self._ensure_registered()
        if handler_var.get() is None:
            raise ValueError(...)  # keep the existing actionable message
        ...
```

Since the factory already checks `LANGFUSE_ENABLED` before instantiating `LangfuseTracingContext`, registration becomes flag-gated *by construction*: flag off → class never constructed → no handler, no client, no export threads, no auth warning, no per-call dispatch, no import of the langfuse/OTel stack. Keys sitting in the environment become inert.

**2. Name-based dedup guard** (`_find_configure_hook_by_name` above): scan `langchain_core.tracers.context._configure_hooks` for an entry whose `ContextVar` is named `"langfuse_handler"` and adopt it instead of registering a second handler. LangChain's identity-only dedup is what allowed the studio plugin's second handler to double every span; with this guard the failure mode degrades from "every span twice" to "works fine". Read-only access to a private attribute, with a graceful fallback if it ever disappears.

**3. `ContextVar(default=handler)` instead of `ContextVar(default=None)` + `.set(handler)`.** A `.set()` is only visible in contexts descended from the setting thread; a default is visible everywhere. Removes a latent "traces vanish in worker threads" edge without changing observable behavior today.

**4. Keep the SDK's own kill switch in agreement.** The langfuse SDK has `LANGFUSE_TRACING_ENABLED` (default `true`, read once at client construction). When constructing the handler, derive it from `LANGFUSE_ENABLED` (e.g. `os.environ.setdefault(...)` or the `tracing_enabled` constructor arg), respecting an explicitly set value as an override. Note the two must not be merged into one variable: the SDK var is opt-out (unset = on) while `LANGFUSE_ENABLED` is opt-in (unset = off), and renaming would break existing deployments — master/derived is the right relationship.

## Backward-compatibility notes

- Environments that today receive "bare" traces from keys-only (flag unset) will stop exporting. That is the point of the fix, but it is observable and deserves a release-note line.
- The existing error path is preserved: `LANGFUSE_ENABLED=true` with langfuse not installed still raises the actionable `ValueError` from `__init__` (handler resolves to `None`).
- Multi-process serving (`AGENT_HTTP_SERVER_INSTANCES`): the lock + class attribute give exactly one registration per process, on the first enabled request.

## Environment

- neuro-san 0.6.88 (mechanism verified against the wheel; also present in earlier 0.6.x)
- langfuse 4.6.1 (also reproduced on 4.14.x)
- langchain-core 1.3.0

## Related

- cognizant-ai-lab/neuro-san-studio#1292 — duplicate traces from the studio plugin's second handler (studio side fixed by removing the plugin and making the langfuse dependency opt-in)
- cognizant-ai-lab/neuro-san-studio#1256 — same duplication observed as "agent runs twice" in Langfuse

A PR implementing the suggested fix is in progress.

Noravee:
## Description

`LangfuseTracingContext` registered its `langfuse.langchain.CallbackHandler` with LangChain's global `register_configure_hook` at import time, gated only on whether the `langfuse` package was importable — not on `LANGFUSE_ENABLED`. Consequences (details in #1191):

1. Any process with langfuse installed and `LANGFUSE_*` keys in the environment exported trace data even with `LANGFUSE_ENABLED=false`/unset — the documented off switch didn't stop export.
2. LangChain dedupes configure-hook handlers by object identity only, so a deployment that registered its own Langfuse handler got **every observation in every trace exactly twice** (cognizant-ai-lab/neuro-san-studio#1292, also reported as cognizant-ai-lab/neuro-san-studio#1256).
3. langfuse-installed-but-keyless environments paid an import-time auth warning plus the langfuse/OTel cold-import cost on every process start.

## The fix

All in `langfuse_tracing_context.py`:

- **Lazy, flag-gated registration.** The class-load `register()` side effect is gone. Registration happens in a lock-guarded `_ensure_registered()` on first construction — and the factory only constructs `LangfuseTracingContext` when `LANGFUSE_ENABLED=true`. Flag off → no handler, no client, no export threads, langfuse never imported; keys in the environment are inert.
- **Name-based dedupe / adoption.** Before registering, scan LangChain's configure hooks for an existing ContextVar named `langfuse_handler` (named constant `LANGFUSE_HANDLER_VAR_NAME`, the cross-component contract). If a foreign component already registered one:
  - handler visible → re-carry the *same instance* as the default of our own ContextVar (visible from every thread; identity dedupe keeps dispatch single),
  - handler not visible (populated via `.set()` in another context, or a lazy `handler_class` hook) → adopt the foreign hook as-is, log a warning, and do **not** raise the misleading "pip install langfuse" error.
- **`ContextVar(default=handler)` instead of `.set()`** so the handler is visible on all threads, not just descendants of the registering one.
- **Failed resolution is not memoized** — a transient import failure no longer poisons tracing until restart; the next construction retries.
- **SDK kill switch kept in agreement**: `LANGFUSE_TRACING_ENABLED` is defaulted to `true` on the enabled path, respecting an explicitly set value.

The `LANGFUSE_ENABLED=true`-but-langfuse-missing path still raises the actionable `ValueError`.

## Backward compatibility

- Environments that previously received "bare" traces from keys-only (flag unset) stop exporting. That is the point of the fix, but it is observable — worth a release-note line.
- Behavior caveats (env vars are read at process start; registration is process-lifetime) are documented in the code comments.

## Testing

- New suite `tests/neuro_san/internals/run_context/langchain/tracing/test_langfuse_tracing_context.py` (12 tests): import-has-no-side-effects, register-once, cross-thread visibility, visible/invisible foreign-hook adoption, no-double-dispatch through a real chain invoke, kill-switch derivation/override, actionable error, resolution-failure retry, construction+clone sharing one registration, factory-flag-off registers nothing. Runs in keyless environments (scoped dummy `OPENAI_API_KEY`).
- Verified green both without langfuse installed and against real `langfuse==4.6.1`; pylint 10.00/10, flake8 clean; `tests/neuro_san/internals` regression: 357 passed.
- End-to-end check with real langfuse: keys set + flag off + chat-session import + chain invoke → 0 hooks, langfuse absent from `sys.modules`; enabled construction incl. `clone()` → exactly 1 hook.

Fixes #1191

Related: cognizant-ai-lab/neuro-san-studio#1294 removes the studio-side duplicate registrant and waits on this change (and a `neuro-san` pin bump) for the flag to become a real off switch.


neuro_san/internals/run_context/langchain/tracing/langfuse_tracing_context.py
```diff
@@ -34,40 +39,164 @@
...
+            if cls.HANDLER_CONTEXT_VAR is not None:
+                # Already done
+                return cls.HANDLER_CONTEXT_VAR
+
+            # If some other component in this process already registered a
+            # langfuse handler hook, reuse its handler instead of registering
+            # a second one.  langchain dedupes hook handlers by object
+            # identity only, so a second handler instance would make every
+            # span get reported twice.
+            existing: ContextVar = cls._find_existing_handler_hook()
```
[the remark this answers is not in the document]
Noravee: Good catch — fixed in 3a13335b: the local annotation is now `Optional[ContextVar]`, matching the function's return type.

neuro_san/internals/run_context/langchain/tracing/langfuse_tracing_context.py
```diff
@@ -34,40 +39,164 @@
...
+
+            # See if we can create a new langfuse handler instance.
+            callback_handler_type: Type[BaseCallbackHandler] = \
+                ResolverUtil.create_type("langfuse.langchain.CallbackHandler",
+                                         raise_if_not_found=False)
+
+            callback_handler: BaseCallbackHandler = None
+            if callback_handler_type is not None:
+                # Create the callback handler instance
+                callback_handler = callback_handler_type()
```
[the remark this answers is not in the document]
Noravee: Agreed — fixed in 3a13335b. The `setdefault` now lives inside the `callback_handler_type is not None` branch, so a failed resolution leaves the environment untouched, while still running before the handler is instantiated (the SDK reads the var once, at client construction). Added an assertion to `test_missing_langfuse_raises_actionable_error` that the error path leaves no env mutation behind.

neuro_san/internals/run_context/langchain/tracing/langfuse_tracing_context.py
```diff
@@ -34,40 +39,164 @@
...
+    HANDLER_CONTEXT_VAR: Optional[ContextVar] = None
+
+    # True when HANDLER_CONTEXT_VAR was adopted from another component and its
+    # handler is not visible from this context.  In that case get() returning
+    # None does not mean langfuse is missing, so the constructor must not
+    # raise its "pip install langfuse" error.
+    _ADOPTED_FOREIGN_HOOK: bool = False
+
+    # Guards the one-time registration per process.
+    _REGISTER_LOCK: threading.Lock = threading.Lock()
```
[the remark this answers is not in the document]
Noravee: Fixed in 3a13335b — dropped the annotation; `threading.Lock` is indeed a factory function on the Python versions this repo supports, not a type.

tests/neuro_san/internals/run_context/langchain/tracing/test_langfuse_tracing_context.py
```diff
@@ -0,0 +1,332 @@
...
+        """
+        _install_fake_langfuse(monkeypatch)
+        monkeypatch.setenv("LANGFUSE_PUBLIC_KEY", "pk-lf-test")
+        monkeypatch.setenv("LANGFUSE_SECRET_KEY", "sk-lf-test")
+
+        hooks_before = _count_langfuse_hooks()
+        reloaded = importlib.reload(ltc_module)
+
+        assert reloaded.LangfuseTracingContext.HANDLER_CONTEXT_VAR is None
+        assert _count_langfuse_hooks() == hooks_before
```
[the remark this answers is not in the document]
Noravee: Fair point — fixed in 3a13335b. The test now executes the module source into an isolated module object via `importlib.util.spec_from_file_location` / `module_from_spec`, so the canonical module and every existing reference to `LangfuseTracingContext` stay untouched and no test-order dependence is introduced.

Noravee (commented):
## TL;DR (author's guide for reviewers)

**The bug (#1191):** importing this module registered a langfuse `CallbackHandler` with langchain's *global* configure hooks whenever the langfuse package happened to be importable. Two consequences:

1. `LANGFUSE_ENABLED` was not a real switch — any process with langfuse installed and `LANGFUSE_*` keys in the environment exported traces even with the flag false or unset.
2. Any other component registering its own langfuse handler (e.g. neuro-san-studio's `LangfusePlugin`) became a **second observer**: langchain dedupes hook handlers by object identity only, so every observation in every trace appeared exactly twice (cognizant-ai-lab/neuro-san-studio#1292).

**The fix (all in `langfuse_tracing_context.py`):** registration now happens lazily, on first construction — and the factory only constructs this class when `LANGFUSE_ENABLED=true`. Flag off ⇒ langfuse is never even imported. Before registering, we adopt any already-registered `langfuse_handler` hook instead of adding a second one; the handler rides as the ContextVar *default* so it is visible from every thread; a failed langfuse resolution is retried instead of memoized.

**Verification:** 13 new unit tests (they run in keyless environments); the duplicate-trace scenario was A/B-reproduced and re-tested with the demos Docker image; CI green; `langfuse` pinned to the tested range in build requirements.

The inline comments below walk the decisions that deserve the most attention, roughly in reading order.

neuro_san/internals/run_context/langchain/tracing/langfuse_tracing_context.py
```diff
@@ -34,40 +39,181 @@
 from neuro_san.internals.interfaces.tracing_context import TracingContext
 from neuro_san.internals.run_context.langchain.tracing.langchain_tracing_context import LangChainTracingContext
 
+# Conventional name for the ContextVar carrying a langfuse CallbackHandler in
+# langchain's configure hooks.  Other components (e.g. the neuro-san-studio
+# LangfusePlugin) use the same name; it is the contract by which competing
+# registrations detect each other and avoid reporting every span twice.
+LANGFUSE_HANDLER_VAR_NAME: str = "langfuse_handler"
```
Noravee, line 46: This name is the cross-component contract. Both this class and the studio `LangfusePlugin` name their ContextVar `langfuse_handler`, and that shared name is how `_find_existing_handler_hook()` below detects a competing registration and avoids the double-reporting bug.

neuro_san/internals/run_context/langchain/tracing/langfuse_tracing_context.py
```diff
@@ -34,40 +39,181 @@
...
+    # handler is not visible from this context.  In that case get() returning
+    # None does not mean langfuse is missing, so the constructor must not
+    # raise its "pip install langfuse" error.
+    _ADOPTED_FOREIGN_HOOK: bool = False
+
+    # Guards the one-time registration per process.
+    _REGISTER_LOCK = threading.Lock()
+
+    @classmethod
+    def _ensure_registered(cls) -> ContextVar:
```
Noravee, line 64: This replaces the old class-load-time `register()` side effect — the heart of the PR. It is only reachable from `__init__`, and the factory only constructs this class when `LANGFUSE_ENABLED=true`, so with the flag off nothing registers and langfuse is never imported. The docstring's one-way-door caveat is worth a careful read: once the first enabled request registers, only a process restart truly turns tracing off.

neuro_san/internals/run_context/langchain/tracing/langfuse_tracing_context.py
```diff
@@ -34,40 +39,181 @@
...
+                return cls.HANDLER_CONTEXT_VAR
+
+            # If some other component in this process already registered a
+            # langfuse handler hook, reuse its handler instead of registering
+            # a second one.  langchain dedupes hook handlers by object
+            # identity only, so a second handler instance would make every
+            # span get reported twice.
+            existing: Optional[ContextVar] = cls._find_existing_handler_hook()
+            if existing is not None:
+                return cls._adopt_existing_handler_hook(existing)
```
Noravee, line 105: This is the fix for the duplicated traces (cognizant-ai-lab/neuro-san-studio#1292): langchain dedupes hook handlers by object identity only, so a second `CallbackHandler` instance means every span reported twice. If someone else already registered one, we reuse theirs instead of creating our own.

neuro_san/internals/run_context/langchain/tracing/langfuse_tracing_context.py
```diff
@@ -34,40 +39,181 @@
...
+                # handler will actually be constructed, so a failed resolution
+                # does not leave a stray env mutation behind.
+                # Caveats: an explicitly set LANGFUSE_TRACING_ENABLED=false
+                # only mutes export - the SDK makes every span non-recording,
+                # but the handler below is still registered and dispatched on
+                # every langchain event, and create_main_span() still runs.
+                # Changing the var in a running process has no effect.  To
+                # turn tracing fully off, set LANGFUSE_ENABLED=false and
+                # restart the process.
+                os.environ.setdefault("LANGFUSE_TRACING_ENABLED", "true")
```
Noravee: Keeps the langfuse SDK's own kill switch (`LANGFUSE_TRACING_ENABLED`, which defaults to true when unset) in agreement with `LANGFUSE_ENABLED`, without clobbering an explicitly set value. Deliberately inside the `callback_handler_type is not None` branch: a failed resolution must not leave a stray env mutation, and the SDK reads this var once, at client construction — so it must be set before the handler is instantiated on the next line.

neuro_san/internals/run_context/langchain/tracing/langfuse_tracing_context.py
```diff
@@ -34,40 +39,181 @@
...
+
+                # Create the callback handler instance
+                callback_handler = callback_handler_type()
+
+            # Carry the handler as the ContextVar default rather than via set():
+            # a set() is only visible in contexts descended from the setting
+            # thread, while a default is visible in every thread.
+            context_var = ContextVar(LANGFUSE_HANDLER_VAR_NAME, default=callback_handler)
```
Noravee, line 136: `default=` rather than `.set()` is load-bearing: a `.set()` is only visible in contexts descended from the setting thread, and our servers dispatch requests across thread-pool event loops that are *siblings* of the registering thread. With `.set()`, traces would silently vanish for every request handled off the first thread; a default is visible everywhere. Covered by `test_handler_visible_from_other_threads`.

neuro_san/internals/run_context/langchain/tracing/langfuse_tracing_context.py
```diff
@@ -34,40 +39,181 @@
...
+            if callback_handler is None:
+                # Leave HANDLER_CONTEXT_VAR unset so a later construction
+                # retries the resolution - a transient import failure must not
+                # poison tracing for the rest of the process.
+                return context_var
```
Noravee: Failure is intentionally not cached: if langfuse can't be resolved right now, we leave `HANDLER_CONTEXT_VAR` unset so the next construction retries, instead of a transient import problem disabling tracing until restart. Covered by `test_resolution_failure_is_not_memoized`.

neuro_san/internals/run_context/langchain/tracing/langfuse_tracing_context.py
```diff
@@ -34,40 +39,181 @@
...
+
+        # The foreign handler is not visible from this context: it was either
+        # populated via set() in another thread, or langchain creates it
+        # lazily from the hook's handler_class.  Use the foreign hook as-is,
+        # and remember that get() returning None here does not mean langfuse
+        # is missing.
+        getLogger(cls.__name__).warning(
+            "Adopted an existing langfuse handler hook whose handler is not "
+            "visible from this context; tracing may be unavailable on some threads.")
+        cls._ADOPTED_FOREIGN_HOOK = True
```
Noravee: The degraded branch: a foreign hook exists but its handler isn't visible from here (it was `.set()` in another context, or langchain builds it lazily from `handler_class`). We adopt the hook as-is and warn. The `_ADOPTED_FOREIGN_HOOK` flag matters below in `__init__`: in this state `get()` returning None does NOT mean langfuse is missing, so the constructor must not raise its "pip install langfuse" error.

neuro_san/internals/run_context/langchain/tracing/langfuse_tracing_context.py
```diff
@@ -87,28 +233,31 @@ def __init__(self, run_target: RunTarget,
...
-    export LANGFUSE_ENABLED=false
-""")
+        # Register the langfuse handler with langchain (idempotent; first
+        # construction in the process does the work) and see if we actually
+        # got a handler instance.  When the hook was adopted from another
+        # component, get() can legitimately return None here (the handler
+        # lives in a context we cannot see), so that case must not be
+        # mistaken for "langfuse is not installed".
+        handler_var: ContextVar = self._ensure_registered()
+        if handler_var.get() is None and not self._ADOPTED_FOREIGN_HOOK:
```
Noravee: The `_ADOPTED_FOREIGN_HOOK` exemption keeps this check honest: on the invisible-foreign-hook path, a None handler is expected and langfuse may well be installed — raising the "pip install langfuse" message there would send an operator debugging in exactly the wrong direction.

neuro_san/internals/run_context/langchain/tracing/langfuse_tracing_context.py
```diff
@@ -87,28 +233,31 @@ def __init__(self, run_target: RunTarget,
...
+        try:
+            from langfuse import Langfuse
+            from langfuse import get_client
+            from opentelemetry.util._decorator import _AgnosticContextManager
+        except ImportError as error:
+            # Reachable when a foreign "langfuse_handler" hook was adopted
+            # (which skips the check above) but langfuse itself is not
+            # importable. Keep the failure as actionable as the check above
+            # instead of leaking a raw ImportError.
+            raise ValueError(_MISSING_LANGFUSE_MESSAGE) from error
```
Noravee: Backstop for the one way the exemption above could mask a genuinely missing package: an adopted hook with langfuse actually uninstalled would otherwise surface as a raw `ImportError` here. Wrapping keeps every "langfuse can't deliver" failure funneled into the same actionable message.

tests/neuro_san/internals/run_context/langchain/tracing/test_langfuse_tracing_context.py
```diff
@@ -0,0 +1,358 @@
...
+
+
+class TestLangfuseTracingContextRegistration:
+    """
+    Test cases for the lazy, once-per-process registration of the Langfuse
+    CallbackHandler (https://github.com/cognizant-ai-lab/neuro-san/issues/1191).
+    """
+
+    @pytest.fixture(autouse=True)
+    def clean_registration_state(self):
```
Noravee, line 102: Everything under test is process-global state (class attributes, langchain's hook list, an env var), so this autouse fixture snapshots and restores all of it around every test — including stripping any langfuse hook a *previous* test in the pytest session may have registered. This is what keeps the suite order-independent.

tests/neuro_san/internals/run_context/langchain/tracing/test_langfuse_tracing_context.py
```diff
@@ -0,0 +1,358 @@
...
+        assert adopted.get() is foreign_handler
+        assert fake_handler_class.instances_created == 0
+
+        seen_in_thread = []
+        thread = threading.Thread(target=lambda: seen_in_thread.append(adopted.get()))
+        thread.start()
+        thread.join()
+        assert seen_in_thread[0] is foreign_handler
+
+    def test_adoption_does_not_double_dispatch(self, monkeypatch):
```
Noravee, line 207: The regression test for the original bug: after adopting a foreign handler, two hooks may carry the same instance, and this drives a real `Runnable.invoke()` to prove langchain's identity dedupe dispatches each event exactly once. If someone reverts to constructing a second handler instance, this fails.

requirements-build.txt
```diff
@@ -16,7 +16,11 @@ langchain-openrouter>=0.2.3,<1.0
 openfga-sdk
 
 # Optional Observability
-langfuse
+# Floor is the verified version (langfuse.langchain.CallbackHandler path,
+# get_client() per-key singleton, LANGFUSE_TRACING_ENABLED unset == "true");
+# major-version cap because langfuse majors have moved the handler import
+# path before (v2 -> v3).
+langfuse>=4.6.1,<5.0
```
Noravee, line 23: Floor = the version whose internals the fix was verified against (handler import path, `get_client()` per-key singleton, `LANGFUSE_TRACING_ENABLED` unset ⇒ true). Cap at the next major because langfuse majors have moved the handler import path before (v2 → v3), which would break `ResolverUtil.create_type("langfuse.langchain.CallbackHandler")`.

neuro_san/internals/run_context/langchain/tracing/langfuse_tracing_context.py
```diff
@@ -34,40 +39,181 @@
...
+            # identity only, so a second handler instance would make every
+            # span get reported twice.
+            existing: Optional[ContextVar] = cls._find_existing_handler_hook()
+            if existing is not None:
+                return cls._adopt_existing_handler_hook(existing)
+
+            # See if we can create a new langfuse handler instance.
+            callback_handler_type: Optional[Type[BaseCallbackHandler]] = \
+                ResolverUtil.create_type("langfuse.langchain.CallbackHandler",
+                                         raise_if_not_found=False)
```
d1donlydfink: This method has a very handy "install_if_missing" keyword arg which will add what package to install to the error message if the class is missing.

neuro_san/internals/run_context/langchain/tracing/langfuse_tracing_context.py
```diff
@@ -34,40 +39,181 @@
...
+# but the langfuse package cannot deliver a handler.
+_MISSING_LANGFUSE_MESSAGE: str = """
+Failed to create Langfuse CallbackHandler. Try one of the following:
+
+If you really wanted to use langfuse for observability, you can install it with
+    pip install langfuse
+
+If you didn't mean to use langfuse for observability, you can do this:
+    export LANGFUSE_ENABLED=false
+"""
```
d1donlydfink: This message is better and more consistently handled with ResolverUtil.create_type()'s install_if_missing keyword argument.

neuro_san/internals/run_context/langchain/tracing/langfuse_tracing_context.py
d1donlydfink: That is reasonable.

d1donlydfink (changes requested):
Use ResolverUtil.create_type()'s install_if_missing keyword arg for consistent lazy-loading messaging.

neuro_san/internals/run_context/langchain/tracing/langfuse_tracing_context.py
Noravee: Done — switched to `create_type("langfuse.langchain.CallbackHandler", install_if_missing="langfuse")` in default raise-mode. Nice side effect: since the standard error now propagates before any state is cached, the no-memoization-of-failures behavior comes for free and the whole `callback_handler is None` branch disappeared.

neuro_san/internals/run_context/langchain/tracing/langfuse_tracing_context.py
Noravee: Done — the custom message constant is gone; all missing-langfuse failures now funnel through ResolverUtil's standard "Try pip installing the package langfuse" error. That made the constructor's None-handler check unreachable, so it and the `_ADOPTED_FOREIGN_HOOK` flag were removed too (net −25 lines). One consequence to be aware of: the old message also suggested `export LANGFUSE_ENABLED=false` as the alternative remedy; that hint now lives only in the `_ensure_registered` docstring.

neuro_san/internals/run_context/langchain/tracing/langfuse_tracing_context.py
```diff
@@ -34,40 +39,160 @@
...
+                return cls._adopt_existing_handler_hook(existing)
+
+            # See if we can create a new langfuse handler instance.
+            # Use lazy loading to prevent installing the world.  When the
+            # optional package is missing this raises the standard actionable
+            # error, before anything below mutates the environment or caches
+            # registration state.
+            callback_handler_type: Type[BaseCallbackHandler] = \
+                ResolverUtil.create_type("langfuse.langchain.CallbackHandler",
+                                         install_if_missing="langfuse")
```
Noravee, line 114: Default raise-mode is deliberate: `install_if_missing`'s message is only surfaced when the resolver raises (with `raise_if_not_found=False` it is discarded), and raising *here* — before the env `setdefault` below and before anything is cached on the class — is what preserves the no-stray-env-mutation and retry-on-next-construction behaviors without their previous explicit branches.

neuro_san/internals/run_context/langchain/tracing/langfuse_tracing_context.py
```diff
@@ -87,28 +212,30 @@ def __init__(self, run_target: RunTarget,
...
+            from langfuse import get_client
+            from opentelemetry.util._decorator import _AgnosticContextManager
+        except ImportError:
+            # Reachable when a foreign "langfuse_handler" hook was adopted
+            # (which skips handler resolution above) but langfuse itself is
+            # not importable.  Re-run the standard resolution so the failure
+            # carries the consistent actionable message; if langfuse resolves
+            # (i.e. something else failed to import), re-raise as-is.
+            ResolverUtil.create_type("langfuse.langchain.CallbackHandler",
+                                     install_if_missing="langfuse")
```
Noravee, line 237: The adopted-foreign-hook path skips handler resolution, so langfuse can still be missing by the time these local imports run. Re-running the standard resolution here yields the same actionable ValueError for that case; if langfuse itself resolves (i.e. some *other* import in the try block failed, e.g. opentelemetry), the bare `raise` re-raises the original error untouched. Covered by `test_adopted_hook_without_langfuse_raises_actionable_error`.

Noravee (commented):
## Update

All missing-langfuse failures now go through `ResolverUtil.create_type(install_if_missing="langfuse")`, consistent with the other optional-dependency call sites (openfga-sdk, langchain-ollama, …). The hand-rolled `_MISSING_LANGFUSE_MESSAGE` is gone.

Switching `create_type` to its default raise-mode also *simplified* the registration logic rather than just re-wording an error (net −25 lines):

- A failed resolution raises **before** any state is cached or the environment is touched, so "failures are not memoized" is now a property of the control flow instead of an explicit `callback_handler is None` branch.
- The constructor's None-handler check became provably unreachable (the non-adopted path raises during resolution; adoption either supplies a handler or is the flagged-invisible case) — so the check **and the `_ADOPTED_FOREIGN_HOOK` class flag were deleted entirely**.
- The one edge case the old check covered (an adopted foreign `langfuse_handler` hook while langfuse is genuinely not installed) is handled in the constructor's `ImportError` handler by re-running the same standard resolution, so it fails with the identical actionable message.

Behavior note: the previous custom error offered `export LANGFUSE_ENABLED=false` as a second remedy alongside `pip install langfuse`; the standard message only suggests the install. The flag semantics remain documented in the `_ensure_registered` docstring.

d1donlydfink (approved):

PR #1195, opened 2026-08-03
Make storage dependencies optional for both S3 and Azure.

d1donlydfink:
* Have S3ReservationsStorage and its ilk use ResolverUtil to resolve a couple of types so that the appropriate install_if_missing error comes up if appropriate boto packages are not installed
* Similar deal with AzureBlobReservationsStorage and install_if_missing
* Move botocore aiobotocore azure-storage-blob and azure-identity dependencies to requirements-build.txt from requirements.txt as optional dependencies.
* Use latest leaf-common==1.2.50 whose create_instance() behavior better surfaces missing pip packages in its reporting of failures.

### DIFF
--- a/neuro_san/deploy/Dockerfile
+++ b/neuro_san/deploy/Dockerfile
@@ -384,6 +384,7 @@ ENV AGENT_EXTERNAL_SERVER_URL=""
 # When not set, this defaults to a null implementation.
 # A simple default implementation is provided by the class
 # "neuro_san.service.watcher.temp_networks.s3.s3_reservations_storage.S3ReservationsStorage"
+# Note that you will need to install appropriate botocore and aiobotocore packages to use this class.
 ENV AGENT_EXTERNAL_RESERVATIONS_STORAGE=""
 
 # When the AGENT_EXTERNAL_RESERVATIONS_STORAGE (immediately above) is set to the
--- a/neuro_san/service/watcher/temp_networks/azure/azure_blob_reservations_expiration.py
+++ b/neuro_san/service/watcher/temp_networks/azure/azure_blob_reservations_expiration.py
@@ -36,7 +36,7 @@ DefaultAzureCredential: Type[Any] = ResolverUtil.create_type("azure.identity.Def
 ContainerClient: Type[Any] = ResolverUtil.create_type("azure.storage.blob.ContainerClient",
                                                       install_if_missing="azure-storage-blob")
 
-from neuro_san.service.watcher.temp_networks.azure.azure_blob_util import AzureBlobUtil
+from neuro_san.service.watcher.temp_networks.azure.azure_blob_util import AzureBlobUtil     # noqa: E402
 
 
 class AzureBlobReservationsExpiration:
--- a/neuro_san/service/watcher/temp_networks/azure/azure_blob_reservations_expiration.py
+++ b/neuro_san/service/watcher/temp_networks/azure/azure_blob_reservations_expiration.py
@@ -15,7 +15,9 @@
 #
 # END COPYRIGHT
 
+from typing import Any
 from typing import Optional
+from typing import Type
 
 from time import time
 from json import loads
@@ -24,11 +26,16 @@ from logging import getLogger
 from logging import Logger
 from os import getenv
 
-from azure.core.exceptions import AzureError
-from azure.identity import DefaultAzureCredential
-from azure.storage.blob import ContainerClient
+from leaf_common.config.resolver_util import ResolverUtil
 
-# pylint: disable=wrong-import-order
+# Lazily resolve here because this is one of the first places the import happens
+AzureError: Type[Any] = ResolverUtil.create_type("azure.core.exceptions.AzureError",
+                                                 install_if_missing="azure-core")
+DefaultAzureCredential: Type[Any] = ResolverUtil.create_type("azure.identity.DefaultAzureCredential",
+                                                             install_if_missing="azure-identity")
+ContainerClient: Type[Any] = ResolverUtil.create_type("azure.storage.blob.ContainerClient",
+                                                      install_if_missing="azure-storage-blob")
+
 from neuro_san.service.watcher.temp_networks.azure.azure_blob_util import AzureBlobUtil
 
 
--- a/neuro_san/service/watcher/temp_networks/s3/s3_reservations_expiration.py
+++ b/neuro_san/service/watcher/temp_networks/s3/s3_reservations_expiration.py
@@ -35,10 +35,10 @@ from leaf_common.config.resolver_util import ResolverUtil
 BaseClient: Type[Any] = ResolverUtil.create_type("botocore.client.BaseClient", install_if_missing="botocore")
 ClientError: Type[Any] = ResolverUtil.create_type("botocore.exceptions.ClientError", install_if_missing="botocore")
 
-from neuro_san.interfaces.reservationist import Reservationist
-from neuro_san.service.watcher.temp_networks.s3.aws_sync_client_worker import AwsSyncClientWorker
-from neuro_san.service.watcher.temp_networks.s3.s3_reservations_retriever import S3ReservationsRetriever
-from neuro_san.service.watcher.temp_networks.s3.s3_util import S3Util
+from neuro_san.interfaces.reservationist import Reservationist                                              # noqa: E402
+from neuro_san.service.watcher.temp_networks.s3.aws_sync_client_worker import AwsSyncClientWorker           # noqa: E402
+from neuro_san.service.watcher.temp_networks.s3.s3_reservations_retriever import S3ReservationsRetriever    # noqa: E402
+from neuro_san.service.watcher.temp_networks.s3.s3_util import S3Util                                       # noqa: E402
 
 
 class S3ReservationsExpiration:
--- a/neuro_san/service/watcher/temp_networks/s3/s3_reservations_expiration.py
+++ b/neuro_san/service/watcher/temp_networks/s3/s3_reservations_expiration.py
@@ -20,6 +20,7 @@ from typing import Callable
 from typing import Dict
 from typing import Iterable
 from typing import Optional
+from typing import Type
 
 from time import time
 from functools import partial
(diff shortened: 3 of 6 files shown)


neuro_san/service/watcher/temp_networks/azure/azure_blob_reservations_expiration.py
```diff
@@ -24,12 +26,17 @@
...
 
-# pylint: disable=wrong-import-order
-from neuro_san.service.watcher.temp_networks.azure.azure_blob_util import AzureBlobUtil
+# Lazily resolve here because this is one of the first places the import happens
+AzureError: Type[Any] = ResolverUtil.create_type("azure.core.exceptions.AzureError",
+                                                 install_if_missing="azure-core")
+DefaultAzureCredential: Type[Any] = ResolverUtil.create_type("azure.identity.DefaultAzureCredential",
+                                                             install_if_missing="azure-identity")
+ContainerClient: Type[Any] = ResolverUtil.create_type("azure.storage.blob.ContainerClient",
+                                                      install_if_missing="azure-storage-blob")
```
d1donlydfink, line 37: Resolve the types dynamically so we can issue an appropriate "Install package X" warning for optional packages.

neuro_san/service/watcher/temp_networks/azure/azure_blob_reservations_expiration.py
```diff
@@ -24,12 +26,17 @@
...
+
+from neuro_san.service.watcher.temp_networks.azure.azure_blob_util import AzureBlobUtil     # noqa: E402
```
d1donlydfink, line 39: As it happens this one file covers all the packages.
Anything later could undergo the same lazy type creation, but it's not necessary unless the filenames get moved around and there is a new "first to install". For the rest in this file the extra "noqa" note for flake8 will get the build to pass.

neuro_san/service/watcher/temp_networks/s3/s3_reservations_expiration.py
```diff
@@ -28,13 +29,16 @@
...
-from botocore.exceptions import ClientError
+from leaf_common.config.resolver_util import ResolverUtil
 
-from neuro_san.interfaces.reservationist import Reservationist
-from neuro_san.service.watcher.temp_networks.s3.aws_sync_client_worker import AwsSyncClientWorker
-from neuro_san.service.watcher.temp_networks.s3.s3_reservations_retriever import S3ReservationsRetriever
-from neuro_san.service.watcher.temp_networks.s3.s3_util import S3Util
+# Lazily resolve here because this is one of the first places the import happens
+BaseClient: Type[Any] = ResolverUtil.create_type("botocore.client.BaseClient", install_if_missing="botocore")
+ClientError: Type[Any] = ResolverUtil.create_type("botocore.exceptions.ClientError", install_if_missing="botocore")
```
d1donlydfink, line 36: Similar deal for botocore here.

neuro_san/service/watcher/temp_networks/s3/s3_reservations_writer.py
```diff
@@ -27,12 +28,17 @@
...
-from aiobotocore.client import AioBaseClient
+from leaf_common.config.resolver_util import ResolverUtil
 
-from neuro_san.interfaces.reservation import Reservation
-from neuro_san.internals.reservations.reservation_dictionary_converter import ReservationDictionaryConverter
-from neuro_san.service.watcher.temp_networks.s3.aws_async_client_worker import AwsAsyncClientWorker
-from neuro_san.service.watcher.temp_networks.s3.s3_util import S3Util
+# Lazily resolve here because this is one of the first places the import happens
+AioBaseClient: Type[Any] = ResolverUtil.create_type("aiobotocore.client.AioBaseClient",
+                                                    install_if_missing="aiobotocore")
```
d1donlydfink, line 35: Similar deal for aiobotocore here.

requirements-build.txt
```diff
@@ -12,6 +12,14 @@ langchain-nvidia-ai-endpoints>=1.0.0,<2.0
 langchain-ollama>=1.0.0,<2.0
 langchain-openrouter>=0.2.3,<1.0
 
+# Optional ReservationsStorage
+# ... for S3
+botocore>=1.34.51
+aiobotocore>=3.7.0
+# ... for Azure
+azure-storage-blob>=12.19.0
+azure-identity>=1.14.0
```
d1donlydfink, line 21: Now the aws and azure dependencies are left to requirements-build.txt so that tests (existing and future) can be happy.

requirements.txt
```diff
@@ -13,8 +13,6 @@ leaf-server-common>=0.1.25
 pyhocon>=0.3.60
 pyOpenSSL>=24.0.0
 
-botocore>=1.34.51
-aiobotocore>=3.7.0
```
d1donlydfink, line 17: Remove botobusiness that not everyone needs from requirements.txt

requirements.txt
```diff
@@ -86,7 +84,3 @@ langchain-mcp-adapters>=0.1.7,<0.2.0
 # Ceiling <2.0: avoid breaking changes from the upcoming 2.x line.
 # See https://github.com/modelcontextprotocol/python-sdk/releases/ for changelogs.
 mcp>=1.23.0,<2.0
-
-# Azure Blob Storage for reservations backend
-azure-storage-blob>=12.19.0
-azure-identity>=1.14.0
```
d1donlydfink, line 92: Remove azure business that not everyone needs from requirements.txt

neuro_san/deploy/Dockerfile
```diff
@@ -384,6 +384,7 @@ ENV AGENT_EXTERNAL_SERVER_URL=""
 # When not set, this defaults to a null implementation.
 # A simple default implementation is provided by the class
 # "neuro_san.service.watcher.temp_networks.s3.s3_reservations_storage.S3ReservationsStorage"
+# Note that you will need to install appropriate botocore and aiobotocore packages to use this class.
 ENV AGENT_EXTERNAL_RESERVATIONS_STORAGE=""
```
d1donlydfink, line 388: Add note about extra installs for Dockerfile.

neuro_san/service/watcher/temp_networks/s3/s3_reservations_expiration.py
Noravee, line 36: These guards only produce the friendly error when this module is imported directly. Through the documented path (`AGENT_EXTERNAL_RESERVATIONS_STORAGE` → `ResolverUtil.create_instance`), the hint gets swallowed and the user sees a misleading `Class 'S3ReservationsStorage' ... not found in PYTHONPATH` instead — see the review body for the mechanism and repro. Moving these `create_type` calls into `temp_networks/s3/__init__.py` fixes that, guarantees the guard fires regardless of which module in the package is imported first, and lets this file revert to plain imports (dropping the `# noqa: E402` lines).

Nit: "Lazily resolve" is a slight misnomer — `create_type` resolves eagerly at module-import time; it's the module import itself that's deferred.

neuro_san/service/watcher/temp_networks/azure/azure_blob_reservations_expiration.py
Noravee, line 37: Same `__init__.py` suggestion as on the S3 side applies here.

Additionally, resolution order creates an install whack-a-mole for a user with no azure packages: `AzureError` fails first ("pip install azure-core"), then on the next run `DefaultAzureCredential` ("azure-identity"), then `ContainerClient` ("azure-storage-blob") — three round trips. Resolving `ContainerClient` first would help, since installing `azure-storage-blob` transitively brings in `azure-core`; or use a combined hint like `install_if_missing="azure-storage-blob azure-identity"`.

Noravee (commented):
One main concern and a few secondary ones:

## Main concern: the install hint never surfaces through the documented usage path

I simulated a missing `botocore`/`azure` and ran the exact production path (`TempNetworkStorageUpdater` resolving `AGENT_EXTERNAL_RESERVATIONS_STORAGE` via `ResolverUtil.create_instance`). The user sees:

```
ValueError: Class 'S3ReservationsStorage' from AGENT_EXTERNAL_RESERVATIONS_STORAGE env var not found in PYTHONPATH
```

…not the pip-install hint. Why: leaf-common's `Resolver._find_module` first tries to import `...s3.s3_reservations_storage` (fails — the hint is recorded in a `messages` list that is only logged when `verbose=True`), then falls back to importing the parent *package*. The empty `temp_networks/s3/__init__.py` imports fine, so `resolve_class_in_module` returns the package, `getattr(package, "S3ReservationsStorage")` raises `AttributeError`, and `ResolverUtil.create_class` converts that into the misleading "not found in PYTHONPATH" error. Same story for Azure. The hint only appears when the expiration module is imported directly, which is not the documented usage.

**Suggested fix (verified to work):** move the `create_type` guard calls into the package `__init__.py` files (`temp_networks/s3/__init__.py` and `temp_networks/azure/__init__.py`). Then the parent-package fallback import also fails, `found_module` stays `None`, and the raised `ValueError` includes the pip-install hint. Bonus: the three touched modules can revert to plain imports (no more `# noqa: E402`), and the guard fires no matter which module in the package gets imported first. Longer-term, leaf-common's `create_class` could chain the recorded `messages` into its `AttributeError` path so import failures inside a resolved module are never masked.

## Secondary concerns

- **Fragile import-order contract.** The guards rely on `*_reservations_expiration` being the first module imported in each package — true today only because the imports in `s3_reservations_storage.py` / `azure_blob_reservations_storage.py` happen to be alphabetical. All other modules in these packages (retriever, reader, writer, util, both client workers) still import boto/azure directly, so a direct import of any of them — or a future import reordering — yields a raw `ImportError` with no hint. The `__init__.py` approach above fixes this categorically.
- **Breaking change for existing deployments.** Anyone using S3/Azure reservations storage who upgrades will now break at server startup until they add the packages. The Dockerfile comment covers the S3 case; this deserves a release-notes callout too.

neuro_san/service/watcher/temp_networks/s3/s3_reservations_expiration.py
d1donlydfink: Hmm. Thanks. I haven't tested that path.

d1donlydfink: This will be fixed with a new leaf-common version continaing new ResolverUtil.create_instance() plumbing.
Actually quite a nice find.

neuro_san/service/watcher/temp_networks/azure/azure_blob_reservations_expiration.py
d1donlydfink: This will be fixed with a new leaf-common version continaing new ResolverUtil.create_instance() plumbing.

d1donlydfink:
As for:
* Fragile import-order contract. 
   Yes. I agree. At the very least further renaming is not expected much at this time.
* Breaking change for existing deployments.
  Yes.  Best to do these refoactorings early before too many external dependencies set in.

requirements.txt
```diff
@@ -6,15 +6,13 @@
 # requirements-build.txt
 
 # In-house dependencies
-leaf-common>=1.2.49
+leaf-common>=1.2.50
```
d1donlydfink, line 9: Use latest leaf-common=1.2.50 with new create_instance() error surfacing.

Noravee (approved):
lgtm

PR #1201, opened 2026-08-04
NS-1199 Fix macOS Obj-C fork-safety crash in multi-instance mode

andreidenissov-cog:
Description of the changes

  Running neuro-san with AGENT_HTTP_SERVER_INSTANCES > 1 on macOS aborts a forked worker with objc[...]: +[NSNumber initialize] may have been in progress in another thread when fork() was called 
  ... Crashing instead. The crash surfaces on the first streaming_chat request (the LLM call reaches into macOS SystemConfiguration/Security via httpx), because Tornado's server.start(N) forks
  workers and Apple's Obj-C runtime is not fork()-safe.

  This change sets OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES from the entrypoint on macOS. Because libobjc reads that variable once at interpreter launch, setting it in-process is too late — so the
  guard sets the flag and re-execs the interpreter (preserving the original -m … <args> invocation) so it takes effect. It is a no-op on non-macOS platforms and a no-op when the flag is already
  set (which also prevents any re-exec loop).

  Added to neuro_san/service/main_loop/server_main_loop.py:
  - import sys
  - ServerMainLoop.ensure_macos_fork_safety() static method (with a comment explaining the fork-unsafety and the re-exec rationale)
  - a call to it as the first statement in the __main__ block

  No production/Linux behavior change.

  Testing

  - Verified the re-exec path (launched without the flag): re-execs once, same PID, lands with the flag set, argv preserved, no loop; and the already-set path returns without re-exec.
  - pylint (repo .pylintrc): 10.00/10 on the changed file.

### DIFF
--- a/neuro_san/service/main_loop/server_main_loop.py
+++ b/neuro_san/service/main_loop/server_main_loop.py
@@ -109,7 +109,7 @@ class ServerMainLoop:
         preserves the original "-m ... <args>" invocation across the re-exec.
 
         No-op on non-Darwin platforms, and no-op (no re-exec) when the flag is
-        already set -- which is also what stops the re-exec from looping.
+        already set to "YES" -- which is also what stops the re-exec from looping.
         """
         if sys.platform != "darwin":
             return
--- a/neuro_san/service/main_loop/server_main_loop.py
+++ b/neuro_san/service/main_loop/server_main_loop.py
@@ -19,6 +19,7 @@ from typing import Dict
 from typing import List
 
 import os
+import sys
 
 from argparse import ArgumentParser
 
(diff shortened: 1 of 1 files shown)


neuro_san/service/main_loop/server_main_loop.py
```diff
@@ -82,6 +83,41 @@ def __init__(self):
...
+
+        No-op on non-Darwin platforms, and no-op (no re-exec) when the flag is
+        already set to "YES" -- which is also what stops the re-exec from looping.
+        """
+        if sys.platform != "darwin":
+            return
+        if os.environ.get("OBJC_DISABLE_INITIALIZE_FORK_SAFETY") == "YES":
+            return
+        os.environ["OBJC_DISABLE_INITIALIZE_FORK_SAFETY"] = "YES"
+        os.execve(sys.executable, sys.orig_argv, os.environ)
```
d1donlydfink, line 119: At some point we might want to farm this out to a utility class with os-specific policy.

d1donlydfink (approved):
LGTM AFAICT

PR #1204, opened 2026-08-04
NS-1188 Dockerfile for neuro-san/Python3.14T platform.

andreidenissov-cog:
This PR adds Dockerfile and build.sh versions for neuro-san P3.14T environment.
Also added to run.sh script: option to specify and pass in run-time env variables through envfile,
so we are able to customize a specific docker run more easily.
Tested by building and running P3.14T neuro-san docker image on Linux EC2 instance.


neuro_san/deploy/run.sh
```diff
@@ -34,10 +34,26 @@ function run() {
 
     check_directory
 
-    CONTAINER_VERSION="0.0.1"
+    CONTAINER_VERSION="${1:-0.0.1}"
     echo "Using CONTAINER_VERSION ${CONTAINER_VERSION}"
```
andreidenissov-cog, line 38: So we can pass specific container version to run.sh script.

neuro_san/deploy/run.sh
```diff
@@ -34,10 +34,26 @@ function run() {
...
+    # This is optional, but if it is set and exists, we will use it.
+    # Environment file should be a simple text file with lines of the form:
+    #   VAR_NAME=VAR_VALUE
+    # This allows us to pass in any collection of run-specific values
+    env_file_cmd=""
+    if [[ -n "${SERVICE_ENV_FILE:-}" && -f "$SERVICE_ENV_FILE" ]]; then
+        echo "Using service environment file: $SERVICE_ENV_FILE"
+        cat "$SERVICE_ENV_FILE"
+        env_file_cmd="--env-file $SERVICE_ENV_FILE"
+    elif [[ -z "${SERVICE_ENV_FILE:-}" ]]; then
```
andreidenissov-cog, line 50: env_file_cmd string will be used below when we issue docker run command.

neuro_san/deploy/run.sh
```diff
@@ -73,6 +89,7 @@ function run() {
         -e ANTHROPIC_API_KEY \
         -e AGENT_SESSION_REQUIRE_HTTPS=false \
         -e LEAF_LOG_SENSITIVE=true \
+        ${env_file_cmd} \
         -p $SERVICE_HTTP_PORT:$SERVICE_HTTP_PORT \
```
andreidenissov-cog, line 92: Use envfile here to pass in any collection of additional env var settings.

neuro_san/deploy/build_py314t.sh
```diff
@@ -0,0 +1,136 @@
...
+# repo publishes no free-threaded tags), so this build provisions the
+# free-threaded interpreter with uv rather than via a base image.
+
+export SERVICE_TAG=${SERVICE_TAG:-neuro-san}
+export SERVICE_VERSION=${SERVICE_VERSION:-0.0.1-py314t}
+
+PYTHON_VERSION=${PYTHON_VERSION:-3.14t}
+BASE_IMAGE=${BASE_IMAGE:-debian:trixie-slim}
+UV_IMAGE=${UV_IMAGE:-ghcr.io/astral-sh/uv:latest}
+echo ">>>>>>>>>>>>>>UV_IMAGE = ${UV_IMAGE}"
```
d1donlydfink, line 52: Nit: Might be nice to be able to lock down the uv image by env var (and not take latest all the time)
Doesn't have to be this PR.

neuro_san/deploy/Dockerfile.py314t
```diff
@@ -0,0 +1,647 @@
...
+    && curl --proto '=https' --tlsv1.2 -sSfL "${url}" -o /tmp/rustup-init \
+    && echo "${sha256}  /tmp/rustup-init" | sha256sum -c - \
+    && chmod +x /tmp/rustup-init \
+    && /tmp/rustup-init -y --no-modify-path --profile minimal --default-toolchain "${RUST_TOOLCHAIN}" \
+    && rm -f /tmp/rustup-init \
+    && rustc --version \
+    && cargo --version
+
+# Bring in the uv binary (static, no runtime deps).
+COPY --from=uvsrc /uv /usr/local/bin/uv
```
d1donlydfink, line 146: Doesn't have to be for now: If there was a way to re-use the existing Dockerfile such that there didn't need to be so much copy/paste into this guy.   I get that something working is better than nothing working, though.  Long term it will not be sustainable to keep updating both Dockerfiles.

d1donlydfink (approved):

Issue #1205, opened 2026-08-05
LoggingConfigRestorer parses HOCON without key sanitization, so dotted logger names silently never apply


Noravee:
## Summary

With leaf-common 1.2.51, all config restorers that go through `AbstractAsyncConfigRestorer` now parse HOCON with `HoconSerializationFormat(sanitize_keys=True)`, which removes the quotation marks pyhocon embeds in keys that had to be quoted in HOCON source (see cognizant-ai-lab/leaf-common#172, follow-up to #451).

`LoggingConfigRestorer` (`neuro_san/service/http/logging/logging_config_restorer.py`) is the one remaining HOCON parse path that does not: it reads `logging.hocon` via `EasyHoconPersistence`, which in leaf-common 1.2.51 has no way to pass `sanitize_keys` (its `PersistenceFactory` path constructs `HoconPersistence` without forwarding the flag, even though `HoconPersistence` itself accepts it).

## Reproduction

`logging.hocon` using the standard Python logging idiom for per-module levels:

```hocon
loggers {
    "neuro_san.internals" {
        level = "DEBUG"
    }
}
```

```python
from leaf_common.persistence.easy.easy_hocon_persistence import EasyHoconPersistence

cfg = EasyHoconPersistence().restore(file_reference="logging.hocon")
print(list(cfg["loggers"].keys()))
# ['"neuro_san.internals"']  <- quotes baked into the key
```

`logging.config.dictConfig(cfg)` then registers a logger literally named `"neuro_san.internals"` (with quotes), so the intended `neuro_san.internals` logger stays NOTSET and the configured level silently never applies.

## Impact

- Any user logging config that sets per-module logger levels (dotted names must be quoted in HOCON) is silently ignored.
- Inconsistent behavior: the same quoted-key HOCON syntax now works in every other config file (agent networks, manifests, llm info, toolbox, MCP servers info) but not in `logging.hocon`.
- The shipped `neuro_san/deploy/logging.hocon` is unaffected (its logger names are plain, and `"()"` parses clean since parentheses are not in pyhocon's quote-retaining character set).

This is not a regression — the behavior predates the sanitization work — but it is now the sole unsanitized path.

## Proposed fix

[...]

Issue #451, opened 2025-10-08
HOCON parsing for keys with forbidden characters return additional quotation signs


d1donlydfink:
For example, keys like “llama3.1” and “llama3:8b“ will be parsed to ‘“llama3.1”’ and ‘“llama3:8b”’, respectively.

Dupe of: UN-3054

PR #1206, opened 2026-08-05
Sanitize HOCON keys at parse time via leaf-common 1.2.51

Noravee:
## Summary

Adopts leaf-common 1.2.51's `HoconSerializationFormat(sanitize_keys=True)` (cognizant-ai-lab/leaf-common#172, follow-up to #451) in `AbstractAsyncConfigRestorer`, so keys that had to be quoted in HOCON source (e.g. `"llama3.1"`, `"llama3:8b"`, url/file-path keys) no longer come back with pyhocon's quotation marks baked into the string — at every nesting level, for every `.hocon` restorer.

This makes three downstream workarounds for the same parsing artifact unnecessary:

- `DefaultLlmFactory.sanitize_keys()` / `strip_outer_quotes()` — removed
- `ManifestKeyConfigFilter` quote replacement — reduced to whitespace normalization
- `McpServersInfoRestorer` quote replacement — reduced to whitespace normalization

Also bumps the leaf-common floor to 1.2.51 (the constructor kwarg does not exist earlier) and adds a test pinning the sanitized key shape at all nesting levels.

## Behavior changes worth calling out

- **Sanitization is now deep and universal.** Previously only top-level llm-info/manifest/MCP keys were cleaned; nested keys and agent-network/toolbox hocons retained quoted keys. Any deployment that adapted to the old quoted form (e.g. coded tools looking up `args['"user.name"']`, or clients sending quoted sly_data keys to match `allow` entries) should drop those workarounds when upgrading.
- **User llm-info overrides in `.json` format now deep-merge with default entries** for quoted-key model names, instead of accidentally replacing them wholesale (the old post-overlay sanitize collapsed mismatched quoted/clean keys with last-wins semantics). This matches what `.hocon` user files always did; documented in a comment at the overlay site.
- **Quote-embedded keys in `.json` config files are no longer normalized.** The removed workarounds ran regardless of source format; the parse-time sanitization is HOCON-only. JSON parsing never produces such keys, and no documented case exists — JSON keys are now taken exactly as written.

## Follow-up

- #1205 tracks the one remaining unsanitized HOCON path (`LoggingConfigRestorer` via `EasyHoconPersistence`, which cannot pass `sanitize_keys` in leaf-common 1.2.51).

## Testing

- New test: quoted dotted/url keys in a `.hocon` restore come back clean at all nesting levels.
- Full unit suite green (85 passed; remaining skips are environment-based, same as main).
- Verified end-to-end: `DefaultLlmFactory().load()` yields 115 model keys with no embedded quotes and dotted-name lookups resolve.

### DIFF
--- a/neuro_san/internals/graph/persistence/manifest_key_config_filter.py
+++ b/neuro_san/internals/graph/persistence/manifest_key_config_filter.py
@@ -54,9 +54,10 @@ class ManifestKeyConfigFilter(ConfigFilter):
         for key, value in basis_config.items():
 
             # Key here is an agent name in a form that we choose.
-            # Keys sometimes come with quotes.
-            manifest_key: str = key.replace(r'"', "")
-            manifest_key = manifest_key.strip()
+            # Quotes that pyhocon embeds in keys are already removed by
+            # AbstractAsyncConfigRestorer (sanitize_keys=True), so only
+            # whitespace normalization is left to do here.
+            manifest_key: str = key.strip()
 
             filtered[manifest_key] = value
 
--- a/neuro_san/internals/graph/persistence/manifest_key_config_filter.py
+++ b/neuro_san/internals/graph/persistence/manifest_key_config_filter.py
@@ -54,9 +54,8 @@ class ManifestKeyConfigFilter(ConfigFilter):
         for key, value in basis_config.items():
 
             # Key here is an agent name in a form that we choose.
-            # Quotes that pyhocon embeds in keys are already removed by
-            # AbstractAsyncConfigRestorer (sanitize_keys=True), so only
-            # whitespace normalization is left to do here.
+            # Keys are quote-sanitized at parse time (sanitize_keys=True),
+            # so only whitespace normalization is left to do here.
             manifest_key: str = key.strip()
 
             filtered[manifest_key] = value
--- a/neuro_san/internals/persistence/abstract_async_config_restorer.py
+++ b/neuro_san/internals/persistence/abstract_async_config_restorer.py
@@ -163,7 +163,14 @@ class AbstractAsyncConfigRestorer(Restorer, ConfigFilter):
         if file_path.endswith(".hocon"):
             # Worth noting that includes within hocon files can only be done synchronously
             # The core pyhocon library does not support async includes.
-            serialization = HoconSerializationFormat()
+            #
+            # sanitize_keys=True removes the quotation marks that pyhocon embeds
+            # in keys that had to be quoted in the HOCON source because they
+            # contain characters that are special in HOCON keys (e.g. "." or ":"
+            # in model names like "llama3.1", or in url/file-path keys).
+            # Without it, such keys come back as '"llama3.1"' with the quotes
+            # baked into the string. Applies at all nesting levels.
+            serialization = HoconSerializationFormat(sanitize_keys=True)
             # pyhocon parsing mutates process-global pyparsing state and is not
             # thread-safe. See HoconParseLock.
             parse_lock = HoconParseLock()
--- a/neuro_san/internals/persistence/abstract_async_config_restorer.py
+++ b/neuro_san/internals/persistence/abstract_async_config_restorer.py
@@ -164,12 +164,10 @@ class AbstractAsyncConfigRestorer(Restorer, ConfigFilter):
             # Worth noting that includes within hocon files can only be done synchronously
             # The core pyhocon library does not support async includes.
             #
-            # sanitize_keys=True removes the quotation marks that pyhocon embeds
-            # in keys that had to be quoted in the HOCON source because they
-            # contain characters that are special in HOCON keys (e.g. "." or ":"
-            # in model names like "llama3.1", or in url/file-path keys).
-            # Without it, such keys come back as '"llama3.1"' with the quotes
-            # baked into the string. Applies at all nesting levels.
+            # sanitize_keys=True: without it, keys that had to be quoted in the
+            # HOCON source come back as '"llama3.1"' with the quotes baked into
+            # the string (see the HoconSerializationFormat docstring for details).
+            # Applies at all nesting levels.
             serialization = HoconSerializationFormat(sanitize_keys=True)
             # pyhocon parsing mutates process-global pyparsing state and is not
             # thread-safe. See HoconParseLock.
--- a/neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
+++ b/neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
@@ -127,9 +127,6 @@ class DefaultLlmFactory(ContextTypeLlmFactory, LangChainLlmFactory):
             extra_llm_infos: Dict[str, Any] = restorer.restore(file_reference=self.llm_info_file)
             self.llm_infos = self.overlayer.overlay(self.llm_infos, extra_llm_infos)
 
-        # sanitize the llm_infos keys
-        self.llm_infos = self.sanitize_keys(self.llm_infos)
-
         # Resolve any new llm factories
         extractor = DictionaryExtractor(self.llm_infos)
         llm_factory_classes: List[str] = []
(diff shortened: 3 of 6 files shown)


Noravee:
**TL;DR for reviewers**

There is exactly one functional line in this PR: `AbstractAsyncConfigRestorer` now parses `.hocon` files with `HoconSerializationFormat(sanitize_keys=True)` (new in leaf-common 1.2.51), so keys that had to be quoted in HOCON source (`"llama3.1"`, urls, file paths) arrive without pyhocon's baked-in quotes — at every nesting level, for all five restorers (llm info, manifest, agent networks, toolbox, MCP servers info). Everything else is deleting or simplifying the three per-consumer workarounds that existed for that artifact, plus the version floor and a test.

Worth knowing while reviewing:

- Sanitization is now **deep and universal** where it used to be three top-level-only patches. Any deployment that adapted to the old quoted keys must drop that workaround — flagged in the PR description for release notes.
- User llm-info overrides in `.json` format switch from *accidental replace* to *deep-merge* semantics for dotted model names (explained in a new code comment in `load()`).
- `logging.hocon` is the one HOCON path left unsanitized (`EasyHoconPersistence` can't pass the flag) — tracked separately in #1205.

Inline comments below mark the load-bearing spots.

Noravee (commented):
Author annotations on the load-bearing parts — see the TL;DR comment for the overview.

neuro_san/internals/persistence/abstract_async_config_restorer.py
```diff
@@ -163,7 +163,12 @@ def deserialize_file_contents(self, file_path: str, file_contents: bytes) -> Dic
         if file_path.endswith(".hocon"):
             # Worth noting that includes within hocon files can only be done synchronously
             # The core pyhocon library does not support async includes.
-            serialization = HoconSerializationFormat()
+            #
+            # sanitize_keys=True: without it, keys that had to be quoted in the
+            # HOCON source come back as '"llama3.1"' with the quotes baked into
+            # the string (see the HoconSerializationFormat docstring for details).
+            # Applies at all nesting levels.
+            serialization = HoconSerializationFormat(sanitize_keys=True)
```
Noravee, line 171: **The one functional change.** Every `.hocon` restore in the codebase funnels through this line, so all five restorers get clean keys at every nesting level. Everything else in the PR is removal of workarounds this makes redundant.

Note the scope: this branch only — the `.json` branch below stays untouched, so JSON keys are now taken exactly as written (the removed workarounds used to strip quotes from JSON-restored dicts too; no documented case relied on that).

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
```diff
@@ -125,11 +125,15 @@ def load(self):
...
         if self.llm_info_file:
             extra_llm_infos: Dict[str, Any] = restorer.restore(file_reference=self.llm_info_file)
+            # Each user entry DEEP-MERGES with any same-named default entry, so a
+            # sparse override like {"gpt-4.1": {"class": "x"}} inherits the default
+            # entry's remaining fields (including a "use_model_name" alias redirect).
+            # This holds for .json user files too now that keys are sanitized at
+            # parse time: their dotted model names ("llama3.1") match the default
+            # keys. Before sanitization, such keys missed the quoted default keys
+            # ('"llama3.1"') and accidentally replaced the default entry wholesale.
             self.llm_infos = self.overlayer.overlay(self.llm_infos, extra_llm_infos)
```
Noravee, line 135: **Behavior change documented in the comment above:** a `.json` user llm-info entry for a dotted model name previously *replaced* the default entry wholesale — an accident of the quoted/clean key mismatch at overlay time — and now deep-merges with it, matching what `.hocon` user files always did. A sparse override inherits the default entry's remaining fields, including a `use_model_name` alias redirect.

neuro_san/internals/run_context/langchain/llms/default_llm_factory.py
```diff
@@ -561,33 +565,6 @@ def get_max_prompt_tokens(self, config: Dict[str, Any]) -> Optional[int]:
...
-            quote_char = s[0]
-            unquoted = s[1:-1]
-
-            # Only strip quotes if inner value does not contain the same quote char
-            # and does not start with whitespace
-            if quote_char not in unquoted and not unquoted.startswith(" "):
-                return unquoted
-        return s
-
-    def sanitize_keys(self, d: Dict[str, Any]) -> Dict[str, Any]:
```
Noravee, line 580: These two methods were only ever called from `load()`. An org-wide code search found no external callers, and the documented LLM-factory extension point (`LangChainLlmFactory` / `StandardLangChainLlmFactory`) never had them — hence removal without a deprecation shim.

neuro_san/internals/graph/persistence/manifest_key_config_filter.py
```diff
@@ -54,9 +54,9 @@ def filter_config(self, basis_config: Dict[str, Any]) -> Dict[str, Any]:
         for key, value in basis_config.items():
 
             # Key here is an agent name in a form that we choose.
-            # Keys sometimes come with quotes.
-            manifest_key: str = key.replace(r'"', "")
-            manifest_key = manifest_key.strip()
+            # Keys are quote-sanitized at parse time (sanitize_keys=True),
+            # so only whitespace normalization is left to do here.
+            manifest_key: str = key.strip()
```
Noravee, line 59: The whitespace strip is kept deliberately: pyhocon preserves padding inside quoted keys (`" foo "` sanitizes to `' foo '`), so this still repairs padded manifest keys. Consolidating the whitespace handling across all restorers (it exists only here and in the MCP restorer) was considered and deferred — it's a behavior decision for a separate PR.

neuro_san/internals/run_context/langchain/mcp/mcp_servers_info_restorer.py
```diff
@@ -49,11 +49,12 @@ def filter_config(self, basis_config: Dict[str, Any], file_path: str = None) ->
...
 
-        # Now, MCP endpoints urls could put in quotes, so strip them out.
+        # Keys (MCP endpoint urls, quoted in HOCON source) are quote-sanitized
+        # at parse time (sanitize_keys=True), so only whitespace normalization
+        # is left to do here.
         result_dict: Dict[str, Any] = {}
         for key, value in basis_config.items():
-            use_key: str = key.replace(r'"', "")
-            use_key = use_key.strip()
+            use_key: str = key.strip()
```
Noravee, line 57: Same rationale as `ManifestKeyConfigFilter`: quote removal moved to parse time, whitespace normalization kept.

requirements.txt
```diff
@@ -6,7 +6,9 @@
 # requirements-build.txt
 
 # In-house dependencies
-leaf-common>=1.2.50
+# Floor 1.2.51: HoconSerializationFormat(sanitize_keys=True) used by
+# AbstractAsyncConfigRestorer to strip pyhocon's embedded quotes from keys.
+leaf-common>=1.2.51
```
Noravee, line 11: Hard floor: the `sanitize_keys` kwarg does not exist before 1.2.51, so an environment holding a stale leaf-common fails at the first `.hocon` restore with a `TypeError`. Environments that don't re-resolve deps on upgrade (editable installs, cached image layers) need a `pip install -r requirements.txt`.

tests/neuro_san/internals/persistence/test_abstract_async_config_restorer.py
```diff
@@ -200,6 +200,35 @@ def test_deserialize_parses_valid_hocon(self) -> None:
         )
         assert result == VALID_DICT
 
+    def test_deserialize_hocon_sanitizes_quoted_keys(self) -> None:
```
Noravee, line 203: Pins the neuro-san-owned contract — `deserialize_file_contents()` passes `sanitize_keys=True` — via a quoted dotted key at top level and a quoted URL key nested two levels deep. The flag's internals (root-level arrays, constructor default, etc.) are covered by leaf-common's own suite from leaf-common#172.

neuro_san/internals/graph/persistence/manifest_key_config_filter.py
d1donlydfink: Theoretically we could also push that down to leaf-common.

neuro_san/internals/persistence/abstract_async_config_restorer.py
d1donlydfink: This is probably reasonable as JSON may have tighter requirements for keys.

requirements.txt
d1donlydfink: The comment will probably be short lived, but it's fine.

d1donlydfink (approved):
Looks good.

I'd like to recommend conveying a little more of the API change in the PR title, as this is what shows up in the automatic release notes.  Absent that being complete (granted, space is limited in a PR title), please write a small blurb in this PR that should accompany the release notes.

PR #1208, opened 2026-08-06
ActivationPreppers

d1donlydfink:
@hshahrzad is primary.

* Add a notion of an ActivationPrepper interface.
* The new interface allows the ActivationFactory to operate a bit more abstractly and farms specific policy logic out to ActivationPrepper implementations, who in turn create CallableActivations when they are activated in the graph.
* Create ActivationPrepper implementations for External, Toolbox, Class, Branch and FrontMan Activations
* A new env var AGENT_ACTIVATION_PREPPER_CLASSES allows extensions to the standard tool node schema for experimentation in other repos.  At the moment, this is experimental and will not be documented in the Dockerfile.


neuro_san/internals/chat/data_driven_chat_session.py
```diff
@@ -238,6 +238,19 @@ async def run_it(self, inputs: AgentFrameworkMessage) -> AgentFrameworkMessage:
                 # This can happen if we have problems with LLM clients API keys:
                 # Construct a message to send back to the client with the error information.
                 message = AgentFrameworkMessage(content=str(exc))
```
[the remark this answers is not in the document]
d1donlydfink: No. This gives the agent a chance to handle the error.

neuro_san/internals/graph/preppers/activation_prepper.py
```diff
@@ -0,0 +1,83 @@
...
+
+from neuro_san.internals.interfaces.agent_tool_factory import AgentToolFactory
+from neuro_san.internals.interfaces.callable_activation import CallableActivation
+from neuro_san.internals.run_context.interfaces.run_context import RunContext
+
+
+class ActivationPrepper:
+    """
+    Interface for policy objects which prepare a particular kind of activation
+    """
```
d1donlydfink, line 32: New interface class for policy that decides whether and how to construct a particular CallableActivation given a particular tool in the agent spec dictionary.

neuro_san/internals/graph/preppers/activation_prepper.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def is_applicable(self, agent_tool_spec: Dict[str, Any]) -> bool:
+        """
+        :param agent_tool_spec: the agent tool spec dictionary
+        :return: True if this ActivationPrepper is applicable to the given agent tool spec
+        """
+        raise NotImplementedError
```
d1donlydfink, line 39: This method answers the question "Given this agent spec, should I use this Prepper?"

neuro_san/internals/graph/preppers/activation_prepper.py
```diff
@@ -0,0 +1,83 @@
...
+    # pylint: disable=too-many-arguments, too-many-positional-arguments
+    def prepare_activation(self,
+                           name: str,
+                           agent_tool_spec: Dict[str, Any],
+                           parent_agent_spec: Dict[str, Any],
+                           arguments: Dict[str, Any],
+                           sly_data: Dict[str, Any],
+                           parent_run_context: RunContext,
+                           factory: AgentToolFactory,
+                           invocation: str) -> CallableActivation:
```
d1donlydfink, line 50: This method basically says, "Construct the particular CallableActivation that this Prepper pertains to"

neuro_san/internals/graph/preppers/branch_activation_prepper.py
```diff
@@ -0,0 +1,65 @@
...
+from neuro_san.internals.graph.preppers.activation_prepper import ActivationPrepper
+from neuro_san.internals.interfaces.agent_tool_factory import AgentToolFactory
+from neuro_san.internals.interfaces.callable_activation import CallableActivation
+from neuro_san.internals.run_context.interfaces.run_context import RunContext
+
+
+class BranchActivationPrepper(ActivationPrepper):
+    """
+    ActivationPrepper implementation for BranchActivations.
+    """
```
d1donlydfink, line 31: Logic for this guy comes from ActivationFactory.create_agent_activation()

neuro_san/internals/graph/preppers/class_activation_prepper.py
```diff
@@ -0,0 +1,65 @@
...
+    """
+    ActivationPrepper implementation for ClassActivations (coded tools).
+    """
+
+    def is_applicable(self, agent_tool_spec: Dict[str, Any]) -> bool:
+        """
+        :param agent_tool_spec: the agent tool spec dictionary
+        :return: True if this ActivationPrepper is applicable to the given agent tool spec
+        """
+        return agent_tool_spec.get("function") is not None and agent_tool_spec.get("class") is not None
```
d1donlydfink, line 38: Logic for this guy comes from ActivationFactory.create_agent_activation()

neuro_san/internals/graph/preppers/external_activation_prepper.py
```diff
@@ -0,0 +1,99 @@
...
+from neuro_san.internals.interfaces.callable_activation import CallableActivation
+from neuro_san.internals.messages.sly_data_redactor import SlyDataRedactor
+from neuro_san.internals.run_context.interfaces.run_context import RunContext
+from neuro_san.internals.run_context.utils.external_agent_parsing import ExternalAgentParsing
+
+
+class ExternalActivationPrepper(ActivationPrepper):
+    """
+    ActivationPrepper implementation for external agent networks.
+    """
```
d1donlydfink, line 35: Logic for this guy comes from ActivationFactory.create_agent_activation()

neuro_san/internals/graph/preppers/front_man_activation_prepper.py
```diff
@@ -0,0 +1,64 @@
...
+    """
+    ActivationPrepper implementation for FrontManActivations.
+    """
+
+    def is_applicable(self, agent_tool_spec: Dict[str, Any]) -> bool:
+        """
+        :param agent_tool_spec: the agent tool spec dictionary
+        :return: True if this ActivationPrepper is applicable to the given agent tool spec
+        """
+        return agent_tool_spec.get("function") is None
```
d1donlydfink, line 38: Logic for this guy comes from ActivationFactory.create_agent_activation()

neuro_san/internals/graph/preppers/toolbox_activation_prepper.py
```diff
@@ -0,0 +1,65 @@
...
+    """
+    ActivationPrepper implementation for toolbox activations.
+    """
+
+    def is_applicable(self, agent_tool_spec: Dict[str, Any]) -> bool:
+        """
+        :param agent_tool_spec: the agent tool spec dictionary
+        :return: True if this ActivationPrepper is applicable to the given agent tool spec
+        """
+        return agent_tool_spec.get("toolbox") is not None
```
d1donlydfink, line 38: Logic for this guy comes from ActivationFactory.create_agent_activation()

neuro_san/internals/graph/registry/activation_factory.py
```diff
@@ -45,6 +43,15 @@ class ActivationFactory(AgentToolFactory):
...
     """
 
+    # Basic list of stateless ActivationPreppers. Order matters.
+    PREPPERS: List[ActivationPrepper] = [
+        ExternalActivationPrepper(),
+        ToolboxActivationPrepper(),
+        ClassActivationPrepper(),
+        BranchActivationPrepper(),
+        FrontManActivationPrepper(),
+    ]
```
d1donlydfink, line 59: A list of base ActivationPreppers  to use.

neuro_san/internals/graph/registry/activation_factory.py
```diff
@@ -147,49 +154,30 @@ def create_agent_activation(self, parent_run_context: RunContext,
...
-            redacted_sly_data: Dict[str, Any] = self._redact_sly_data(parent_run_context, sly_data)
-
-            # Get the spec for allowing upstream data
-            extractor = DictionaryExtractor(parent_agent_spec)
-            empty = {}
-            allow_from_downstream: Dict[str, Any] = extractor.get("allow.from_downstream", empty)
-
-            agent_activation = ExternalActivation(parent_run_context, factory, name, arguments, redacted_sly_data,
-                                                  allow_from_downstream, invocation)
-            return agent_activation
```
d1donlydfink, line 169: Logic for the above block moves to ExternalActivationPrepper.

neuro_san/internals/graph/registry/activation_factory.py
```diff
@@ -147,49 +154,30 @@ def create_agent_activation(self, parent_run_context: RunContext,
...
-
-        # Merge the arguments coming in from the LLM with those that were specified
-        # in the hocon file for the agent.
-        use_args: Dict[str, Any] = self._merge_args(arguments, agent_tool_spec)
-
-        if agent_tool_spec.get("toolbox") is not None:
-            # If a toolbox is in the spec, this is a shared coded tool where tool's description and
-            # args schema are defined in either AGENT_TOOLBOX_INFO_FILE or toolbox_info.hocon.
-            agent_activation = ToolboxActivation(parent_run_context, factory, use_args, agent_tool_spec, sly_data)
```
d1donlydfink, line 178: Logic for the above block moves to ToolboxActivationPrepper.

neuro_san/internals/graph/registry/activation_factory.py
```diff
@@ -147,49 +154,30 @@ def create_agent_activation(self, parent_run_context: RunContext,
...
-            return agent_activation
-
-        if agent_tool_spec.get("function") is not None:
-            # If we have a function in the spec, the agent has arguments
-            # it wants to be called with.
-            if agent_tool_spec.get("class") is not None:
-                # Agent specifically requested a python class to be run,
-                # and tool's description and args schema are defined in agent network hocon.
-                agent_activation = ClassActivation(parent_run_context, factory, use_args, agent_tool_spec, sly_data)
```
d1donlydfink, line 187: Logic for the above block moves to ClassActivationPrepper.

neuro_san/internals/graph/registry/activation_factory.py
```diff
@@ -147,49 +154,30 @@ def create_agent_activation(self, parent_run_context: RunContext,
...
-            else:
-                agent_activation = BranchActivation(parent_run_context, factory, use_args, agent_tool_spec, sly_data)
```
d1donlydfink, line 189: Logic for the above block moves to BranchActivationPrepper.

neuro_san/internals/graph/registry/activation_factory.py
```diff
@@ -147,49 +154,30 @@ def create_agent_activation(self, parent_run_context: RunContext,
...
-        else:
-            # Get the tool to call from the spec.
-            agent_activation = FrontManActivation(parent_run_context, factory, agent_tool_spec, sly_data)
```
d1donlydfink, line 192: Logic for the above block moves to FrontManActivationPrepper.

neuro_san/internals/graph/registry/activation_factory.py
```diff
@@ -147,49 +154,30 @@ def create_agent_activation(self, parent_run_context: RunContext,
...
+        agent_activation: CallableActivation = prepper.prepare_activation(
+            name,
+            agent_tool_spec,
+            parent_agent_spec,
+            arguments,
+            sly_data,
+            parent_run_context,
+            factory,
+            invocation
+        )
```
d1donlydfink, line 215: New, more abstract logic using the Preppers.

neuro_san/internals/graph/registry/activation_factory.py
```diff
@@ -227,38 +215,3 @@ def get_name_from_spec(self, agent_spec: Dict[str, Any]) -> str:
...
-        :param agent_tool_spec: The dictionary representing the spec registered agent
-        """
-        config_args: Dict[str, Any] = agent_tool_spec.get("args")
-        if config_args is None:
-            # Nothing to override
-            return llm_args
-
-        overlay = DictionaryOverlay()
-        merged_args: Dict[str, Any] = overlay.overlay(llm_args, config_args)
-        return merged_args
```
d1donlydfink, line 248: merge_args() moves to a utility method on the ActivationPrepper interface.

neuro_san/internals/graph/registry/activation_factory.py
```diff
@@ -227,38 +215,3 @@ def get_name_from_spec(self, agent_spec: Dict[str, Any]) -> str:
...
-        :param sly_data: The internal representation of the sly_data to be redacted
-        :return: A new sly_data dictionary, redacted as per the parent spec
-        """
-        parent_spec: Dict[str, Any] = None
-        if parent_run_context is not None:
-            parent_spec = parent_run_context.get_agent_tool_spec()
-
-        redactor = SlyDataRedactor(parent_spec, config_keys=["allow.sly_data", "allow.to_downstream.sly_data"])
-        redacted: Dict[str, Any] = redactor.filter_config(sly_data)
-        return redacted
```
d1donlydfink, line 264: redact_sly_data() moves to be a private method on the ExternalActivationPrepper.

neuro_san/internals/graph/registry/activation_factory.py
```diff
@@ -116,6 +128,34 @@ def _determine_agent_tool_path(self) -> str:
 
         return agent_tool_path
 
+    def _create_preppers(self) -> List[ActivationPrepper]:
+        """
+        Create a list of ActivationPrepper instances to use using BASE_PREPPERS as a basis,
+        then add any externally defined preppers.
+
+        :return: A list of ActivationPrepper instances
+        """
```
d1donlydfink, line 139: Use the BASE_PREPPERS list as a basis, add externally defined peppers as classes are specified in the new env var.  This allows for extension of graph functionality.

neuro_san/internals/graph/registry/activation_factory.py
```diff
@@ -116,6 +128,34 @@ def _determine_agent_tool_path(self) -> str:
...
+        preppers: List[ActivationPrepper] = shallow_copy(self.BASE_PREPPERS)
+
+        # Then read the agent network's prepper list
+        external_preppers: List[ActivationPrepper] = []
+
+        prepper_class_name: str = None
+        agent_prepper_classes: List[str] = environ.get("AGENT_ACTIVATION_PREPPER_CLASSES") or ""
+        for prepper_class_name in agent_prepper_classes.split(" "):
+            prepper_class_name = prepper_class_name.strip()
+            prepper: ActivationPrepper = ResolverUtil.create_instance(prepper_class_name,
```
Noravee, line 149: Since `_create_preppers()` runs in the constructor, and an `ActivationFactory` is created for every `DataDrivenChatSession` (via `AgentToolRegistry`), the classes in `AGENT_ACTIVATION_PREPPER_CLASSES` get re-resolved and re-instantiated on every conversation.

The env var can't change after the process starts, so this repeats the same work every time and always produces the same list. It also gives custom preppers a different lifecycle from the `BASE_PREPPERS` (created once per process), even though both must be stateless. And if a custom prepper's constructor does anything heavy, that cost is paid per conversation instead of once.

Suggestion: resolve the env var once and cache the result (e.g., lazily fill a class-level list on first use), so custom preppers are per-process just like `BASE_PREPPERS`.

neuro_san/internals/graph/registry/activation_factory.py
```diff
@@ -45,6 +46,16 @@ class ActivationFactory(AgentToolFactory):
     That is, this is where neuro-san tools are made real.
     """
 
+    # Basic list of stateless ActivationPreppers. Order matters.
+    BASE_PREPPERS: List[ActivationPrepper] = [
+        ExternalActivationPrepper(),
+        # Note that external prepper definitions get added at this point.
```
Noravee, line 53: My read of the ordering here: because `ExternalActivationPrepper` is always first and matches any `None` spec, custom preppers from the env var can never intercept external agent references — they only get a shot at internal tool nodes.

If I understand correctly, that's intentional and actually a good safety property: the external path is where sly_data redaction happens, so letting a custom prepper run ahead of it could send unredacted sly_data to an external network.

If that's the intent, maybe we should expand this comment to say so explicitly — so a future change doesn't make the insertion point configurable and quietly break that guarantee.

neuro_san/internals/graph/preppers/front_man_activation_prepper.py
```diff
@@ -0,0 +1,64 @@
...
+from typing import Dict
+
+from neuro_san.internals.graph.activations.front_man_activation import FrontManActivation
+from neuro_san.internals.graph.preppers.activation_prepper import ActivationPrepper
+from neuro_san.internals.interfaces.agent_tool_factory import AgentToolFactory
+from neuro_san.internals.interfaces.callable_activation import CallableActivation
+from neuro_san.internals.run_context.interfaces.run_context import RunContext
+
+
+class FrontManActivationPrepper(ActivationPrepper):
```
Noravee, line 28: If I understand correctly, the front man is normally created via `create_front_man()`, which constructs `FrontManActivation` directly and never goes through the prepper chain — so this prepper only fires as a fallback when a spec has no `function`.

That leaves two construction paths for `FrontManActivation`, and it also means custom preppers from the env var can't affect front-man creation, which someone extending via that mechanism might expect to work.

If bypassing the chain in `create_front_man()` is intentional, maybe we should add a short comment there saying so. Otherwise, routing it through this prepper would leave a single construction path.

Noravee (dismissed):
Nice refactor — approving. I checked the new prepper chain against the old if/else dispatch for every spec shape (`None` → External, `toolbox` → Toolbox, `function`+`class` → Class, `function` only → Branch, no `function` → FrontMan) and it is behaviorally equivalent, including the sly_data redaction, arg merging, and error paths.

Main (non-blocking) discussion points are in the inline comments:
1. Custom preppers from `AGENT_ACTIVATION_PREPPER_CLASSES` get re-resolved and re-instantiated per conversation (the factory is created per `DataDrivenChatSession`) — suggest resolving once and caching so they match the per-process lifecycle of `BASE_PREPPERS`.
2. Custom preppers can never intercept external agent references since `ExternalActivationPrepper` is always first — likely intentional (it protects sly_data redaction); suggest documenting that explicitly.
3. `create_front_man()` bypasses the prepper chain, leaving two construction paths for `FrontManActivation` — suggest a clarifying comment if intentional, or routing through the prepper.

**Nits — correctness / robustness**
- `activation_factory.py:144` — `agent_prepper_classes: List[str] = environ.get(...) or ""` is annotated `List[str]` but holds a `str`. Should be `str` (the `List[str]` is what `.split()` produces).
- `activation_factory.py:145` — `.split(" ")` matches existing repo convention, but bare `.split()` would be strictly better: it handles tabs/newlines/multiple spaces and returns `[]` for the empty string, removing the reliance on `ResolverUtil` swallowing empty tokens.
- `activation_factory.py:143` — `prepper_class_name: str = None` is a dead initialization; the loop variable shadows it immediately.
- `activation_prepper.py:43` — `is_applicable()` is typed `Dict[str, Any]` but the contract includes `None` (that's exactly what `ExternalActivationPrepper` matches on). Worth documenting in the interface docstring that the spec can be `None` for external agents, since external implementers will code against this.

**Nits — style / churn**
- The `os.environ` → `from os import environ` change is unrelated churn and leaves the file importing `os` both ways (`activation_factory.py:22-23`), since `os.sep`/`os.pathsep` still need `os`. Suggest picking one style.

neuro_san/internals/graph/registry/activation_factory.py
d1donlydfink: Yup. Almost all of the time the env var path will not be activated due to its experimental nature.
We resolve other classes at speed in other places all the time.
At this point I'm not worried about performance.

neuro_san/internals/graph/preppers/front_man_activation_prepper.py
d1donlydfink: There is such a thing as a front man without a function definition and this catches that.
It's not an ideal situation, but it can theoretically exist. 
For this PR I am trying not to change any logic, merely refactoring, so I wanted to avoid any assumptions that could cause other things to break.

neuro_san/internals/graph/registry/activation_factory.py
d1donlydfink: Yes. Added to the comment as to the intended scope.

d1donlydfink:
Took @Noravee 's advice except for this one:
> activation_factory.py:143 — prepper_class_name: str = None is a dead initialization; the loop variable shadows it immediately.

Yes. My beef with Python is that in the iterator loops you don't get typing of your iterations variable at all.
This is not about the initialization. It's about the typing.

Noravee (approved):
LGTM.

neuro_san/internals/graph/registry/activation_factory.py
```diff
@@ -116,6 +130,34 @@ def _determine_agent_tool_path(self) -> str:
...
+
+        :return: A list of ActivationPrepper instances
+        """
+        preppers: List[ActivationPrepper] = shallow_copy(self.BASE_PREPPERS)
+
+        # Then read the agent network's prepper list
+        external_preppers: List[ActivationPrepper] = []
+
+        prepper_class_name: str = None
+        agent_prepper_classes: str = environ.get("AGENT_ACTIVATION_PREPPER_CLASSES") or ""
```
d1donlydfink, line 146: @hshahrzad : It's possible this env var might not be the best way to extend to do what we want to do.
We'll see.

hshahrzad: Agreed, but it's okay for now since we need something simple to move forward.

hshahrzad (approved):
Looks good to me.

Issue #1209, opened 2026-08-07
Efficiency follow-ups from the pydantic v2 migration: args_schema model caching and JSON-schema title noise


Noravee:
Two related efficiency items surfaced during the code review of the pydantic v2 migration of `BaseModelDictionaryConverter` (follow-up to #1207). Neither is a correctness bug; both waste resources on hot paths and got measurably more expensive with the migration.

## 1. No caching of dynamically created args_schema models

`args_schema` pydantic models are rebuilt from scratch for every tool of every agent activation on every request:

- `AsyncDirectAgentSession.streaming_chat` builds a fresh `DataDrivenChatSession` per request, whose setup runs `LangChainRunContext.create_resources` -> `BaseToolFactory.create_base_tool` -> `LangChainOpenAIFunctionTool.from_function_json` -> `BaseModelDictionaryConverter.from_dict` -> `create_model()` for every tool, plus once more per downstream agent activation (`branch_activation.py`).
- `PydanticParametersNetworkValidator` also builds and discards a model per agent at registry load (and again on dynamic deploys), so large manifests pay the full cost at startup just to throw the models away.

pydantic v2's `create_model()` compiles a Rust core schema + `SchemaValidator` + `SchemaSerializer` per call — benchmarked at ~1.8x the cost of the v1 shim it replaced (~0.34ms vs ~0.19ms per 5-field model), multiplied by nesting depth since each nested object is its own `create_model()` call.

Since internal specs are static registry data, a module-level cache keyed on a canonical serialization of the parameters dict (e.g. `json.dumps(spec, sort_keys=True)`) would collapse this to one compile per distinct spec, and validation at load would warm the cache for the first request. Design considerations:

- `BaseToolFactory.create_function_tool` mutates `function_json["name"] = name` before conversion — the cache key must account for that (or key on the parameters block only, which is what the converter actually consumes).
- External-agent `function_json` arrives over the network per request; a spec-keyed (content-addressed) cache handles that naturally, but needs a size bound.
- Model classes are immutable once created and safe to share across threads/requests.

Possibly relevant to the load-test "unreachable" investigation (150-user EC2 runs), where per-request setup cost is a suspect.

## 2. Pydantic "title" keys leak into anyOf branches of the provider-facing schema

[...]

PR #1210, opened 2026-08-07
Migrate BaseModelDictionaryConverter to native pydantic v2

### LINKED ISSUE #1207: Pydantic v1 UserWarning on Python 3.14: migrate base_model_dictionary_converter to native pydantic v2
## Problem

Running the agent CLI under Python 3.14 emits this warning on every startup:

```
neuro_san/internals/run_context/langchain/core/base_model_dictionary_converter.py:25: UserWarning: Core Pydantic V1 functionality isn't compatible with Python 3.14 or greater.
  from pydantic.v1 import Field
```

**Environment:** Python 3.14.2, pydantic 2.12.5, langchain-core 1.2.9

## Root cause

`base_model_dictionary_converter.py` (lines 25-26) imports `Field` and `create_model` from `pydantic.v1` — the pydantic-v1 compatibility shim bundled inside pydantic 2. Starting with pydantic 2.12, any import of `pydantic.v1` on Python >= 3.14 unconditionally emits this UserWarning (`pydantic/v1/__init__.py`).

The `pydantic.v1` import is a langchain < 0.3 legacy: old langchain required pydantic v1 models for `args_schema`. langchain-core 1.x is pydantic-v2 native, so there is no remaining reason to build v1 models. This is the only `pydantic.v1` import in the neuro-san codebase.

The v1 shim still *functions* on 3.14 today (verified: model creation, validation, and `convert_to_openai_function` all pass), so this is currently noise rather than breakage — but pydantic has declared v1 unsupported on 3.14+ ("minimal support" only as of pydantic 2.13), and this code path builds the args schema for every tool in every agent network.

## Proposed fix

Migrate the one file to native pydantic v2:

```diff
-from pydantic import BaseModel
-from pydantic.v1 import Field
-from pydantic.v1 import create_model
+from typing import Optional
+
+from pydantic import BaseModel
+from pydantic import Field
+from pydantic import create_model
```

Plus one semantic-preserving change in `openai_function_to_pydantic()`:

```diff
             if field_name not in required:
                 field_kwargs["default"] = None
+                # pydantic v1 implicitly made fields with a None default Optional;
+                # v2 does not, so an explicit null from an LLM would fail validation.
+                field_type = Optional[field_type]
             field = Field(**field_kwargs)
```

The `Optional` wrap matters: v1 silently made any `default=None` field nullable; v2 dropped that, so without it an LLM sending an explicit `"arg": null` for an optional parameter would raise a ValidationError at tool-invocation time.

## Verified

Tested the ported converter on both Python 3.14.2 (pydantic 2.12.5) and Python 3.13.5 (pydantic 2.11.7):

- Nested-object recursion, arrays, optional fields with explicit nulls
- `dict(obj)` iteration and `parse_obj` duck-check used by `PydanticArgumentDictionaryConverter` work on v2 instances
- `langchain_core.utils.function_calling.convert_to_openai_function` produces a valid spec
- Downstream consumers (`PydanticParametersNetworkValidator`, `ParametersSchemaNetworkValidator`) use no v1-specific APIs

## Behavior changes to be aware of (all Python versions)

1. **Wire schema for optional fields** changes from `{"type": "integer"}` to `"anyOf": [{"type": "integer"}, {"type": "null"}], "default": null`. This is the standard shape all modern langchain v2 tools produce and provider adapters handle it, but given the Gemini-strictness comment in `langchain_openai_function_tool.py` (~line 164), a regression run against Gemini is worthwhile.
2. **Stricter validation of LLM-supplied args:** v2 lax mode still coerces `"5"` -> int, but rejects lossy coercions v1 performed silently (`1.5` -> int field, `5` -> str field). Rejections surface as validation errors the calling LLM can retry, instead of silent truncation/coercion.

## Related upstream context

- langchain-core itself also imports `pydantic.v1` internally, so one copy of the same warning (attributed to `langchain_core/_api/deprecation.py:25`) remains even after this fix. That one is resolved by upgrading to pydantic >= 2.13, whose `pydantic.v1` namespace ships v1.10.26 with Python 3.14 support and the warning removed. The current `requirements.txt` pin (`pydantic>=2.9.2,<3.0`) already permits 2.13.x.
- https://github.com/langchain-ai/langchain/issues/33926 (closed as external — langchain keeps its compat imports)
- https://github.com/pydantic/pydantic/issues/12618 (pydantic 2.13 removes the warning)

Noravee:
Fixes #1207

## Why

Running the agent CLI under Python 3.14 with pydantic 2.12+ emits `UserWarning: Core Pydantic V1 functionality isn't compatible with Python 3.14 or greater` on every startup, because `BaseModelDictionaryConverter` imported `Field`/`create_model` from the `pydantic.v1` compatibility shim — a langchain <0.3 legacy. langchain 1.x is pydantic-v2 native, and pydantic has declared v1 unsupported on 3.14+, so this migrates the converter (the only `pydantic.v1` import in the codebase) to native v2 and addresses the fallout found in review.

## What

**Core migration**
- `Field`/`create_model` now come from native pydantic v2. Non-required field types are wrapped in `Optional[...]` (v1 made None-default fields implicitly nullable; v2 does not, and LLMs do send explicit nulls). Unrecognized type strings raise explicitly, since v2 accepts the `None` annotation that v1 rejected — without this, bad specs would validate silently.

**Review fixes on top**
- **Standard JSON Schema type names accepted**: `integer` → `int`, `number` → `Union[int, float]` (preserves whichever the LLM sent), `bool` → `bool`. Without aliases, array specs like `{"type": "array", "items": {"type": "integer"}}` that loaded under v1 (as `List[None]`) would hard-fail at registry load and skip the whole network.
- **Invalid parameter names rejected up front**: pydantic v2 crashes with cryptic TypeError/NameError on names v1 tolerated (`model_config`, `model_dump*`/`model_validate*` shadows, leading underscores). `validate_field_name()` rejects them with clear messages before `create_model()` sees them.
- **New `ToolSpecError(ValueError)`**: spec problems are now distinguishable from connectivity errors. Previously a bad external-agent spec was misreported as "Agent/tool was unreachable", and toolbox tools (which get no load-time validation) were dropped with a generic message.
- **Bare `"object"` params map to `Dict[str, Any]` instead of `Any`**: bare `Any` produces an empty JSON schema that langchain-google-genai (every version tested, 2.1.8–4.3.2) degrades to a STRING declaration — Gemini then sends JSON-encoded strings where tools expect dicts. `Dict[str, Any]` yields a real `{"type": "object"}` schema that converts to OBJECT everywhere. Well-formed dict args are unaffected; no shipped registry uses a bare object param.
- **Malformed-spec hardening**: `from_dict(None)` honors the documented None→None contract; explicit-null `properties`/`required`, non-list `required` (whose string form substring-matched field names), missing array `items`, missing/non-string `type` keys (anyOf/enum/union-list schemas) all get clean `ToolSpecError`s instead of raw AttributeError/TypeError crashes. All reachable from hocon registries or unvalidated external-agent JSON.
- **Nested-arg flattening fixed**: `is_pydantic_object()` uses `isinstance(value, BaseModel)` instead of duck-typing on `parse_obj` (deprecated in v2, removed in v3, and defeatable by a field literally named `parse_obj`).

**Documented behavior changes** (in code comments at the causing lines)
- Optional fields serialize as `anyOf: [X, null], default: null` instead of flat `{"type": X}` (OpenAI *strict* mode note included; neuro-san never enables it).
- v2 rejects v1's lossy coercions (int→str, fractional float→int) — surfaces as a correctable tool error to the LLM, not a crashed run.
- Required-ness is now enforced for bare-object params (v1 implicitly made `Any` fields optional even when listed in `required`).
- Names shadowing BaseModel attrs outside protected namespaces (`schema`, `json`, ...) now build with a warning where v1's NameError made validation reject them.

Deferred efficiency follow-ups (args_schema model caching, `_rm_titles` anyOf title noise): #1209.

## Testing

- 20 new unit tests: type aliases, invalid names, bare-object contract (dict passthrough, non-dict rejection, `"type": "object"` schema shape), malformed-spec cases, and `parse_obj`-field flattening regression.
- Full suite: 424 passed; the 17 failures in `zzz_hocons` are pre-existing on `main` (verified by stash-run) and require live LLM keys.
- Verified on Python 3.14.2 + pydantic 2.12.5 with warnings-as-errors: no pydantic.v1 warning, conversion works end to end.
- All 32 shipped registries pass the real validation pipeline; property-name scan of all 95 hocons builds cleanly under v2.
- Gemini schema conversion verified empirically on langchain-google-genai 4.1.2 (repo floor) and 4.3.2 (latest): bare objects now declare OBJECT (was STRING).

## Note: a similar-looking warning remains on Python 3.14 (not from neuro-san)

This PR removes neuro-san's own `pydantic.v1` usage — the original warning source (`base_model_dictionary_converter.py:25`). On Python 3.14 you may still see the same warning *text* from `langchain_core/tools/base.py:38`. That one is langchain-core's backwards-compatibility import of `pydantic.v1`: pydantic < 2.13 warns on that import even though no v1 models are actually used (pydantic/pydantic#12618, discussed in langchain-ai/langchain#34441 — no langchain-core release avoids the import, including latest).

**Remedy**: install `pydantic>=2.13` (it bundles a Python-3.14-compatible v1), verified warning-free with warnings-as-errors. The requirements floor intentionally stays at `>=2.9.2`: the openrouter SDK (pulled in via `langchain-openrouter` in requirements-build.txt) pins `pydantic<2.13`, so forcing the floor would make dev/CI installs unresolvable — pip currently resolves pydantic to 2.12.x when both requirements files are installed together. Cross-referenced comments in both requirements files document this; revisit when openrouter lifts its cap.

## Release note draft

> **Agent network parameter schemas now build on pydantic v2** (was: the deprecated `pydantic.v1` compatibility shim, which Python 3.14 no longer supports). User-visible changes:
>
> - **New**: standard JSON Schema type spellings are accepted in `parameters`/`sly_data_schema`: `"integer"` (= `"int"`), `"number"` (accepts both integers and floats), `"bool"` (= `"boolean"`) — OpenAI/MCP-style function specs now work as-is.
> - **Stricter, with clear errors instead of silent misbehavior or cryptic crashes**:
>   - Parameter names pydantic v2 cannot accept (`model_config`, names shadowing `model_dump*`/`model_validate*`, leading underscores) are rejected at validation/tool-build time. Under v1, underscore names were silently dropped; the others crashed at request time with errors like `'FieldInfo' object is not iterable`.
>   - Unrecognized scalar types now fail validation (v1 silently built the field as untyped).
>   - `"array"` properties must declare `"items"`; `"required"` must be a list of strings (a string value previously required fields by accidental substring match).
> - **Type coercion is stricter** (pydantic v2 semantics): lossy coercions like int→string or 5.5→int are rejected and surface to the calling LLM as correctable tool errors instead of being silently coerced.
> - **Wire format**: optional parameters serialize as `anyOf: [<type>, null], default: null` instead of a flat type declaration. All supported providers accept this; it only matters for schema-strict gateways (e.g. OpenAI strict structured-output mode, which neuro-san does not enable).
> - Bare `"type": "object"` parameters (no `properties`) now reach Gemini as `OBJECT` instead of degrading to `STRING`.
>
> All 32 shipped registries pass unchanged; no hocon changes are required unless a network relies on one of the malformed patterns above, which now gets a precise validation message.


neuro_san/internals/run_context/langchain/core/base_model_dictionary_converter.py
```diff
@@ -80,6 +112,11 @@ def from_dict(self, obj_dict: Dict[str, Any]) -> BaseModel:
                 If obj_dict is not the correct type, it is also reasonable
                 to return None.
         """
+        # Honor the DictionaryConverter contract stated above.  None reaches
+        # here when an external agent reports "parameters": null.
+        if obj_dict is None:
+            return None
```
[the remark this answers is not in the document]
Noravee: Good catch — fixed in 1c561e53. `schema_source` is now normalized from `None` to `{}` at this call site, so an explicit "parameters": null builds the same explicit empty args_schema as a missing parameters block, and the `_arun(*args, **kwargs)` inference path stays disabled. Kept `from_dict(None) -> None` as-is since the DictionaryConverter interface contract documents None -> None. Added regression tests covering both the missing and explicit-null parameters cases.

neuro_san/internals/run_context/langchain/core/base_model_dictionary_converter.py
```diff
@@ -42,13 +46,41 @@ class BaseModelDictionaryConverter(DictionaryConverter):
...
-        #       you need to turn it into one using foo = dict(my_object_arg)
-        #       See: https://docs.pydantic.dev/1.10/usage/exporting_models/#dictmodel-and-iteration
-        #       This is what the PydanticArgumentDictionaryConverter is for.
-        "object": Any
+        # Aliases so that standard JSON Schema / OpenAI function spec type
+        # names work as well as the HOCON-native names above.
+        # "number" maps to a union because JSON Schema's number covers both
+        # integers and floats, and pydantic's smart-mode union preserves
+        # whichever one the caller actually sent.
+        "integer": int,
```
Noravee, line 54: **Why aliases and not just the friendly error**: under v1, an unknown item type inside an array built as `List[None]` and the network *loaded* (the arg worked when omitted). The new explicit raise would have turned specs using the standard `"integer"`/`"number"` spellings into networks silently skipped at registry load — an availability regression. With the aliases, standard OpenAI/JSON-Schema specs now simply work, which is strictly better than both v1 and the unaliased migration. `number` maps to `Union[int, float]` so an integer sent by the LLM stays an `int` (plain `float` would coerce `5` → `5.0` before the CodedTool sees it).

neuro_san/internals/run_context/langchain/core/base_model_dictionary_converter.py
```diff
@@ -42,13 +46,41 @@ class BaseModelDictionaryConverter(DictionaryConverter):
...
+        #         {"type": "object", "additionalProperties": true},
+        #         which converts to a proper OBJECT declaration everywhere.
+        #
+        #       For well-formed arguments the two behave identically: a
+        #       dictionary sent by the LLM reaches the tool as the same
+        #       plain dict.  Non-dict values (e.g. a JSON-encoded string),
+        #       which Any silently passed through with the wrong type, now
+        #       fail pydantic validation and are reported back to the LLM
+        #       as a correctable tool error instead.
+        "object": Dict[str, Any]
```
Noravee, line 83: **The Gemini fix** — this is the change with wire-format impact. Bare `Any` serializes to an empty `{}` JSON schema, which langchain-google-genai (every version tested, 2.1.8→4.3.2) degrades to a `STRING` declaration; Gemini then sends JSON-encoded strings where tools expect dicts. `Dict[str, Any]` emits `{"type": "object", "additionalProperties": true}` → verified `OBJECT` on both our adapter floor (4.1.2) and latest (4.3.2). Only properties-less objects are affected, and no shipped registry has one; objects with properties keep the nested-model path untouched.

neuro_san/internals/run_context/langchain/core/base_model_dictionary_converter.py
```diff
@@ -119,17 +186,94 @@ def openai_function_to_pydantic(self, name: str, function_dict: Dict[str, Any])
...
+                # providers for every non-required field: pydantic v1 emitted a
+                # flat {"type": X}, while v2 emits
+                # {"anyOf": [{"type": X}, {"type": "null"}], "default": null}.
+                # Provider adapters handle the anyOf-with-null shape (Gemini's
+                # converts it to a nullable declaration), but note that
+                # OpenAI's *strict* structured-output mode does not accept the
+                # "default" keyword and langchain does not strip it.  neuro-san
+                # never enables strict mode itself; this only matters for
+                # schema-strict gateways or user code that opts into it.
+                field_type = Optional[field_type]
```
Noravee, line 212: **The one behavior change every provider sees**: optional fields now serialize as `anyOf: [X, null], default: null` instead of v1's flat `{"type": X}`. Verified that Gemini's adapter converts this to a nullable declaration; OpenAI/Anthropic accept it in default mode. The known edge: OpenAI *strict* structured-output mode rejects the `default` keyword — neuro-san never enables strict, so this only matters for schema-strict gateways. The wrap itself is required for v1 parity: without it, an LLM sending an explicit `null` for an optional arg fails validation.

neuro_san/internals/run_context/langchain/core/base_model_dictionary_converter.py
```diff
@@ -95,14 +132,44 @@ def openai_function_to_pydantic(self, name: str, function_dict: Dict[str, Any])
...
+        if not isinstance(required, List):
+            # Also guards the "in" test below: a string here (e.g.
+            # "required": "city") would substring-match unrelated field
+            # names instead of naming one required field.
+            raise ToolSpecError(f"'required' must be a list of field names, got {type(required).__name__}")
+        if any(not isinstance(entry, str) for entry in required):
+            # Non-string entries match nothing in the "in" test below, so
+            # every field would silently become optional; they also crash
+            # downstream validators that use the entries as keys.
+            raise ToolSpecError(f"'required' must contain only field-name strings, got {required!r}")
```
Noravee, line 160: Subtle pre-existing bug this also fixes: a *string* `required` (e.g. `"required": "city"`) used to substring-match unrelated field names — a field named `"it"` would become required because `"it" in "city"`. Non-string entries (`[{}]`) matched nothing, silently making every field optional. Both now fail with clear spec errors.

neuro_san/internals/run_context/langchain/core/base_model_dictionary_converter.py
```diff
@@ -119,17 +186,94 @@ def openai_function_to_pydantic(self, name: str, function_dict: Dict[str, Any])
...
+        try:
+            model: BaseModel = create_model(name, **fields)
+        except (NameError, TypeError) as exception:
+            # Safety net for anything validate_field_name() does not cover:
+            # report whatever create_model() rejects as the same spec-error
+            # type the rest of the pipeline knows how to handle cleanly.
+            raise ToolSpecError(f"Could not create pydantic model '{name}': {exception}") from exception
         return model
 
+    def validate_field_name(self, field_name: str):
```
Noravee, line 237: These names all built fine under the v1 shim (underscore names were warned-and-dropped). Under v2 they raise `TypeError`/`NameError` from inside `create_model` — exception types no handler on the tool-creation path catches, so a single hocon param named `model_config` would abort agent setup for the whole request with a cryptic `'FieldInfo' object is not iterable`. Rejecting them here with `ToolSpecError` routes them through load-time validation like any other spec problem. The docstring documents exactly which names are rejected and which (e.g. `model_name`) remain fine.

neuro_san/internals/run_context/langchain/core/tool_spec_error.py
```diff
@@ -0,0 +1,30 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# END COPYRIGHT
+
+
+class ToolSpecError(ValueError):
```
Noravee, line 19: Subclassing `ValueError` is deliberate: `PydanticParametersNetworkValidator` and every existing broad handler keep working unchanged. Where the distinct *type* gets used shifted with the #1231 merge: `create_external_tool` now separates connectivity from definition errors structurally (scoped try blocks), so a `ToolSpecError` there simply rides the `ValueError` catch into the "invalid function definition" report. The type-based `except ToolSpecError` survives on the toolbox path, which has no such structure and no load-time validation. Historical context: under v1, converter failures were `ConfigError` (a `RuntimeError`), so they escaped `except ValueError` handlers loudly — the migration's switch to `ValueError` semantics is what created the misreporting risk this class guards against.

neuro_san/internals/run_context/langchain/core/base_tool_factory.py
```diff
@@ -115,6 +116,16 @@ async def create_external_tool(self, name: Union[str, Dict[str, Any]]) -> Union[
         try:
             function_json: Dict[str, Any] = await adapter.get_function_json(self.invocation_context)
             return self.create_function_tool(function_json, name)
+        except ToolSpecError as exception:
```
Noravee: *(Updated after the merge with #1231, which restructured this method.)* Originally this PR added `except ToolSpecError` here so a bad *spec* from a reachable external agent wouldn't be journaled as "was unreachable" — at the time a single broad `except ValueError` covered both connectivity and spec problems. #1231 fixed that same misreporting structurally: `get_function_json` (connectivity → "unreachable") and `ensure_external_parameters` + `create_function_tool` (→ "invalid function definition") now sit in separate try blocks, and `ToolSpecError`, being a `ValueError`, lands in the second one. What remains of this PR's change in this file is the toolbox path: `create_toolbox_tool` catches `ToolSpecError` ahead of its generic `ValueError` handler, because toolbox specs aren't covered by registry-load validation — that handler is the first place a bad toolbox spec can be reported.

neuro_san/internals/run_context/langchain/core/langchain_openai_function_tool.py
```diff
@@ -171,6 +178,14 @@ def from_function_json(function_json: Dict[str, Any],
         # function again later on in langchain agent-land.
         converter = BaseModelDictionaryConverter("parameters")
         schema_source: Dict[str, Any] = use_function_json if use_function_json != function_json else {}
+        if schema_source is None:
```
Noravee, line 180: Copilot round-1 catch: once `from_dict(None)` honors the converter's documented None→None contract, a `"parameters": null` spec would have left `args_schema` unset — re-enabling langchain's `_arun(*args, **kwargs)` schema inference, which emits the `args`-array-without-`items` spec Gemini rejects (the exact thing this block's comment says it exists to avoid; previously this input crashed outright). Normalizing at the call site keeps the converter contract intact. *Post-#1231 note*: external agents no longer reach this code with `null` — `ensure_external_parameters()` synthesizes the default `inquiry` parameter first — so this normalization now covers every *other* caller (internal and toolbox specs).

neuro_san/internals/run_context/langchain/core/pydantic_argument_dictionary_converter.py
```diff
@@ -82,5 +82,13 @@ def is_pydantic_object(self, value: Any) -> bool:
...
+        An isinstance() check is correct now that BaseModelDictionaryConverter
+        creates native pydantic v2 models.  The previous duck-typed check
+        (hasattr(value, "parse_obj")) dated from when the values here were
+        pydantic v1 models: it keyed off an API that is deprecated in
+        pydantic v2 and slated for removal in v3, and it returned a wrong
+        False whenever a model had a field literally named "parse_obj"
+        (buildable under v2), which let nested models through unflattened.
         """
-        return hasattr(value, "parse_obj") and callable(value.parse_obj)
+        return isinstance(value, BaseModel)
```
Noravee, line 94: Safe to switch to `isinstance` now because this PR is what makes every model reaching here a native v2 model. The old duck-check had two failure modes: it depended on `parse_obj` surviving (deprecated since 2.0, removed in pydantic 3), and a model with a field literally named `parse_obj` — buildable under v2, verified — made `getattr` return the field value, so `callable()` failed and the nested model reached CodedTools unflattened (`args["nested"]["key"]` → `TypeError`). Regression test covers that exact case.

Noravee (commented):
**TLDR**

This PR does three things, in increasing order of subtlety:

1. **The migration itself** (first commit): `pydantic.v1` shim → native v2 in `BaseModelDictionaryConverter` — kills the Python 3.14 `UserWarning` and gets us off an API pydantic no longer supports. Two semantic-parity shims were needed: `Optional[...]` wrapping for non-required fields, and an explicit raise for unrecognized type strings (v2 silently accepts the `None` annotation v1 rejected).
2. **Fallout fixes**: pydantic v2 is stricter in some places (parameter names like `model_config` now crash `create_model`, standard JSON-Schema type spellings inside array `items` went from loads-fine to network-skipped) and laxer in others (unknown scalar types would have validated silently). New `validate_field_name()`, `TYPE_LOOKUP` aliases, and `ToolSpecError` handle all of it with actionable messages routed to the right handlers.
3. **Hardening**: malformed specs (explicit nulls, missing `items`, missing/union `type` keys, non-list/non-string `required`) now produce clean spec errors instead of raw `AttributeError`/`TypeError` — all reachable from hocon registries, toolbox specs, or external-agent JSON (whose *content* is never validated at registry load, so the converter is the first line of defense there).

**Post-merge integration** (the branch has since absorbed #1231, #1232, and #1234 from main): #1231's restructured `create_external_tool` separates connectivity from definition errors with scoped try blocks, so `ToolSpecError` — a `ValueError` — routes into its "invalid function definition" report on the external path; the type-based catch remains on the toolbox path only. #1231's synthesized default `inquiry` parameter passes the stricter v2 converter, and its new factory test was updated from the v1 `.required` attribute to v2's `.is_required()`. The leaf-common merges (#1232/#1234) don't interact with converter logic.

**Evidence backing the risky bits**: bare-object → `Dict[str, Any]` schema conversion verified empirically on langchain-google-genai 4.1.2 (our floor) and 4.3.2 (latest); all 32 shipped registries pass the real validation pipeline; property names across all 95 hocons build cleanly under v2; full suite 450 green in CI. Deferred efficiency items are tracked in #1209.

Inline notes below point at the decisions most worth scrutinizing.

docs/agent_hocon_reference.md
```diff
@@ -570,8 +570,12 @@ This dictionary has the following keys:
 | type         | A string describing the type of the property. (See below) |
 | default      | An optional default value for the property |
 
-Scalar types here can be "int", "float", "string", "bool".
+Scalar types here can be "int", "float", "string", "bool" or "boolean".
+The standard JSON Schema spellings are also accepted: "integer" (same as "int")
+and "number" (accepts both integers and floats), so function specs copied from
+OpenAI- or MCP-style tool definitions work as-is.
```
d1donlydfink, line 576: Should we add "integer" and "number" to this list in the docs as well given additions below?

neuro_san/internals/run_context/langchain/core/base_model_dictionary_converter.py
d1donlydfink: This is a nice update actually.

neuro_san/internals/run_context/langchain/core/base_model_dictionary_converter.py
d1donlydfink: Might want a release note info prepared about this.

d1donlydfink (dismissed):
In general, Approve. One question about adding to docs.

neuro_san/internals/run_context/langchain/core/pydantic_argument_dictionary_converter.py
d1donlydfink: Much less hacky

docs/agent_hocon_reference.md
Noravee: Done in f1dcc59e — folded the JSON Schema spellings into the enumerated list itself: `"int" or "integer", "float", "number" (accepts both integers and floats), "string", and "bool" or "boolean"`. (Checked the other reference docs — this was the only type-vocabulary list.)

neuro_san/internals/run_context/langchain/core/base_model_dictionary_converter.py
Noravee: Good idea — added a **Release note draft** section to the PR description covering this change (the newly rejected parameter names and what happened to them under v1) plus the other user-visible behavior changes: the new type aliases, stricter coercion, the optional-field wire-format change, and the Gemini bare-object fix. Ready to lift into the release notes when this ships.

d1donlydfink (approved):

PR #1212, opened 2026-08-07
NS-1211 Run external storage expiration checks by only one service instance.

andreidenissov-cog:
Summary

  In multi-instance (multi-process) server deployments, every worker was
  independently running the external-storage expiration/cleanup checks against
  the same shared S3 reservations store. That is redundant work that scales with
  the number of workers. This change makes the expiration checks run on exactly
  one worker, while all workers continue to serve requests normally.

  To do this, the server now forks its worker processes explicitly so it can
  learn each worker's identity, records that identity in the ServerContext, and
  uses it to disable the external-storage check on every worker but one.

  Changes
  
  Explicit worker forking and worker identity (http_server.py, server_context.py):
  - Replaced server.start(http_server_instances) with an explicit
    tornado.process.fork_processes(num_workers) call, resolving num_workers from
    http_server_instances and falling back to os.cpu_count() when it is 0.
  - Capture the returned worker_id and store (worker_id, num_workers) in
    ServerContext via the new set_worker_info(), get_worker_id(), and
    get_num_workers() accessors.
  - Log worker startup (worker_id, pid, count).
  
  Disable expiration on all but one worker (temp_network_storage_updater.py,
  external_storage_util.py):
  - On start(), when num_workers > 1 and worker_id > 0, set the external storage
    check interval to 0 (disabled) so only worker 0 performs expiration.
  - Added ExternalStorageUtil.set_check_interval_seconds(), and hoisted the
    env-var name into the class constant
    EXTERNAL_STORAGE_CHECK_PERIOD_SECONDS_ENV_VAR.
  - TempNetworkStorageUpdater.__init__ now takes a single ServerContext instead
    of network_storage_dict + queues, pulling the storage dict and queues from it,
    so the updater can read worker info.

  S3 check-interval timing fix (s3_reservations_storage.py):
  - Moved the get_check_interval_seconds() read out of __init__ and into start(),
    so it observes env-var overrides applied later in startup (such as the
    per-worker disabling above) instead of latching the value at construction time.

  Related issue: NS-1211

  Testing
  - Verified manually with a multi-worker deployment that expiration checks run on worker 0 only.


neuro_san/service/http/server/http_server.py
```diff
@@ -190,8 +190,18 @@ def start(self, startables: List[Startable]):
...
+        num_workers: int = self.server_config.http_server_instances
+        if num_workers <= 0:
+            num_workers = os.cpu_count() or 1
+
+        # Do not create or access an asyncio/Tornado event loop before this call.
+        worker_id: int = tornado.process.fork_processes(num_workers)
+        self.logger.info({}, "Starting %d worker processes (worker_id=%d, pid=%d)",
+                         num_workers, worker_id, os.getpid())
+        # Register this worker's ID and total number of workers in the server context for use by other components.
+        self.server_context.set_worker_info(worker_id, num_workers)
```
andreidenissov-cog, line 214: Do the processes fork on lower-level API so we can get a unique worker ID for each running instance.
Save this information to instance's ServerContext.

neuro_san/service/main_loop/server_main_loop.py
```diff
@@ -303,7 +303,7 @@ def main_loop(self):
 
         # Another component to start is the temporary networks updater
         temp_networks_updater: TempNetworkStorageUpdater =\
-            TempNetworkStorageUpdater(self.server_context.get_network_storage_dict(), self.server_context.get_queues())
+            TempNetworkStorageUpdater(self.server_context)
         components_to_start.append(temp_networks_updater)
```
andreidenissov-cog, line 307: Change in constructor signature.

neuro_san/service/utils/server_context.py
```diff
@@ -64,6 +64,11 @@ def __init__(self):
         self.server_port: int = AgentSessionConstants.DEFAULT_HTTP_PORT
         self.event_work_queue: AsyncCollatingQueue = AsyncCollatingQueue()
 
+        # Number of worker processes for the server. To be set by the server initialization code.
+        self.num_workers: int = 0
+        # Id of the current worker process. To be set by the server initialization code.
+        self.worker_id: int = 0
+
```
andreidenissov-cog, line 72: For server context of a running neuro-san instance, add this worker unique ID and total number of workers.
Below - setter and getters for new members.

neuro_san/service/watcher/temp_networks/common/external_storage_util.py
```diff
@@ -52,3 +54,16 @@ def get_check_interval_seconds(logger: Logger) -> float:
...
+        """
+        Set expiration interval defined by environment variable
+        to value of check_interval_seconds paarmeter.
+        :param check_interval_seconds: The check interval in seconds to set in the env var.
+        :logger: Logger instance for logging.
+        :return: Nothing.
+        """
+        envvar_name: str = ExternalStorageUtil.EXTERNAL_STORAGE_CHECK_PERIOD_SECONDS_ENV_VAR
+        environ[envvar_name] = str(check_interval_seconds)
+        logger.info("%s is set to %s seconds.", envvar_name, check_interval_seconds)
```
andreidenissov-cog, line 74: Added a setter for external storage expiration check period.
We are getting this parameter from env variable, so we also set it through env variable.
Note that after fork, each worker process has its own copy of env variables, so setting variable value after fork doesn't affect other workers.

neuro_san/service/watcher/temp_networks/s3/s3_reservations_storage.py
```diff
@@ -61,15 +61,16 @@ def __init__(self, bucket_name: str = "", prefix: str = S3Util.DEFAULT_RESERVATI
...
         """
         Initialize the S3 client and validate connection to the bucket.
 
         This method can be called to re-initialize the connection if needed.
         """
+        # Set check interval seconds when we start: value could be overridden by now.
+        # This can throw ValueError if env var is invalid
+        self._check_interval_seconds = ExternalStorageUtil.get_check_interval_seconds(self.logger)
+
         self.reader.start()
```
andreidenissov-cog, line 74: Move this code from constructor to start() method, which is called after the fork and so will access
its own possibly modified env variable.

neuro_san/service/watcher/temp_networks/updater/temp_network_storage_updater.py
```diff
@@ -39,32 +39,30 @@
...
+    def __init__(self, server_context: ServerContext):
         """
         Constructor
 
-        :param network_storage_dict: A dictionary of string (descripting scope) to
-                    AgentNetworkStorage instance which keeps all the AgentNetwork instances
-                    of a particular grouping.
-        :param queues: A Queue of AsyncCollatingQueues for temp network deployment
+        :param server_context: The ServerContext instance for this service.
         """
```
andreidenissov-cog, line 60: Change this constructor parameters: we take all we need from server context anyway,
and now we need even more. So pass in the server context itself.

neuro_san/service/watcher/temp_networks/updater/temp_network_storage_updater.py
```diff
@@ -113,6 +111,16 @@ def start(self):
...
+        # The optimization thing we are doing here:
+        # for multi-instance server deployments, we want to only run expiration checks for external storage
+        # on one of the instances, not all of them, to avoid redundant work.
+        # To do that, we have to set external storage expiration period to 0 for all instances except one.
+        worker_id: int = self.server_context.get_worker_id()
+        num_workers: int = self.server_context.get_num_workers()
+        if num_workers > 1 and worker_id > 0:
+            # Disable expiration checks for all but one worker.
+            ExternalStorageUtil.set_check_interval_seconds(0.0, self.logger)
+
```
andreidenissov-cog, line 127: And here is the meat: if our current worker is one of several in multi-instance setup, but NOT the first one,
disable periodic expiration check for it. Otherwise we'll have several processes banging on the same S3 bucket.

neuro_san/service/watcher/temp_networks/common/external_storage_util.py
```diff
@@ -52,3 +54,16 @@ def get_check_interval_seconds(logger: Logger) -> float:
...
         return check_interval_seconds
+
+    @staticmethod
+    def set_check_interval_seconds(check_interval_seconds: float, logger: Logger) -> None:
+        """
+        Set expiration interval defined by environment variable
+        to value of check_interval_seconds paarmeter.
+        :param check_interval_seconds: The check interval in seconds to set in the env var.
+        :logger: Logger instance for logging.
+        :return: Nothing.
```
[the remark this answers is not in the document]
andreidenissov-cog: Fixed.

neuro_san/service/http/server/http_server.py
```diff
@@ -190,8 +190,18 @@ def start(self, startables: List[Startable]):
         # Bind the socket with a custom backlog
         server.bind(self.http_port, backlog=self.server_config.http_connections_backlog)
 
-        # Start N child processes (0 = one per CPU core)
-        server.start(self.server_config.http_server_instances)
+        # Determine the number of worker processes to start.
+        # If http_server_instances is 0, use the number of CPU cores.
+        num_workers: int = self.server_config.http_server_instances
+        if num_workers <= 0:
+            num_workers = os.cpu_count() or 1
```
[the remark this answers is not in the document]
andreidenissov-cog: Fixed.

neuro_san/service/http/server/http_server.py
```diff
@@ -190,8 +190,18 @@ def start(self, startables: List[Startable]):
...
+
+        # Do not create or access an asyncio/Tornado event loop before this call.
+        worker_id: int = tornado.process.fork_processes(num_workers)
+        self.logger.info({}, "Starting %d worker processes (worker_id=%d, pid=%d)",
+                         num_workers, worker_id, os.getpid())
+        # Register this worker's ID and total number of workers in the server context for use by other components.
+        self.server_context.set_worker_info(worker_id, num_workers)
```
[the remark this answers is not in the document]
andreidenissov-cog: Good catch. Fixed.

neuro_san/service/main_loop/server_main_loop.py
d1donlydfink: This is much saner.  I think we should have all the watcher Startables take a ServerContext as a common constructor pattern, then no one has to be special and watchers can be started more abstractly.

neuro_san/service/watcher/temp_networks/common/external_storage_util.py
d1donlydfink: That is worth a code comment

neuro_san/service/watcher/temp_networks/s3/s3_reservations_storage.py
```diff
@@ -61,15 +61,16 @@ def __init__(self, bucket_name: str = "", prefix: str = S3Util.DEFAULT_RESERVATI
...
-
     def start(self):
         """
         Initialize the S3 client and validate connection to the bucket.
 
         This method can be called to re-initialize the connection if needed.
         """
+        # Set check interval seconds when we start: value could be overridden by now.
+        # This can throw ValueError if env var is invalid
+        self._check_interval_seconds = ExternalStorageUtil.get_check_interval_seconds(self.logger)
```
d1donlydfink, line 72: This same kind of fix will need to be done in AzureBlobReservationsStorage as well.

d1donlydfink (changes requested):
* Also need to make change in AzureBlobReservationsStorage similar to S3
* Put comment about processes in code in ExternalStorageUtil
* Something similar will need to be done with the PeriodicEventInitiator. Doesn't have to be this PR.

Future:
Consider more abstraction for starting watcher processes by passing ServerContext into all of them on construct time (only).

Still have multi-process questions about the queue-of-queues:
* Is this guy getting created before or after MP fork?
* If before, how does fork effect sanity of existing approach?
* If after, we are probably OK (I guess?)

neuro_san/service/watcher/temp_networks/common/external_storage_util.py
andreidenissov-cog: Done.

neuro_san/service/watcher/temp_networks/s3/s3_reservations_storage.py
andreidenissov-cog: Moved this to start().

neuro_san/service/http/server/http_server.py
```diff
@@ -187,16 +188,37 @@ def start(self, startables: List[Startable]):
...
+        if num_workers <= 0:
+            num_workers = os.cpu_count() or 1
+
+        # Bind the listening socket(s) BEFORE forking so every worker inherits
+        # the same bound fd. Do NOT create/touch an IOLoop before fork_processes.
+        sockets = tornado.netutil.bind_sockets(
+            self.http_port, backlog=self.server_config.http_connections_backlog)
+
+        # Do not create or access an asyncio/Tornado event loop before this call.
+        worker_id: int = tornado.process.fork_processes(num_workers)
```
Noravee: One behavior change worth a decision here: the old `server.start(1)` does **not** fork — `num_processes=1` runs in the current process — while an unconditional `fork_processes(num_workers)` forks even for a single worker, turning pid 1 into a tornado supervisor with one child. For any deployment running `http_server_instances=1` (e.g. one instance per pod), two things quietly change:

1. **SIGTERM/graceful shutdown** — k8s signals pid 1, which is now the supervisor; it doesn't forward signals or drain requests, and when it exits the container teardown SIGKILLs the worker mid-request. Previously the serving process received SIGTERM directly.
2. **Crash visibility** — `fork_processes` respawns dead children in place (default `max_restarts=100`), so a crash-looping worker no longer surfaces as container restarts: no k8s restart counter, no crashloop backoff, liveness looks healthy. Previously a crash was a visible container restart.

Suggestion: skip the fork for the single-worker case to preserve the existing process model —

```python
if num_workers > 1:
    worker_id = tornado.process.fork_processes(num_workers)
else:
    worker_id = 0
```

(worker 0 keeps expiration duty either way) — or, if the supervisor model is intended for 1 worker too, document the SIGTERM/restart semantics and verify draining on a pod. Do we know what `http_server_instances` is set to in prod?

andreidenissov-cog: Good find. Put in this change.

d1donlydfink (approved):

PR #1213, opened 2026-08-07
fix macOS multi-instance service startup

### LINKED ISSUE #1199: neuro-san crash when running multi-instance service.
That's the message:

objc[57832]: +[NSNumber initialize] may have been in progress in another thread when fork() was called. We cannot safely call it or ignore it in the fork() child process. Crashing instead. Set a breakpoint on objc_initializeAfterForkError to debug.

arnabnandy7:
## Summary

Fixes #1199 by replacing the fork-based multi-instance service startup with a portable process supervisor model that does not use `fork()`.

Multi-instance mode (`http_server_instances > 1` or `0` for CPU count) now:

- Starts each worker in a fresh Python interpreter using `WorkerSupervisor`.
- Runs one Tornado instance per worker with `SO_REUSEPORT` enabled so workers share the configured HTTP port.
- Assigns worker IDs (`0` to `N-1`) and total worker count via environment variables (`NEURO_SAN_WORKER_ID`, `NEURO_SAN_NUM_WORKERS`), ensuring singleton background tasks (such as S3 reservation expiration checks from #1212) run strictly on Worker 0.
- Completely removes `tornado.process.fork_processes` across supported platforms.
- Forwards termination signals (`SIGINT`, `SIGTERM`) from the supervisor to all workers.
- Propagates worker failures through the supervisor’s exit status.
- Single-instance startup (`http_server_instances=1`) runs directly in `ServerMainLoop` without supervisor overhead.
- Multi-instance mode on Windows raises an explicit `NotImplementedError` explaining `SO_REUSEPORT` limitations.

## Why this change is needed

Tornado’s legacy `fork_processes` implementation creates worker processes using `os.fork()`. In modern Python:

- On macOS, forking after system frameworks (Objective-C runtime, Security, SystemConfiguration) have initialized causes workers to crash with `+[NSNumber initialize] may have been in progress in another thread when fork() was called`. The previous workaround (`OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES`) suppressed the runtime warning but retained the unsafe process model.
- On Linux, `fork()` without `exec()` carries well-documented risks of inheriting pre-fork locks, background threads, and C-extension states (e.g. SSL, gRPC, tokenizers).
- Windows does not support `os.fork()`.

## Why this solution is better

This change replaces `fork()` with a clean, portable supervisor across target platforms:

- **Clean Interpreter State:** Each worker process starts with clean runtime and interpreter state.
- **Port Sharing via `SO_REUSEPORT`:** Follows Tornado's recommended multi-process architecture where independent processes bind the same port using `reuse_port=True`.
- **Worker Coordination:** Workers know their identity, preserving the background task coordination introduced in #1212 where only Worker 0 runs maintenance tasks.
- **Process Lifecycle & Signal Forwarding:** The supervisor tracks all workers, forwards termination signals cleanly, and ensures worker failures are reflected in the supervisor's exit code.
- **Single-instance Continuity:** Single-instance mode (`http_server_instances=1`) remains direct and untouched.

### Automated Tests
- **Worker Supervisor Tests:** `python -m unittest -v tests/neuro_san/service/main_loop/test_worker_supervisor.py` (6/6 passed)
- **Service Test Suite:** `python -m pytest tests/neuro_san/service/` (53 passed, 8 subtests passed)
- **Code Quality:** `flake8` clean, `pylint` rated 10.00/10

### Manual Verification
Smoke test on macOS / Linux:

```bash
AGENT_HTTP_SERVER_INSTANCES=2 \
python -m neuro_san.service.main_loop.server_main_loop
```


d1donlydfink (changes requested):
Thank you for your contribution!
Some recommendations for this PR:

1) Fix your pylint.  From the test Action:

  Running custom pylint command: build_scripts/run_pylint.sh
  Running pylint on directories ' neuro_san quick-start tests':
  ************* Module neuro_san.service.http.config.http_server_config
  neuro_san/service/http/config/http_server_config.py:33:0: R0902: Too many instance attributes (8/7) (too-many-instance-attributes)
  ************* Module neuro_san.service.main_loop.server_main_loop
  neuro_san/service/main_loop/server_main_loop.py:120:19: R1732: Consider using 'with' for resource-allocating operations (consider-using-with)

... you should have been able to do this before submitting. (We also use flake8 on this repo, but you cleared that).

2) You should know your work is a little premature.  At the moment, the server is not entirely safe to run with HTTP_INSTANCES > 1, so we are not yet advertising it in top-level docs like you are.  There are a few background processes that run that would be redundant and/or tripping over each other if we recommend this now.

3) It so happens that we are currently working to make HTTP_INSTANCES > 1 a safe and sane thing. We believe it will come with great performance improvements.    Among this is this PR: https://github.com/cognizant-ai-lab/neuro-san/pull/1212 ... which you will eventually have to merge with.  It will not be the last.  

4) FYI: ServerMainLoop is starting to be bloated with OS-specific junk.  We need to factor that out into its own set of interfaces + implementations.

arnabnandy7:
@d1donlydfink I've made the suggested changes, also merged earlier commits into a squash.

docs/service.md
```diff
@@ -0,0 +1,58 @@
+# Experimental multi-process service operation
```
d1donlydfink: Unnecessary top-level documentation.

We currently document all the env vars in Dockerfiles.
Admitting this document would create yet another place to resist doc-rot for the rest of eternity

Please remove.

neuro_san/service/main_loop/macos_worker_supervisor.py
```diff
@@ -0,0 +1,101 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# END COPYRIGHT
+"""
+Supervises fresh worker processes for multi-instance service startup on macOS.
+"""
```
d1donlydfink: File comments are not necessary when class comments are doing the work.
There should only ever be one class per file anyway. If there is salient detail here, move to class comment.

neuro_san/service/main_loop/worker_supervisor.py
```diff
@@ -0,0 +1,101 @@
...
+        """Return whether this process was started as a supervised macOS worker."""
+        return sys.platform == "darwin" and os.environ.get(MacOsWorkerSupervisor.WORKER_ENV) == "1"
+
+    @staticmethod
+    def run() -> int:
+        """
+        Start fresh macOS worker processes instead of forking this process.
+
+        :return: The supervisor exit code, or -1 when no supervisor is needed.
+        """
```
d1donlydfink, line 93: There are no comments in this method, so no one knows WTF is going on.
There were a bunch of comments in server_main_loop method that this seems to replace that have disappeared. While consice information is good, removing information is generally not good.

d1donlydfink (changes requested):
* Remove service.md doc.  This superfluous doc creates yet another opportunity for doc-rot.
* Comments that were in server_main_loop macOS method as to why special macOS handling was required have disappeared
* There are no comments describing what is going on in MacOsWorkerSupervisor. No one will understand.
* The new test case. A number of things wrong here:

    1) Why does this not derive from TestCase like all the other tests in neuro-san?

    2) This test class name is not reflected in the file name.

    3) What is the value of having this test when pretty much absolutely everything is patched and we are testing macOS-business when the tests themselves run for everyone else on a linux container?  We will never see anything actually break methinks.

I appreciate your ethic for attempting to have a test and docs for everything new you do.
It is noble, and in many ways better than my own practice.
Please examine the gestalt/vibe of what is already going on in the repo before acting on that though.

tests/neuro_san/service/test_server_main_loop.py
```diff
@@ -0,0 +1,76 @@
...
+
+import signal
+import sys
+
+from unittest.mock import patch
+
+from neuro_san.service.main_loop.macos_worker_supervisor import MacOsWorkerSupervisor
+
+
+class TestMacOsWorkerSupervisor:
```
d1donlydfink: A number of things wrong here:

1) Why does this not derive from TestCase like all the other tests in neuro-san?

2) This test class name is not reflected in the file name.

3) What is the value of having this test when pretty much absolutely everything is patched and we are testing macOS-business when the tests themselves run for everyone else on a linux container?  We will never see anything actually break methinks.

docs/service.md
arnabnandy7: removed `docs/service.md` entirely so the Dockerfiles remain the single source of truth for these environment variables.

neuro_san/service/main_loop/macos_worker_supervisor.py
arnabnandy7: removed the module-level docstring and moved the relevant context into the `MacOsWorkerSupervisor` class documentation.

neuro_san/service/main_loop/worker_supervisor.py
arnabnandy7: Moved the implementation out of `ServerMainLoop` should not have removed the reasoning behind it.

tests/neuro_san/service/test_server_main_loop.py
arnabnandy7: Renaming the file and deriving from `TestCase` would address the style concerns, but not the more important issue: the test mocked the platform and subprocess behavior so heavily that it did not meaningfully verify the macOS failure mode.

I removed the test rather than retaining a low-value test solely for coverage.

arnabnandy7:
I restored that context in the class documentation and added focused comments throughout `run()`. It now explains:

- Why Tornado’s normal fork-based startup is unsafe on macOS.
- Why disabling Objective-C fork safety only suppresses the symptom.
- Why fresh Python interpreters are used for workers.
- How the worker marker prevents recursive supervision.
- How `0` is translated into one worker per CPU.
- Why workers use shared-port binding.
- How subprocess cleanup, signal forwarding, and failure propagation work.

d1donlydfink (dismissed):

andreidenissov-cog:
Good PR @arnabnandy7 !
Here is my general comment though - if we are going for more hands-on control of multi-instance service,
why limit ourselves to separate OS-specific branch?
Could you extend this approach to be general and portable across all our target platforms?
It could be follow-up PR though, not necessarily this one.

arnabnandy7:
> Good PR @arnabnandy7 ! Here is my general comment though - if we are going for more hands-on control of multi-instance service, why limit ourselves to separate OS-specific branch? Could you extend this approach to be general and portable across all our target platforms? It could be follow-up PR though, not necessarily this one.

@andreidenissov-cog let me try here, in case needed I'll ask for follow up PR.

andreidenissov-cog (dismissed):
Approved with the expectation that this approach will be extended to all our target platforms.

arnabnandy7:
@andreidenissov-cog @d1donlydfink based on request above, I've adjusted the changes to incorporate OS agnostic approach.

neuro_san/service/main_loop/worker_supervisor.py
```diff
@@ -0,0 +1,166 @@
...
+                workers.append(
+                    stack.enter_context(subprocess.Popen(worker_command, env=worker_env))
+                )
+
+            def stop_workers(_signal_number, _frame):
+                """Terminate every worker which is still running."""
+                for worker in workers:
+                    if worker.poll() is None:
+                        worker.terminate()
+
```
andreidenissov-cog: Can this function be class-level, not nested?
Our code rules are to refrain from nested functions unless really necessary.

arnabnandy7: @andreidenissov-cog it's updated! 

I moved `stop_workers` to be a `class-level` method on `WorkerSupervisor` and added dedicated unit test coverage.

andreidenissov-cog (approved):
Approved to move things forward.

andreidenissov-cog:
I've merged your PR, but please look at follow-up issue in neuro-san kanban board: #1343.

PR #1214, opened 2026-08-07
Add unit coverage for load-test minimal chat filter

arnabnandy7:
## Summary

Adds focused unit coverage for the load-test `--minimal` chat-filter option already implemented on the base branch.

## Coverage

- Verifies the default filter remains `maximal`
- Verifies `--minimal` selects the minimal filter
- Verifies the minimal filter is forwarded to `agent_cli`


vince-leaf (approved):
LGTM

vince-leaf (changes requested):
Reasons:

Name collision with an existing concept. The load test already has --level min (min/norm/adv). A bare --minimal reads like "minimal level", not "minimal chat filter". Two unrelated "minimal"s in the same CLI is the confusing part.
It hides a behavioral side effect. --chat-filter minimal at least names the thing it changes; --minimal silently turns off client-side LLM/token reporting (the exact surprise you just hit).
No real ergonomic win. It saves 14 characters on a flag you set once in a wrapper script, and it violates the team's long-form-flag doctrine (--yes not -y) — an alias is the same trade-off in the opposite direction.
Two ways to say one thing means help text, README, and tests all have to cover both forever.

arnabnandy7:
> Reasons:
> 
> Name collision with an existing concept. The load test already has --level min (min/norm/adv). A bare --minimal reads like "minimal level", not "minimal chat filter". Two unrelated "minimal"s in the same CLI is the confusing part. It hides a behavioral side effect. --chat-filter minimal at least names the thing it changes; --minimal silently turns off client-side LLM/token reporting (the exact surprise you just hit). No real ergonomic win. It saves 14 characters on a flag you set once in a wrapper script, and it violates the team's long-form-flag doctrine (--yes not -y) — an alias is the same trade-off in the opposite direction. Two ways to say one thing means help text, README, and tests all have to cover both forever.

I've made the changes based on observations, please review.

vince-leaf (approved):
LGTM

vince-leaf (commented):
Nits (none blocking):

- The reviewer thread history is stale: your CHANGES_REQUESTED asked for --chat-filter minimal instead of --minimal, but the base branch itself settled on --minimal, so the PR correctly follows the base now. Nothing left to act on there.
- test_minimal_filter_is_forwarded_to_agent_cli passes chat_filter_type="MINIMAL" (uppercase) and only asserts "--minimal" in command; there's no negative case asserting MAXIMAL does not add --minimal, and no HTTP-client-mode coverage — description says "forwarded to agent_cli" which is only the subprocess path. Minor coverage gaps, not correctness issues.
- Class docstring says "minimal chat filter" while one test checks the maximal default — trivial.

andreidenissov-cog (approved):
LGTM.

vince-leaf:
@arnabnandy7, I have added comments above. Just wondering, do you want to add that enhancement now or later?
Note: I have made some updates related to the matter, and they are already in main.

arnabnandy7:
> @arnabnandy7, I have added comments above. Just wondering, do you want to add that enhancement now or later? Note: I have made some updates related to the matter, and they are already in main.

Looking into this @vince-leaf thanks for the feedbacks

arnabnandy7:
@vince-leaf I've added the test as requested

vince-leaf:
> @vince-leaf I've added the test as requested

@arnabnandy7 
FYI: Minor observation, not blocking: the HTTP test's FakeSession.streaming_chat returns an empty iterator that is never consumed because CapturingProcessor.process_once short-circuits — fine for asserting request-state plumbing; it just doesn't exercise the time-to-first-response wrapper in http_client.py. No action needed.

Good enough — merge as is. The test's purpose is to prove the --minimal filter lands in the HTTP request state, and it does that; the time-to-first-response timing wrapper is a separate concern that's already covered by test_http_client_timeout.py.

PR #1215, opened 2026-08-07
Recover from InvalidToken credential rejections in temp-network S3 access

Noravee:
## The problem (production, happening now)

Prod pods sporadically report just-created temp networks as **404 / not found** even though the reservation object is safely in S3 — cognizant-ai-lab/neuro-san-studio#1310. From the Datadog logs:

```
S3ReservationsReader: S3 error processing reservation object <id> during sync:
An error occurred (InvalidToken) when calling the GetObject operation:
The provided token is malformed or otherwise invalid.
```

Mechanism:

1. The S3 workers cache a **frozen credential snapshot** and sign every request with it.
2. When that snapshot goes stale, S3 rejects requests — in prod's case with **`InvalidToken`**, not `ExpiredToken`.
3. The credential-retry gate (`S3Util.is_expired_token_error`) matches **only `ExpiredToken`**, so the stale snapshot is never invalidated.
4. The reader's error handler logs the `ClientError` and returns "not found" — so the API 404s, on every read, until a pod restart.

It looks sporadic because the pod that created the network still serves it from memory (`/function` works), while the other pods must read S3 and fail (`/connectivity` 404s).

## The fix (minimal, recovery-only)

New `S3Util.is_credential_rejection_error()` classifies all three of S3's temporary-token rejection codes as *reset credentials and retry*:

- `ExpiredToken` / `ExpiredTokenException` (what the gate already matched)
- `InvalidToken` — "malformed or otherwise invalid" (what prod is actually getting)
- `TokenRefreshRequired` — "must be refreshed" (the third member of the same family)

Both workers' retry loops and the expiration sweep's re-raise now classify through it. `InvalidAccessKeyId` / `SignatureDoesNotMatch` deliberately still surface — those mean rotated long-lived key pairs or signing bugs, which re-resolution cannot fix and retrying would mask. The retry warning now names the rejection code it is recovering from instead of assuming expiry.

Why this is enough for prod: prod runs on IRSA (`serviceAccount.roleArn`), so the credential chain re-resolves to fresh role credentials whenever the reset fires. The wedge was never that recovery was impossible — it was that the gate never fired for `InvalidToken`.

## What this deliberately does NOT do

This fixes **recovery**, not **prevention**: pods still take the failed call + retry hit whenever the snapshot goes stale, and a rejection that outlasts the retry budget can still surface as a 404. Eliminating the failure mode itself — dropping the frozen-snapshot caching entirely so botocore's credential stack auto-refreshes at request-signing time — is the redesign in #1157, kept separate so this small fix can ship while that design is discussed.

## Tests

- `TestCredentialRejectionGate` — pins the gate's boundary in both directions: the four codes that must reset, the four (`InvalidAccessKeyId`, `SignatureDoesNotMatch`, `AccessDenied`, `NoSuchKey`) that must surface.
- `TestReaderInvalidTokenRecovery` — reproduces the #1310 failure mode end-to-end on the read path: requests signed with the first (stale) credential snapshot are rejected with `InvalidToken`; the test asserts the read *succeeds* after the gate resets and re-resolves. **Verified red against the current gate**: without the widening it fails with a `(None, None)` read after a single credential resolution — logging the exact prod error line above.

Fixes the recovery half of cognizant-ai-lab/neuro-san-studio#1310. Related: #1153, #1157.

### DIFF
--- a/neuro_san/service/watcher/temp_networks/s3/aws_sync_client_worker.py
+++ b/neuro_san/service/watcher/temp_networks/s3/aws_sync_client_worker.py
@@ -85,13 +85,13 @@ class AwsSyncClientWorker:
                 return retval
 
             except ClientError as err:
-                # S3Util.is_expired_token_error() is used instead of raw
-                # DictionaryExtractor access: the extractor returns a stored
-                # None in preference to its default, and
-                # '"ExpiredToken" not in None' would raise TypeError inside
+                # S3Util.is_credential_rejection_error() is used instead of
+                # raw DictionaryExtractor access: the extractor returns a
+                # stored None in preference to its default, and substring
+                # checks against a None code would raise TypeError inside
                 # this handler, masking the original ClientError
                 # (see S3Util.get_error_code for details).
-                if not S3Util.is_expired_token_error(err):
+                if not S3Util.is_credential_rejection_error(err):
                     raise
 
                 last_err = err
@@ -100,13 +100,17 @@ class AwsSyncClientWorker:
                 #             profiles have token-based credentials that may expire.
                 #             See: https://docs.aws.amazon.com/boto3/latest/guide/configuration.html
 
-                # Reset the cached credentials as they are likely expired and try again.
+                # Reset the cached credentials - S3 rejected them (expired,
+                # or malformed/mismatched, e.g. captured mid-rotation; see
+                # S3Util.is_credential_rejection_error) - and try again.
                 self.frozen_credentials = None
                 if source is None:
                     source = self.name
-                self.logger.warning("%s (%d): %s credentials seem to have expired. Retrying. "
-                                    "If you believe you have non-expiring %s credentials, be sure they are correct.",
-                                    source, attempt, self.aws_service, self.aws_service)
+                self.logger.warning("%s (%d): %s rejected the request's credentials (%s). Retrying with "
+                                    "re-resolved credentials. If you believe you have valid non-expiring "
+                                    "%s credentials, be sure they are correct.",
+                                    source, attempt, self.aws_service, S3Util.get_error_code(err),
+                                    self.aws_service)
 
         # Exhausted retries
         if last_err is not None:
--- a/neuro_san/service/watcher/temp_networks/s3/s3_reservations_expiration.py
+++ b/neuro_san/service/watcher/temp_networks/s3/s3_reservations_expiration.py
@@ -243,13 +243,14 @@ class S3ReservationsExpiration:
             if error_code in ("NoSuchKey", "404"):
                 self.logger.debug("%s: Reservation %s was already removed by another process", self.name, obj_key)
                 expired = True  # Object is gone, which is the desired outcome for expiration
-            elif S3Util.is_expired_token_error(exception):
-                # Credential expiry must NOT be swallowed here. The client this sweep
+            elif S3Util.is_credential_rejection_error(exception):
+                # Credential rejection (ExpiredToken / InvalidToken /
+                # TokenRefreshRequired) must NOT be swallowed here. The client this sweep
                 # uses was created from frozen credentials (see AwsSyncClientWorker),
                 # which never auto-refresh, and the only code that can refresh them is
                 # retry_with_new_client() wrapping expire_any_reservations() at the top
                 # of the sweep - and it can only react to ClientErrors that actually
-                # reach it. If ExpiredToken were merely logged like the codes below,
+                # reach it. If these errors were merely logged like the codes below,
                 # every remaining key in the sweep would make one doomed S3 call,
                 # nothing would be expired, and the sweep would still report success;
                 # credentials would only refresh when a later sweep's list_objects_v2
--- a/neuro_san/service/watcher/temp_networks/s3/s3_util.py
+++ b/neuro_san/service/watcher/temp_networks/s3/s3_util.py
@@ -98,6 +98,40 @@ class S3Util:
         return "ExpiredToken" in S3Util.get_error_code(err)
 
     @staticmethod
+    def is_credential_rejection_error(err: ClientError) -> bool:
+        """
+        :param err: The ClientError exception to evaluate
+        :return: True if S3 rejected the credentials the request was signed
+                 with, in a way that discarding the cached credential state
+                 and re-resolving it can plausibly fix:
+
+                 * ExpiredToken / ExpiredTokenException - the session token
+                   has expired.
+                 * InvalidToken - "malformed or otherwise invalid", e.g. a
+                   credential state captured mid-rotation or a revoked role
+                   session. Observed in production (neuro-san-studio issue
+                   #1310): a stale cached credential snapshot was rejected
+                   with InvalidToken on every read, and because the retry
+                   gate matched only ExpiredToken, the snapshot was never
+                   invalidated - every S3 read of the affected network
+                   failed (surfacing as a 404 for a network that exists in
+                   S3) until a pod restart. Exact match on purpose; there
+                   is no InvalidTokenException variant for S3, and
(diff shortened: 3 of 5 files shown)


neuro_san/service/watcher/temp_networks/s3/s3_util.py
```diff
@@ -97,6 +97,40 @@ def is_expired_token_error(err: ClientError) -> bool:
         """
         return "ExpiredToken" in S3Util.get_error_code(err)
 
+    @staticmethod
+    def is_credential_rejection_error(err: ClientError) -> bool:
```
Noravee, line 101: **The whole fix, one predicate.** Both workers' retry loops and the expiration sweep's re-raise classify through here, so the policy has a single edit point. The docstring records the #1310 evidence for `InvalidToken` and why `InvalidAccessKeyId`/`SignatureDoesNotMatch` stay excluded. `InvalidToken`/`TokenRefreshRequired` are exact matches on purpose — S3 has no `...Exception` variants of them, and near-misses like `InvalidClientTokenId` belong to other services. (#1157's version of this same function additionally handles the local `NoCredentialsError`/`PartialCredentialsError` cases, which only matter under its keyless-client design.)

neuro_san/service/watcher/temp_networks/s3/aws_sync_client_worker.py
```diff
@@ -85,13 +85,13 @@ def retry_with_new_client(self, work_function: Callable, *, source: str = None)
...
-                # None in preference to its default, and
-                # '"ExpiredToken" not in None' would raise TypeError inside
+                # S3Util.is_credential_rejection_error() is used instead of
+                # raw DictionaryExtractor access: the extractor returns a
+                # stored None in preference to its default, and substring
+                # checks against a None code would raise TypeError inside
                 # this handler, masking the original ClientError
                 # (see S3Util.get_error_code for details).
-                if not S3Util.is_expired_token_error(err):
+                if not S3Util.is_credential_rejection_error(err):
```
Noravee, line 94: **Where prod got wedged.** This gate previously read `is_expired_token_error(err)`, so an `InvalidToken` rejection re-raised here on attempt 1 — the reset below never ran, the reader logged-and-swallowed, and the stale snapshot survived to fail the next read too. The recovery machinery was always right there; it just never fired for the code prod was actually getting.

neuro_san/service/watcher/temp_networks/s3/aws_sync_client_worker.py
```diff
@@ -100,13 +100,17 @@ def retry_with_new_client(self, work_function: Callable, *, source: str = None)
                 #             profiles have token-based credentials that may expire.
                 #             See: https://docs.aws.amazon.com/boto3/latest/guide/configuration.html
 
-                # Reset the cached credentials as they are likely expired and try again.
+                # Reset the cached credentials - S3 rejected them (expired,
+                # or malformed/mismatched, e.g. captured mid-rotation; see
+                # S3Util.is_credential_rejection_error) - and try again.
                 self.frozen_credentials = None
```
Noravee, line 106: **The recovery mechanism (unchanged by this PR).** Nulling the snapshot makes the next attempt re-freeze from `session.get_credentials()` — on IRSA that's a refreshable credential object, so the re-freeze picks up fresh role credentials and the retry succeeds. The warning below now names the actual rejection code instead of assuming expiry, so the next Datadog search can tell recovery from failure.

neuro_san/service/watcher/temp_networks/s3/s3_reservations_expiration.py
```diff
@@ -243,13 +243,14 @@ def expire_one_reservation(self, obj_key: str, current_time: float,
             if error_code in ("NoSuchKey", "404"):
                 self.logger.debug("%s: Reservation %s was already removed by another process", self.name, obj_key)
                 expired = True  # Object is gone, which is the desired outcome for expiration
-            elif S3Util.is_expired_token_error(exception):
-                # Credential expiry must NOT be swallowed here. The client this sweep
+            elif S3Util.is_credential_rejection_error(exception):
```
Noravee, line 246: Same widening for the sweep: a credential rejection must re-raise to the outer `retry_with_new_client()` wrapper (the only code that can reset the snapshot) rather than fall into the log-and-continue else branch — otherwise every remaining key makes one doomed S3 call and the sweep reports success having expired nothing.

tests/neuro_san/service/watcher/temp_networks/s3/test_invalid_token_recovery.py
```diff
@@ -0,0 +1,211 @@
...
+                "Code": code,
+                "Message": f"{code} (test)",
+            },
+            "ResponseMetadata": {"HTTPStatusCode": 400},
+        },
+        operation_name,
+    )
+
+
+class TestCredentialRejectionGate(TestCase):
```
Noravee, line 83: Gate boundary pinned in both directions: the four codes that must trigger a credential reset, and the four (`InvalidAccessKeyId`, `SignatureDoesNotMatch`, `AccessDenied`, `NoSuchKey`) that must keep surfacing. A future "helpful" widening or narrowing of the gate breaks one of these on purpose.

tests/neuro_san/service/watcher/temp_networks/s3/test_invalid_token_recovery.py
```diff
@@ -0,0 +1,211 @@
...
+                    "expiration_time_in_seconds": time.time() + 3600.0,
+                },
+                "stored_at": time.time(),
+            },
+        }
+        self.fake_s3.objects[f"reservations/{reservation_id}.json"] = \
+            json.dumps(spec).encode("utf-8")
+        return reservation_id
+
+    def test_invalid_token_triggers_credential_reset_and_read_succeeds(self):
```
Noravee, line 150: **The #1310 reproduction.** Requests signed with the first (stale) credential resolution are rejected with `InvalidToken`; the re-frozen snapshot works. Against main's gate this test fails exactly the way prod does — it logs the character-for-character error line from the Datadog report and `get_one_reservation()` returns `(None, None)` after a single resolution. With the widening, the read recovers on the retry.

Noravee (commented):
**TL;DR — minimal recovery fix for the prod `InvalidToken` 404s** (cognizant-ai-lab/neuro-san-studio#1310)

* **The bug:** the workers sign requests with a cached frozen-credential snapshot. When it goes stale, prod S3 rejects reads with `InvalidToken` — but the retry gate matched only `ExpiredToken`, so the snapshot was never invalidated. The reader swallowed the error and returned 404 for networks that exist in S3, on every read, until a pod restart. It looked sporadic because the creator pod served from memory while the other pods had to read S3 and fail.
* **The fix:** one classification predicate (`S3Util.is_credential_rejection_error`) covering all three of S3's temporary-token rejection codes — `ExpiredToken`/`ExpiredTokenException`, `InvalidToken`, `TokenRefreshRequired` — wired into both workers' retry gates and the expiration sweep's re-raise. On IRSA (prod), the reset re-freezes from a refreshable credential object, so recovery is one retry away; the wedge was only that the gate never fired.
* **Deliberately excluded:** `InvalidAccessKeyId`/`SignatureDoesNotMatch` — rotated key pairs or signing bugs, which re-resolution can't fix and retrying would mask.
* **Scope:** recovery only, smallest possible diff. Preventing the failure mode entirely (dropping snapshot caching so botocore auto-refreshes at signing time) is the redesign in #1157, kept separate for the design discussion.
* **Tests:** gate boundary pinned in both directions, plus an end-to-end read-path recovery test — verified red against the current gate, where it fails by logging the character-for-character prod error line. 23 passed, pylint 10.00/10, flake8 clean.

andreidenissov-cog (approved):
Looks good as a quick fix.

PR #1216, opened 2026-08-07
Messages refactor

d1donlydfink:
This looks like a lot, but this is mostly some files moving around and the corresponding imports being changed.
Packages moved:
* neuro_san.internals.messages -> neuro_san.message.types
* neuro_san.internals.filters -> neuro_san.message.filters
* neuro_san.message_processing -> neuro_san.message.processors

Classes moved:
* neuro_san.internals.messages.origination -> neuro_san.internals.journals.origination
* neuro_san.internals.run_context.utils.external_agent_parsing -> neuro_san.intermals.utils.external_agent_parsing
* neuro_san.internals.messages.sly_data_redactor -> neuro_san.message.utils.sly_data_redactor

This actually straightens out some import tangles at the package level where old message_processing was reaching into internals and internals was reaching back out to message_processing. Now neuro_san.messages is clearly a lower-level package than neuro_san.internals.

There is a new DeprecationRedirect class which handles state for class rename redirections and warning notices
given an old class name and a new class name.  This is employed in the top level `neuro_san/__init__.py` file in the form of a dictionary.   The DeprecationRedirect class will eventually move to leaf-common.

Basically the idea is that old references will still work until they can be repaired, but will emit a warning message the first time (and only the first time) they are used.

Tested manually by doing stuff like:
`from neuro_san.internals.messages.chat_message_type import ChatMessageType`
a few times to 1st see the warning message only once, and then again to not.
Worth noting that this was what Copilot was erroneously worried about.

Also tried AND and it's still fine.


neuro_san/client/simple_one_shot.py
```diff
@@ -21,7 +21,7 @@
 from neuro_san.client.agent_session_factory import AgentSessionFactory
 from neuro_san.client.streaming_input_processor import StreamingInputProcessor
 from neuro_san.interfaces.agent_session import AgentSession
-from neuro_san.message_processing.basic_message_processor import BasicMessageProcessor
+from neuro_san.message.processors.basic_message_processor import BasicMessageProcessor
```
d1donlydfink, line 24: Typical (boring) import change.

neuro_san/message/filters/answer_message_filter.py
```diff
@@ -18,8 +18,8 @@
 from typing import Dict
 from typing import List
 
-from neuro_san.internals.filters.message_filter import MessageFilter
-from neuro_san.internals.messages.chat_message_type import ChatMessageType
+from neuro_san.message.filters.message_filter import MessageFilter
+from neuro_san.message.types.chat_message_type import ChatMessageType
```
d1donlydfink, line 22: Typical file moved to the top-level neuro_san.messages package.  Also boring.
It too necessarily has import changes.

neuro_san/internals/utils/deprecation_redirect.py
```diff
@@ -0,0 +1,145 @@
...
+    """
+    Utilities for redirecting deprecated classes based on a single dictionary of the format:
+        {
+            "<fully_qualified_old_class_name>": "<fully_qualified_new_class_name>",
+            ...
+        }
+    The idea is that one global instance of this class lives in a module's __init__.py file
+    and is initialized with a dictionary described above and manages class redirections and
+    deprecation warnings from there.
+    """
```
d1donlydfink, line 40: This class is the meat of the logic change.

neuro_san/__init__.py
```diff
@@ -16,10 +14,43 @@
...
+            "neuro_san.internals.journals.origination.Origination",
+        "neuro_san.internals.parsers.structure.json_structure_parser.JsonStructureParser":
+            "neuro_san.message.parsers.structure.json_structure_parser.JsonStructureParser",
+        "neuro_san.message_processing.message_processor.MessageProcessor":
+            "neuro_san.message.processors.message_processor.MessageProcessor",
+        "neuro_san.message_processing.basic_message_processor.BasicMessageProcessor":
+            "neuro_san.message.processors.basic_message_processor.BasicMessageProcessor",
+    },
+    next_version="0.7.0"
+)
```
d1donlydfink, line 49: The DeprecationRedirect instance is initialize with class mappings from old -> new.
There actually are a lot more classes that have moved, but I am particularly cataloging the ones that are used in neuro-san-studio and nsflow.

neuro_san/__init__.py
```diff
@@ -16,10 +14,43 @@
...
+        "neuro_san.internals.messages.chat_message_type.ChatMessageType":
+            "neuro_san.message.types.chat_message_type.ChatMessageType",
+        "neuro_san.internals.messages.origination.Origination":
+            "neuro_san.internals.journals.origination.Origination",
+        "neuro_san.internals.parsers.structure.json_structure_parser.JsonStructureParser":
+            "neuro_san.message.parsers.structure.json_structure_parser.JsonStructureParser",
+        "neuro_san.message_processors.message_processor.MessageProcessor":
+            "neuro_san.message.processors.message_processor.MessageProcessor",
+        "neuro_san.message_processors.basic_message_processor.BasicMessageProcessor":
+            "neuro_san.message.processors.basic_message_processor.BasicMessageProcessor",
```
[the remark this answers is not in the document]
d1donlydfink: Addressed.

neuro_san/internals/utils/deprecation_redirect.py
```diff
@@ -0,0 +1,142 @@
...
+
+    def redirect_modules(self):
+        """
+        Redirect deprecated classes and their modules
+        """
+        old_class: str = None
+        for old_class in self.old_class_to_new_class.keys():
+            old_module: str = self.get_module_from_fully_qualified(old_class)
+            modules[f"{self.module_name}.{old_module}"] = modules[self.module_name]
+
```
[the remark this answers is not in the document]
d1donlydfink: No they will not.  That has been tested manually and it works just fine.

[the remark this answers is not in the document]
Noravee: Copilot is actually right on this one — the manual test passes only because `neuro_san` was already imported (warm process). In a fresh interpreter the import machinery imports the parent packages before `__init__` gets to install the leaf aliases:

```
$ python -c "from neuro_san.internals.messages.chat_message_type import ChatMessageType"
ModuleNotFoundError: No module named 'neuro_san.internals.messages'
```

5 of the 6 mapped paths fail like this cold (`external_agent_parsing` survives only because its old parent dir still exists on disk). nsflow's `agent_log_processor.py` does exactly this import at module top, so whether it crashes depends on import order. Even warm, only the `from <leaf> import <Class>` spelling is redirected — `import x.y.z as m`, package-level imports, `mock.patch("old.path...")`, unpickling ≤0.6.x data, and pylint/IDE resolution all still fail.

Suggestion: real stub modules at the old locations (each one a re-export from the new path plus the DeprecationWarning) — fixes all of the above with less machinery than the `sys.modules` aliasing.

neuro_san/internals/utils/deprecation_redirect.py
```diff
@@ -0,0 +1,145 @@
...
+            "<fully_qualified_old_class_name>": "<fully_qualified_new_class_name>",
+            ...
+        }
+    The idea is that one global instance of this class lives in a module's __init__.py file
+    and is initialized with a dictionary described above and manages class redirections and
+    deprecation warnings from there.
+    """
+
+    def __init__(self, module_name: str, old_class_to_new_class: Dict[str, str], next_version: str = None,
+                 do_not_redirect_modules_for_testing: bool = False):
```
Noravee: Secondary: DeprecationRedirect has no unit tests, and warm manual testing can't catch the cold-import case — worth a subprocess-based test running each mapped old import in a fresh interpreter, plus a loop asserting every map value resolves. Also `do_not_redirect_modules_for_testing` is currently unused.

neuro_san/internals/utils/deprecation_redirect.py
```diff
@@ -0,0 +1,145 @@
...
+        self.next_version: str = next_version
+        self.do_not_redirect_modules_for_testing: bool = do_not_redirect_modules_for_testing
+
+        self.warned: Set[str] = set()
+
+        self.clean_keys()
+        if not self.do_not_redirect_modules_for_testing:
+            self.redirect_modules()
+
+    def clean_keys(self):
```
Noravee, line 59: Nits: mutates the caller's dict in place; `key.replace(...)` strips every occurrence rather than just the prefix (`removeprefix` is safer); and iterating a `set` randomizes map order, so if two entries ever share a bare class name, which one wins `find_old_class_key`'s `endswith` match is nondeterministic per process.

neuro_san/__init__.py
```diff
@@ -16,10 +14,45 @@
...
+
+_DEPRECATION_REDIRECT = DeprecationRedirect(
+    __name__,
+    # A map from old class name to new class name for compatibility
+    {
+        "neuro_san.internals.messages.chat_message_type.ChatMessageType":
+            "neuro_san.message.types.chat_message_type.ChatMessageType",
+        "neuro_san.internals.messages.origination.Origination":
+            "neuro_san.internals.journals.origination.Origination",
+        "neuro_san.internals.parsers.structure.json_structure_parser.JsonStructureParser":
```
Noravee, line 39: Two mismatches with the PR description: the `internals.parsers.structure` → `message.parsers.structure` move isn't mentioned there, and `SlyDataRedactor` is listed as a headline move but has no entry in this map — intended?

tests/neuro_san/message/utils/test_sly_data_redactor.py
```diff
@@ -19,7 +19,7 @@
 
 from unittest import TestCase
 
-from neuro_san.internals.messages.sly_data_redactor import SlyDataRedactor
+from neuro_san.message.utils.sly_data_redactor import SlyDataRedactor
```
Noravee, line 22: Minor: this file (and `tests/neuro_san/internals/parsers/structure/test_json_structure_parser.py`) stayed at the old tree location while the source moved to `neuro_san/message/` — move them so the test tree keeps mirroring the source tree?

Noravee:
**Main compat concern** (also raised in the now-resolved thread on `redirect_modules()` — reposting top-level so it doesn't get lost):

The redirect only works if `neuro_san` is already imported. The `sys.modules` aliases are installed when `neuro_san/__init__.py` runs, but on a cold import of an old path Python imports the parent packages first — and `neuro_san.internals.messages` no longer exists on disk:

```
$ python -c "from neuro_san.internals.messages.chat_message_type import ChatMessageType"
ModuleNotFoundError: No module named 'neuro_san.internals.messages'
```

5 of the 6 mapped paths fail this way in a fresh process (`external_agent_parsing` works only because its old parent dir still exists).

Real examples from the repos this map is targeting — nsflow, current main:

- `nsflow/backend/utils/agentutils/agent_log_processor.py:24-25`
  `from neuro_san.internals.messages.chat_message_type import ChatMessageType`
  `from neuro_san.message_processing.message_processor import MessageProcessor`
- `nsflow/backend/utils/agentutils/async_streaming_input_processor.py:26`
  `from neuro_san.internals.messages.origination import Origination`

Whether those lines survive under this PR depends on whether *some other* module happened to import `neuro_san` earlier in the same process. That's probably true in the full nsflow app today — but a direct script import, pytest collecting one of these files first, or any import-order refactor turns it into the `ModuleNotFoundError` above.

Even when warm, only the `from <old.leaf> import <Class>` spelling is redirected: `import x.y.z as m`, package-level forms (`from neuro_san.internals.messages import chat_message_type`), `mock.patch("neuro_san.message_processing...")`, unpickling data saved under ≤0.6.x, and pylint/IDE resolution (E0401 on the old paths) all still fail.

Suggested fix: real stub modules at the old locations — e.g. `neuro_san/internals/messages/chat_message_type.py` re-exporting from the new path and emitting the DeprecationWarning. Real parent packages on disk fix every case above (cold imports, all import spellings, mock.patch, pickles, static analysis) with less machinery than the sys.modules aliasing; `DeprecationRedirect` could stay for the warn-once bookkeeping.

neuro_san/__init__.py
d1donlydfink: The intention is to let imports in neuro-san-studio and nsflow work without modification.
I wasn't keen on mapping absolutely everything. Most clients are not reaching into neuro_san.internals

tests/neuro_san/message/utils/test_sly_data_redactor.py
d1donlydfink: Nice catch. Done.

neuro_san/internals/utils/deprecation_redirect.py
d1donlydfink: Made it deterministic with removeprefix() and created a new dictionary as it went along.

neuro_san/internals/utils/deprecation_redirect.py
d1donlydfink: Added simple tests. Removed do_not_redirect_modules_for_testing.

d1donlydfink:
> The redirect only works if neuro_san is already imported. The sys.modules aliases are installed when neuro_san/__init__.py runs, but on a cold import of an old path Python imports the parent packages first — and neuro_san.internals.messages no longer exists on disk:

This should be fixed now

d1donlydfink:
>Real examples from the repos this map is targeting

All of these examples work just fine under both cold and warm starts from fresh python interpreter.

> mock.patch("neuro_san.message_processing..." ...

If you are patching the old reference, you are doing something wrong.
This is typically something that happens in tests within neuro_san, and whoever is moving stuff around should be able to get all those references within their PR because those tests will fail.

Yes people could be patching stuff in external tests. At this point it's highly unlikely because the way the framework is set up, you trust the framework and concentrate on tests for your own classes.

> Suggested fix: real stub modules at the old locations ...

Yes this is always an option, especially when class names are the same in 2 places, however it's not a necessity this time around. In general, I like the idea of being able to keep all the deprecations in a single place, as opposed to having to go hunting for them.  What I really wanted this time around was to be able to delete the old package references, which I should be able to do this time around given the circumstances.

If I were moving one class from a package that remained to another package (new or otherwise), I'd have to stub the module in the old location, otherwise this scheme would not work.

Noravee (approved):
LGTM.

Verified the fix against the new head, each in a fresh interpreter:

- All 6 mapped old paths import cold, including the nsflow pair from `agent_log_processor.py`
- `mock.patch("neuro_san.message_processing...")` now works
- Pickles saved under ≤0.6.x paths load cold, with the correct one-time DeprecationWarning
- Real packages (e.g. `neuro_san.internals.journals`) are not shadowed by the aliasing

Main concern resolved — approving.

Issue #1217, opened 2026-08-07
Per-request langfuse flush blocks request completion; +112% completion latency at 200 concurrent requests


Noravee:
## Summary

With `LANGFUSE_ENABLED=true`, request **completion latency degrades severely under concurrency** — up to **+112% average / +142% P0** in a 200-concurrent-request load test — while first-response latency is unaffected. The root cause is that neuro-san calls the langfuse SDK's **blocking `flush()` at the end of every request**, which stalls the finishing request (and its executor's event loop thread) until the **process-wide shared span queue** has been delivered to the Langfuse server over sequential HTTPS round-trips.

The flush buys almost nothing: the SDK's background batch exporter ships the same data within ≤5 seconds on its own, and the SDK registers an `atexit` flush for process exit. The per-request wait converts telemetry delivery time into user-facing request latency.

## Production-scale observation (what surfaced this)

200 identical requests, same container (`neuro-san==0.6.90`, `neuro-san-studio==0.3.16`), only `LANGFUSE_ENABLED` differing:

| Metric | LF=false | LF=true | Delta |
|---|---|---|---|
| Wall time | 179.29s | 289.34s | **+61.4%** |
| Average duration | 126s | 267s | **+111.9%** |
| P0 (fastest request) | 101.2s | 245.5s | **+142.6%** |
| P50 | 124.0s | 265.7s | +114.3% |
| P90 | 153.1s | 271.8s | +77.5% |
| P100 | 179.2s | 289.2s | +61.4% |
| First response min/avg/max | 33s / 48s / 63s | 31s / 50s / 61s | ~unchanged |
| Success | 200/200 | 200/200 | same |

Note the signature: the **floor (P0) inflates the most** and the latency spread **compresses** (P0–P100 narrows from 78s to 44s). That is a convoy pattern — all completions queueing behind one shared resource — not a uniform slowdown.

## Root cause — code trace

1. `DataDrivenChatSession.close_of_work()` (`neuro_san/internals/chat/data_driven_chat_session.py`) runs at the end of every chat request and awaits `self.tracing_context.flush()`.
2. `LangfuseTracingContext.flush()` (`neuro_san/internals/run_context/langchain/tracing/langfuse_tracing_context.py`) calls the **synchronous** `self.langfuse_client.flush()` directly on the executor's asyncio event loop thread — no `run_in_executor`, so the loop thread is pinned for the duration:
   ```python
   async def flush(self):
[...]

PR #1218, opened 2026-08-07
Improvements for Langfuse performance

d1donlydfink:
Found by @vince-leaf 
Yields ~10% performance improvement.

### DIFF
--- a/neuro_san/internals/run_context/langchain/tracing/langfuse_tracing_context.py
+++ b/neuro_san/internals/run_context/langchain/tracing/langfuse_tracing_context.py
@@ -357,7 +357,7 @@ class LangfuseTracingContext(LangChainTracingContext):
         # At the end of every request, a call to Langfuse's SDK flush(), is a blocking HTTP upload.
         # It drains the whole process's span queue, not just that request's.
         # It runs on a shared event loop, so while it runs, other requests freeze.
-        # The more concurrent requests, the bigger the queue, the longer everyone waited.
+        # The more concurrent requests, the bigger the queue, the longer everyone waits.
         # What you lose: the guarantee that a request's traces are uploaded before the request is marked done.
         # The Langfuse SDK's background thread uploads them anyway; the only real exposure is the final batch if
         # the process is killed — which is what a shutdown drain would cover.
--- a/neuro_san/internals/run_context/langchain/tracing/langfuse_tracing_context.py
+++ b/neuro_san/internals/run_context/langchain/tracing/langfuse_tracing_context.py
@@ -131,8 +131,8 @@ class LangfuseTracingContext(LangChainTracingContext):
             callback_handler: BaseCallbackHandler = callback_handler_type()
             # Langfuse's handler is synchronous, so LangChain routes every trace event through
             # a thread pool. Each of those ThreadPoolExecutor.submit calls grabs a process-global lock,
-            # and with thousands of events × 50 requests, threads just queued on it — 76% of active time
-            # in the profile. This run_inline = True tells LangChain to call the handler directly,
+            # and with thousands of events, threads just queue on it.
+            # This run_inline = True tells LangChain to call the handler directly,
             # skipping the thread pool entirely.
             callback_handler.run_inline = True
 
@@ -161,6 +161,11 @@ class LangfuseTracingContext(LangChainTracingContext):
             # Registering a second hook with the same instance is safe:
             # langchain adds a hook's handler only if that exact object is
             # not already among the run's handlers.
+
+            # Also set run_inline = True to skip the thread pool entirely with the foreign handler.
+            # This is safe since run_inline only changes how langchain dispatches events, not what it does.
+            # See other run_inline comment in _ensure_registered() above.
+            foreign_handler.run_inline = True
             context_var = ContextVar(LANGFUSE_HANDLER_VAR_NAME, default=foreign_handler)
             register_configure_hook(context_var, inheritable=True)
             cls.HANDLER_CONTEXT_VAR = context_var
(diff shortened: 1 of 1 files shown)


neuro_san/internals/run_context/langchain/tracing/langfuse_tracing_context.py
```diff
@@ -129,6 +129,12 @@ def _ensure_registered(cls) -> ContextVar:
 
             # Create the callback handler instance
             callback_handler: BaseCallbackHandler = callback_handler_type()
+            # Langfuse's handler is synchronous, so LangChain routes every trace event through
+            # a thread pool. Each of those ThreadPoolExecutor.submit calls grabs a process-global lock,
+            # and with thousands of events × 50 requests, threads just queued on it — 76% of active time
+            # in the profile. This run_inline = True tells LangChain to call the handler directly,
+            # skipping the thread pool entirely.
+            callback_handler.run_inline = True
```
d1donlydfink, line 137: Vince's Fix #2

neuro_san/internals/run_context/langchain/tracing/langfuse_tracing_context.py
```diff
@@ -347,5 +353,13 @@ async def flush(self):
...
+        # Do nothing.
+        # At the end of every request, a call to Langfuse's SDK flush(), is a blocking HTTP upload.
+        # It drains the whole process's span queue, not just that request's.
+        # It runs on a shared event loop, so while it runs, other requests freeze.
+        # The more concurrent requests, the bigger the queue, the longer everyone waited.
+        # What you lose: the guarantee that a request's traces are uploaded before the request is marked done.
+        # The Langfuse SDK's background thread uploads them anyway; the only real exposure is the final batch if
+        # the process is killed — which is what a shutdown drain would cover.
+        # if self.langfuse_client is not None:
+        #    self.langfuse_client.flush()
```
d1donlydfink: A different way to do Vince's Fix #1.

vince-leaf: Fine!

vince-leaf (dismissed):
LGTM

vince-leaf (dismissed):
LGTM

neuro_san/internals/run_context/langchain/tracing/langfuse_tracing_context.py
Noravee, line 137: Suggestion: also set this on the adoption path. `_adopt_existing_handler_hook()` re-registers a *foreign* handler instance (the coexistence case from #1192), and that instance won't have `run_inline` set — so a process that adopts an existing `langfuse_handler` hook silently keeps the slow per-event thread-pool dispatch. One line in the visible-adoption branch covers it:

```python
foreign_handler.run_inline = True
```

(plus a short comment noting we deliberately mutate the foreign handler — safe, since `run_inline` only changes how langchain dispatches to it, not what it does). The invisible-hook branch can't be helped since we have no handler reference there.

Also a nit on this comment block: the profile anecdotes ("thousands of events × 50 requests", "76% of active time") won't mean much to future readers, and the lock is per-executor (contention across per-loop executor threads + GIL) rather than process-global. The durable statement is: langchain dispatches sync handlers via `await run_in_executor(...)` per event unless `run_inline` is set; with `run_inline=True` the handler's in-memory span bookkeeping runs on the event loop (uploads stay on the SDK's background thread). See #1217 for the measurements.

neuro_san/internals/run_context/langchain/tracing/langfuse_tracing_context.py
```diff
@@ -347,5 +353,13 @@ async def flush(self):
         """
         Flush the tracing context.
         """
-        if self.langfuse_client is not None:
-            self.langfuse_client.flush()
+        # Do nothing.
+        # At the end of every request, a call to Langfuse's SDK flush(), is a blocking HTTP upload.
+        # It drains the whole process's span queue, not just that request's.
+        # It runs on a shared event loop, so while it runs, other requests freeze.
```
Noravee: The behavior change is right, but this comment describes a different mechanism than the one verified in #1217 — worth correcting since future maintainers will reason from it:

- `flush()` isn't itself "a blocking HTTP upload" — it **waits** until the SDK's background exporter has drained the **process-wide shared queue** (`tracer_provider.force_flush()`, default 30s timeout, plus two unbounded `queue.join()`s). The uploads always happen on the SDK's background thread.
- "It runs on a shared event loop, so while it runs, other requests freeze" — the server runs one executor/event loop per request (`AsyncioExecutorPool`), so there is no shared loop to freeze. The cross-request coupling is the shared span queue + single exporter thread: every finishing request waits for *everyone's* pending spans, which is why in the 200-request data the *floor* inflated most (P0 +142%) while the tail grew least (P100 +61%). There is also a capacity cost: pinned executors can't return to the pool.
- "…which is what a shutdown drain would cover" — reads like missing work, but the langfuse SDK already registers an `atexit` shutdown flush, so graceful exits are covered today. The only true loss window is a hard kill (SIGKILL/OOM) eating the last ≤5s of spans.

Suggest replacing the block with a couple of accurate lines + `See https://github.com/cognizant-ai-lab/neuro-san/issues/1217`, and deleting the commented-out code below (git history and the issue preserve it).

Noravee (dismissed):
Approving — both changes are the right fixes for the langfuse latency regression (#1217), and they match what we validated empirically: removing the per-request flush recovered essentially the whole regression in an A/B/C container test (LF=true within ~2% of LF=false at 100 concurrent, traces still delivered), and the `run_inline` change addresses the executor-dispatch contention found in profiling. The flush fix is also at the right layer — `close_of_work()`'s `TracingContext.flush()` hook stays intact and only the langfuse implementation changes.

Two non-blocking suggestions inline: one small coverage gap on the adopted-handler path, and a correction to the mechanism described in the `flush()` comment so future readers reason from the verified behavior. Full analysis and measurements are in #1217.

andreidenissov-cog (dismissed):
Comments do need fixing, but otherwise LGTM.

neuro_san/internals/run_context/langchain/tracing/langfuse_tracing_context.py
d1donlydfink: Advice taken.

neuro_san/internals/run_context/langchain/tracing/langfuse_tracing_context.py
d1donlydfink: Both bits of advice taken.

Noravee (approved):
LGTM.

Issue #1222, opened 2026-08-09
Phase 1: Internal content-block safety (zero wire-schema change)


Noravee:
Part of #1221. **Zero wire-schema change — a bug-fix phase, delivered as six small PRs, each individually revertible.**

Non-string LLM content is currently destroyed (or crashes) at several fixed flatten points (see #1221 for the full findings). This phase makes the entire in-server pipeline block-safe without touching the wire format. Because Anthropic `thinking` and OpenAI `reasoning`/`reasoning_effort` are already whitelisted in `llm_config`, these are live-bug fixes.

### Decisions taken

- The `_arun` repr fix ships (as its own PR — see PR 6). **Done: merged in #1226.**
- Intra-request replay of reasoning blocks into provider history is accepted: it only affects agents whose hocon opted into thinking, and it is required for correct Anthropic extended-thinking + tool-use (signed thinking blocks must be preserved in history).
- The `text`-value corrections (see PR 2) ship as release-noted bug fixes, not behind a flag.
- #1216 (messages refactor) has merged; paths below reflect the post-refactor layout (`neuro_san/message/{types,filters,parsers,processors,utils}`).

### PR 1 — `content_utils` + test infrastructure *(no behavior change)* — **Done: merged in #1276.**

- [ ] New `neuro_san/message/utils/content_utils.py` (a single `ContentUtils` class per repo convention) — the single flatten/normalize policy replacing today's two divergent flattens:
  - `flatten_to_text(message_or_content) -> str` — `BaseMessage.text` semantics (concatenate top-level strings + `type=="text"` blocks); handles str, list-of-str, list-of-dict without crashing; does NOT strip (callers that strip today keep stripping).
  - `standard_blocks(message) -> List[dict]` — wraps `.content_blocks` + JSON sanitizer (bytes→base64 str, non-JSON scalars→str). Must **exclude `tool_call`/`tool_call_chunk`/`invalid_tool_call` blocks** (merged in from `.tool_calls` by the property) — otherwise every existing Anthropic tool-calling deployment (list content with `tool_use` blocks, no reasoning enabled) starts emitting blocks in Phase 2 and golden parity breaks.
  - `is_trivial(blocks) -> bool` — single text block, no annotations/extras (the emission gate, used in Phase 2).
[...]

PR #1226, opened 2026-08-09
Return tool message content instead of message object from _arun

Noravee:
Part of #1222 (PR 6 of the Phase 1 ladder — independent of the others, does not depend on #1216).

## Problem

`LangChainOpenAIFunctionTool._arun()` returns the `AgentToolResultMessage` **object**. Langchain passes `str` (and content-block list) tool returns through to the calling LLM's `ToolMessage` as-is, but falls back to `str()` for any other object — and `str()` of a pydantic message is its repr. So the calling LLM literally sees

```
content='<the actual answer>' additional_kwargs={} response_metadata={} type='agent_tool_result' name=None id=None tool_calls=[] invalid_tool_calls=[] usage_metadata=None tool_result_origin=[{'tool': ...}]
```

as the sub-agent's answer, on **every** sub-agent call: wasted tokens, prompt noise the parent LLM has to unwrap, and internal field names (`tool_result_origin`) leaking into prompts.

## Fix

Return `the_answer.content` (with a `None` guard — a `Run` can legitimately carry no tool message, see `submit_tool_outputs()`). This is deliberately `.content` rather than `.content_blocks` or `.text`: `.content` is the carried payload, so when tool answers later become content-block lists, langchain passes the list through natively and this method needs no further change.

## Behavior change note (for release notes)

The `ToolMessage` text every calling LLM sees changes from the pydantic repr wrapper to the clean answer text. Networks whose prompts were (accidentally) calibrated against the repr format may shift behavior.

## Testing

- New `tests/.../core/test_langchain_openai_function_tool.py`: answer comes back as content (a `str`, not the message object); `None` tool message returns `None` without dereferencing; exception path still returns `str(exception)`.
- `music_nerd_pro_multi_agents`

### DIFF
--- a/neuro_san/internals/run_context/langchain/core/langchain_openai_function_tool.py
+++ b/neuro_san/internals/run_context/langchain/core/langchain_openai_function_tool.py
@@ -209,7 +209,11 @@ It's function_json is described thusly:
         ("a" is for asynchronous).
 
         :return: The content of the BaseMessage that tells us the "answer"
-                 from the tool. Can be None if the tool produced no message.
+                 from the tool - a string today, though langchain also allows
+                 lists of content blocks.
+                 Can be None if the tool produced no message.
+                 On failure, returns the string form of the exception so the
+                 calling LLM can verbally recognize the problem.
         """
         run: LangChainRun = None
         try:
--- a/neuro_san/internals/run_context/langchain/core/langchain_openai_function_tool.py
+++ b/neuro_san/internals/run_context/langchain/core/langchain_openai_function_tool.py
@@ -208,8 +208,8 @@ It's function_json is described thusly:
         calls to another llm/agent instance happen asynchronously.
         ("a" is for asynchronous).
 
-        :return: The result of the tool in the form of the BaseMessage
-                 that tells us the "answer" from the tool.
+        :return: The content of the BaseMessage that tells us the "answer"
+                 from the tool. Can be None if the tool produced no message.
         """
         run: LangChainRun = None
         try:
@@ -246,4 +246,12 @@ It's function_json is described thusly:
 
         # Assume the answer is latest tool message in the Run.
         the_answer: BaseMessage = run.get_tool_message()
-        return the_answer
+        if the_answer is None:
+            return None
+
+        # Return the message's content rather than the message object itself.
+        # Langchain passes str (and content-block list) tool returns through
+        # to the calling LLM's ToolMessage as-is, but falls back to str() for
+        # any other object - which for a BaseMessage is its pydantic repr
+        # ("content='...' additional_kwargs={} ..."), not the answer.
+        return the_answer.content
--- a/tests/neuro_san/internals/run_context/langchain/core/test_langchain_openai_function_tool.py
+++ b/tests/neuro_san/internals/run_context/langchain/core/test_langchain_openai_function_tool.py
@@ -1,0 +1,100 @@
+
+# Copyright © 2023-2026 Cognizant Technology Solutions Corp, www.cognizant.com.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# END COPYRIGHT
+
+from unittest.mock import AsyncMock
+from unittest.mock import MagicMock
+
+import pytest
+
+from neuro_san.internals.messages.agent_tool_result_message import AgentToolResultMessage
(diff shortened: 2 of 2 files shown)


neuro_san/internals/run_context/langchain/core/langchain_openai_function_tool.py
```diff
@@ -208,8 +208,8 @@ async def _arun(
         calls to another llm/agent instance happen asynchronously.
         ("a" is for asynchronous).
 
-        :return: The result of the tool in the form of the BaseMessage
-                 that tells us the "answer" from the tool.
+        :return: The content of the BaseMessage that tells us the "answer"
+                 from the tool. Can be None if the tool produced no message.
```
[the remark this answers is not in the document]
Noravee: Fixed in c80ba724 — the `:return:` docstring now covers all three paths: the message content (a string today, though langchain also allows content-block lists), `None` when the Run carries no tool message, and the string form of the exception on failure.

Noravee:
**TLDR:** one behavioral line changed — `_arun()` now returns `the_answer.content` instead of the `AgentToolResultMessage` object, so the calling LLM sees the sub-agent's answer instead of the message's pydantic repr. Everything else is a `None` guard that preserves existing behavior, a completed docstring, and three unit tests locking the contract (content-not-object, `None` passthrough, exception-string path). Inline notes on the key lines below.

neuro_san/internals/run_context/langchain/core/langchain_openai_function_tool.py
```diff
@@ -246,4 +250,12 @@ async def _arun(
...
-        return the_answer
+        if the_answer is None:
+            return None
+
+        # Return the message's content rather than the message object itself.
+        # Langchain passes str (and content-block list) tool returns through
+        # to the calling LLM's ToolMessage as-is, but falls back to str() for
+        # any other object - which for a BaseMessage is its pydantic repr
+        # ("content='...' additional_kwargs={} ..."), not the answer.
+        return the_answer.content
```
Noravee, line 261: **This is the one behavioral change in the PR.** Langchain packs a tool's return value into the `ToolMessage` the calling LLM sees: `str` and `list` pass through as-is, anything else falls back to `str()` — so returning the message object leaked its pydantic repr into the parent agent's prompt on every sub-agent call.

Returning `.content` (the carried payload) rather than `.content_blocks` (a re-parsing *view* with translator side effects) is deliberate: today this is a plain string, and when multimodal tool answers arrive (#1222 phases 4+), a content-block list stored in `.content` passes through langchain unchanged — this line should never need to change again.

neuro_san/internals/run_context/langchain/core/langchain_openai_function_tool.py
```diff
@@ -246,4 +250,12 @@ async def _arun(
 
         # Assume the answer is latest tool message in the Run.
         the_answer: BaseMessage = run.get_tool_message()
-        return the_answer
+        if the_answer is None:
```
Noravee, line 253: **Preserves the pre-existing `None` behavior exactly.** A `Run` can legitimately carry no tool message (see `submit_tool_outputs()` when there are no parseable tool outputs). Before this PR, `_arun()` returned that `None` as-is; the guard keeps the same return value while preventing the new `.content` dereference from raising `AttributeError`.

tests/neuro_san/internals/run_context/langchain/core/test_langchain_openai_function_tool.py
```diff
@@ -0,0 +1,100 @@
...
+        the_message = AgentToolResultMessage(content="the answer",
+                                             tool_result_origin=[{"tool": "test_tool",
+                                                                  "instantiation_index": 0}])
+        run = LangChainRun("tool_base", [], tool_message=the_message)
+        tool = self.make_tool(run_to_return=run)
+
+        result = await tool._arun()   # pylint: disable=protected-access
+
+        assert result == "the answer"
+        assert isinstance(result, str)
```
Noravee, line 74: **The regression test for the repr leak.** Asserts the calling LLM receives the sub-agent's answer as plain content — not an `AgentToolResultMessage` object, whose `str()` form (`content='...' additional_kwargs={} ... tool_result_origin=[...]`) langchain would otherwise substitute as the tool output.

tests/neuro_san/internals/run_context/langchain/core/test_langchain_openai_function_tool.py
```diff
@@ -0,0 +1,100 @@
...
+
+    Whatever _arun() returns becomes the ToolMessage content the calling LLM
+    sees: langchain passes str and content-block list returns through as-is,
+    but falls back to str() for any other object.  Returning a BaseMessage
+    would therefore expose its pydantic repr ("content='...'
+    additional_kwargs={} ...") to the calling LLM instead of the answer.
+    """
+
+    @staticmethod
+    def make_tool(run_to_return: LangChainRun = None,
```
Noravee, line 41: **`model_construct()` bypasses pydantic field validation** so the tool can be wired to a `MagicMock` ToolCaller without satisfying the full constructor contract — same pattern as `test_run_context_runnable.py`. The `AsyncMock` stands in for `make_tool_function_calls()`, the only ToolCaller method `_arun()` touches.

d1donlydfink (approved):
Nice find.
This one is very old.

Issue #1230, opened 2026-08-11
create_function_tool mutates the shared registry spec of a same-server external agent


Noravee:
## Summary

`BaseToolFactory.create_function_tool()` writes into the function spec dictionary it is handed:

```python
function_json["name"] = name
```

For a same-server (direct-session) external agent reference, that dictionary is the **live registry spec** of the referenced network, shared by every session — so creating the tool mutates another network's registry entry.

## Trace

1. Network A references network B as `/B`, both served by the same process.
2. `BaseToolFactory.create_external_tool()` → `ExternalToolAdapter.get_function_json()` → `AsyncDirectAgentSession.function()` returns `spec.get("function", {})` — B's registry `function` dict **by reference**, no copy anywhere along the chain (`AgentNetwork.get_agent_tool_spec()` returns the `agent_spec_map` entry directly).
3. `create_function_tool()` sets `function_json["name"] = "/B"` on that shared dict.

Result: B's registry spec now carries a `"name"` key that was never in its hocon, holding the *caller's* reference string. Subsequent `function()` responses for B include it, and two callers using different reference forms (`/B` vs a host-qualified URL) overwrite each other on the same key.

## Impact

- Cross-request mutation of shared registry state; caller-dependent data visible to every other session of the referenced network.
- Latent race under concurrent serving (two callers writing different values between set and use).
- Nothing crashes today only because every current consumer of `"name"` overwrites it before reading — but anything that serializes or inspects the registry spec (connectivity reporting, future spec caching) observes the contamination.

## Where it came from

Pre-existing: the mutation dates back to an old commit (`a875359f`), well before the fix for #1228. Noted during review of that fix: `ensure_external_parameters()` now copies the dict on its substitution branch (shielding the registry in the parameterless case) while the pass-through branch does not, so the two paths currently have inconsistent aliasing behavior.

## Suggested fix

Copy at the point of mutation, which protects every path through it (internal, external, and toolbox tools all funnel here):

```python
function_json = dict(function_json)
function_json["name"] = name
```

A shallow copy suffices — only the top-level `"name"` key is written. A regression test can assert that the dict passed to `create_function_tool()` is not mutated after tool creation.

arnabnandy7:
@Noravee please review #1290

PR #1231, opened 2026-08-11
Synthesize default parameters for parameterless external front-men

### LINKED ISSUE #1228: [BUG] An external subnetwork whose front-man declares no `parameters` silently receives `{}` instead of the user's query
## Summary

When agent network A references agent network B as an external tool
(`"tools": ["/some_dir/network_b"]`), and B's front-man `function` block declares
no `parameters`, B is invoked with an empty argument object. It never receives
the user's question.

Nothing in the call fails — there is no error, no warning, and no log entry that
distinguishes this from a normal invocation. The subnetwork simply answers as if
it had been handed an empty JSON object, and the orchestrator relays or
rationalizes that non-answer. The only symptom visible to a user is a
plausible-looking but useless response.

## Environment

| | |
| --- | --- |
| `neuro-san` | 0.6.91 |
| `neuro-san-studio` | 0.3.17 |
| `nsflow` | 0.6.17 |
| `langchain` | 1.3.14 |
| Model | `openai/gpt-5.4-mini` (fallback `claude-sonnet`) |
| OS | macOS (darwin 25.5.0), Python 3.13 |

## Reproduction

Repository: <https://github.com/kaushik-cognizant/neuro-san-bug>

```bash
ns run
```

The repo contains two orchestrators and two Music Nerd subnetworks. Each
orchestrator/subnetwork pair is identical apart from the subnetwork's
`function.parameters` block:

| | Working path | Failing path |
| --- | --- | --- |
| Orchestrator | [music_orchestrator_a.hocon](registries/music_networks/music_orchestrator_a.hocon) | [music_orchestrator_b.hocon](registries/music_networks/music_orchestrator_b.hocon) |
| Subnetwork | [music_nerd_with_params.hocon](registries/music_networks/music_nerd_with_params.hocon) | [music_nerd_no_params.hocon](registries/music_networks/music_nerd_no_params.hocon) |
| Front-man `parameters` | `inquiry` (string, required) | *absent* |

Both orchestrators use the same instructions, both subnetworks use the same
`description` and `instructions`, and both share
[config/llm_config.hocon](config/llm_config.hocon). The declared parameters are
the only difference.

### Steps

1. `ns run` from the repository root.
2. Ask `music_orchestrator_a` → `MusicOrchestrator`
    → `music_nerd_with_params`: 
   `Which band wrote Yellow Submarine?` 
   — it answers correctly.
3. Ask `music_orchestrator_b` → `MusicOrchestrator`
  → `music_nerd_no_params`: `Which band wrote Yellow Submarine?`
  - Tool error: I couldn’t retrieve the music data needed to answer. Please try again.
4. Compare `logs/thinking_dir/`.

## Actual result

Orchestrator B calls the subnetwork with no arguments:

```json
{
    "invoked_agent_name": "__music_networks__music_nerd_no_params",
    "invoking_start": true,
    "params": {}
}
```

`MusicOrchestrator.__music_networks__music_nerd_no_params` logs:

```json
{
    "tool_args": {},
    "tool_start": true
}
```

and returns:

> It looks like you sent an empty JSON object: `{}`
>
> If you want, I can help with anything music-related—song meanings, band
> histories, album recommendations, hidden gems, live recordings, or who played
> on what.

This reply is the tell: the subnetwork's LLM is describing the human message it
was actually handed. The orchestrator, whose instructions say to report failures,
then answers the user:

> Tool error: I couldn't retrieve the music data needed to answer. Please try
> again.

An orchestrator without that defensive instruction would instead pass the
subnetwork's "you sent an empty JSON object" text straight through to the user,
or hallucinate over it.

### Working comparison

Orchestrator A, calling the parameterized subnetwork:

```json
{
    "tool_args": {
        "inquiry": "Which band wrote Yellow Submarine?"
    },
    "tool_start": true
}
```

> "Yellow Submarine" was written by **Lennon–McCartney** and recorded by **The
> Beatles**. …

Identical prompts on both sides, so the failure is isolated to the external
invocation path, not to the Music Nerd instructions or the orchestrator wording.

## Expected result

Either of these would resolve the problem:

1. **Fail loudly.** If a front-man reachable as an external tool is required to
   declare parameters, that should be validated at network-load time with a
   clear, actionable error naming the offending network — not left to produce
   garbage answers at runtime.
2. **Forward the query.** When the argument object resolves to empty, the
   subnetwork should still receive the user's question.

Behavior for subnetworks that already declare parameters must be unchanged.

## Impact

- Affects any external subnetwork whose front-man omits `function.parameters`,
  on every invocation.
- The failure is silent. The call succeeds, the thinking logs show a normal
  `tool_start` / `tool_end` pair, and the only evidence is the subnetwork's own
  confused prose — which is easy to mistake for a prompt-quality problem.
- Several sample networks shipped with neuro-san define parameterless front-men
  and carry a comment stating that the *absence* of parameters is "the primary
  way to identify this tool as a front-man." Copying that pattern and then
  referencing the network from another network reproduces this bug directly,
  making it an easy trap for new users.

## Expected Behaviour

- Referencing a parameterless external subnetwork either produces a clear error
  at load time, or the subnetwork receives the user's query.

Noravee:
## Summary

Fixes #1228.

When agent network A references network B as an external tool and B's front-man declares no `function.parameters`, B was presented to A's LLM as a zero-argument tool. The LLM invoked it with `{}`, and since the tool-call arguments are the only message channel into an external network, B never received the caller's request — silently producing plausible-looking garbage answers.

Root cause: the calling side builds the external tool directly from the spec that `function()` reports, and an empty parameter schema is faithfully presented as a zero-argument tool. The load-time guard that used to reject parameterless externals was relaxed long ago to support internal no-argument coded tools, which legalized this case by side effect.

## What this does

- **Forward the query.** When an external agent's function spec declares no parameters, `BaseToolFactory` synthesizes a single required `inquiry` string parameter so the calling LLM passes its request through — the same shape as an explicitly declared single-parameter front-man. External agents that declare parameters pass through byte-identical; internal parameterless tools (no-argument coded tools, toolbox tools) are untouched, including the explicit empty args_schema they need for Gemini, now locked in by a regression test.
- **Not silently.** The substitution is reported to the journal and server logs, once per process per agent. Observing the agent healthy re-arms the warning, so a hocon that is fixed via hot-reload and later regresses warns anew.
- **Only when nothing was declared.** A `parameters` block carrying anything beyond an empty `properties` declaration (e.g. `additionalProperties`, `anyOf`, `$ref`) is a declared schema in an unsupported dialect: it is rejected loudly rather than silently replaced. Only the `object` + `properties` form is supported, now stated in the docs.
- **Honest error reporting.** `create_external_tool` now distinguishes a transport failure ("was unreachable") from a spec that cannot be made into a tool ("reported an invalid function definition"), and the synthesis message is never journaled for a spec that will fail validation anyway. A malformed non-dict `parameters` is reported as invalid instead of crashing the calling agent's request with an `AttributeError`.
- **Docs and examples.** `agent_hocon_reference.md` explains the external-parameter contract and the synthesized fallback; the stale "no parameters is the primary way to identify a front-man" comment (front-man detection has been positional for a while) is corrected in all 21 example registries; the unreachable-agent hint no longer claims parameters are mandatory.

## Testing

New unit tests cover: synthesis for absent and empty-properties parameters, pass-through for declared parameters, once-per-process warning with hot-reload re-arm, unsupported-dialect rejection, non-dict parameters, no-description specs, unreachable vs invalid-definition reporting, and the empty-args_schema regression guard for internal tools. Full run_context/graph suites pass; flake8 and pylint clean.

## Related

During review, a pre-existing shared-state bug was found adjacent to this path (same-server external references mutate the referenced network's registry spec via `function_json["name"] = name`). Tracked separately in #1230.

### DIFF
--- a/docs/agent_hocon_reference.md
+++ b/docs/agent_hocon_reference.md
@@ -534,8 +534,17 @@ For a user-facing front-man, what is contained in this description often suffice
 Parameters contains an optional [JSON Schema](https://json-schema.org) dictionary describing what
 specific information the agent needs as input arguments when it is called.
 
-A front-man typically does not need parameters defined, unless the agent network being described
-is anticipated as being called from other agent networks.
+A front-man typically does not need parameters defined when its agent network is only
+ever talked to directly by a user.
+
+When the agent network is anticipated as being called from other agent networks
+(that is, referenced as an external tool), the front-man's parameters are how a
+calling agent passes its request along — the tool-call arguments are the only channel
+through which anything reaches the external network. Declare at least one parameter
+(for example, a single required `inquiry` string) for any network intended to be
+referenced externally. If no parameters are declared, neuro-san synthesizes a default
+required `inquiry` string parameter on the calling side (and logs a warning) so the
+caller's request still gets through.
 
 ##### type
 
--- a/docs/agent_hocon_reference.md
+++ b/docs/agent_hocon_reference.md
@@ -533,6 +533,9 @@ For a user-facing front-man, what is contained in this description often suffice
 
 Parameters contains an optional [JSON Schema](https://json-schema.org) dictionary describing what
 specific information the agent needs as input arguments when it is called.
+Only the `type: object` + `properties` form described below is supported.
+A schema expressed with other JSON Schema constructs (such as `additionalProperties`,
+`anyOf`, or `$ref`) is rejected as invalid when the agent is called as a tool.
 
 A front-man typically does not need parameters defined when its agent network is only
 ever talked to directly by a user.
--- a/docs/agent_hocon_reference.md
+++ b/docs/agent_hocon_reference.md
@@ -534,8 +534,10 @@ For a user-facing front-man, what is contained in this description often suffice
 Parameters contains an optional [JSON Schema](https://json-schema.org) dictionary describing what
 specific information the agent needs as input arguments when it is called.
 Only the `type: object` + `properties` form described below is supported.
-A schema expressed with other JSON Schema constructs (such as `additionalProperties`,
-`anyOf`, or `$ref`) is rejected as invalid when the agent is called as a tool.
+JSON Schema constructs outside this form (such as `additionalProperties`,
+`anyOf`, or `$ref`) are not honored: a schema without a `properties` dictionary
+is rejected as invalid when the agent is called as a tool, and unsupported
+constructs appearing alongside a `properties` dictionary are ignored.
 
 A front-man typically does not need parameters defined when its agent network is only
 ever talked to directly by a user.
--- a/docs/agent_hocon_reference.md
+++ b/docs/agent_hocon_reference.md
@@ -542,8 +542,10 @@ ever talked to directly by a user.
 
 When the agent network is anticipated as being called from other agent networks
 (that is, referenced as an external tool), the front-man's parameters are how a
-calling agent passes its request along — the tool-call arguments are the only channel
-through which anything reaches the external network. Declare at least one parameter
+calling agent passes its request along — the tool-call arguments are the only message
+channel through which a caller's request reaches the external network.
+(Private data can separately flow through the opt-in `sly_data` channel —
+see [to_downstream](#to_downstream).) Declare at least one parameter
 (for example, a single required `inquiry` string) for any network intended to be
 referenced externally. If no parameters are declared, neuro-san synthesizes a default
 required `inquiry` string parameter on the calling side (and logs a warning) so the
--- a/neuro_san/internals/run_context/langchain/core/base_tool_factory.py
+++ b/neuro_san/internals/run_context/langchain/core/base_tool_factory.py
@@ -194,13 +194,27 @@ class BaseToolFactory:
             return function_json
 
         parameters: Dict[str, Any] = function_json.get("parameters") or {}
+        if not isinstance(parameters, Dict):
+            # Not a schema we can reason about.
+            # Let verify_function_json() report it as invalid.
+            return function_json
+
         properties: Dict[str, Any] = parameters.get("properties") or {}
         if properties:
             return function_json
 
+        # A parameters block carrying anything beyond an empty properties
+        # declaration is a declared schema in an unsupported dialect
+        # (e.g. additionalProperties, anyOf, $ref). Let verify_function_json()
+        # report it rather than silently replacing the declared contract
+        # with the synthesized one.
+        if parameters and not {"type", "properties", "required"}.issuperset(parameters.keys()):
+            return function_json
+
         message: str = (
-            f"The front-man of external agent {name} declares no function.parameters, "
-            f"so a single required '{self.DEFAULT_EXTERNAL_PARAMETER_NAME}' string parameter "
+            f"The front-man of external agent {name} declares no parameters "
+            f"in its function definition, so a single required "
+            f"'{self.DEFAULT_EXTERNAL_PARAMETER_NAME}' string parameter "
             "is being synthesized for it to receive the calling agent's request. "
             "To control what this agent network receives, declare at least one parameter "
             "in the function definition of its front-man."
--- a/neuro_san/internals/run_context/langchain/core/base_tool_factory.py
+++ b/neuro_san/internals/run_context/langchain/core/base_tool_factory.py
@@ -20,6 +20,7 @@ from typing import List
 from typing import Set
 from typing import Union
 
+from copy import deepcopy
 from logging import Logger
 from logging import getLogger
 
@@ -45,6 +46,25 @@ class BaseToolFactory:
     """
     Creates langchain BaseTools.
     """
+
+    # The tool-call arguments are the only channel through which a calling
+    # agent passes anything to an external agent network.  A front-man that
+    # declares no function.parameters would be presented to the calling LLM
+    # as a zero-argument tool, which the LLM would then invoke with {} -
+    # the external network would never receive the caller's request.
+    # When that happens we substitute this schema so the request still
+    # gets through.
+    DEFAULT_EXTERNAL_PARAMETER_NAME: str = "inquiry"
+    DEFAULT_EXTERNAL_PARAMETERS: Dict[str, Any] = {
+        "type": "object",
+        "properties": {
+            DEFAULT_EXTERNAL_PARAMETER_NAME: {
+                "type": "string",
+                "description": "The request to send to this agent network."
+            }
+        },
+        "required": [DEFAULT_EXTERNAL_PARAMETER_NAME]
+    }
 
     # pylint: disable=too-many-arguments, too-many-positional-arguments
     def __init__(self,
@@ -113,6 +133,7 @@ class BaseToolFactory:
         adapter = ExternalToolAdapter(session_factory, name)
         try:
             function_json: Dict[str, Any] = await adapter.get_function_json(self.invocation_context)
+            function_json = await self.ensure_external_parameters(function_json, name)
             return self.create_function_tool(function_json, name)
         except ValueError as exception:
             # Could not reach the server for the external agent, so tell about it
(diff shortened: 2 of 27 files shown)


neuro_san/internals/run_context/langchain/core/base_tool_factory.py
```diff
@@ -113,15 +136,107 @@ async def create_external_tool(self, name: Union[str, Dict[str, Any]]) -> Union[
...
+                    Can be None when the agent was unreachable.
+        :param name: The name of the external agent, for reporting.
+        :return: The function_json as-is when it already declares parameters,
+                    otherwise a copy with DEFAULT_EXTERNAL_PARAMETERS substituted in.
+        """
+        if function_json is None:
+            # Unreachable external agent. create_function_tool() reports this case.
+            return None
+
+        if function_json.get("description") is None:
```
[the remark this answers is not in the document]
Noravee: Half of this was a real catch — fixed in d59eaf54; the other half is intentional. Taking the two parts separately:

**Falsy non-dict values: correct, fixed.** The `or {}` coalescing ran before the isinstance guard, so only truthy malformed values (like a string) were routed to validation, while `[]`, `""`, `false`, and `0` were silently repaired with the synthesized default. The guard now checks the raw value before coalescing — matching the equivalent guard in `verify_function_json` — so any non-dict `parameters` is reported as an invalid function definition. The test is parametrized over all five shapes.

**Schemas with a non-empty `properties` plus unsupported constructs: intentional pass-through.** `ensure_external_parameters` deliberately answers only one question — "did the author declare anything?" — and leaves all validity judgment to `verify_function_json`, which has always been the single owner of that rule. Duplicating shape validation here would create two independent definitions of "valid schema" that can drift apart. A schema like `{"type": "array", "properties": {...}}` or one carrying `additionalProperties` alongside `properties` passed validation the same way before this PR (only the presence of `type` is checked, and extra keywords are ignored by the converter), so this PR preserves that behavior unchanged. Tightening it would be a change to `verify_function_json` affecting internal tools too, and belongs in its own PR. The docs sentence did overclaim on this point, though — it now says unsupported constructs alongside `properties` are ignored rather than rejected, which matches the actual (pre-existing) behavior.

neuro_san/internals/run_context/langchain/core/base_tool_factory.py
```diff
@@ -46,6 +47,28 @@ class BaseToolFactory:
     Creates langchain BaseTools.
     """
 
+    # Parameters substituted for an external agent whose front-man declares
+    # none of its own. See ensure_external_parameters() for the full rationale.
+    DEFAULT_EXTERNAL_PARAMETER_NAME: str = "inquiry"
```
Noravee, line 52: The synthesized fallback schema. `inquiry` matches the parameter name used by the sample networks, so a subnetwork receives `{"inquiry": "..."}` — byte-for-byte the shape of the already-working declared-parameter path (e.g. `music_nerd_with_params` in the issue repro).

neuro_san/internals/run_context/langchain/core/base_tool_factory.py
```diff
@@ -46,6 +47,28 @@ class BaseToolFactory:
...
+        "required": [DEFAULT_EXTERNAL_PARAMETER_NAME]
+    }
+
+    # Class-level because BaseToolFactory instances are per-request: remembers
+    # which external agents this process has already warned about synthesizing
+    # parameters for, so the warning is not repeated on every request.
+    # An agent later observed with declared parameters is removed again, so a
+    # network that is fixed and then regresses warns anew - hocon files can be
+    # edited and hot-reloaded without a server restart.
+    synthesis_warned: Set[str] = set()
```
Noravee, line 70: Class-level on purpose: `BaseToolFactory` is rebuilt on every request (tool resources are torn down per turn), so instance state can't remember what was already warned. Without this, one misconfigured network emits an identical journal message + log warning per calling agent per turn — thousands of duplicate lines under load for a static config condition.

neuro_san/internals/run_context/langchain/core/base_tool_factory.py
```diff
@@ -113,15 +136,109 @@ async def create_external_tool(self, name: Union[str, Dict[str, Any]]) -> Union[
...
             message += str(exception)
+            await self.report_tool_exclusion(message)
+            return None
+
+        try:
+            function_json = await self.ensure_external_parameters(function_json, name)
+            return self.create_function_tool(function_json, name)
+        except ValueError as exception:
+            # The agent was reachable, but what it reported cannot be made into a tool.
+            message: str = f"Agent/tool {name} reported an invalid function definition. " + \
```
Noravee, line 151: New failure class: previously every `ValueError` here — including schema problems — was reported as "was unreachable", sending users to debug connectivity for what is actually a hocon authoring error. The fetch (genuinely unreachable) and validate/create (invalid definition) steps now have separate handlers.

neuro_san/internals/run_context/langchain/core/base_tool_factory.py
```diff
@@ -113,15 +136,109 @@ async def create_external_tool(self, name: Union[str, Dict[str, Any]]) -> Union[
...
+                    Can be None when the agent was unreachable.
+        :param name: The name of the external agent, for reporting.
+        :return: The function_json as-is when it already declares parameters,
+                    otherwise a copy with DEFAULT_EXTERNAL_PARAMETERS substituted in.
+        """
+        if function_json is None:
+            # Unreachable external agent. create_function_tool() reports this case.
+            return None
+
+        if function_json.get("description") is None:
```
Noravee, line 194: Ordering matters here: a spec with no description fails `verify_function_json()` regardless of parameters, so synthesizing first would journal a promise that the request will get through, immediately followed by the tool being dropped. Gating on description keeps the synthesis message truthful — it only ever appears for a tool that is actually created.

neuro_san/internals/run_context/langchain/core/base_tool_factory.py
```diff
@@ -113,15 +136,109 @@ async def create_external_tool(self, name: Union[str, Dict[str, Any]]) -> Union[
...
+            # A spec with no description fails verify_function_json() no matter
+            # what parameters it has. Leave it alone so that validation reports
+            # the real problem, instead of journaling a promise here that a
+            # synthesized parameter will get the request through, immediately
+            # followed by the tool being dropped.
+            return function_json
+
+        raw_parameters: Any = function_json.get("parameters")
```
Noravee, line 202: Checked on the raw value, before any `or {}` coalescing, so falsy malformed values (`[]`, `""`, `false`, `0`) are routed to validation the same as truthy ones — mirroring the guard in `verify_function_json`. This method never judges validity itself; it only answers "did the author declare anything?" and leaves validity to `verify_function_json`, the single owner of that rule.

neuro_san/internals/run_context/langchain/core/base_tool_factory.py
```diff
@@ -113,15 +136,109 @@ async def create_external_tool(self, name: Union[str, Dict[str, Any]]) -> Union[
...
+            # warning in case the network regresses later.
+            BaseToolFactory.synthesis_warned.discard(name)
+            return function_json
+
+        # A parameters block carrying anything beyond an empty properties
+        # declaration is a declared schema in an unsupported dialect
+        # (e.g. additionalProperties, anyOf, $ref). Let verify_function_json()
+        # report it rather than silently replacing the declared contract
+        # with the synthesized one.
+        if parameters and not {"type", "properties", "required"}.issuperset(parameters.keys()):
```
Noravee, line 222: The line that keeps synthesis honest: a `parameters` block whose keys go beyond `{type, properties, required}` is a *declared* schema in an unsupported dialect, not an absent one. It falls through to validation and is rejected loudly, rather than having the author's contract silently replaced with the synthesized one. Unknown future keywords fail safe (loud) by construction.

neuro_san/internals/run_context/langchain/core/base_tool_factory.py
```diff
@@ -113,15 +136,109 @@ async def create_external_tool(self, name: Union[str, Dict[str, Any]]) -> Union[
...
+            # (e.g. a string, a list, a boolean).
+            # Let verify_function_json() report it as invalid.
+            return function_json
+
+        parameters: Dict[str, Any] = raw_parameters or {}
+        properties: Dict[str, Any] = parameters.get("properties") or {}
+        if properties:
+            # The network declares its own parameters. Re-arm the synthesis
+            # warning in case the network regresses later.
+            BaseToolFactory.synthesis_warned.discard(name)
```
Noravee, line 214: The re-arm half of the warning dedup: observing the agent healthy removes it from the warned set. Hocon files hot-reload without a server restart, so without this, a network that is fixed and later regresses back to parameterless would stay silent until the next restart — recreating a small version of the original silent failure.

neuro_san/internals/run_context/langchain/core/langchain_openai_function_tool.py
```diff
@@ -101,7 +101,9 @@ def verify_function_json(function_json: Dict[str, Any]):
             message = f"Function for {name} has no description.\n"
 
         parameters: Dict[str, Any] = function_json.get("parameters")
-        if parameters:
+        if parameters is not None and not isinstance(parameters, Dict):
+            message = f"Function for {name} needs its parameters to be a dictionary.\n"
```
Noravee, line 105: Closes a long-standing crash: a non-dict `parameters` used to raise `AttributeError` at `parameters.get("type")`, which escaped the `except ValueError` upstream and aborted the calling agent's entire resource setup. Raising `ValueError` routes it into the invalid-definition report like any other schema problem.

docs/agent_hocon_reference.md
```diff
@@ -533,9 +533,25 @@ For a user-facing front-man, what is contained in this description often suffice
 
 Parameters contains an optional [JSON Schema](https://json-schema.org) dictionary describing what
 specific information the agent needs as input arguments when it is called.
-
-A front-man typically does not need parameters defined, unless the agent network being described
-is anticipated as being called from other agent networks.
+Only the `type: object` + `properties` form described below is supported.
```
Noravee, line 536: States the boundary the code actually enforces: no `properties` → rejected as invalid; unsupported constructs alongside `properties` → ignored (pre-existing converter behavior, unchanged by this PR).

Noravee (commented):
## TL;DR

**Problem (#1228):** referencing an external network whose front-man declares no `function.parameters` presented it to the calling LLM as a zero-argument tool. The LLM called it with `{}`, the subnetwork never saw the user's question, and nothing failed loudly — the only symptom was a plausible-but-useless answer.

**Fix:** when (and only when) an external front-man genuinely declares nothing, the calling side synthesizes a single required `inquiry` string parameter so the request flows through — the same shape as the already-working declared-parameter path. Everything else fails loudly and accurately:

- Declared parameters pass through byte-identical; internal parameterless tools (no-arg coded tools) are untouched, including the empty `args_schema` Gemini needs (regression-tested).
- Declared-but-unsupported schemas (`additionalProperties`, `anyOf`, `$ref`, non-dict values) are rejected under a new, accurate "invalid function definition" report instead of being mislabeled "unreachable" or silently replaced.
- The synthesis is announced in the journal and server logs once per process per agent, re-arming if the network's hocon is fixed via hot-reload and later regresses.
- Docs and all 21 example registries drop the outdated "no parameters identifies the front-man" guidance (front-man detection is positional).

d1donlydfink (approved):

PR #1234, opened 2026-08-11
Use latest leaf-common==1.2.53 and use logging imports from there

d1donlydfink:
Use LoggingSetup from leaf_common which now has the setup_logging() method on it.

This now means that the only places including leaf_server_common are under neuro_san.service,
which will enable a cleaving of service packages into a separate repo.

### DIFF
--- a/neuro_san/service/generic/agent_server_logging.py
+++ b/neuro_san/service/generic/agent_server_logging.py
@@ -22,7 +22,7 @@ from typing import List
 import logging
 import os
 
-from leaf_server_common.logging.logging_setup import setup_logging
+from leaf_common.logging.logging_setup import LoggingSetup
 from leaf_server_common.server.grpc_metadata_forwarder import GrpcMetadataForwarder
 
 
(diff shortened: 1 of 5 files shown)


neuro_san/service/generic/agent_server_logging.py
```diff
@@ -22,7 +22,7 @@
 import logging
 import os
 
-from leaf_server_common.logging.logging_setup import setup_logging
+from leaf_common.logging.logging_setup import LoggingSetup
```
d1donlydfink, line 25: Use LoggingSetup from leaf_common, which now has the setup_logging() method on it.

requirements.txt
```diff
@@ -6,7 +6,7 @@
 # requirements-build.txt
 
 # In-house dependencies
-leaf-common>=1.2.52
+leaf-common>=1.2.53
```
d1donlydfink, line 9: Use latest leaf-common.

andreidenissov-cog (approved):
Looks good.

neuro_san/session/session_invocation_context.py
```diff
@@ -32,7 +32,7 @@
 
 from leaf_common.asyncio.asyncio_executor import AsyncioExecutor
 from leaf_common.asyncio.asyncio_executor_pool import AsyncioExecutorPool
-from leaf_server_common.logging.logging_setup import setup_extra_logging_fields
+from leaf_common.logging.logging_setup import LoggingSetup
```
d1donlydfink, line 35: This is the file that made this worth doing

PR #1242, opened 2026-08-13
Remove need for leaf-server-common

d1donlydfink:
Remove unused files and methods that used leaf-server-common.


neuro_san/service/generic/agent_server_logging.py
```diff
@@ -48,13 +47,6 @@ def __init__(self, server_name_for_logs: str,
         self.forwarded_request_metadata: List[str] = forwarded_request_metadata_str.split(" ")
         self.logging_config: Dict[str, Any] = logging_config
 
-    def get_forwarder(self) -> GrpcMetadataForwarder:
-        """
-        :return: A GrpcMetadataForwarder instance initialized with
-                 the list of forwarded request metadata keys
-        """
-        return GrpcMetadataForwarder(self.forwarded_request_metadata)
```
d1donlydfink, line 56: No one calls this method.

neuro_san/service/generic/agent_service.py
d1donlydfink, line 1: This file is no longer used.

neuro_san/service/generic/agent_service_provider.py
d1donlydfink, line 1: This file is no longer used.

requirements.txt
```diff
@@ -7,7 +7,6 @@
 
 # In-house dependencies
 leaf-common>=1.2.53
-leaf-server-common>=0.1.25
```
d1donlydfink, line 10: Remove need for leaf-server-common as a dependency

andreidenissov-cog (dismissed):
Finally :-)
LGTM.

Noravee (dismissed):
LGTM — the deleted `AgentService`/`AgentServiceProvider` have no remaining in-repo references, no `.py` file imports `leaf_server_common` anymore, and the packages it pulled in transitively (grpcio extras, opentelemetry) aren't needed: `grpcio` still comes via `leaf-common`, and the one `opentelemetry` import in `langfuse_tracing_context.py` is guarded and only exercised when langfuse is installed.

One minor concern (non-blocking, doesn't surface in CI): the Python 3.14t build files still smoke-test the import of `leaf_server_common`, so they'll fail once this merges:

- `neuro_san/deploy/Dockerfile.py314t` (~line 187): `python -c "import leaf_common, leaf_server_common; ..."` will fail the stage-1 image build.
- `build_scripts/make_venv_py314t.sh` (~lines 206-208): the venv verification heredoc imports `leaf_server_common` and will always exit with an error.

There are also stale comments mentioning leaf-server-common in those two files plus `neuro_san/deploy/build_py314t.sh`. Fine to fix here or in a small follow-up.

d1donlydfink:
Good catches @Noravee : I didn't grep in those places.

PR #1243, opened 2026-08-13
Enforce tool selection at execution time in LlmConfigToolSelectorMiddleware

### LINKED ISSUE #1239: Enforce tool selection at execution time in LlmConfigToolSelectorMiddleware
Part of #1237 (GHSA-3xg4-wfwr-gc2g, Finding 1).

## Problem

`LlmConfigToolSelectorMiddleware` (subclass of LangChain's `LLMToolSelectorMiddleware`) narrows only the tool list *advertised* to the model per call, via `request.override(tools=[subset])` in `wrap_model_call`. The agent executor is built with the **full** tool list (`create_agent(tools=self.tools, ...)` in `neuro_san/internals/run_context/langchain/core/langchain_run_context.py`), and the tool-execution node validates tool calls against that full registry — not against the narrowed advertisement.

Consequently, a tool call for a de-selected tool still executes if it reaches the executor. Realistic vectors:

- The model emits a tool name it saw advertised in an earlier turn (tool names persist in chat history across turns even when the current turn's selection dropped them).
- Replayed/checkpointed state containing pending tool calls (we pass a `checkpointer` to `create_agent`).
- Co-resident middleware mutating messages.

Tool selection is therefore advertise-only and is not an execution-time boundary.

## Fix plan

1. **Post-filter the model response** in `LlmConfigToolSelectorMiddleware.wrap_model_call` / `awrap_model_call`: after the handler returns, check the AIMessage's `tool_calls` against the selected set ∪ `always_include`. Any tool call for a non-advertised tool is not executed — replace it with an error `ToolMessage` ("tool not available for this request") so the agent loop lets the model retry. This is race-free because selection and response live in the same call.
2. **Add a `wrap_tool_call` backstop** in the same middleware: record the current selection into agent state (custom state key via the middleware's `state_schema`), and reject any tool call whose name is not in the recorded selection. This covers replayed checkpointed state and message-injecting middleware, which post-filtering alone does not.
3. **Documentation**: state explicitly (class docstring + user docs) that LLM-based tool selection is a token/latency optimization, not access control. The set of tools configured in the agent network is the actual capability boundary; genuinely sensitive tools must not be wired into networks exposed to untrusted input, or must be gated by execution-time policy (e.g. HITL middleware).
4. **Tests**: a scripted (non-LLM) model that emits a tool call for a de-selected tool; assert the tool does not execute and the model receives the error ToolMessage. Cover both sync and async paths.
5. Optionally raise the desync upstream with LangChain — every user of `LLMToolSelectorMiddleware` has the same gap.

## Non-goal / caveat

Even with enforcement, the selector chooses tools based on the untrusted last user message, so an attacker can steer selection toward a sensitive tool. Enforcement closes the advertised-vs-executable desync; it does not make the selector a security policy. That is the documentation point in item 3.

Noravee:
## Summary

Fixes #1239 (GHSA-3xg4-wfwr-gc2g Finding 1): LangChain's `LLMToolSelectorMiddleware` only narrows the tool list *advertised* to the model per call, while the executor keeps every configured tool callable — so a tool call naming a de-selected tool (e.g. a name the model remembered from an earlier turn) still executed.

`LlmConfigToolSelectorMiddleware` now enforces the selection at execution time:

- `awrap_model_call()` records, per tool call id, the tool names that were advertised on the model call that produced it.
- `awrap_tool_call()` rejects any tool call whose name was not among those advertised, returning an error `ToolMessage` naming the valid tools so the model can retry.

## Design notes

- **Bookkeeping lives in sly_data, not message metadata.** Message-rewriting middleware (langchain's PIIMiddleware redaction, summarization) rebuild AIMessages without copying `response_metadata`, which would silently disable enforcement. sly_data is outside langgraph agent state, request-scoped, server-side only, and redacted from clients/external tools by default. When the `sly_data`/`origin_str` special args aren't configured, a private per-instance dict is used (namespacing prevents tool-call-id collisions between agents).
- **`unadvertised_policy` constructor arg** ("allow" default / "deny") controls what happens to tool calls with no recorded advertisement (e.g. fabricated by co-resident middleware that bypasses the model). "deny" is recommended for agents exposed to untrusted input.
- Denial handles tool calls without ids (ToolCall.id is Optional, ToolMessage.tool_call_id requires str), and a tool call's bookkeeping entry is cleaned up after successful execution.
- Docs state explicitly that LLM-based tool selection is a token/latency optimization, not a security boundary — the configured tool set remains the capability boundary.

## Scope

Enforcement covers the static tool set the agent gets from its HOCON configuration and is a feature of `LlmConfigToolSelectorMiddleware` specifically — both documented:

- Dynamic tools are unsupported: middleware listed after the selector must not modify the tool list, since the record reflects this middleware's layer.
- Other tool-selection middleware configured by class name (including langchain's upstream `LLMToolSelectorMiddleware`) only narrow what the model is shown and get no execution-time enforcement.

## Tests

18 tests including async end-to-end through `create_agent` with scripted models: the selection model picks only `safe_echo`, the main model emits a call to the de-selected `canary` tool, and the executor must not run it. A regression test wires a PIIMiddleware-style metadata-stripping middleware and asserts enforcement survives. Negative control verified separately: upstream `LLMToolSelectorMiddleware` executes the de-selected tool under the identical scenario.


neuro_san/middleware/llm_config_tool_selector_middleware.py
```diff
@@ -15,17 +15,43 @@
...
+# The value is a dictionary keyed by a per-middleware-instance namespace,
+# whose values map tool call ids to the list of tool names that were
+# advertised to the model on the call that produced them.
+#
+# This bookkeeping deliberately lives OUTSIDE of langgraph agent state:
+# message-rewriting middleware (summarization, PII redaction, etc.) can
+# rebuild AIMessages and would silently destroy any bookkeeping carried
+# on the messages themselves. sly_data is request-scoped, server-side only,
+# and redacted from clients and external tools by default.
+ADVERTISED_TOOLS_KEY: str = "neuro_san_advertised_tools"
```
Noravee, line 53: **Why sly_data and not message metadata:** the first iteration stamped advertised tools onto `AIMessage.response_metadata`, but middleware that rebuild messages (langchain's PIIMiddleware redaction constructs a new `AIMessage(content, id, name, tool_calls)` without copying `response_metadata`) would silently strip the stamp and disable enforcement exactly on redacted turns. sly_data sits outside langgraph agent state, so no middleware can touch it, and `SlyDataRedactor` is default-deny toward clients and external tools.

neuro_san/middleware/llm_config_tool_selector_middleware.py
```diff
@@ -92,3 +174,191 @@ def __init__(
...
+                        is used to avoid tool call id collisions between agents.
+        :param unadvertised_policy: "allow" or "deny" for tool calls with no
+                        recorded advertisement.
+        """
+        if unadvertised_policy not in ("allow", "deny"):
+            raise ValueError(f"unadvertised_policy must be 'allow' or 'deny', got '{unadvertised_policy}'")
+        self.unadvertised_policy: str = unadvertised_policy
+
+        holder: Dict[str, Any] = sly_data if sly_data is not None else {}
+        namespace: str = origin_str if origin_str else f"instance-{id(self)}"
```
Noravee, line 200: The namespace matters when sly_data is shared: every agent in the network shares one sly_data dict, and each agent has its own selector instance. Without namespacing, two agents whose models emit the same tool call id (scripted models often use `call_1`) would collide. `origin_str` gives a readable namespace when configured; the per-instance fallback guarantees isolation otherwise. When sly_data isn't passed at all (it's a HOCON-opt-in special arg), a private per-instance dict keeps enforcement working identically — sharing only adds observability.

neuro_san/middleware/llm_config_tool_selector_middleware.py
```diff
@@ -92,3 +174,191 @@ def __init__(
...
+        per_request: Dict[str, Dict[Optional[str], List[str]]] = holder.setdefault(ADVERTISED_TOOLS_KEY, {})
+
+        # Maps tool call id -> list of tool names advertised on the model call
+        # that produced that tool call. Keys can be None, as providers may omit
+        # tool call ids (ToolCall.id is Optional).
+        self.advertised_tools: Dict[Optional[str], List[str]] = per_request.setdefault(namespace, {})
+
+    async def awrap_model_call(
```
Noravee, line 208: Pattern note: we don't reimplement any upstream selection logic. `super().awrap_model_call()` runs the LLM-based selection and calls the handler with the narrowed request; we only interpose a handler wrapper so whatever tool list actually reached the model gets recorded against the tool calls it produced.

neuro_san/middleware/llm_config_tool_selector_middleware.py
```diff
@@ -92,3 +174,191 @@ def __init__(
...
+        result: Union[ToolMessage, Command] = await handler(request)
+
+        # The tool call is done with its bookkeeping entry, so clean it up to keep
+        # the map from growing over long requests and to avoid stale context should
+        # a tool call id ever be reused. None ids are kept, as multiple tool calls
+        # can share None. The entry is popped only after successful execution so
+        # that a retrying outer middleware still finds it.
+        call_id: Optional[str] = request.tool_call.get("id")
+        if call_id is not None:
+            self.advertised_tools.pop(call_id, None)
```
Noravee, line 276: **Deliberate asymmetry — please don't 'fix' this:** entries are popped only after *successful* execution. Denied calls keep their entry on purpose: if we popped on denial, a repeated call with the same id would fall through to the unadvertised path, which under the default `allow` policy would *execute* the tool — a security regression. Growth from denied entries is bounded by the request lifetime (sly_data resets every turn) and the agent loop's recursion limit. (Copilot suggested popping on denial in three review rounds; rejected each time for this reason.)

neuro_san/middleware/llm_config_tool_selector_middleware.py
```diff
@@ -92,3 +174,191 @@ def __init__(
...
+        """
+        advertised: List[str] = []
+        for tool in narrowed_request.tools:
+            if isinstance(tool, dict):
+                # Provider-specific tool dicts may carry the name at the top level
+                # or nested in OpenAI function format ({"function": {"name": ...}}).
+                name: Optional[str] = tool.get("name")
+                function_spec: Any = tool.get("function")
+                if name is None and isinstance(function_spec, dict):
+                    name = function_spec.get("name")
```
Noravee, line 297: Future-proofing only: every tool source in neuro-san today (internal agents, function tools, toolbox, MCP via langchain-mcp-adapters) produces `BaseTool` instances, so dict-format tools are currently unreachable. This nested lookup exists so that if provider-native dict tools (OpenAI `{"function": {"name": ...}}` shape) are ever added, they'd be recorded correctly instead of falsely denied.

neuro_san/middleware/llm_config_tool_selector_middleware.py
```diff
@@ -92,3 +174,191 @@ def __init__(
...
+        if messages is None:
+            messages = [response]
+
+        for message in messages:
+            if not isinstance(message, AIMessage):
+                continue
+            for tool_call in message.tool_calls or []:
+                self.advertised_tools[tool_call.get("id")] = advertised
+
+    def _deny_unadvertised_tool_call(self, request: ToolCallRequest) -> Optional[ToolMessage]:
```
Noravee, line 313: Two distinct paths here: (1) a call whose model call *was* recorded but whose tool wasn't in the advertised list is always denied — that's the vulnerability being fixed; (2) a call with *no record at all* (e.g. fabricated by a co-resident middleware that bypassed the model) falls to `unadvertised_policy` — `allow` by default for compatibility, `deny` recommended for untrusted-input deployments.

tests/neuro_san/middleware/test_llm_config_tool_selector_middleware.py
```diff
@@ -0,0 +1,380 @@
...
+
+    def with_structured_output(self, schema: Any, **kwargs: Any) -> RunnableLambda:
+        """
+        The middleware calls this to get structured {"tools": [...]} output.
+        """
+        _ = schema, kwargs
+        return RunnableLambda(lambda _input: {"tools": list(self.selected)})
+
+
+class MetadataStrippingMiddleware(AgentMiddleware):
```
Noravee, line 73: This mimics exactly what langchain's `PIIMiddleware.after_model` does on redaction: rebuild the AIMessage with the same id and tool_calls but without `response_metadata`, letting `add_messages` replace the original in state. It's the regression test for the design decision to keep bookkeeping in sly_data — under the earlier metadata-based design, this middleware combination silently disabled enforcement.

tests/neuro_san/middleware/test_llm_config_tool_selector_middleware.py
```diff
@@ -0,0 +1,380 @@
...
+        Assert the de-selected tool did not run and the model got the error ToolMessage.
+        """
+        self.assertEqual(self.canary_ran, [], "De-selected tool must not execute")
+
+        tool_messages = [message for message in result["messages"] if isinstance(message, ToolMessage)]
+        self.assertEqual(len(tool_messages), 1)
+        self.assertEqual(tool_messages[0].status, "error")
+        self.assertIn("was not among the tools selected", tool_messages[0].content)
+
+    async def test_deselected_tool_does_not_execute_in_agent(self):
```
Noravee, line 317: This is the advisory scenario end-to-end through real `create_agent` wiring: selection picks only `safe_echo`, the scripted main model nonetheless calls the de-selected tool, and the executor must not run it. Negative control (run separately, not in CI): the identical scenario against upstream `LLMToolSelectorMiddleware` *does* execute the de-selected tool, confirming both the vulnerability and that this subclass is what fixes it.

docs/agent_hocon_reference.md
```diff
@@ -1197,3 +1197,46 @@ A ProgressReporter instance that can be used to report progress dictionaries via
...
+tools added by such middleware are falsely rejected, and further narrowing is not enforced. If you
+combine the tool selector with other tool-modifying middleware, list those middleware before the
+tool selector, never after it.
+
+Execution-time enforcement is a feature of `LlmConfigToolSelectorMiddleware` specifically. Other
+tool-selection middleware configured by class name (including langchain's own
+`LLMToolSelectorMiddleware`) only narrow what the model is shown and provide no execution-time
+enforcement.
+
+Note that LLM-based tool selection is a token/latency optimization, **not** a security boundary.
```
Noravee: The key operator takeaway: even with execution-time enforcement, the selection LLM chooses tools based on the (untrusted) last user message, so an attacker can steer it toward any configured tool. Enforcement makes advertised == executable; it does not make the selector an access-control mechanism. Sensitive tools must not be configured on agents exposed to untrusted input.

Noravee (commented):
**TL;DR**

LangChain's tool selector only narrows what the model is *shown*; the executor keeps every configured tool callable, so a call naming a de-selected tool still ran (GHSA-3xg4-wfwr-gc2g Finding 1). This PR closes that gap in `LlmConfigToolSelectorMiddleware` with a record-then-gate design:

1. **Record** (`awrap_model_call`): after the superclass narrows the tool list, store `tool_call_id -> [advertised tool names]` for every tool call the model emits.
2. **Gate** (`awrap_tool_call`): before a tool executes, look up its call id; if its name wasn't advertised on the model call that produced it, return an error `ToolMessage` (naming the valid tools) instead of executing.

Key decisions to review:
- Bookkeeping lives in **sly_data, not message metadata** — message-rewriting middleware (e.g. langchain PIIMiddleware) rebuild AIMessages without `response_metadata` and would silently disable enforcement.
- **`unadvertised_policy`** ("allow" default / "deny") governs tool calls with no record at all.
- **Scope**: static HOCON tool set, this middleware only; dynamic-tool middleware and third-party selectors are documented as unsupported/unenforced.

Suggested reading order: the two end-to-end tests (`test_deselected_tool_does_not_execute_in_agent`, `test_enforcement_survives_message_rewriting_middleware`), then `awrap_model_call`/`awrap_tool_call`, then the inline notes below.

docs/agent_hocon_reference.md
d1donlydfink: I'm not sure this belongs here.  This is more hoc on-reference, not how-to-use-middleware.
Maybe this nice explanation belongs in the Middleware class itself.

d1donlydfink (changes requested):
I don’t think we need the documentation in the agent_hocon_reference.md.  That would be scope-creep for that documentation. What you have in the class is plenty.

docs/agent_hocon_reference.md
Noravee: The documentation here has been removed.

d1donlydfink (approved):

Issue #1244, opened 2026-08-13
Validation does not catch dangling toolbox references; ToolboxNetworkValidator is unwired and buggy


Noravee:
## Problem

`ManifestNetworkValidator` includes no toolbox validation — `manifest_network_validator.py` has a placeholder comment `# No ToolBoxNetworkValidator yet.` Both `hocon_validator_cli` and server-side manifest restoration build their validation from this composite, so an agent network whose `"toolbox"` value names a tool with no toolbox entry passes validation cleanly.

At runtime such a reference fails tool creation, but the resulting `ValueError` is caught in `BaseToolFactory.create_toolbox_tool`, journaled as a warning, and the tool is skipped — the agent starts and serves traffic without it. With the former default `requests_*` tools removed from the toolbox (#1241), stale references in existing networks become a realistic upgrade hazard, and pre-flight validation is the right place to catch them: operators who validate before deploying should see the failure there, not in production behavior.

## The dormant validator has bugs

`neuro_san/internals/validation/network/toolbox_network_validator.py` exists but was never wired into the composite, and it cannot be wired as-is:

1. **Wrong membership test.** It checks the agent's *name* against the toolbox keys instead of the agent spec's `"toolbox"` value:

   ```python
   elif agent_name not in self.tools:
       errors.append(f"Toolbox agent '{agent_name}' has no matching tool in toolbox.")
   ```

   An agent `{"name": "web_browse_tool", "toolbox": "requests_get"}` would be flagged as missing even when `requests_get` exists, while `{"name": "my_tool", "toolbox": "no_such_tool"}` would pass whenever the agent happens to share a name with any toolbox entry — false positives and false negatives.

2. **Wrong "is a toolbox agent" heuristic.** It treats every agent without `instructions` as a toolbox agent (`if agent.get("instructions") is None:`), which also sweeps in class-based coded tools (agents with a `"class"` key) and would flag them all as having no matching toolbox entry.

## Suggested fix

- Identify toolbox agents by the presence of the `"toolbox"` key, and validate `agent.get("toolbox")` against the toolbox names.
- Validate against the names the server would actually resolve: the default toolbox info file overlaid with any user file from `AGENT_TOOLBOX_INFO_FILE` / the network's `toolbox_info_file` key (i.e. `ToolboxFactory.load()` semantics). Consider exposing a name-listing method on `ToolboxFactory` rather than duplicating the overlay logic in the validator.
[...]

PR #1245, opened 2026-08-14
Remove langchain-community requests tools from the default toolbox

### LINKED ISSUE #1132: Remove (deprecated) langchain-community
## Background

`langchain-community` has been sunset by the LangChain team
(https://github.com/langchain-ai/langchain-community/issues/674). The repository
was archived on June 19, 2026 and is now read-only. The package remains
installable from PyPI but is frozen: no bug fixes, no security patches, no new
model/tool support, and no compatibility guarantees with future
`langchain-core` releases.

Integrations were expected to move to dedicated packages, but only some did.
This issue tracks the effect on neuro-san, which currently pins
`langchain-community>=0.4,<1.0` in `requirements.txt`.

## Impact 1: Token cost fallback tables are dead ends

`neuro_san/internals/run_context/langchain/token_counting/llm_token_callback_handler.py`
imports fallback price lookup tables for OpenAI/Azure OpenAI and
Anthropic/Bedrock Anthropic models:

- `langchain_community.callbacks.openai_info` — `get_openai_token_cost_for_model`, `TokenType`
- `langchain_community.callbacks.bedrock_anthropic_callback` — `MODEL_COST_PER_1K_INPUT_TOKENS`, `MODEL_COST_PER_1K_OUTPUT_TOKENS`

These were never moved to `langchain-openai` / `langchain-aws` and have no
successor package. The tables are frozen at 2024/early-2025-era models and were
already returning $0 for newer models before the sunset
(https://github.com/langchain-ai/langchain-community/issues/503). They cannot
be relied on for cost accounting anymore.

**Resolution:** drop the fallback tables and use the price information in the
LLM info hocon (`price_per_1k_input_tokens` / `price_per_1k_output_tokens` in
`default_llm_info.hocon`) as the only price source. When a model has no price
info, default the cost to 0 and log a warning.

## Impact 2: Default toolbox tools are built on langchain-community classes

Every langchain tool/toolkit shipped in the default toolbox
(`neuro_san/internals/run_context/langchain/toolbox/toolbox_info.hocon`)
instantiates classes from `langchain_community`:

| Toolbox entry | Class |
|---|---|
| `requests_toolkit` | `langchain_community.agent_toolkits.openapi.toolkit.RequestsToolkit` |
| `requests_get` / `requests_post` / `requests_patch` / `requests_put` / `requests_delete` | `langchain_community.tools.requests.tool.Requests*Tool` |
| (constructor arg of all of the above) | `langchain_community.utilities.requests.TextRequestsWrapper` |

These tools have to be removed or migrated to dedicated integration packages.
The problem: dedicated packages exist for only some former community tools —
e.g. Tavily search has `langchain-tavily` — but the requests tools have none
(`langchain-requests` does not exist on PyPI). Options for the requests tools:

1. Remove them from the default toolbox and document alternatives.
2. Reimplement them as neuro-san shared coded tools (no langchain dependency).
3. Vendor the small `RequestsToolkit`/`Requests*Tool`/`TextRequestsWrapper`
   code (MIT-licensed) into neuro-san.

## Other affected references

- `neuro_san/internals/run_context/langchain/toolbox/toolbox_factory.py:30` —
  `from langchain_community.agent_toolkits.base import BaseToolkit`.
  Direct replacement exists: `from langchain_core.tools import BaseToolkit`.
- `toolbox_factory.py:155` — user-facing error message cites
  `langchain_community.tools.bing_search.BingSearchResults` as an example class.
- `docs/toolbox_info_hocon_reference.md:73,100` — examples reference
  `langchain_community` classes.
- `tests/.../toolbox/test_toolbox_factory.py` — tests reference the above.

### LINKED ISSUE #1136: Remove requests tools from toolbox since they are from langchain community

### LINKED ISSUE #1241: Remove langchain-community requests_* tools shipped with allow_dangerous_requests=true from default toolbox
Part of #1237 (GHSA-3xg4-wfwr-gc2g, dangerous default).

## Problem

The default toolbox (`neuro_san/internals/run_context/langchain/toolbox/toolbox_info.hocon`) ships `requests_get` / `requests_post` / `requests_put` / `requests_patch` / `requests_delete` and `requests_toolkit`, all built on langchain-community classes with `allow_dangerous_requests = true` hardcoded. Any deployment that wires these into an agent exposed to untrusted or retrieved (RAG) input allows that input to drive arbitrary server-side HTTP requests (SSRF to internal services or cloud metadata endpoints, arbitrary POST exfiltration) — amplified by the tool-selection desync in #1239.

Removal is already signposted: `toolbox_factory.py` logs a deprecation warning for `langchain_community.*` tool classes, and langchain-community itself has been sunset (repository archived).

## Fix plan

1. Remove the `requests_get/post/put/patch/delete` and `requests_toolkit` entries from `toolbox_info.hocon`.
2. Remove or rewrite the shipped example registry `neuro_san/registries/requests_get.hocon` (and any docs/examples referencing the `requests_*` toolbox tools).
3. Migration note in release notes / docs: operators who need HTTP tools must add them explicitly via `AGENT_TOOLBOX_INFO_FILE`, making `allow_dangerous_requests` an explicit, informed opt-in in their own config.
4. If langchain-community was a dependency only for these tools, drop it; otherwise leave dependency cleanup to the broader langchain-community removal.
5. Docs guidance for any future built-in HTTP tool: default-deny link-local/metadata (169.254.169.254) and RFC1918 addresses, with explicit opt-in for internal ranges.

Noravee:
## Description

Removes the six `requests_*` entries (`requests_get`, `requests_post`, `requests_put`, `requests_patch`, `requests_delete`, and `requests_toolkit`) from the default toolbox. They were built on the deprecated [langchain-community](https://github.com/langchain-ai/langchain-community/issues/674) package and shipped with `allow_dangerous_requests=true`, so every default deployment carried HTTP tools with dangerous requests pre-enabled. Operators who need HTTP tools now define them in their own toolbox info file (`AGENT_TOOLBOX_INFO_FILE` environment variable or the `toolbox_info_file` hocon key), which makes any dangerous-request opt-in an explicit, informed decision.

Fixes #1241
Fixes #1136
Fixes #1132 — these toolbox entries were the last functional use of `langchain-community` in neuro-san (the token-cost fallback tables and the `requirements.txt` pin were already removed). What remains after this PR are prose mentions and the factory's sunset warning for user-configured community classes; nothing in the library imports or depends on the sunset package.

## What changed

- **Default toolbox** (`toolbox_info.hocon`): the six langchain entries are removed; only coded tools ship by default. The file's comments now state the removal briefly and point at the schema reference instead of duplicating it.
- **Reference cleanup**: the `requests_get.hocon` example registry and its manifest entry are removed, the example test fixture is deleted, and `date_time_timezone.hocon` points at the toolbox info file instead of hand-maintaining a tool list.
- **Migration errors**: a `"toolbox"` reference to a removed tool now raises targeted guidance (why it was removed and how to re-enable it deliberately) instead of a generic `Tool 'requests_get' is not defined in None.` A user toolbox file that overrode only the `args` of a removed entry (inheriting `class` from the bundled default via the overlay) gets the same guidance instead of a nameless missing-`class` error. Unknown tool names now report the searched sources by name, and behave identically across both toolbox entry points — `get_shared_coded_tool_class` previously crashed with a bare `AttributeError` on names with no entry.
- **Docs** (`toolbox_info_hocon_reference.md`): the langchain `class` example is fixed (`langchain_tavily.tavily_search.TavilySearch` — the previous two-segment re-export form is rejected by the resolver) and the three-segment `<package_name>.<module_name>.<ClassName>` requirement is documented; the nested-`args` section documents the `from*_api_wrapper` classmethod preference; intros no longer advertise HTTP tools as shipping by default.
- **Tests**: unmocked coverage of the langchain toolbox path (a real `BaseTool` subclass with a real nested wrapper class, no API key needed), an assertion that the langchain-community sunset warning actually logs, env-hermetic connectivity tests, and tests for the new migration errors.

## Migration

Existing networks that reference the removed names keep starting (tool-creation failures are journaled, matching the behavior for any unknown tool), but the journal warning now explains the removal and points at `AGENT_TOOLBOX_INFO_FILE` / `toolbox_info_file`. See [toolbox_info_hocon_reference.md](https://github.com/cognizant-ai-lab/neuro-san/blob/main/docs/toolbox_info_hocon_reference.md) for defining tools in your own toolbox info file.

## Follow-up

- #1244 tracks fixing and wiring `ToolboxNetworkValidator` so dangling toolbox references fail pre-flight validation instead of only surfacing at runtime.


neuro_san/internals/run_context/langchain/toolbox/toolbox_factory.py
```diff
@@ -232,6 +249,42 @@ def create_tool_from_toolbox(
...
+        """
+        tool_info: Dict[str, Any] = self.toolbox_infos.get(tool_name)
+        if not tool_info:
+            if tool_name in self.REMOVED_TOOLS:
+                raise self._removed_tool_error(tool_name)
+            sources: str = "the default toolbox info file"
+            if self.toolbox_info_file:
+                sources += f" or in {self.toolbox_info_file}"
+            raise ValueError(f"Tool '{tool_name}' is not defined in {sources}.")
+        return tool_info
```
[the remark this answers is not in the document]
Noravee: Agreed — fixed in 352604f4. Absence is now checked with `is None`, and the dictionary-shape and `class`-presence validation moved into `_require_tool_info()` itself. That keeps the two entry points consistent: applying `is None` only at the lookup would have let an empty entry reach `get_shared_coded_tool_class()` and return `None` for the class, crashing downstream — the exact failure mode this PR removed. An empty-but-present entry now gets `Tool 'x' is missing required key: 'class'` from both paths (or the migration guidance when the name is a removed tool), with a test covering both.

neuro_san/internals/run_context/langchain/toolbox/toolbox_info.hocon
```diff
@@ -21,189 +21,33 @@
 {
     # ------------------------------------------------------------------------------------------
     # Overview:
-    # - This file defines configurations for langchain's tools and toolkits as well as shared coded tools.
-    # - More details about LangChain tools and toolkits can be found at:
-    #   https://docs.langchain.com/oss/python/integrations/tools
+    # - This file defines the tools that ship with the default neuro-san toolbox. The format
+    #   supports both langchain tools/toolkits and shared coded tools; the default entries
+    #   below are all coded tools.
     # - See https://github.com/cognizant-ai-lab/neuro-san/blob/main/docs/agent_hocon_reference.md for more informations on coded tools.
```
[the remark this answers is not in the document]
Noravee: Fixed in 352604f4.

neuro_san/internals/run_context/langchain/toolbox/toolbox_factory.py
```diff
@@ -232,6 +232,65 @@ def create_tool_from_toolbox(
...
+        tool_info: Dict[str, Any] = self.toolbox_infos.get(tool_name)
+        if tool_info is None:
+            if tool_name in self.REMOVED_TOOLS:
+                raise self._removed_tool_error(tool_name)
+            sources: str = "the default toolbox info file"
+            if self.toolbox_info_file:
+                sources += f" or in {self.toolbox_info_file}"
+            raise ValueError(f"Tool '{tool_name}' is not defined in {sources}.")
+
+        if not isinstance(tool_info, Dict):
```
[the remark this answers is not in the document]
Noravee: Changed to the builtin `dict` in 3f512508. One correction to the premise: the unsubscripted `typing.Dict` alias is valid in `isinstance` checks at runtime — the "Subscripted generics" TypeError only applies to forms like `Dict[str, Any]` — so malformed entries were being detected. The builtin is still the right idiom since `typing.Dict` is deprecated, hence applied.

Noravee:
## TL;DR

**What:** removes the six `requests_*` entries (`requests_get/post/put/patch/delete`, `requests_toolkit`) from the default toolbox — the default now ships only coded tools. Closes #1241 and #1136, and completes #1132: these entries were the last functional use of `langchain-community` in the library.

**Why:** they were built on the deprecated langchain-community package and shipped with `allow_dangerous_requests=true`, so every default deployment carried HTTP tools with dangerous requests pre-enabled. Re-enabling is now an explicit opt-in via a user toolbox info file (`AGENT_TOOLBOX_INFO_FILE` / `toolbox_info_file`).

**Suggested reading order:**

1. `toolbox_info.hocon` — the removal itself.
2. `toolbox_factory.py` — the migration-error work: `REMOVED_TOOLS` plus `_require_tool_info()`, which consolidates all entry validation into one choke point shared by both toolbox entry points.
3. `docs/toolbox_info_hocon_reference.md` — migration note, corrected `class` example (the old one could never resolve), and the `from*_api_wrapper` classmethod note.
4. Tests — unmocked real-path coverage replacing the deleted fixture (no API key needed), migration-error tests, env-hermeticity fixes.
5. Registry/manifest/fixture cleanups.

**Key decisions to sanity-check:**

- A removed tool behaves exactly like any unknown tool at runtime: tool creation fails, the error is journaled, and the agent starts without the tool. Only the message quality changed (actionable migration guidance instead of `not defined in None`). No special fatal handling — deliberate.
- Runtime error messages cite the langchain-community deprecation and the re-enable mechanism, no issue numbers.
- The docs deliberately do **not** include a copy-paste re-enable entry for the dangerous tools.
- The validation gap (networks with dangling `toolbox` references pass `hocon_validator_cli`) is split out to #1244 — wiring a validator changes validation behavior repo-wide and deserves its own review.

**Verification:** 452 unit tests green locally, flake8/pylint 10/10, pymarkdown clean, CI green, and Copilot review looped to convergence (5 rounds; every comment fixed or answered in-thread).

neuro_san/internals/run_context/langchain/toolbox/toolbox_info.hocon
```diff
@@ -21,189 +21,33 @@
...
     #
     # Customization via Agent HOCON File:
-    # - In the agent HOCON file, users can define an "args" section to **add or override** 
+    # - In the agent HOCON file, users can define an "args" section to **add or override**
     #   arguments from this tool configuration file dynamically at runtime.
     # ------------------------------------------------------------------------------------------
 
     # --------------------------- Langchain's Tools and Toolkits -------------------------------
-    # Requirements:
+    # No langchain tools ship in this default toolbox. The former requests_* tools and
```
Noravee, line 35: The removal itself. The short comment here is deliberate: the full rationale, migration steps, and SSRF guidance live in one place (docs/toolbox_info_hocon_reference.md) so future wording changes don't have to chase copies — an earlier revision had near-identical paragraphs in both files. Operators who define their own toolbox info file get it overlaid on this one, so opting the requests tools back in works exactly as before, minus the dangerous default.

neuro_san/internals/run_context/langchain/toolbox/toolbox_factory.py
```diff
@@ -77,6 +79,19 @@ class ToolboxFactory(ContextTypeToolboxFactory):
         "neuro_san/internals/run_context/langchain/toolbox/toolbox_info.hocon"
     """
 
+    # Tools that used to ship in the default toolbox info file but were removed
+    # because they were built on the deprecated langchain-community package.
+    # A stale reference to one of these gets a targeted migration message
+    # instead of the generic "not defined" error.
+    REMOVED_TOOLS: FrozenSet[str] = frozenset({
```
Noravee, line 86: This hardcoded list exists purely for error quality: a stale `"toolbox": "requests_get"` reference gets migration guidance instead of the generic (and previously baffling) `Tool 'requests_get' is not defined in None.` It also catches the subtler case where a user toolbox file overrode only the `args` of a bundled entry and inherited `class` via the overlay — post-removal such an entry is class-less and would otherwise die on a nameless missing-'class' error. The set can be dropped once migrations from this release stop mattering.

neuro_san/internals/run_context/langchain/toolbox/toolbox_factory.py
```diff
@@ -232,6 +232,65 @@ def create_tool_from_toolbox(
         instance.tags = ["langchain_tool"]
         return instance
 
+    def _require_tool_info(self, tool_name: str) -> Dict[str, Any]:
```
Noravee, line 235: Single validation choke point shared by both toolbox entry points. Presence, dict shape, 'class' presence, and non-empty-string all lived inline in create_tool_from_toolbox() only; get_shared_coded_tool_class() had none of it. Runtime severity is unchanged: these ValueErrors are still caught and journaled by BaseToolFactory.create_toolbox_tool, so an agent referencing a removed/unknown tool starts and serves without it, exactly like any unknown tool — a removed tool is just a nonexistent tool now. Only the message quality changed.

neuro_san/internals/run_context/langchain/toolbox/toolbox_factory.py
```diff
@@ -302,9 +361,11 @@ def get_shared_coded_tool_class(self, tool_name: str) -> str:
         Get class of the shared coded tool
 
         :param tool_name: The name of the tool
-        :return: The class of the coded tool
+        :return: The class of the coded tool.
+                Can raise a ValueError if the tool_name is unknown to the toolbox,
+                just like create_tool_from_toolbox().
         """
-        tool_info: Dict[str, Any] = self.toolbox_infos.get(tool_name)
+        tool_info: Dict[str, Any] = self._require_tool_info(tool_name)
```
Noravee, line 368: Before this PR this method did `self.toolbox_infos.get(tool_name).get("class")` — an unknown name crashed with a bare `AttributeError: 'NoneType' object has no attribute 'get'`. The path is reachable outside create_tool_from_toolbox()'s guards: ToolboxActivation.get_full_class_ref() lands here, e.g. when the model emits a tool name whose binding was dropped at setup (calling_activation validates the call against the spec's tool list, not the successfully bound tools). It now fails with the same actionable ValueError as everywhere else.

docs/toolbox_info_hocon_reference.md
```diff
@@ -63,24 +65,31 @@ The value for each key is a dictionary describing the tool's properties. The sch
...
+Fully qualified class name of the tool. It must exist in the server's `PYTHONPATH` and must have at least
+three dot-separated segments: `<package_name>.<module_name>.<ClassName>`. Point at the module that actually
+defines the class — the shorter top-level re-export that many integration packages offer
+(e.g. `langchain_tavily.TavilySearch`) is rejected.
 
 Example:
 
 ```hocon
-"class": "langchain_community.tools.requests.tool.RequestsGetTool"
+"class": "langchain_tavily.tavily_search.TavilySearch"
```
Noravee, line 89: The previous example here (`langchain_tavily.TavilySearch`, a top-level re-export) could never work: `_resolve_class` requires at least 3 dot segments because it splits the path into package/module/class for leaf-common's `Resolver(packages).resolve_class_in_module(...)`. The corrected path is verified against the actual langchain-tavily 0.2.18 wheel: `langchain_tavily/tavily_search.py` defines `class TavilySearch(BaseTool)`.

docs/toolbox_info_hocon_reference.md
```diff
@@ -93,26 +102,37 @@ Example:
...
         }
     }
 }
 ```
 
-This instantiates `RequestsGetTool(requests_wrapper=TextRequestsWrapper(headers={}), allow_dangerous_requests=True)`.
+This instantiates `SomeTool(api_wrapper=SomeApiWrapper(timeout=30))`.
+
+> [!NOTE]
+> If the tool class defines a `from..._api_wrapper` classmethod (several langchain toolkits do, e.g.
```
Noravee, line 132: This note exists because the factory prefers any `from*...api_wrapper` classmethod over the constructor (see `_get_from_api_wrapper_method`), and argument validation runs against the chosen callable's signature. Concretely: `GitHubToolkit.from_github_api_wrapper` takes `github_api_wrapper`, so a literal `api_wrapper` kwarg would be rejected with a signature error the constructor-only story couldn't explain. The replaced requests example had no such classmethod, so the old docs never hit this.

tests/neuro_san/internals/run_context/langchain/toolbox/test_toolbox_factory.py
```diff
@@ -101,9 +105,133 @@ def test_create_toolbox_returns_single_base_tool(self, factory):
...
+
+        factory.toolbox_info_file = None
+        with pytest.raises(ValueError, match="not defined in the default toolbox info file."):
+            factory.create_tool_from_toolbox("no_such_tool", {})
+
+        factory.toolbox_info_file = "/path/to/user_toolbox.hocon"
+        with pytest.raises(ValueError, match="or in /path/to/user_toolbox.hocon"):
+            factory.create_tool_from_toolbox("no_such_tool", {})
+
+    def test_create_toolbox_real_tool_unmocked(self, factory):
```
Noravee, line 165: The deleted requests_get fixture was the only unmocked coverage of the langchain branch of create_tool_from_toolbox() — every other test patches Resolver/ArgumentValidator, and the surviving coded tool short-circuits before that branch. This test replaces that coverage without needing an API key: a real BaseTool subclass and a real nested wrapper class resolved from real dotted paths, exercising class resolution, the nested dict-with-class branch of _resolve_args(), real argument validation, user-args overlay precedence, and name/tag assignment.

neuro_san/registries/date_time_timezone.hocon
```diff
@@ -85,23 +85,11 @@
...
-            #   - "requests_put": Makes an HTTP PUT request.
-            #   - "requests_delete": Makes an HTTP DELETE request.
-            #   - "requests_toolkit": Contains all of the request tools.
-            #
-            # Coded Tools
-            #
-            #   - "get_current_date_time": Return current date and time for a specified timezone.
-            #
-            # Each of these tools may require specific arguments in the "args" section.
+            # AVAILABLE TOOLS
```
Noravee, line 88: The hand-maintained 'CURRENTLY AVAILABLE TOOLS' list is gone on purpose: it had already drifted (it listed the six removed tools until this PR) and would drift again with every toolbox change. Pointing at the toolbox info file keeps a single source of truth.

neuro_san/internals/run_context/langchain/toolbox/toolbox_factory.py
d1donlydfink: What will happen if someone later purposely re-adds these tools? 
(Should we care?)

d1donlydfink (approved):
What will happen if someone later purposely re-adds these Request tools?
(Should we care?)

It is a shame to lose a web-services example. It'd be good to look for a replacement.

neuro_san/internals/run_context/langchain/toolbox/toolbox_factory.py
Noravee: Purposely re-adding them works — `REMOVED_TOOLS` doesn't block anything. It's only consulted when the lookup finds no entry (or a class-less one), so a user toolbox info file that defines `requests_get` with a full entry (class + args) takes normal precedence via the overlay and resolves like any other tool. The operator owns the consequences: they install langchain-community themselves (it's no longer a neuro-san dependency, though the frozen package remains on PyPI), and they get the sunset warning at tool-creation time. A partial re-add (args without class) gets the migration error pointing at the docs. So: no, I don't think we should care beyond what already happens — the removal is about not *shipping* the dangerous default, not about preventing an informed opt-in.

Noravee:
> What will happen if someone later purposely re-adds these Request tools? (Should we care?)

Answered in the inline thread: a full re-add in a user toolbox info file just works — `REMOVED_TOOLS` only shapes the error for missing/partial entries, it doesn't block an informed opt-in. The operator installs langchain-community themselves and gets the sunset warning at tool creation.

> It is a shame to lose a web-services example. It'd be good to look for a replacement.

Agreed — we can have a replacement, with one constraint: anything in the *default* toolbox must have its class importable by the server, so a langchain-based replacement means adding its package to `requirements.txt` (or `requirements-build.txt` if it's only exercised by tests/examples). Re-introducing langchain-community as that dependency is off the table since it's deprecated, and no maintained dedicated package exists for the generic requests tools (`langchain-requests` isn't on PyPI — see #1132). Two candidate shapes:

1. **A neuro-san shared coded tool** (e.g. `http_get`) built on `aiohttp`/`httpx`, which are already runtime dependencies — no new requirement at all, and we could bake in the SSRF hardening (deny link-local/metadata and RFC 1918 by default) that the langchain requests tools never had.
2. **A maintained integration package** (e.g. `langchain-tavily` for web search) — the closest maintained "fetch things from the web" example, but it's search rather than raw HTTP, and needs an API key plus a requirements entry.

Option 1 seems like the better fit for a default example. Can open a follow-up issue if that sounds right.

PR #1247, opened 2026-08-14
Use latest leaf-common==1.3.1

d1donlydfink:
This lighter version of leaf-common also lightens the transient dependency load.
Use leaf-common==1.3.1 which removes the deprecation redirects for Resolver and ResolverUtil - clean break.


Noravee (approved):
LGTM.

PR #1248, opened 2026-08-14
NS-1246 Add record/playback LLM proxy for neuro-san load & regression testing.

andreidenissov-cog:
## Summary
Adds a standalone, OpenAI-compatible HTTP proxy under
tests/record_playback_llm_server/ that records a neuro-san session against a
real LLM host and plays it back offline. It solves two problems with our
load/regression tests: they cost real tokens, and LLM responses are
non-deterministic. Recording once and replaying from disk makes those tests
free and repeatable.

It is wire-compatible with the OpenAI Chat Completions API, so any agent
network configured for class = "openai" is redirected at it with a single
openai_api_base change -- the same seam the existing tests/mock_llm_server uses.
This is test-only tooling; no production code is changed.

## Modes (selected with --mode)
- record: forward every request to the real host (endpoint + key from env
  vars), relay the real response back, and tee it into a cassette file,
  capturing wall-clock latency.
- playback: serve responses from the cassette by matching a canonical request
  signature. No network, no tokens. An unmatched request fails hard with 504,
  so tests surface gaps deterministically instead of silently faking data.
- hybrid: like playback, but a cache miss falls through to the real host (when
  configured), records the result into the current cassette, and returns it --
  a self-healing cassette.

## Additional capabilities
- Multi-response (--multi-response): record accumulates every distinct response
  for a request; playback serves them round-robin.
- Playback delay (--delay-mode none|recorded|fixed|random): an up-front
  per-response delay before serving a cache hit, emulating LLM latency.
  "recorded" uses each response's own recorded latency (first_byte_seconds for
  streams, latency_seconds for one-shot); in multi-response each variant uses
  its own.
- Latency capture: each recorded response stores wall-clock latency_seconds
  (and first_byte_seconds for streams).
- Success-only recording: non-2xx upstream responses (429/401/5xx) are relayed
  to the caller but never cached, so a transient failure cannot poison a
  cassette.
- Cassette cleaner (cleanup_cassette.py): a standalone tool that strips any
  non-2xx responses from an existing cassette (dropping single-response entries,
  trimming multi-response variants) to make an older cassette safe for playback.
- TLS verification anchored to the certifi CA bundle (matches the OpenAI/httpx
  SDKs) so recording against public hosts works on macOS python.org builds.
- Configurable upstream timeouts and max in-flight requests via env vars, to
  avoid spurious timeouts under load.

## How matching works
The response is non-deterministic but the request is a deterministic function of
the agent network and its inputs. In multi-turn flows each request embeds the
prior responses (assistant messages, tool_call ids, tool results); because
playback returns byte-identical recorded responses, downstream requests
reconstruct identically, so a sha256 of the canonicalized request body is a
stable key across a whole conversation.

## Configuration (record and hybrid modes), via environment variables
- RECORD_PLAYBACK_UPSTREAM_BASE_URL, RECORD_PLAYBACK_UPSTREAM_API_KEY
- RECORD_PLAYBACK_UPSTREAM_REQUEST_TIMEOUT_SECONDS,
  RECORD_PLAYBACK_UPSTREAM_CONNECT_TIMEOUT_SECONDS,
  RECORD_PLAYBACK_UPSTREAM_MAX_CLIENTS

## Layout
One class per file: RecordPlaybackLlmServer (CLI entry point), ProxyState,
ProxyHandler (shared record/playback/hybrid logic) with ChatCompletionsHandler
and ModelsHandler endpoints, HealthHandler, UpstreamClient, Cassette,
PlaybackDelay, RequestCanonicalizer, and the CassetteCleaner tool. A README
documents usage for every mode, the cleaner, and the cassette format.

## Testing
- flake8 clean and pylint 10.00/10 on the whole package.
- Committed pytest at tests/record_playback_llm_server/test_record_playback_llm_server.py
  (21 tests) runs in the default unit suite with no server or LLM key required.
  It covers: request canonicalization key stability; Cassette round-trip and
  multi-response de-duplication; all PlaybackDelay policies (including stream
  first-byte vs one-shot latency); CassetteCleaner drop/trim rules; and
  in-process record/playback/hybrid integration against a fake upstream --
  json + stream byte-identical replay with no upstream contact on hits, hard 504
  on a playback miss, multi-response round-robin, hybrid record-on-miss then hit,
  and non-2xx relayed but not recorded.

## Related issue
NS-1246

## Notes for reviewers
- Test-only: everything lives under tests/, no neuro_san/ production code is
  modified.
- The proxy is OpenAI-wire only (hosts reachable via an OpenAI-compatible
  base_url); native Anthropic/Bedrock/Gemini wire formats are out of scope.


andreidenissov-cog:
Superseded by #1261 (branch ASD-NS-1246-boombox01), which contains the identical change set (including the CodeQL exception-exposure fix). Continuing review on #1261.

Issue #1237, opened 2026-08-13
Address Security Advisory.


d1donlydfink:
https://github.com/cognizant-ai-lab/neuro-san/security/advisories/GHSA-3xg4-wfwr-gc2g

Noravee:
All three sub-issues are remediated and released: #1239 (PR #1243) and #1241 (PR #1245) shipped in 0.6.93; #1240 (PR #1251) shipped in 0.6.94. The security advisory will be published separately per coordinated disclosure. Follow-up hardening (toolbox reference validation) continues in #1244.

PR #1251, opened 2026-08-14
Add AGENT_TOOL_PATH_ONLY to restrict CodedTool class resolution

### LINKED ISSUE #1240: Add AGENT_TOOL_PATH_ONLY to restrict CodedTool class resolution to AGENT_TOOL_PATH
Part of #1237 (GHSA-3xg4-wfwr-gc2g, Finding 2).

## Problem

CodedTool class resolution (`AbstractClassActivation._attempt_resolve` in `neuro_san/internals/graph/activations/abstract_class_activation.py`) tries a **Phase 1 direct import of the fully-qualified dotted path** before consulting the `AGENT_TOOL_PATH` hierarchy in Phase 2. A `class` ref in an agent-network HOCON can therefore load any module importable on the server's PYTHONPATH, not just code under the operator's tool directory.

Two consequences:

- Python imports execute module-level code, so even a class ref that later fails the `issubclass(CodedTool)` check has already run the target module's top-level code during resolution.
- Combined with manifest hot-reload, write access to the registry directory alone (no ability to write Python anywhere) is enough to trigger import of any installed module — effectively code execution in the server process, at runtime, without a restart.

## Fix plan

1. **New boolean env var `AGENT_TOOL_PATH_ONLY`** (default `false`, naming consistent with `AGENT_MCP_ONLY`). When `true`, `_attempt_resolve` skips Phase 1 entirely; class refs resolve only through Phase 2's `Resolver` rooted under the agent tool path hierarchy. Phase 2 prepends the package root before importing, so a ref cannot escape it.
2. **Toolbox coded tools are governed by the same flag.** `ToolboxActivation` resolves through the same `AbstractClassActivation.build()` → `resolve_class()` path, so gating Phase 1 covers them automatically. Verify that shipped built-in coded tools under `neuro_san/coded_tools/` (e.g. `get_current_date_time.GetCurrentDateTime`) remain resolvable in strict mode — if needed, include that known-safe package as an additional Phase 2 root rather than an exception mechanism.
3. **Fail loudly at load time.** Add a validation pass at registry load/hot-reload that checks each network's `class` refs under the strict rules (pure string/prefix check — no import attempts), so a network that would break is rejected at load with a clear error instead of failing on first tool invocation. Rejecting one network must not affect already-loaded networks.
4. **Scope note in docs**: the flag governs CodedTool class resolution (registry `class` refs and toolbox coded tools). Langchain-side toolbox entries (`"class": "langchain_community..."` style base tools) are instantiated by `toolbox_factory` through a different path and are operator startup config; they are out of scope for this flag.
5. **Backward compatibility**: default `false` because fully-qualified refs to pip-installed packages are currently legal. Log a one-time startup notice when the flag is off (resolution is unrestricted). Consider flipping the default in a future major version.
6. **Documentation of the trust model**: agent registries and coded tools are code-equivalent. Anyone with write access to the registry directory or `AGENT_TOOL_PATH` can run code as the server. Recommend restricted write permissions, review of registry changes, and least-privilege/containerized deployment.
7. **Tests**: strict mode blocks a fully-qualified ref outside the tool path (and the module is never imported); strict mode still resolves tools under `AGENT_TOOL_PATH` and shipped toolbox coded tools; default mode preserves current behavior; hot-reload of a violating network is rejected loudly without disturbing existing networks.

Noravee:
## Description

Adds a boolean environment variable `AGENT_TOOL_PATH_ONLY` (default `false`) that restricts CodedTool `class` resolution to the `AGENT_TOOL_PATH` hierarchy.

Fixes #1240 (part of #1237 / GHSA-3xg4-wfwr-gc2g, Finding 2).

## Problem

`AbstractClassActivation` resolved a CodedTool `class` reference by first attempting a direct import of the fully-qualified dotted path (Phase 1), before falling back to the `AGENT_TOOL_PATH` hierarchy (Phase 2). Because importing a module executes its top-level code, a `class` reference in an agent-network hocon could run import-time code from **any** module on the server's `PYTHONPATH` — even one that later fails the `issubclass(CodedTool)` check. Combined with manifest hot-reload, write access to the registry directory alone was enough to trigger arbitrary module import in the server process at runtime.

## What changed

- **New flag `AGENT_TOOL_PATH_ONLY`.** When `true`, Phase 1 (fully-qualified direct import) is skipped entirely, so a CodedTool `class` resolves only through the `Resolver` rooted under the `AGENT_TOOL_PATH` hierarchy, which a reference cannot escape. Parsed with `leaf_common` `ConfigUtil.get_bool`, so boolean-like spellings (`true`/`yes`, case-insensitive, trimmed) all enable it — a security control should not silently fail open on `AGENT_TOOL_PATH_ONLY=1`.
- **Toolbox shared coded tools are covered by the same flag**, since they resolve through the same `AbstractClassActivation.build()` path. Shipped `neuro_san/coded_tools` tools remain resolvable in strict mode through the existing hierarchy walk (tested).
- **Default `false`** for backwards compatibility (fully-qualified refs to installed packages are currently legal). A one-time notice is logged at first CodedTool resolution when running unrestricted, and the strict-mode resolution error explains the restriction and the relative form to use instead.
- **Docs**: the trust model (registry write access is code-equivalent) plus a precise statement of what the flag does and does **not** cover.

## Scope

Per the issue, this flag governs **CodedTool class resolution** (registry `class` refs and toolbox shared coded tools). It deliberately does not restrict `llm_config.class`, middleware classes, or langchain toolbox tool classes: those are expected to be installed provider/framework packages with no `AGENT_TOOL_PATH`-equivalent to confine them to, so the flag is the wrong mechanism for them. The docs state this explicitly rather than implying the flag closes those vectors. The registry-write trust boundary remains the primary control there.

## Follow-up

The fix plan's load-time validation pass (reject networks whose `class` refs would fail under strict mode, up front instead of on first invocation) is deferred to a follow-up, as it overlaps the network-validator work in #1244.

## Tests

Run unmocked against real fixture modules (no `Resolver` mocks): strict mode blocks a fully-qualified reference outside the tool path **without ever importing the module** (asserted via `sys.modules`), still resolves network-specific, shared, and shipped toolbox coded tools, boolean-like flag values enable it, and default mode preserves current behavior.


docs/toolbox_info_hocon_reference.md
```diff
@@ -158,7 +158,9 @@ These tools extend from the `CodedTool` class.
 #### [`class`](./agent_hocon_reference.md#class)
 
 Fully qualified class name in the format `tool_module.ToolClass`. The `class` must point to a module available in your
-`AGENT_TOOL_PATH` and server `PYTHONPATH`.
+`AGENT_TOOL_PATH` and server `PYTHONPATH`. When the `AGENT_TOOL_PATH_ONLY` environment variable is `true`, resolution
+is restricted to the `AGENT_TOOL_PATH` hierarchy: fully-qualified references to modules elsewhere on the `PYTHONPATH`
+are rejected.
```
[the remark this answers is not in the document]
Noravee: Fixed in e9aee91f. Reworded to match the implementation: strict mode does not resolve fully-qualified references at all (even ones inside AGENT_TOOL_PATH), so a shared coded tool's class must be given relative to the hierarchy as `tool_module.ToolClass`.

neuro_san/internals/graph/activations/abstract_class_activation.py
```diff
@@ -228,6 +233,15 @@ def resolve_class(self, class_name: str, module_name: str):
...
         agent_name: str = self.factory.get_name_from_spec(self.agent_tool_spec)
         agent_tool_path: str = ".".join(this_agent_tool_path_parts[:-len(agent_network_name_parts)])
+        strict_note: str = ""
+        if self.is_agent_tool_path_only():
+            strict_note = """
+Note: AGENT_TOOL_PATH_ONLY is enabled, so fully-qualified class references are
+not resolved at all — even ones that point inside AGENT_TOOL_PATH. Reference the
+class relative to the AGENT_TOOL_PATH hierarchy instead, e.g. "<module>.<ClassName>"
+for a shared tool or "<agent_network>.<module>.<ClassName>" for a network-specific
+tool, rather than by its full package path.
```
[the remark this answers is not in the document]
Noravee: Good catch — fixed in e9aee91f. You're right that resolve_class() searches the network's own directory first (from get_agent_tool_path()) and walks up to the shared root, so the network name should not be in the reference; it was also wrong for hierarchical '/'-containing network names. The hint now recommends the relative `<module>.<ClassName>` form for both cases and explains the search order instead of implying the network name must be included.

Noravee:
## TL;DR

**What:** adds `AGENT_TOOL_PATH_ONLY` (default `false`). When `true`, a CodedTool `class` reference resolves only under the `AGENT_TOOL_PATH` hierarchy — the fully-qualified direct-import phase is skipped. Closes #1240 (part of #1237 / GHSA-3xg4-wfwr-gc2g, Finding 2).

**Why:** class resolution tried a direct import of the fully-qualified dotted path *before* falling back to `AGENT_TOOL_PATH`. Because importing a module executes its top-level code, a `class` ref in a registry hocon could run import-time code from any module on the server's `PYTHONPATH` — and with manifest hot-reload, write access to the registry directory alone was enough to trigger it at runtime.

**Suggested reading order:**

1. `abstract_class_activation.py` — the gate: `is_agent_tool_path_only()` and the Phase-1 skip in `_attempt_resolve()`.
2. `docs/agent_hocon_reference.md` — the trust model and the precise scope of what the flag does and does not cover.
3. Tests — unmocked resolution against real fixture modules, including the "never imported" assertion.

**Key decisions to sanity-check:**

- **Scope is deliberately CodedTool-only.** The flag governs CodedTool `class` refs and toolbox *shared coded tools*. It does **not** gate `llm_config.class`, middleware classes, or langchain toolbox tool classes — those are expected to be installed provider/framework packages with no `AGENT_TOOL_PATH`-equivalent to confine them to, so the flag would be the wrong mechanism. The docs state this rather than implying the flag closes those vectors; the registry-write-is-code-equivalent trust boundary remains the primary control there.
- **Default `false`** for backwards compatibility (fully-qualified refs to installed packages are currently legal). Flipping the default is a future major-version consideration.
- **Parsed with `ConfigUtil.get_bool`**, so `true`/`yes`/case-variants/whitespace all enable it — a security flag shouldn't silently fail open on `AGENT_TOOL_PATH_ONLY=1`.
- **Load-time validation deferred** to a follow-up (overlaps the validator work in #1244). Runtime enforcement still fails safe, and the strict-mode error explains the restriction.

neuro_san/internals/graph/activations/abstract_class_activation.py
```diff
@@ -269,13 +284,22 @@ def _attempt_resolve(self, class_name: str, module_name: str,
...
-        python_class: Type[Any] = ResolverUtil.create_type(fully_qualified_name, raise_if_not_found=False)
-        if python_class is not None:
-            return python_class
+        # Phase 1 - Try the simplest thing first - a direct import from a fully-qualified path.
+        # This is skipped when AGENT_TOOL_PATH_ONLY is set: importing a module executes its
+        # top-level code, so a fully-qualified "class" reference in an agent network hocon
+        # could otherwise run import-time code from any module on the server's PYTHONPATH.
+        # Phase 2 below roots every import under the AGENT_TOOL_PATH hierarchy, so a
+        # reference cannot escape it.
+        if not self.is_agent_tool_path_only():
```
Noravee, line 293: The core of the fix. In default mode Phase 1 does a direct import of the fully-qualified '<module>.<class>' and returns on success. Under the flag that whole block is skipped, so resolution falls straight through to Phase 2 below, which only ever imports packages rooted at the AGENT_TOOL_PATH hierarchy (verified: leaf_common's Resolver always prefixes the package root — there is no unrooted fallback). Skipping is what prevents the import: a rejected class (e.g. one that fails the later issubclass(CodedTool) check) has already executed its module's top-level code by the time it's rejected, so the only safe move is to not import it at all.

neuro_san/internals/graph/activations/abstract_class_activation.py
```diff
@@ -307,6 +331,32 @@ def _attempt_resolve(self, class_name: str, module_name: str,
...
+        :return: True if the AGENT_TOOL_PATH_ONLY environment variable restricts
+                CodedTool class resolution to the AGENT_TOOL_PATH hierarchy only.
+                Defaults to False for backwards compatibility, where fully-qualified
+                references can resolve to any class on the server's PYTHONPATH.
+        """
+        # ConfigUtil.get_bool accepts the boolean-like spellings used elsewhere for
+        # neuro-san env flags (true/yes, case-insensitive, whitespace-trimmed), so an
+        # operator setting this security flag the way other flags are set gets the
+        # behavior they expect rather than a silent fail-open.
+        return ConfigUtil.get_bool(environ, "AGENT_TOOL_PATH_ONLY", default=False)
```
Noravee, line 346: Uses leaf_common ConfigUtil.get_bool rather than a hand-rolled 'value == "true"' check. This matters specifically because it's a security control: get_bool accepts the boolean-like spellings other neuro-san flags accept (true/yes, case-insensitive, whitespace-trimmed), so an operator who sets AGENT_TOOL_PATH_ONLY the way they set every other flag gets the protection they expect instead of silently failing open. (Residual: the string '1' is still not recognized — that's get_bool's behavior repo-wide, not specific to this flag.)

neuro_san/internals/graph/activations/abstract_class_activation.py
```diff
@@ -228,6 +233,16 @@ def resolve_class(self, class_name: str, module_name: str):
         # If we exhausted all levels without success, warn and raise an error
         agent_name: str = self.factory.get_name_from_spec(self.agent_tool_spec)
         agent_tool_path: str = ".".join(this_agent_tool_path_parts[:-len(agent_network_name_parts)])
+        strict_note: str = ""
+        if self.is_agent_tool_path_only():
+            strict_note = """
```
Noravee, line 238: Strict-mode failures append this note to the standard resolution error. Two things it gets right after review: (1) it says fully-qualified refs aren't resolved AT ALL in strict mode — even ones pointing inside AGENT_TOOL_PATH — because Phase 1 is skipped wholesale; (2) it tells the operator to use the relative '<module>.<ClassName>' form with NO agent-network prefix, since resolution already searches the network's own directory first and walks up to the shared root (putting the network name in the ref would double-prefix and fail, and would break for hierarchical '/'-containing names).

neuro_san/internals/graph/activations/abstract_class_activation.py
```diff
@@ -66,6 +67,10 @@ class AbstractClassActivation(AbstractCallableActivation):
     - ToolboxActivation: looks up the class reference from a predefined toolbox.
     """
 
+    # Process-wide guard so the notice that class resolution is unrestricted
+    # is logged only once, not on every tool invocation.
+    _unrestricted_notice_logged: bool = False
```
Noravee, line 72: Process-wide guard so the 'resolution is unrestricted' notice logs once, not on every tool invocation. Known limitation: the check-then-set isn't synchronized, so under concurrent first-resolutions the notice could log more than once. Left as-is deliberately — it's a cosmetic INFO line, and a lock on every resolution to dedup a log message isn't worth it.

tests/neuro_san/internals/graph/activations/test_abstract_class_activation.py
```diff
@@ -404,3 +408,117 @@ def test_reservationist_initialization_when_allowed(self, mock_run_context, mock
...
+    """
+
+    def test_default_mode_resolves_fully_qualified_ref(self, fixture_activation, monkeypatch):
+        """Test that with the flag off, a fully-qualified ref to a module outside
+        AGENT_TOOL_PATH resolves by direct import (backwards-compatible behavior)."""
+        monkeypatch.delenv("AGENT_TOOL_PATH_ONLY", raising=False)
+        cls = fixture_activation.resolve_class("CanaryTool", CANARY_MODULE)
+        assert cls.__name__ == "CanaryTool"
+
+    def test_tool_path_only_blocks_fully_qualified_ref_without_importing(
```
Noravee, line 468: The security-critical test. It doesn't just assert strict mode raises — it asserts the target module is NOT in sys.modules afterward (popped first), proving resolution never imported it. That's the real guarantee: import-time code execution is the vulnerability, so 'rejected' isn't enough; 'never imported' is the property that matters. Runs unmocked against a real canary module deliberately placed outside the tool path.

tests/neuro_san/internals/graph/activations/test_abstract_class_activation.py
```diff
@@ -404,3 +408,117 @@ def test_reservationist_initialization_when_allowed(self, mock_run_context, mock
...
+        cls = fixture_activation.resolve_class("NetworkTool", "network_tool")
+        assert cls.__name__ == "NetworkTool"
+
+    def test_tool_path_only_resolves_shared_tool(self, fixture_activation, monkeypatch):
+        """Test that strict mode still resolves a shared tool one level up the hierarchy."""
+        monkeypatch.setenv("AGENT_TOOL_PATH_ONLY", "true")
+        cls = fixture_activation.resolve_class("SharedTool", "shared_tool")
+        assert cls.__name__ == "SharedTool"
+
+    def test_tool_path_only_accepts_boolean_like_values(self, fixture_activation, monkeypatch):
```
Noravee, line 494: Regression guard for the fail-open concern behind the ConfigUtil.get_bool choice: 'true'/'True'/'TRUE'/'yes'/' true ' must all enable strict mode (each asserted to block + not import the canary). Without this, a future refactor back to a strict '== "true"' comparison would silently reopen the fail-open hole for boolean-like values.

tests/neuro_san/internals/graph/activations/test_abstract_class_activation.py
```diff
@@ -404,3 +408,117 @@ def test_reservationist_initialization_when_allowed(self, mock_run_context, mock
...
+        """Test that common boolean-like spellings enable the flag, so an operator
+        setting it like other neuro-san flags is not silently left unrestricted."""
+        sys.modules.pop(CANARY_MODULE, None)
+        for truthy in ("true", "True", "TRUE", "yes", " true "):
+            monkeypatch.setenv("AGENT_TOOL_PATH_ONLY", truthy)
+            with pytest.raises(ValueError):
+                fixture_activation.resolve_class("CanaryTool", CANARY_MODULE)
+            assert CANARY_MODULE not in sys.modules
+
+    def test_tool_path_only_resolves_shipped_toolbox_coded_tool(self, mock_run_context, monkeypatch):
```
Noravee, line 504: Proves the scope boundary from the other direction: strict mode must still resolve the coded tools shipped under neuro_san/coded_tools (referenced by the default toolbox info file, e.g. get_current_date_time.GetCurrentDateTime), via the normal hierarchy walk — no special-casing or allow-list. If strict mode broke shipped tools, the default toolbox would break.

tests/neuro_san/internals/graph/activations/resolution_canary.py
```diff
@@ -0,0 +1,31 @@
...
+# END COPYRIGHT
+"""
+Canary module deliberately OUTSIDE any AGENT_TOOL_PATH used in tests.
+
+Resolution tests reference it fully-qualified two ways:
+- default mode resolves it (proving Phase 1 imports from anywhere on PYTHONPATH);
+- strict mode must NOT import it (importing executes top-level code, which is
+  the vulnerability the flag closes), asserted via absence from sys.modules.
+
+Nothing else may import this module, or the strict test's never-imported
```
Noravee, line 25: This module carries a 'nothing else may import it' contract: the strict test's never-imported assertion is only meaningful if no other code (or test collection) imports it first. It's safe because pytest's default collection only picks up test_*.py, and a grep confirms nothing imports resolution_canary. One shared canary now serves both the default-mode (resolves it) and strict-mode (must not import it) tests.

docs/agent_hocon_reference.md
```diff
@@ -853,6 +853,23 @@ If the agent is called `math_guy` and the class is valued as `calculator.Calcula
...
+> Because agent network hocon files can name classes to load, hocon files and coded tools are
+> code-equivalent: anyone who can write to the registry directory or `AGENT_TOOL_PATH` can
+> effectively run code as the server process. Restrict write access to both locations, review
+> registry changes as you would code changes, and prefer least-privilege deployments.
+>
+> Setting the `AGENT_TOOL_PATH_ONLY` environment variable to `true` (default `false`) hardens the
+> CodedTool `class` path specifically: fully-qualified references are no longer resolved, so a
+> CodedTool `class` resolves only under the `AGENT_TOOL_PATH` hierarchy described above and cannot
+> import an arbitrary module elsewhere on the `PYTHONPATH`. The same flag governs shared coded
+> tools referenced through the [toolbox](#toolbox). It does **not** restrict other class references
```
Noravee, line 867: The scope statement is the load-bearing sentence for reviewers. The flag hardens the CodedTool class path (and toolbox shared coded tools) only; llm_config.class, middleware classes, and langchain toolbox tool classes are still resolved fully-qualified. This is intentional — those are installed provider/framework packages with no AGENT_TOOL_PATH-equivalent to confine them to, so this flag is the wrong mechanism for them. The IMPORTANT block above frames the real boundary: registry write access is code-equivalent regardless of the flag.

d1donlydfink: You should add this new env var to the sample Dockerfiles (there are 2 now) with documentation. Try to turn the hardened setting on if possible. Here and eventually in studio.

d1donlydfink (changes requested):
Please add the new env var + docs to both Dockerfiles.
This will also need to happen in studio when the version flips there.

docs/agent_hocon_reference.md
Noravee: Done in 2f1d6698 — both sample Dockerfiles now document AGENT_TOOL_PATH_ONLY in their Server Hardening sections, and the hardening is turned ON in both. Turning it on surfaced one thing: math_guy.hocon deliberately used a fully-qualified class reference ('a test for fully-qualified class name' per its old comment), which is exactly the form strict mode refuses — so the sample container would have served a broken network. Migrated it to the relative `calculator.Calculator` form; the fully-qualified resolution path keeps its coverage in the unit tests (default-mode canary tests), and the hocon keeps a commented-out fully-qualified example for when the flag is off. Studio: agreed, to be done there when the neuro-san version flips.

Noravee:
## Change summary — Dockerfile hardening (2f1d6698)

Addressing the requested change:

- **Both sample Dockerfiles** (`neuro_san/deploy/Dockerfile`, `neuro_san/deploy/Dockerfile.py314t`): added `AGENT_TOOL_PATH_ONLY` to the Server Hardening section with full documentation — what it restricts (CodedTool `class` references, including toolbox shared coded tools, confined to the `AGENT_TOOL_PATH` hierarchy; fully-qualified references are never even imported), what it deliberately does not restrict (`llm_config.class`, middleware classes, langchain toolbox tool classes), a pointer to the docs, and migration guidance for client repos that riff on these Dockerfiles.
- **Hardening is ON in both sample containers** (`ENV AGENT_TOOL_PATH_ONLY="true"`), per "try to turn the hardened setting on if possible".
- **`math_guy.hocon`**: turning the flag on would have broken this bundled network — it deliberately used a fully-qualified class reference as "a test for fully-qualified class name", which is exactly the form strict mode refuses. Migrated to the relative `calculator.Calculator` form. The fully-qualified path keeps its test coverage in this PR's unit tests (default-mode canary resolution), and the hocon comment now teaches both forms, with the fully-qualified variant kept as a commented-out example usable when the flag is off.

Verification: both hocons parse; `calculator.Calculator` resolves through the exact hierarchy-walk machinery strict mode uses; all 43 activation tests pass.

Studio follow-up: the same env var + docs (and default-on if its example networks allow) to be applied in neuro-san-studio when the version flips there.

d1donlydfink (approved):

PR #1259, opened 2026-08-16
Fix duplicate descendant journaling events

### LINKED ISSUE #1256: Sub-agent events are journaled twice: parent agent's JournalingCallbackHandler receives descendant runs' events and mints phantom tool origins (-02)
## Summary

When an agent network has **two or more levels of internal LLM agents** (front man → sub-agent), every event inside the sub-agent — its LLM turns and its tool invocations — is journaled **twice** into the `/streaming_chat` stream and thinking files:

1. Once by the sub-agent's own `JournalingCallbackHandler` (correct origin), and
2. Once by the **parent** agent's handler, which reports the event from its own perspective: parent-level origin, and for `langchain_tool`-tagged tools (MCP + toolbox) a **freshly minted instantiation index**, producing phantom origins like `front_man.some-mcp-tool-02`.

Nothing actually runs twice — verified with a counting MCP server (1 physical `tools/call` per request while the stream reports two full invocation sequences). But the duplicated messages make users believe their agents are making excessive/duplicate calls, and the phantom instantiation index can land on the **legitimate** copy (the two handlers race), corrupting origins in otherwise-correct messages.

## Reproduction

Any 2-level network shows it. In-repo: run `music_nerd_pro_multi_agents` with `chat_filter: MAXIMAL` — the **front man's** thinking file contains `Invoking: accounting_tool`, although `accounting_tool` belongs to the `Accountant` sub-agent and the front man never calls it.

Minimal deterministic repro (no live LLM needed) in the collapsed section below: scripted tool-calling mock LLM + `front → child → MCP tool` network + counting FastMCP server. Output shows the full duplicated sequence:

```
AGENT | front                                  | Invoking: `get-member-details-tool` with:   <- leaked copy
AGENT | front.get-member-details-tool          | Received arguments:                          <- phantom origin
AGENT | front.child                            | Invoking: `get-member-details-tool` with:   <- real copy
AGENT | front.child.get-member-details-tool-02 | Received arguments:                          <- real copy got -02!
...
MCP server tool executions: 1
```

## What duplicates, by node type (all verified)

| Under the sub-agent | Duplicated? | Details |
|---|---|---|
| Sub-agent's own LLM turns | Yes | Text as AGENT message under the parent's origin; often masked by `write_message_if_next_not_dupe()` for single-turn sub-agents, visible otherwise |
| MCP tools | Fully | `Invoking:` / `Received arguments:` / result / `Got result:` all twice + phantom `-0N` origin (tagged `langchain_tool` → full tool-journal path incl. `add_spec_name_to_origin`) |
| Toolbox langchain tools | Fully | Same tag (`toolbox_factory.py`), same behavior |
| Coded tools | Partially | Only the `Invoking:` line duplicates; `Received arguments:`/`Got result:` are activation-journaled (not callback-driven) and untagged |
| External agents | `Invoking:` line only | Their internals run in a separate session/run tree |

Scales with depth: 3 agent levels → leaf events journaled 3×. Client-visible with `MAXIMAL` filter; always present in thinking files. Not related to middleware, LLM provider, or transport (reproduces with a mock LLM over a direct session).

## Root cause

The parent's `JournalingCallbackHandler` is **inheritable in the child's langchain run tree**. The child agent chain is invoked inline inside the parent's ToolNode coroutine, where the ambient `var_child_runnable_config` holds the parent's `AsyncCallbackManager`. `run_context_runnable.py`'s `merge_configs(ensure_config(), local_config)` then folds the child's handler **into** that parent manager (`merge_configs` manager+list → `add_handler(inherit=True)`), so the child graph runs with both handlers, and inheritable handlers fire for every nested run.

Note: this is **not** simply a regression from commit 33b4b3ce ("Fix trace parent/child propagation", needed for Langfuse nesting). With both halves of that commit experimentally neutralized, the duplication persists — langchain-core 1.x propagates the ambient parent config into the child anyway (e.g. `langchain_run_context.py` invokes the child chain with a config that has no `callbacks` key, so `ensure_config` fills it from the ambient contextvar). The inheritance itself is wanted for tracing; the bug is that the journaling handler processes events that belong to descendant agents.

Same mechanism, related symptom: the parent's `LlmTokenCallbackHandler` is inherited identically, which is why per-agent token totals are subtree-cumulative.

## Proposed fix

Scope-guard the handler rather than fight the attachment:

1. Each agent chain adds an inheritable tag `neuro-san:scope:<full_origin_name>` to its runnable config in `run_it()` (full origin names include instantiation indices, so they are unique per activation). Tags accumulate down the run tree, so the deepest scope tag on any event names the agent it belongs to.
2. `JournalingCallbackHandler.on_llm_end` / `on_tool_start` / `on_tool_end` return early when the event's tags contain a scope that **extends** the handler's own scope (`tag.startswith(my_scope + ".")`). Verified that all three callbacks receive `tags` and that config-level tags propagate into every nested run's events.

Properties: fail-open (no scope tags → current behavior), returns before `add_spec_name_to_origin()` so no phantom indices get minted, does not touch manager composition or `parent_run_id` so Langfuse/tracing nesting is unaffected. Token accounting deliberately left unchanged (cumulative totals may be relied upon — separate decision).

A regression test needs a tool-calling mock LLM (the existing `ChatMockLlm` only echoes); the scripted model below could be contributed alongside a two-level test network asserting exactly-once journaling.

<details>
<summary>Minimal repro assets</summary>

**scripted_llm.py** (put on PYTHONPATH):

```python
from typing import Any, Dict, List, Optional

from langchain_core.callbacks import CallbackManagerForLLMRun
from langchain_core.language_models import BaseChatModel
from langchain_core.messages import AIMessage, BaseMessage, ToolMessage
from langchain_core.outputs import ChatGeneration, ChatResult
from langchain_core.utils.function_calling import convert_to_openai_tool
from pydantic import ConfigDict, Field


class ScriptedLlm(BaseChatModel):
    model_name: str = Field(default="scripted", alias="model")
    model_config = ConfigDict(populate_by_name=True)

    def bind_tools(self, tools, **kwargs):
        return self.bind(tools=[convert_to_openai_tool(t) for t in tools])

    def _generate(self, messages: List[BaseMessage], stop: Optional[List[str]] = None,
                  run_manager: Optional[CallbackManagerForLLMRun] = None, **kwargs: Any) -> ChatResult:
        tools: List[Dict[str, Any]] = kwargs.get("tools") or []
        has_tool_result = any(isinstance(m, ToolMessage) for m in messages)
        if tools and not has_tool_result:
            name = tools[0]["function"]["name"]
            message = AIMessage(content="Let me call the tool now.",
                                tool_calls=[{"name": name, "args": {"inquiry": "hello"}, "id": "call_1"}])
        else:
            message = AIMessage(content=f"FINAL ANSWER (tools bound: {len(tools)})")
        return ChatResult(generations=[ChatGeneration(message=message)])

    @property
    def _llm_type(self) -> str:
        return "scripted"
```

**mcp_server.py** (counts physical calls):

```python
from mcp.server.fastmcp import FastMCP

mcp = FastMCP("member", host="127.0.0.1", port=3123)
CALLS = {"n": 0}

@mcp.tool(name="get-member-details-tool")
def get_member_details(inquiry: str) -> str:
    CALLS["n"] += 1
    print(f"### TOOL CALL #{CALLS['n']}: inquiry={inquiry!r}", flush=True)
    return "**Member Information**: plan A, active"

if __name__ == "__main__":
    mcp.run(transport="streamable-http")
```

**duo_mcp.hocon**:

```hocon
{
    "llm_config": { "class": "scripted_llm.ScriptedLlm", "model_name": "scripted" },
    "tools": [
        {
            "name": "front",
            "function": {"description": "Front man."},
            "instructions": "You are the front man.",
            "tools": ["child"],
        },
        {
            "name": "child",
            "function": {
                "description": "A child agent.",
                "parameters": {
                    "type": "object",
                    "properties": {"inquiry": {"type": "string", "description": "the ask"}},
                    "required": ["inquiry"],
                },
            },
            "instructions": "You are the child agent.",
            "tools": [{"url": "http://127.0.0.1:3123/mcp", "tools": ["get-member-details-tool"]}],
        },
    ],
}
```

Runner: `DirectAgentSessionFactory().create_session("<path>/duo_mcp.hocon")`, send `{"user_message": {"type": "HUMAN", "text": "hi"}, "chat_filter": {"chat_filter_type": "MAXIMAL"}}`, print each chunk's type/origin/text.

Environment: neuro-san main (0.6.94+), langchain-core 1.3.0, langchain 1.2.15, Python 3.13.
</details>

arnabnandy7:
## Summary

Fixes duplicate journaling of events from nested agents.

LangChain propagates callback handlers into descendant runs. As a result, both the descendant agent’s journaling handler and its ancestors’ handlers processed the same LLM and tool events. This produced duplicate streaming/thinking messages and could allocate phantom tool origins such as `-02`.

## Changes

- Track the nearest active journaling handler using a `ContextVar`.
- Ignore descendant events in ancestor handlers for:
  - LLM completion events
  - Tool start events
  - Tool end events
- Apply the ownership check before allocating tool origins.
- Preserve existing behavior when callbacks are invoked without an active agent scope.
- Add focused regression tests covering:
  - Descendant LLM events
  - Descendant tool events
  - Prevention of phantom origin allocation
  - Processing by the owning handler

This follows the existing callback-attribution pattern used by token accounting and does not change callback inheritance or tracing relationships.

## Testing

```bash
python -m pytest -v tests/neuro_san/internals/run_context/langchain/journaling/test_journaling_callback_handler.py
```

Fixes #1256


arnabnandy7:
@Noravee to review

d1donlydfink:
Beware: This is another fork PR that isn't running tests.
Also: I think @Noravee might have done something recently to address a similar problem.

arnabnandy7:
> Beware: This is another fork PR that isn't running tests.
> Also: I think @Noravee might have done something recently to address a similar problem.

I'll fix this.

Noravee (changes requested):
Thank you for taking this on — and for landing on the right pattern. I verified the approach thoroughly (mechanism trace through the ContextVar/task lineage plus empirical runs on 2-level MCP, coded-tool, parallel sub-agents, 3-level nesting, and direct external-agent topologies) and it produces exactly-once journaling with correct origins in all of them, with no legitimate messages lost. Mirroring the `llm_token_callback_var` / `_is_own_call()` attribution pattern from token accounting is exactly the right call. The requests below are about hardening and style, not the approach.

## Requested changes

1. **Don't reach for `callbacks[0]`** in `run_it()`. It works today because the `JournalingCallbackHandler` happens to be constructed first, but it breaks silently if the list order ever changes (e.g. another handler added ahead of it). Assign the handler to a named variable when constructing the list and pass that variable.

2. **Document the context-lineage invariant** in a comment on `journaling_callback_var`. The fix relies on every agent-chain invocation executing in a context descended from its `run_it()` scope. Every boundary in the current path preserves this (`copy_context()` in `count_tokens()`, `AsyncioExecutor.submit()`/`create_task()`, `call_soon_threadsafe`), but if a future refactor ever routes sub-agent invocation through a boundary that does not capture context (a bare thread, `run_in_executor` for the chain itself), the var arrives `None` and — because of the fail-open — the duplication silently returns with no test failing. A comment stating the invariant (and why fail-open is the right degradation) will save whoever trips this later.

3. **Decouple the scope from token counting.** `count_tokens_in_journaling_scope()` sets the journaling scope inside a helper whose name and home are about token counting — two unrelated concerns in one seam. Prefer a plain `try/finally` around the existing `count_tokens()` call in `run_it()`, or a small context manager mirroring `get_llm_token_callback()`, so the journaling scope stands on its own.

4. **Comment the change at the level this codebase does** — `LlmTokenCallbackHandler` and `LangChainTokenCounter.count_tokens()` are the reference. In particular: each `if not self._is_own_call(): return` guard deserves a line explaining why foreign events are skipped (the token handler comments every one of its equivalent checks), and the scope set-site should document the ordering constraint — the var must be set *before* `count_tokens()` runs, because `count_tokens()` captures the context via `copy_context()` when it creates the agent-chain task, and that captured context is how the value reaches every descendant run.

## Style

Please match the repo's conventions (the surrounding files are the reference):

- Type hints on everything, including local variables and return types.
- Docstrings with `:param ...:` and `:return:` lines — the scope helper and the new test helpers currently have one-line docstrings without them.
- One class per module, and no module-level functions. The module-level `journaling_scope()` contextmanager added to the test file violates this — move it into the test class (e.g. a `@staticmethod` or a small helper method), give it a return type hint, and document its param/return.
- Same for `_make_handler()`: return type hint plus `:return:` in the docstring.

## On testing

The unit-test style fixes above are all that's asked of this PR. An integration regression test (verifying the ContextVar actually propagates through `copy_context()`/task creation to real event-dispatch time, which the unit tests can't see) needs new shared test infrastructure, so I'll handle that in a follow-up PR after this merges — no action needed from you there.

Happy to re-review quickly once these are in — the core of the change is solid.

Noravee:
Two additions to the review above, plus a concrete recipe for the integration test so point 2 is actionable.

## 1. Comment density

Beyond the invariant comment in point 3, please comment the change at the level this codebase does — `LlmTokenCallbackHandler` and `LangChainTokenCounter.count_tokens()` are the reference. In particular:

- Each `if not self._is_own_call(): return` guard deserves a line explaining why foreign events are skipped (the token handler comments every one of its equivalent checks).
- The scope set-site should document the ordering constraint: the var must be set *before* `count_tokens()` runs, because `count_tokens()` captures the context via `copy_context()` when it creates the agent-chain task, and that captured context is how the value reaches every descendant run. This is the most load-bearing, least obvious line in the PR.

## 2. About the #1256 repro code

Treat the snippets in #1256 as a sketch, not paste-ready — they predate the style discussion and don't follow repo conventions (missing `:param:`/`:return:` docstrings, unhinted `bind_tools`, module-level functions). Here's what an in-repo version should look like.

### Tool-calling mock LLM

Add e.g. `neuro_san/test/llms/chat_tool_calling_mock_llm.py`, one class, modeled on the existing `ChatMockLlm` (same file layout, docstring style, hints). Behavior:

- Implement `bind_tools()` — `BaseChatModel` has no usable default. `return self.bind(tools=[convert_to_openai_tool(t) for t in tools])` (from `langchain_core.utils.function_calling`) is enough.
- In `_generate()`: if tools were bound (`kwargs.get("tools")`) and the messages contain no `ToolMessage` yet, return an `AIMessage` with `tool_calls=[{"name": <first tool>, "args": {...}, "id": "call_1"}]`; otherwise return a plain final answer. That yields a deterministic two-turn agent loop with zero network.

### Test network

A registry hocon (precedent: `chat_mock_llm_echo.hocon`, which registers the mock via `"llm_config": {"class": "neuro_san.test.llms.chat_mock_llm.ChatMockLlm"}`):

- `front` (front man): `instructions`, `tools: ["child"]`.
- `child`: `instructions`, and a `function` block **with at least one required string parameter** — the front man's LLM needs a schema to emit the tool call.
- Under `child`, a `langchain_tool`-tagged tool so the test covers the origin-minting path (`Received arguments:` / `Got result:` / instantiation index). A toolbox tool gets that tag automatically (`toolbox_factory.py`), and `ToolboxFactory.toolbox_info_file` can be set directly in the test for hermeticity — see `tests/neuro_san/session/test_direct_agent_session_connectivity.py` for that pattern, and note `DirectAgentSession` accepts a `toolbox_factory=` kwarg. A live MCP server is *not* needed — the guard sits in the callback handler, upstream of transport.

### Driver + assertions

Load the network with `AgentNetworkRestorer` and wire a session the way `DirectAgentSessionFactory.create_session()` does (or use the factory with `AGENT_MANIFEST_FILE` pointed at a test manifest fixture). Run `streaming_chat` with `{"chat_filter": {"chat_filter_type": "MAXIMAL"}}`, collect every chunk's `(origin path, text)`, then assert:

1. `"Invoking: \`<tool name>\` with:"` appears **exactly once**.
2. `"Received arguments:"` and `"Got result:"` each appear **exactly once**, with origin `front.child.<tool>`.
3. No origin anywhere has `instantiation_index > 1` (no `-02` suffixes) — this is the phantom-origin regression check and it fails on main today.
4. The child's LLM text appears exactly once, under `front.child` — not under `front`.
5. The front man's own messages are still present (`Invoking: \`child\``, the tool-result message, the final AI answer) — guards must not suppress legitimate events.

### Gotchas

- Any **new** directory under `tests/` needs an `__init__.py`, or the CI pylint gate silently skips it.
- Monkeypatch/set `AGENT_MANIFEST_FILE` (and `AGENT_TOOL_PATH` if you use a coded tool) inside the test so it stays hermetic.

With that in place the test pins the actual propagation path (ContextVar → `copy_context()` → task lineage → `ahandle_event`), which the current unit tests can't see.

arnabnandy7:
@Noravee addressed, please proceed

Noravee (dismissed):
Thank you for the quick and thorough turnaround — every point from the review was addressed, and promoting the scope helper into a production `scope()` context manager on the handler (so the unit tests exercise the real API) is a nicer design than what I asked for.

neuro_san/internals/run_context/langchain/core/run_context_runnable.py
```diff
@@ -148,11 +148,14 @@ async def run_it(self, inputs: Input) -> Output:
...
-        base_journal: Journal = self.invocation_context.get_journal()
-        origination: Origination = self.invocation_context.get_origination()
+        journaling_callback: JournalingCallbackHandler = JournalingCallbackHandler(
+            self.journal,
+            self.invocation_context.get_journal(),
+            self.origin,
+            self.invocation_context.get_origination()
+        )
         callbacks: List[BaseCallbackHandler] = [
-            JournalingCallbackHandler(self.journal, base_journal, parent_origin, origination)
```
d1donlydfink, line 155: Why are we switching the parent_origin for self.origin?
It is not obvious at this level why this change should happen.

neuro_san/internals/run_context/langchain/journaling/journaling_callback_handler.py
```diff
@@ -39,6 +44,18 @@
...
+# Each agent must execute its chain in a context descended from the scope entered
+# in RunContextRunnable.run_it(). The current path preserves that lineage through
+# copy_context(), task creation, and thread-safe task submission. If a future
+# execution boundary stops propagating context, this value becomes None. The
+# handler deliberately fails open in that case to preserve historical journaling
+# rather than silently discard legitimate messages, though duplicate descendant
+# events can then return.
+journaling_callback_var: ContextVar[Optional["JournalingCallbackHandler"]] = (
+    ContextVar("journaling_callback", default=None)
+)
```
d1donlydfink, line 56: I'm not a fan of ContextVars. It's a global cop-out.
How could this be done without context vars?

tests/neuro_san/internals/run_context/langchain/journaling/test_journaling_callback_handler.py
```diff
@@ -53,11 +61,75 @@ async def test_uses_tool_name_when_present(self):
...
         """A serialized tool with no name yields a diagnostic placeholder label
         instead of an empty "Invoking: ``"; the raw value is still reported."""
         handler, journal = self._make_handler()
         await handler.on_tool_start({}, "input", run_id=uuid4(), tags=[], inputs={})
         message = journal.write_message.call_args.args[0]
         assert message.content == "Invoking: `<unnamed tool>` with:"
         assert message.structure["invoked_agent_name"] is None
+
+
+class TestExclusiveAgentAttribution:
```
d1donlydfink: One class per file please.

d1donlydfink (changes requested):
* Why the switch from parent_origin to self.origin?
* How can this be done without using ContextVars?
* One class per file in tests.

neuro_san/internals/run_context/langchain/core/run_context_runnable.py
Noravee: There's no behavior change here — it's a diff illusion. The previous code had a local alias (`parent_origin = self.origin`) that was passed to the constructor; when the `callbacks[0]` fix introduced the named handler variable, the alias got inlined, so the constructor now receives `self.origin` directly — the same value it always received.

Agreed it reads worse without the alias though. @arnabnandy7 — please switch this constructor call to keyword arguments (`calling_agent_journal=self.journal, base_journal=..., parent_origin=self.origin, origination=...`) so the roles stay self-documenting.

neuro_san/internals/run_context/langchain/journaling/journaling_callback_handler.py
Noravee: Some context on why it's built this way: LangChain's callback system itself runs on ContextVars — config, tags, and handler propagation all ride `var_child_runnable_config` and friends — so we can't opt out of the mechanism, only choose whether to ride along with a var of our own. Riding along is the easier option, and it's the pattern we already established for token-accounting attribution in #1187 (`llm_token_callback_var` / `_is_own_call()` in `llm_token_callback_handler.py`); this PR deliberately mirrors it so journaling and token accounting scope the same way.

There is a ContextVar-free design for the ownership check (inheritable scope tags on the run tree, so the signal travels in the event payload instead of task-local state), but it only pays off if journaling and token accounting migrate together — and the token side keeps a var regardless, since LangChain's `register_configure_hook()` attachment API takes one. If we want to pursue that direction, we'll scope it in a separate design issue covering both handlers; for this PR we'll stay consistent with #1187.

No action needed here, @arnabnandy7.

tests/neuro_san/internals/run_context/langchain/journaling/test_journaling_callback_handler.py
Noravee: @arnabnandy7 — please move `TestExclusiveAgentAttribution` into its own module in the same directory (e.g. `test_journaling_exclusive_agent_attribution.py`), leaving the pre-existing `TestOnToolStartInvokingLabel` where it is. With that and the keyword-arguments tweak from the other thread, we should be good to merge.

neuro_san/internals/run_context/langchain/core/run_context_runnable.py
arnabnandy7: @d1donlydfink @Noravee updated the `JournalingCallbackHandler` construction to use explicit keyword arguments, including `parent_origin=self.origin`. This preserves the existing behavior while making each argument’s role clear.

tests/neuro_san/internals/run_context/langchain/journaling/test_journaling_callback_handler.py
arnabnandy7: @d1donlydfink @Noravee moved `TestExclusiveAgentAttribution` into `test_journaling_exclusive_agent_attribution.py`. The pre-existing `TestOnToolStartInvokingLabel` remains in the original module. Both focused test modules pass, and Pylint reports 10.00/10

arnabnandy7:
> * Why the switch from parent_origin to self.origin?

There wasn’t meant to be a behavior change here—`parent_origin` was just a local alias for `self.origin`. I’ve switched the constructor call to keyword arguments so `parent_origin=self.origin` is explicit and easier to follow.

> * How can this be done without using ContextVars?

We could avoid adding our own ContextVar by passing an agent-scope tag through LangChain’s run configuration and checking that tag in each callback. The tradeoff is that LangChain already uses ContextVars for callback propagation, and token accounting already follows this same ownership pattern. 

I kept the two systems consistent for this fix. A move to tag-based scoping would make more sense as a separate change covering both journaling and token accounting.

> * One class per file in tests.

I moved `TestExclusiveAgentAttribution` into its own test module and left the existing `TestOnToolStartInvokingLabel` where it was. The focused tests pass, and Pylint is clean.

cc @d1donlydfink @Noravee

d1donlydfink (approved):
I am now good if @Noravee is also good.

Thanks for your contribution @arnabnandy7! and thanks also for suffering our scrutiny.
You have been prolific lately.

Noravee (approved):
LGTM.

PR #1261, opened 2026-08-18
NS-1246 Add record/playback LLM proxy for neuro-san load & regression testing.

andreidenissov-cog:
## Summary
Adds a standalone, OpenAI-compatible HTTP proxy under
tests/record_playback_llm_server/ that records a neuro-san session against a
real LLM host and plays it back offline. It solves two problems with our
load/regression tests: they cost real tokens, and LLM responses are
non-deterministic. Recording once and replaying from disk makes those tests
free and repeatable.

It is wire-compatible with the OpenAI Chat Completions API, so any agent
network configured for class = "openai" is redirected at it with a single
openai_api_base change -- the same seam the existing tests/mock_llm_server uses.
This is test-only tooling; no production code is changed.

## Modes (selected with --mode)
- record: forward every request to the real host (endpoint + key from env
  vars), relay the real response back, and tee it into a cassette file,
  capturing wall-clock latency.
- playback: serve responses from the cassette by matching a canonical request
  signature. No network, no tokens. An unmatched request fails hard with 504,
  so tests surface gaps deterministically instead of silently faking data.
- hybrid: like playback, but a cache miss falls through to the real host (when
  configured), records the result into the current cassette, and returns it --
  a self-healing cassette.

## Additional capabilities
- Multi-response (--multi-response): record accumulates every distinct response
  for a request; playback serves them round-robin.
- Playback delay (--delay-mode none|recorded|fixed|random): an up-front
  per-response delay before serving a cache hit, emulating LLM latency.
  "recorded" uses each response's own recorded latency (first_byte_seconds for
  streams, latency_seconds for one-shot); in multi-response each variant uses
  its own.
- Latency capture: each recorded response stores wall-clock latency_seconds
  (and first_byte_seconds for streams).
- Success-only recording: non-2xx upstream responses (429/401/5xx) are relayed
  to the caller but never cached, so a transient failure cannot poison a
  cassette.
- Cassette cleaner (cleanup_cassette.py): a standalone tool that strips any
  non-2xx responses from an existing cassette (dropping single-response entries,
  trimming multi-response variants) to make an older cassette safe for playback.
- TLS verification anchored to the certifi CA bundle (matches the OpenAI/httpx
  SDKs) so recording against public hosts works on macOS python.org builds.
- Configurable upstream timeouts and max in-flight requests via env vars, to
  avoid spurious timeouts under load.

## How matching works
The response is non-deterministic but the request is a deterministic function of
the agent network and its inputs. In multi-turn flows each request embeds the
prior responses (assistant messages, tool_call ids, tool results); because
playback returns byte-identical recorded responses, downstream requests
reconstruct identically, so a sha256 of the canonicalized request body is a
stable key across a whole conversation.

## Configuration (record and hybrid modes), via environment variables
- RECORD_PLAYBACK_UPSTREAM_BASE_URL, RECORD_PLAYBACK_UPSTREAM_API_KEY
- RECORD_PLAYBACK_UPSTREAM_REQUEST_TIMEOUT_SECONDS,
  RECORD_PLAYBACK_UPSTREAM_CONNECT_TIMEOUT_SECONDS,
  RECORD_PLAYBACK_UPSTREAM_MAX_CLIENTS

## Layout
One class per file: RecordPlaybackLlmServer (CLI entry point), ProxyState,
ProxyHandler (shared record/playback/hybrid logic) with ChatCompletionsHandler
and ModelsHandler endpoints, HealthHandler, UpstreamClient, Cassette,
PlaybackDelay, RequestCanonicalizer, and the CassetteCleaner tool. A README
documents usage for every mode, the cleaner, and the cassette format.

## Testing
- flake8 clean and pylint 10.00/10 on the whole package.
- Committed pytest at tests/record_playback_llm_server/test_record_playback_llm_server.py
  (21 tests) runs in the default unit suite with no server or LLM key required.
  It covers: request canonicalization key stability; Cassette round-trip and
  multi-response de-duplication; all PlaybackDelay policies (including stream
  first-byte vs one-shot latency); CassetteCleaner drop/trim rules; and
  in-process record/playback/hybrid integration against a fake upstream --
  json + stream byte-identical replay with no upstream contact on hits, hard 504
  on a playback miss, multi-response round-robin, hybrid record-on-miss then hit,
  and non-2xx relayed but not recorded.

## Related issue
NS-1246

## Notes for reviewers
- Test-only: everything lives under tests/, no neuro_san/ production code is
  modified.
- The proxy is OpenAI-wire only (hosts reachable via an OpenAI-compatible
  base_url); native Anthropic/Bedrock/Gemini wire formats are out of scope.

### DIFF
--- a/tests/record_playback_llm_server/README.md
+++ b/tests/record_playback_llm_server/README.md
@@ -1,0 +1,175 @@
+# Record / Playback LLM Server
+
+A standalone OpenAI-compatible HTTP proxy for **recording** a neuro-san
+session against a real external LLM host and **playing it back** offline.
+
+Two goals:
+
+1. **Free** — record once against the paid host, then replay from disk with
+   zero token cost.
+2. **Repeatable** — replaying byte-identical recorded responses removes the
+   random nature of LLM output, so load and regression tests become
+   deterministic.
+
+It is wire-compatible with the OpenAI Chat Completions API, so any neuro-san
+agent network configured for `class = "openai"` can be redirected at it with a
+single `openai_api_base` change — the same seam the sibling `mock_llm_server`
+uses.
+
+There is **no synthetic mode** here: every response is either forwarded from
+the real host (record) or served from a prior recording (playback).
+
+## Modes
+
+| Mode | What it does |
+|---|---|
+| `record` | Forwards each request to the external LLM host (endpoint + key from env vars), relays the real response back to neuro-san, and tees it into the cassette file. |
+| `playback` | Serves responses from the cassette by matching the canonical request signature. No network, no tokens. An unmatched request fails hard with HTTP 504. |
+
+## External host configuration (record mode)
+
+The external LLM host is configured **only** via environment variables:
+
+| Variable | Purpose |
+|---|---|
+| `RECORD_PLAYBACK_UPSTREAM_BASE_URL` | Base URL of the real host, including the version segment. e.g. `https://api.openai.com/v1`. Required in record mode. |
+| `RECORD_PLAYBACK_UPSTREAM_API_KEY` | Bearer credential for that host. Optional (a warning is logged if absent, for hosts that need no auth). |
+| `RECORD_PLAYBACK_UPSTREAM_REQUEST_TIMEOUT_SECONDS` | Whole-request timeout when forwarding to the real host. Default `600`. |
+| `RECORD_PLAYBACK_UPSTREAM_CONNECT_TIMEOUT_SECONDS` | Connection timeout when forwarding to the real host. Default `30`. |
+| `RECORD_PLAYBACK_UPSTREAM_MAX_CLIENTS` | Maximum simultaneous in-flight requests to the real host before further requests queue. Default `100`. |
+
+The incoming request's own `openai_api_key` (pointed at this proxy) is ignored
+and replaced with the real credential above.
+
+All timeout/limit variables accept a positive number; an invalid or
+non-positive value logs a warning and falls back to the default. They apply in
+**record mode only** (playback never contacts the network).
+
+> **If you are seeing timeout errors while recording under load**, raising
+> `RECORD_PLAYBACK_UPSTREAM_REQUEST_TIMEOUT_SECONDS` alone may not be enough:
+> Tornado's client serves only `max_clients` requests at once and queues the
+> rest, and time spent queued counts against the request timeout. Raise
+> `RECORD_PLAYBACK_UPSTREAM_MAX_CLIENTS` to at least your peak concurrency.
+
+## Running
+
+The package is runnable as a module:
+
+```bash
+export PYTHONPATH=$(pwd)
+
+# 1) Record a session against the real host (costs tokens, once).
+export RECORD_PLAYBACK_UPSTREAM_BASE_URL="https://api.openai.com/v1"
+export RECORD_PLAYBACK_UPSTREAM_API_KEY="sk-..."
+python -m tests.record_playback_llm_server.record_playback_llm_server \
+    --mode record --port 8899 --cassette ./session.cassette.json
+
+# ... run your load/integration test against neuro-san, then Ctrl-C ...
+
+# 2) Replay it forever, for free, deterministically (no env vars needed).
+python -m tests.record_playback_llm_server.record_playback_llm_server \
+    --mode playback --port 8899 --cassette ./session.cassette.json
+```
+
+### CLI options
+
+| Flag | Default | Description |
+|---|---|---|
+| `--host` | `localhost` | Bind interface. |
+| `--port` | `8899` | Bind port (differs from the mock's 8888 so both can run at once). |
+| `--mode` | _(required)_ | `record` or `playback`. |
+| `--cassette` | `./llm_cassette.json` | Path to the cassette JSON file. |
+| `--stream-replay-delay` | `0.0` | Seconds between streamed SSE frames during playback, to emulate inter-token cadence. `0` replays as fast as possible. |
+
+## Pointing a neuro-san agent network at the proxy
+
+```hocon
+llm_config {
+    class = "openai"
+    model_name = "gpt-4.1"
+    openai_api_base = "http://localhost:8899/v1"
+    openai_api_key = "not-needed"
+}
+```
+
+- `openai_api_base` must include the `/v1` path segment.
+- Use the **same** agent network and inputs for record and playback — the
+  match key is derived from the request, so a changed prompt is a new
+  (unrecorded) request.
+
+## How matching works
+
+The response of an LLM is non-deterministic, but the **request** is a
+deterministic function of the agent network plus its inputs. In a multi-turn
+agent flow, each request embeds the previous responses (assistant messages,
+`tool_call` ids, tool results). Because playback returns **byte-identical**
+recorded responses, every downstream request reconstructs identically — so a
+hash of the canonicalized request body is a stable key across the whole
+conversation.
+
+Canonicalization (`RequestCanonicalizer`):
+
+- Parses the JSON body and re-serializes it with **sorted keys**, so
+  incidental key ordering does not change the key.
+- Keeps the `stream` flag as part of the key (a streamed request and a
+  one-shot request map to different recorded responses).
+- Drops any fields listed in `VOLATILE_BODY_KEYS` (empty by default; extend it
+  if a client is found to inject a per-run random value into the body).
+
+The key is `sha256(f"{METHOD} {path}\n{canonical_body}")`.
+
+## Cassette format
+
+An ordered, human-diffable JSON array — commit it to git as a test fixture:
+
+```json
+{
+  "version": 1,
+  "entries": [
+    {
+      "key": "<sha256>",
+      "method": "POST",
+      "path": "/chat/completions",
+      "request": "POST /chat/completions\n{...canonical body...}",
+      "response": {
+        "kind": "json",
+        "status": 200,
+        "body": { "id": "chatcmpl-...", "choices": [ ... ] }
+      }
+    }
+  ]
+}
+```
+
+A streamed response is stored with `"kind": "stream"` and a `"chunks"` array
+of the individual SSE frames (`data: {...}\n\n`, terminated by
+`data: [DONE]\n\n`), re-emitted verbatim on playback.
+
+## Internal layout
+
+One class per file:
+
+| File | Class | Responsibility |
+|---|---|---|
+| `record_playback_llm_server.py` | `RecordPlaybackLlmServer` | CLI entry point; reads env config, builds the app, runs the loop. |
+| `proxy_state.py` | `ProxyState` | Process-wide state: mode, cassette, upstream client, replay pacing. |
+| `proxy_handler.py` | `ProxyHandler` | Shared record/playback logic for both proxied endpoints (base class). |
+| `chat_completions_handler.py` | `ChatCompletionsHandler` | `POST /v1/chat/completions`. |
+| `models_handler.py` | `ModelsHandler` | `GET /v1/models`. |
+| `health_handler.py` | `HealthHandler` | `GET /healthz`. |
+| `upstream_client.py` | `UpstreamClient` | Async HTTP client to the real host (record mode), one-shot and streaming. |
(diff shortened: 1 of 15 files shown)


tests/record_playback_llm_server/proxy_handler.py
```diff
@@ -0,0 +1,383 @@
...
+        first_byte: List[Optional[float]] = [None]
+
+        def on_chunk(chunk: bytes) -> None:
+            if first_byte[0] is None:
+                first_byte[0] = time.monotonic() - started
+            accumulated.extend(chunk)
+            self.write(bytes(chunk))
+            # Flush must happen on the event loop; schedule it as a coroutine so
+            # the caller (neuro-san) sees tokens progressively during recording.
+            tornado.ioloop.IOLoop.current().spawn_callback(self._safe_flush)
```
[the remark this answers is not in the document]
andreidenissov-cog: Fixed.

tests/record_playback_llm_server/test_record_playback_llm_server.py
```diff
@@ -0,0 +1,376 @@
...
+        rec_url = start_app(RecordPlaybackLlmServer.build_app(rec))
+        code, body = await _post(rec_url + CHAT_PATH, PAYLOAD)
+        assert code == 200 and _content(body) == "resp-1"
+        assert box["n"] == 1
+
+        pb = ProxyState(mode=ProxyState.MODE_PLAYBACK, cassette=Cassette(path))
+        pb_url = start_app(RecordPlaybackLlmServer.build_app(pb))
+        code2, body2 = await _post(pb_url + CHAT_PATH, PAYLOAD)
+        assert code2 == 200 and json.loads(body2) == json.loads(body)   # byte-identical replay
+        assert box["n"] == 1                                            # upstream NOT hit on playback
```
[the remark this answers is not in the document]
andreidenissov-cog: Comment changed.

tests/record_playback_llm_server/proxy_handler.py
```diff
@@ -0,0 +1,384 @@
...
+
+        try:
+            response: tornado.httpclient.HTTPResponse = await self.state.upstream.fetch_stream(
+                self.upstream_path, method, body_bytes, on_chunk)
+        except (OSError, tornado.httpclient.HTTPError) as exception:
+            self._fail_upstream(exception)
+            return
```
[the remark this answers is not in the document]
andreidenissov-cog: Fixed.

tests/record_playback_llm_server/cassette.py
```diff
@@ -0,0 +1,155 @@
...
+        """
+        entry: Optional[Dict[str, Any]] = self.entries.get(key)
+        if entry is None:
+            entry = dict(meta)
+            entry["key"] = key
+            entry["responses"] = [response]
+            self.entries[key] = entry
+            self._order.append(key)
+        else:
+            responses: List[Dict[str, Any]] = entry.setdefault("responses", [])
```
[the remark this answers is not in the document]
andreidenissov-cog: This is wrong - due to badly formulated comment.

tests/record_playback_llm_server/proxy_handler.py
```diff
@@ -0,0 +1,470 @@
...
+        (LLM error bodies are JSON, not SSE).
+        """
+        code: int = status if status is not None else 200
+        if 200 <= code < 300:
+            self.set_header("Content-Type", "text/event-stream")
+            self.set_header("Cache-Control", "no-cache")
+            self.set_header("X-Accel-Buffering", "no")
+        else:
+            self.set_status(code)
+            self.set_header("Content-Type", content_type or "application/json")
```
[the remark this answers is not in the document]
andreidenissov-cog: Fixed.

tests/record_playback_llm_server/cassette.py
```diff
@@ -0,0 +1,155 @@
...
```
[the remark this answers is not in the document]
andreidenissov-cog: We don't use "response" key. All responses are held in "responses".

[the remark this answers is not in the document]
andreidenissov-cog: Already commented.

tests/record_playback_llm_server/record_playback_llm_server.py
```diff
@@ -0,0 +1,279 @@
...
+        logging.basicConfig(level=logging.INFO, format="%(asctime)s %(levelname)s %(message)s")
+        args = cls.parse_args(argv)
+        try:
+            asyncio.run(cls.run(args))
+        except KeyboardInterrupt:
+            logging.info("record-playback-llm shutting down")
+
+
+if __name__ == "__main__":
+    RecordPlaybackLlmServer.main()
```
d1donlydfink, line 279: Between this and Vince's load-test, we might need a tests/apps directory to distinguish between actual test and non-shipping test apps.  Can be a refactor for later.

tests/record_playback_llm_server/cleanup_cassette.py
```diff
@@ -0,0 +1,206 @@
...
+
+    @classmethod
+    def main(cls, argv: Optional[List[str]] = None) -> None:
+        """CLI entry point: configure logging, parse args, run."""
+        logging.basicConfig(level=logging.INFO, format="%(asctime)s %(levelname)s %(message)s")
+        cls.run(cls.parse_args(argv))
+
+
+if __name__ == "__main__":
+    CassetteCleaner.main()
```
d1donlydfink, line 206: Difficult to distinguish which source is an app's entrypoint and what is not.  
Seems like there are at least 2 apps in this folder.
Strongly consider a reorg of the files here to make that much clearer.
Doesn't have to be now (so @vince-leaf can use it) but soon, please.

tests/record_playback_llm_server/test_record_playback_llm_server.py
```diff
@@ -0,0 +1,391 @@
...
+
+
+def _content(body: bytes) -> str:
+    """Extract the assistant text from a one-shot chat completion body."""
+    return json.loads(body)["choices"][0]["message"]["content"]
+
+
+def _upstream_client(base_url: str) -> UpstreamClient:
+    """Build an UpstreamClient pointed at a fake upstream's /v1 base."""
+    return UpstreamClient(base_url=f"{base_url}/v1", api_key=None)
```
d1donlydfink, line 156: Oh boy. All this stuff above just to make a test.

This is all the more reason to reorg these files into different folders.
Weird in our world to have tests right next to apps.
If these tests were in a separate dir it would read better to have the ancillary test classes in their own files like we do elsewhere.  I get that at the moment, that would be confusing. Have to undo one confusing to undo another.

tests/record_playback_llm_server/test_record_playback_llm_server.py
```diff
@@ -0,0 +1,391 @@
...
+
+    def test_clean_is_idempotent(self):
+        """Cleaning an already-clean cassette removes nothing."""
+        entries = [{"key": "k1", "response": {"kind": "json", "status": 200}}]
+        once, _ = CassetteCleaner.clean_entries(entries)
+        twice, stats = CassetteCleaner.clean_entries(once)
+        assert len(twice) == 1 and stats["dropped_failure"] == 0 and stats["variants_removed"] == 0
+
+
+class TestProxyIntegration:
```
d1donlydfink, line 273: Yeah WTF. Multiple test classes in a single file will have me asking "Where does this live again?" when they fail. File reorg is your next order of business.

tests/record_playback_llm_server/README.md
```diff
@@ -0,0 +1,288 @@
...
+
+## Modes
+
+Selected with `--mode`:
+
+| Mode | What it does |
+|---|---|
+| `record` | Forwards each request to the external LLM host (endpoint + key from env vars), relays the real response back to neuro-san, and tees it into the cassette file (capturing wall-clock latency). |
+| `playback` | Serves responses from the cassette by matching the canonical request signature. No network, no tokens. An unmatched request fails hard with HTTP 504. |
+| `hybrid` | Like `playback`, but a cache **miss** falls through to the real host (when an upstream is configured via the env vars below), records the result into the **current** cassette, and returns it — a self-healing cassette that fills in gaps on demand. With no upstream configured, a miss behaves like plain playback (504). |
```
d1donlydfink, line 27: Maybe another aspect of "hybrid" to make usage simpler:
If multiple requests for the same input hit at the same time, delay them all until a first response comes back from backing LLMs.   Then you wouldn't really need modes at all.

tests/record_playback_llm_server/README.md
```diff
@@ -0,0 +1,288 @@
...
+### Cleaning a cassette
+
+`cleanup_cassette.py` repairs a cassette left with failure responses by an
+interrupted or throttled recording session, making it safe for playback/hybrid:
+
+- Single-response entries with a non-2xx status are dropped.
+- Multi-response entries have their non-2xx variants removed; the entry is
+  dropped only if no successful variant remains.
+- All other data (requests, keys, latencies, unknown fields, `version`) is
+  preserved.
```
d1donlydfink, line 45: With expanded hybrid approach described above, it's possible that in most circumstances touching the filesystem would never be needed.

tests/record_playback_llm_server/README.md
```diff
@@ -0,0 +1,288 @@
...
+agent network configured for `class = "openai"` can be redirected at it with a
+single `openai_api_base` change — the same seam the sibling `mock_llm_server`
+uses.
+
+## Modes
+
+Selected with `--mode`:
+
+| Mode | What it does |
+|---|---|
```
d1donlydfink, line 24: Which of these is the default? Seems like hybrid would make a decent default.

tests/record_playback_llm_server/cleanup_cassette.py
```diff
@@ -0,0 +1,206 @@
...
+            with open(path, "r", encoding="utf-8") as cassette_file:
+                data: Any = json.load(cassette_file)
+        except (OSError, json.JSONDecodeError) as exception:
+            raise SystemExit(f"failed to read cassette {path}: {exception}") from exception
+        if not isinstance(data, dict) or not isinstance(data.get("entries"), list):
+            raise SystemExit(f"{path} is not a valid cassette (expected an object with an 'entries' list)")
+        return data
+
+    @staticmethod
+    def write(path: str, data: Dict[str, Any]) -> None:
```
d1donlydfink, line 137: If all these methods weren't static path could be a member variable to a class and you wouldn't have to pass it around everywhere.

tests/record_playback_llm_server/proxy_handler.py
```diff
@@ -0,0 +1,474 @@
...
+        """Parse a response body as JSON for diff-friendly storage; fall back to text."""
+        if not raw_body:
+            return ""
+        try:
+            return json.loads(raw_body)
+        except (json.JSONDecodeError, ValueError):
+            return raw_body.decode("utf-8", errors="replace")
+
+    @staticmethod
+    def _split_sse_frame_bytes(raw: bytes) -> List[bytes]:
```
d1donlydfink, line 423: Perhaps define SSE somewhere.

tests/record_playback_llm_server/proxy_handler.py
```diff
@@ -0,0 +1,474 @@
...
+    cassette, and returns it -- a self-healing cassette.
+
+    With multi-response enabled, RECORD (and hybrid-mode record-on-miss)
+    accumulate every distinct response for a request and PLAYBACK serves them
+    round-robin; otherwise a request maps to a single response.
+
+    On a genuine playback miss (playback mode, or hybrid mode with no
+    upstream) the handler fails hard with HTTP 504 so a load/regression test
+    surfaces the gap deterministically rather than silently faking data.
+    """
```
d1donlydfink, line 64: The code in this class could sure use some comments so we know what's going on.
Especially in the string parsing/manipulating bits.
Stench of non-vetted vibing implies you might not understand what is going on in here yourself.

d1donlydfink (approved):
What you have here is actually very cool.
Approving for now to get this ball rolling so @vince-leaf and others can use it and save us $.
Please schedule some work for reorganization soon per PR comments.
Also, file that needs the most love is ProxyHandler. Very little idea what is going on in there - smells like vibed slop.

Other PR comments to handler later as well. File tickets if needed.

PR #1264, opened 2026-08-18
Switch ollama smoke test model from qwen3:8b to gemma4:e4b

### LINKED ISSUE #1249: Nightly smoke test music_nerd_pro_llm_ollama times out: qwen3:8b re-calls Accountant in an infinite loop since #1226
## Symptom

The nightly Smoke Tests workflow has failed every run since 2026-08-11 on:

`TestSmokeTestHocons::test_hocon_with_server_non_default_llm_0_music_nerd_pro_llm_ollama_combination_responses_with_history_http` — `leaf_common.time.timeout_reached_exception.TimeoutReachedException`

- Last pass: [run 31387105707](https://github.com/cognizant-ai-lab/neuro-san/actions/runs/31387105707) (Aug 10, test took 109s)
- First fail: [run 31490304734](https://github.com/cognizant-ai-lab/neuro-san/actions/runs/31490304734) (Aug 11, hit the 360s test timeout; server-side request ran 601s = the network's `max_execution_seconds` cap)

## Root cause

#1226 changed `LangChainOpenAIFunctionTool._arun()` to return `the_answer.content` instead of the `AgentToolResultMessage` object. The tool result the calling LLM sees went from the pydantic repr wrapper (`content="{'running_cost': 3.0}" additional_kwargs={} ...`) to the bare value (`{'running_cost': 3.0}`).

For qwen3:8b this is fatal: the Accountant's function description ("pass the current running cost to the API to get the updated running cost") plus a result that mirrors the argument it just passed reads as an **iterative protocol**. The model echoes the returned cost back as the next call's argument (0 → 3 → 6 → 9 → …) forever. The failing runs' agent.log shows `Calling class accountant.Accountant` every ~10s until the 600s execution cap. The old repr wrapper only masked this by accident; #1226's release-note caveat ("networks whose prompts were accidentally calibrated against the repr format may shift behavior") is exactly what happened.

Environment is ruled out: the Aug 10 (pass) and Aug 11 (fail) runs used the same `ollama/ollama:latest` image digest, the same qwen3:8b model blob, and the same relevant pip package versions. Only the repo SHA differed, and the only change in the agent tool-call path between those SHAs is #1226.

## What does NOT fix it (verified by direct `/api/chat` A/B against qwen3:8b)

- Returning JSON instead of Python repr from the coded tool (`json.dumps`)
- Renaming the result key so it doesn't mirror the argument name
- Rewriting the Accountant tool description ("call exactly once, result is final")
- A stop instruction in the frontman prompt *alone* — mostly works but flaky (4/5 seeds at default temperature; one interactive run looped 10× before escaping)

## Fix (verified 20/20 seeds in isolation, plus end-to-end via HTTP server and agent_cli)

Two-part, scoped entirely to the example network:

1. Append a stop instruction to the frontman `instructions` in `neuro_san/registries/music_nerd_pro_llm_ollama.hocon`: *"The Accountant tool result is final. After you receive it, do NOT call any tools again. Immediately respond with the final JSON object."*
2. Add a network-scoped `neuro_san/coded_tools/music_nerd_pro_llm_ollama/accountant.py` (standard `<AGENT_TOOL_PATH>/<network>/` resolution) that subclasses the global Accountant and appends an in-band stop cue to the result dict: `"note": "This value is final. Do not call the Accountant again. Respond with your final JSON answer now."`

The in-band cue is the load-bearing part: it sits in the tool result, which is where the model is looking when it decides whether to call again.

With both in place the test scenario passes: one Accountant call per question, `running_cost` 3.0 / 6.0, answers contain Beatles / Liverpool.

## Longer-term consideration

Nothing in the framework bounds tool calls per turn, so one bad model decision becomes a 600s runaway. A configurable per-agent tool-call limit would turn this class of failure into a bounded, diagnosable error.

Noravee:
## Problem

The nightly Smoke Tests workflow has failed every run since 2026-08-11 on the ollama test case (`music_nerd_pro_llm_ollama`). Root cause analysis in #1249: after #1226 changed sub-agent/tool results from the pydantic repr wrapper to clean content, qwen3:8b re-calls the Accountant tool in an infinite loop — it misreads the tool exchange ("pass the current running cost, get the updated one") as an iterative protocol and echoes the returned cost back as the next call's argument until the network's 600s execution cap kills the request.

## Fix

Swap the smoke-test model to `gemma4:e4b`, which handles the same network unmodified: exactly one Accountant call per question. Prompt-level and tool-result-level workarounds for qwen3:8b were prototyped and proved either flaky or more invasive (details in #1249).

Changes:
- `music_nerd_pro_llm_ollama.hocon`: `model_name` → `gemma4:e4b`
- `default_llm_info.hocon`: new `gemma4:e4b` entry (128K context per the [Gemma 4 model card](https://ai.google.dev/gemma/docs/core/model_card_4); `max_output_tokens` not published for Gemma 4, assumed 8192 from Gemma 3)
- `set_ollama_model.sh`: model name hoisted into a single `OLLAMA_MODEL` variable (was hardcoded in two places) and set to `gemma4:e4b`; shellcheck clean

## Caveat: model size

`gemma4:e4b` is **9.6 GB — almost double qwen3:8b's 5.2 GB**. The model is re-pulled on every nightly run, so expect a longer "Set Ollama model" step, and the runner's ~16 GiB RAM has less headroom while the model is loaded. Offsetting this, e4b is a ~4.5B-effective-parameter model, so inference is substantially faster: the test completes in ~21s locally vs qwen3:8b's 109s on the CI CPU runner even when it passed.

## Testing

- 10/10 consecutive full two-turn scenarios over HTTP (one Accountant call per question, Beatles/`running_cost: 3.0`, Liverpool/`running_cost: 6.0`)
- The actual pytest smoke test passes repeatedly against a local server + local Ollama
- `default_llm_info.hocon` verified to parse via pyhocon with the new entry resolving correctly

After merge, the Smoke Tests workflow can be triggered manually via workflow_dispatch to confirm on CI without waiting for the nightly.

Fixes #1249


neuro_san/internals/run_context/langchain/llms/default_llm_info.hocon
```diff
@@ -1035,6 +1035,18 @@
...
+    "gemma4:e4b": {
+        "class": "ollama",
+        "model_info_url": "https://ollama.com/library/gemma4",
+        "modalities": {
+            "input": [ "text", "image", "audio" ],
+            "output": [ "text" ],
+        },
+        "capabilities": [ "tools" ],
+        "context_window_size": 128000,  # https://ai.google.dev/gemma/docs/core/model_card_4 (E4B: 128K tokens)
+        "max_output_tokens": 8192,      # Not published for Gemma 4; assumed from Gemma 3, which documents 8192
```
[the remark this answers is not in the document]
Noravee: This field must be a number (or `null`), not a string: `DefaultLlmFactory.get_max_prompt_tokens()` multiplies it by `prompt_token_fraction` ([default_llm_factory.py#L547-L559](https://github.com/cognizant-ai-lab/neuro-san/blob/main/neuro_san/internals/run_context/langchain/llms/default_llm_factory.py#L547-L559)), so `"unknown"` would raise `TypeError` at request time. The guard there only handles `null`, which is reserved for models whose limit is truly indeterminate per-request (e.g. OpenRouter meta-routers) — using `null` here would silently disable `prompt_token_fraction` for this model.

Since Gemma 4 doesn't publish an output limit, we assume Gemma 3's documented 8192 (noted in the inline comment). The value only takes effect when a network opts into `prompt_token_fraction`, and the worst case of a too-low assumption is a conservative cap on generation length — not rejected requests or unexpected truncation.

neuro_san/internals/run_context/langchain/llms/default_llm_info.hocon
```diff
@@ -1035,6 +1035,18 @@
         "max_output_tokens": 32768,
         "knowledge_cutoff": "unknown",
     },
+    "gemma4:e4b": {
+        "class": "ollama",
+        "model_info_url": "https://ollama.com/library/gemma4",
+        "modalities": {
+            "input": [ "text", "image", "audio" ],
```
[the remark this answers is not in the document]
Noravee: It does support audio. See the doc in comment.

neuro_san/registries/music_nerd_pro_llm_ollama.hocon
```diff
@@ -29,7 +29,7 @@
         ]
     },
     "llm_config": {
-        "model_name": "qwen3:8b",
+        "model_name": "gemma4:e4b",
```
Noravee, line 32: Change model of `music_nerd_pro_llm_ollama` from `qwen3:8b` to `gemma4:e4b`.

neuro_san/internals/run_context/langchain/llms/default_llm_info.hocon
```diff
@@ -1035,6 +1035,18 @@
         "max_output_tokens": 32768,
         "knowledge_cutoff": "unknown",
     },
+    "gemma4:e4b": {
```
Noravee, line 1038: Add `gemma4:e4b` in the default llm info file.

build_scripts/set_ollama_model.sh
```diff
@@ -1,6 +1,10 @@
 #!/bin/bash
   
-# Script set the ollama model by preforming a pull command. It is expected that Ollama Service is running! 
+# Script set the ollama model by performing a pull command. It is expected that Ollama Service is running!
+
+# Ollama model used by the smoke tests.
+# Must match the model_name in neuro_san/registries/music_nerd_pro_llm_ollama.hocon
+OLLAMA_MODEL="gemma4:e4b"
```
Noravee, line 7: Change CI workflow to use `gemma4:e4b`.

vince-leaf: @Noravee, how much more time does it take to load this model, gemma4:e4b?

Noravee: Download scales with size: qwen3:8b's 5.2 GB took ~15s on the runner (Aug 11 log), so gemma4:e4b's 9.6 GB should take ~30s — about +15s per nightly. Loading into memory measured +1.5s (5.0s → 6.5s). Both are one-time per run and dwarfed by the inference speedup: the e4b model runs the actual test several times faster than qwen3:8b on CPU (which also spent 80s+ of the warmup step generating thinking tokens), so total job time should go down, not up.

vince-leaf (approved):
LGTM, but pending any other team member's objective.

d1donlydfink (approved):

Issue #1225, opened 2026-08-09
Phase 4: Tool block returns, answer surfacing, mcp-adapters unpin, payload knobs


Noravee:
Part of #1221. Depends on Phases 1-3. **Each item independently shippable/optional.**

Rounds out block support for tools and answers, lifts the `langchain-mcp-adapters` pin, and adds payload-control knobs.

### Work items

- [ ] CodedTool block returns: `abstract_class_activation.py:192-194` — returns that `looks_like_blocks` → `AIMessage(content=blocks)` instead of `f"{retval}"` (documented convention; raw bytes without a mime type stay str-ified — refuse to guess). `parse_tool_output` (`langchain_run_context.py:451-541`) list-content guard around `parse_tool_chat_list_string`.
- [ ] `LangChainOpenAIFunctionTool._arun` block-list return path (if the Phase 1 str-case commit was held, ship the list-case now — it is what multimodal sub-agent results require).
- [ ] Answer surfacing: `AnswerMessageFilter` (`answer_message_filter.py:48-53`) accepts non-empty `content_blocks` as a third final-answer criterion (per its own "more options in the future" comment); final `AgentFrameworkMessage` carries answer blocks (`data_driven_chat_session.py:261-285`; its content is already Union-typed).
- [ ] External (server-to-server) agents: `external_activation.py:214,:241` builds the sub-agent answer as flattened `get_compiled_answer()` text → `AIMessage(content=answer)`. Wire it to consume `processor.get_content_blocks()` when present, with the same history-safety projection as tool results (no data blocks in assistant-role history). Until this lands, "lossless multi-hop" applies to the journal stream only. Optional: `gather_input` block passthrough for server-to-server multimodal input.
- [ ] `langchain-mcp-adapters` unpin: lift to `>=0.2.0,<0.3.0` with the coupled `mcp` floor re-verified (the pins are documented as coupled in requirements.txt) — gated on an MCP multi-part-content integration test: image-returning MCP tool, sub-agent called twice in one request, follow-up turn via chat_context, exercising the Phase 1 history-safety projection in `OriginatingJournal`.
- [ ] Payload knobs:
  - `chat_context_elide_block_data` server knob — strips base64 from outbound chat_context data blocks, marking `extras.neuro_san.elided=true` (the Phase 2 restore-side rule already drops sourceless data blocks, so elided contexts restore safely).
[...]

Issue #1224, opened 2026-08-09
Phase 3: Multimodal inbound (mime_data + content_blocks)


Noravee:
Part of #1221. Depends on Phases 1-2. **Shippable alone.**

Lets clients send base64 images/files/audio. Two accepted inbound shapes: the already-published-but-never-read `mime_data` field (simple clients) and full `content_blocks` (power clients). Outbound stays canonical (`content_blocks` only; `mime_data` is never emitted).

### Work items

- [ ] Session layer — the single choke point where inbound multimodal dies today (`async_direct_agent_session.py:181`, sync twin `direct_agent_session.py:171`): read the whole `user_message`; `user_input = blocks_from_chat_message(user_message) or text`. Precedence: `content_blocks` wins when both present.
  - `mime_data` mapping: `[{"type":"text","text":<text>}]` (if non-empty) + per entry `{"mime_type": m, "mime_bytes": <b64 str>}` → `{"type": image|audio|video|file by mime prefix, "base64": ..., "mime_type": m}` (`text/plain` → `text-plain` block). Matches the existing OpenAPI `string/format:byte` contract in `agent_service.json`.
  - `content_blocks`: v1 standard blocks verbatim (supports url/file_id sourcing, PDFs via `{"type":"file","base64":...,"mime_type":"application/pdf","extras":{"filename":...}}`).
  - Shape check (list of dicts each with str `type`; malformed → same early exit as `user_input=None` today) + optional `NEURO_SAN_MAX_CONTENT_BLOCK_BYTES` size guard.
  - Base64 STRINGS only — raw bytes never enter the dicts, so `to_json_safe`'s bytes→None trap is never sprung and direct vs HTTP sessions stay identical.
- [ ] Widen the six-signature `user_input` chain to `Union[str, List[Dict[str, Any]]]`: `data_driven_chat_session.py` (:142, :230, :291, :322), `front_man.py:62`, `front_man_activation.py:38`, `run_context.py:54`, `langchain_run_context.py` `submit_message` (:344-365) — `HumanMessage(user_message)` at :357 natively accepts v1 block lists with pin-compliant partner packages. Truncate the `ValidationError` text so it never embeds base64 payloads.
[...]

Issue #1223, opened 2026-08-09
Phase 2: Additive content_blocks wire field — reasoning outbound + lossless chat_context round-trip


Noravee:
Part of #1221. Depends on Phase 1. **Shippable alone.**

Adds ONE additive wire field carrying LangChain v1 standard content blocks verbatim, exposes reasoning/thinking to clients, and makes chat_context round-trips lossless. `text` remains the always-populated flattened view, so old clients are unaffected.

### Work items

- [ ] `chat.proto`: `repeated google.protobuf.Struct content_blocks = 9 [json_name="content_blocks"];` on `ChatMessage` (field 9 is free) — "LangChain v1 standard content blocks, carried verbatim; text remains the always-populated flattened view". Regenerate `chat_pb2.py` + `agent_service.json` via the existing OpenAPI generation script (verify Struct renders as a free-form object; the regen also refreshes the MCP validator's derived schema) + the schema dump embedded in `docs/mcp_service.md`.
- [ ] `BaseMessageDictionaryConverter.to_dict` outbound mapping: str content → `text` only (unchanged shape). List content → `text = flatten_to_text(...)` (always set, minimum `""`) + `content_blocks = standard_blocks(...)` gated by `is_trivial` (a single-text-block list collapses to plain text with no new key → bit-identical wire bytes for existing traffic). Blocks computed at capture/convert time while `model_provider`/`output_version` metadata is intact.
- [ ] `from_dict` inbound restore (chat_context legs and the `langchain_only=False` branch): `content_blocks` present → `content` = that list, with defensive stripping:
  - **strip `tool_call`/`invalid_tool_call` blocks** — chat_context carries HUMAN/SYSTEM/AI only, never the paired tool results; OpenAI returns 400 for assistant `tool_calls` without following tool messages. (Do NOT use the `content_blocks=` constructor kwarg auto-copy into `.tool_calls` for restore until AGENT_TOOL_RESULT messages are restored as paired results.)
  - **drop/unwrap `non_standard` blocks** (fall back to text) — not valid provider input; they legitimately appear for providers without registered translators (no translators exist for ollama/nvidia/openrouter/deepseek).
  - **drop orphaned/malformed reasoning blocks** (client-edited chat_context breaks Anthropic signature pairing) and **any data block lacking url/base64/file_id** (also covers the Phase 4 eliding knob's output).
[...]

PR #1276, opened 2026-08-21
Add ContentUtils policy class and block-content test infrastructure

Noravee:
Part of #1222 (PR 1 of the Phase 1 ladder). **Pure addition — no production code path changes behavior yet.** This PR defines the API that PRs 2–5 build on, so now is the cheap time to challenge naming and policy decisions: once the converter and journaling PRs consume this class, changing signatures means touching every call site.

## What's here

- **`ContentUtils`** (`neuro_san/message/utils/content_utils.py`) — the single flatten/standardize/inspect policy for message content that can be a plain string (all of today's traffic) or a list of langchain v1 content blocks (thinking/reasoning, multimodal — what the later phases add). Replaces, in later PRs, the two divergent flatten implementations that destroy or crash on block content today (`BaseMessageDictionaryConverter.to_dict` and `RunContextRunnable.parse_chain_result`).
- **`ContentFixtures`** (tests) — provider-exact message shapes (Anthropic thinking-first with signature, Anthropic tool-use, OpenAI Responses reasoning, v1 blocks, list-of-str, MCP multi-part tool content, whitespace, empty-list, multimodal human), shared by this and the follow-up golden-parity suites so every PR tests against the same shapes.
- **Golden-parity tests** for the wire converter — lock the EXACT ChatMessage dicts of today's plain-string traffic (full dict equality, incl. optionals and `from_dict` branches) before PR 2 touches the converter. These must stay green untouched through the whole phase.
- **Block-aware `ChatMockLlm`** — token counting/streaming now goes through the flattened text so list-of-blocks input doesn't crash tiktoken, plus an `emit thinking:` marker that makes `_generate` respond with Anthropic-style thinking-first block content (with `model_provider` stamped), so later PRs can test block handling end-to-end without API keys. Plain-string behavior is byte-identical (verified against the previous implementation: full ChatResult dumps, chunk sequences, token math, callback args, and pydantic `model_fields`).

## Policy decisions to review (locked in docstrings + tests)

These are what Phase 2's emission gate rests on — cheapest to challenge now:

1. `standard_blocks` **excludes `tool_call`/`tool_call_chunk`/`invalid_tool_call` blocks** — they're merged into the `content_blocks` view from `.tool_calls` and are not message content; including them would change the wire shape of every existing Anthropic tool-calling agent.
2. `is_trivial` treats a single text block as collapsible even when it carries provider bookkeeping keys (`id`/`index`) — those messages are plain strings on today's wire, and the collapse keeps that shape.
3. `is_empty_content` documents that a text-omission gate must apply to **list-form content only** — today's wire emits `text=""` for empty string content and omits the key only for empty lists.
4. `normalize_message` stamps `response_metadata["output_version"]="v1"` when storing standardized blocks — without it, a later `.content_blocks` access re-runs the provider translator and wraps reasoning as `non_standard` (verified on langchain-core 1.3.0 and 1.5.0).
5. `history_safe_text` replaces data blocks with `[image attachment: image/png]`-style references — providers reject raw image/file blocks in assistant-role history, and an image-only tool result must not become empty history (needed by PR 5).
6. `looks_like_blocks` rejects v0-era blocks (`source_type` key) and tool-call blocks inbound, so clients can't inject synthetic tool calls via `content_blocks`.

## Testing

- 67 tests in the new trees; full suite 612 passed / 1 skipped.
- pylint 10.00/10, flake8 clean.


neuro_san/message/utils/content_utils.py
```diff
@@ -0,0 +1,417 @@
...
+                return False
+            if item.get("type") in ContentUtils.TOOL_CALL_BLOCK_TYPES:
+                return False
+            if "source_type" in item:
+                return False
+        return True
+
+    # pylint: disable=too-many-return-statements
+    @staticmethod
+    def blocks_from_chat_message(chat_message: Dict[str, Any]) -> Optional[List[Dict[str, Any]]]:
```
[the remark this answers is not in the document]
Noravee: Fixed in 8e1289cd — signature is now `Optional[Dict[str, Any]]`, matching the None handling and the tests.

Noravee:
**TLDR:** one production file, and it isn't wired to anything yet. `ContentUtils` is the flatten/standardize policy that PRs 2–5 of #1222 will consume; everything else here is test infrastructure (provider-exact fixtures, golden-parity locks on the wire converter, a block-safe mock LLM). Zero behavior change for existing traffic — the mock's plain-string output was byte-compared against the previous implementation, and the parity tests assert full dict equality on today's wire shapes. The real review surface is the six policy decisions listed in the description — they're what Phase 2's emission gate rests on, and they're cheapest to change now while the API has no consumers. Sanity-checked live via `chat_mock_llm_echo.hocon`: plain echo unchanged; `emit thinking: hello world` flows through the full pipeline and still yields `hello world`. Inline notes on the key spots below.

neuro_san/message/utils/content_utils.py
```diff
@@ -0,0 +1,417 @@
...
+    or to a standardized block list, must go through this class so that the
+    projections cannot diverge.
+    """
+
+    # Block types that AIMessage.content_blocks merges in from the message's
+    # .tool_calls field (or their chunk/invalid variants). These are excluded
+    # from the standardized view: they are not message *content*, and today's
+    # wire format never carries them, so including them would change the wire
+    # output of every existing tool-calling agent.
+    TOOL_CALL_BLOCK_TYPES: List[str] = ["tool_call", "tool_call_chunk", "invalid_tool_call"]
```
Noravee, line 58: This is the most parity-critical line in the PR. langchain's `content_blocks` property merges `.tool_calls` into its view, and Anthropic tool-calling turns carry `tool_use` in content — so without this exclusion, every existing Anthropic tool-calling agent (no reasoning enabled) would start emitting `content_blocks` once the wire converter starts using this class. With it, their standardized view collapses to a single text block → trivial → plain text → today's exact wire shape.

neuro_san/message/utils/content_utils.py
```diff
@@ -0,0 +1,417 @@
...
+                 otherwise a model_copy with normalized content
+        """
+        normalized: Union[str, List[Dict[str, Any]]] = ContentUtils.normalize_content(message)
+        if isinstance(normalized, str):
+            if normalized == message.content:
+                return message
+            return message.model_copy(update={"content": normalized})
+
+        response_metadata: Dict[str, Any] = dict(message.response_metadata or {})
+        response_metadata["output_version"] = ContentUtils.OUTPUT_VERSION_V1
```
Noravee, line 253: This stamp defuses a verified langchain trap: storing the already-standardized block view into `.content` while `model_provider` metadata is intact means a *later* `.content_blocks` access re-runs the provider translator over standard blocks and wraps reasoning as `non_standard` (reproduced on langchain-core 1.3.0 and 1.5.0). `output_version="v1"` is langchain's documented passthrough contract; `test_normalize_message_stamps_output_version` and the idempotency test lock it.

neuro_san/message/utils/content_utils.py
```diff
@@ -0,0 +1,417 @@
...
+            mime_type: str = str(entry.get("mime_type") or "")
+            mime_bytes: Any = entry.get("mime_bytes")
+            if not isinstance(mime_bytes, str) or not mime_bytes or not mime_type:
+                return None
+            blocks.append(ContentUtils._data_block(mime_type, mime_bytes))
+
+        return blocks
+
+    @staticmethod
+    def history_safe_text(source: Union[BaseMessage, str, List[Any], None]) -> str:
```
Noravee, line 342: Defined now because it's API surface, consumed later when tool results start carrying data blocks (the native-message-preservation step of #1222): the wire/journal copy keeps the full blocks, but the chat_history copy is down-converted to an assistant-role `AIMessage` — and providers reject image/file blocks in assistant content (API 400). This projection keeps text and turns data blocks into `[image attachment: image/png]` references, so an image-only tool result never becomes empty (or illegal) history.

tests/neuro_san/message/types/test_base_message_dictionary_converter.py
```diff
@@ -0,0 +1,169 @@
...
+from langchain_core.messages.system import SystemMessage
+
+from neuro_san.message.types.agent_framework_message import AgentFrameworkMessage
+from neuro_san.message.types.agent_message import AgentMessage
+from neuro_san.message.types.agent_tool_result_message import AgentToolResultMessage
+from neuro_san.message.types.base_message_dictionary_converter import BaseMessageDictionaryConverter
+from neuro_san.message.types.chat_message_type import ChatMessageType
+
+
+class TestBaseMessageDictionaryConverter:
```
Noravee, line 29: These tests are the backward-compatibility contract for the whole content-block effort in executable form: they assert the EXACT dicts (full equality, not subsets) that today's plain-string traffic produces — including the optionals and `from_dict` branches. The follow-up PRs of #1222 must keep every test in this class green untouched. Deliberately absent: list-form content through `to_dict`, because its current behavior (`""` for thinking-first content, a crash for list-of-str) is the bug being fixed, not a shape to lock.

neuro_san/test/llms/chat_mock_llm.py
```diff
@@ -41,6 +44,13 @@ class ChatMockLlm(BaseChatModel):
     Adapted from https://python.langchain.com/docs/how_to/custom_chat_model/
     """
 
+    # When the (flattened) input starts with this marker, _generate responds
+    # with Anthropic-style thinking-first block content instead of an echo,
+    # so block handling can be tested end-to-end without live provider keys.
+    # _generate only: _stream ignores the marker and streams the echo.
+    # ClassVar keeps this a constant rather than a pydantic model field.
+    THINKING_MARKER: ClassVar[str] = "emit thinking:"
```
Noravee, line 52: Opt-in test hook, `_generate`-only. An echo mock can never produce thinking-first *output*, which is the exact shape whose mishandling this effort fixes — and CI has no Anthropic key. Strings starting with this marker get an Anthropic-shaped thinking-first block response (with `model_provider` stamped so translation behaves like production). Nothing in the repo sends the prefix; plain-string behavior was byte-compared against the previous implementation (ChatResult dumps, chunk sequences, token math, callback args, pydantic `model_fields`).

tests/neuro_san/message/content_fixtures.py
```diff
@@ -0,0 +1,143 @@
...
+
+from typing import Any
+from typing import Dict
+from typing import List
+
+from langchain_core.messages.ai import AIMessage
+from langchain_core.messages.human import HumanMessage
+
+
+class ContentFixtures:
```
Noravee, line 26: The literal strings here are arbitrary, but the keys and ordering are load-bearing mimicry of real provider output: thinking-first with `signature`, `model_provider` in `response_metadata` (it selects langchain's translator), Responses-API `summary` lists, mcp-adapters-0.2-style multi-part tool content. One shared catalog so this PR's unit tests and the follow-up parity suites all exercise the same canonical shapes instead of each test file inventing slightly-wrong ones.

neuro_san/message/utils/content_utils.py
```diff
@@ -0,0 +1,417 @@
...
+from typing import Union
+
+import base64
+import math
+
+from langchain_core.messages.base import BaseMessage
+from langchain_core.messages.content import KNOWN_BLOCK_TYPES
+
+
+class ContentUtils:
```
d1donlydfink, line 31: Everything in neuro_san.messages is intended to be consumed by clients.
So far I'm not seeing how this connects with clients of our system.

Because:
a) you are operating on langchain constructs here and not our own message type
b) There is nothing but test code currently accessing this class

... it probably makes sense to have this in internals for now, so it's not sooo public.

d1donlydfink (changes requested):
Maybe we should move ContentUtils to either:
    neuro_san.internals.utils
or perhaps better yet because it's very langchain-y
    neuro_san.internals.run_context.langchain.util
(which we should probably rename to utils to be consistent with other places in codebase).

Reason: the top-level neuro_san.messages stuff is meant to be consumed by clients.
So far I do not see the connection to clients when you are doing very down-and-dirty langchain-y business,
cuz what we tell them is "look at the last AgentFrameworkMessage"

It's possible I'm missing something here, so please explain the use cases other than mock llm.
If it's really only to be used with mock llm stuff, then maybe it should live near there to start.

Noravee:
Totally fair questions — the missing context is that this PR is the first step of a planned ladder in #1222 (Phase 1 of the reasoning/multimodal content plan in #1221), and the class is deliberately landing *before* its consumers so the API can be reviewed while it's still free to change. The mock LLM is not the use case — it's just what lets the later steps be tested without API keys.

**The step-by-step plan for this phase** (each step its own small PR, each revertible):

1. **This PR** — define the single flatten/standardize policy + golden-parity tests that lock today's exact wire dicts + a block-safe mock. No production wiring, by design — hence "only test code touches it today".
2. **Wire converter fix** — `BaseMessageDictionaryConverter.to_dict` (in `neuro_san/message/types/`) replaces its `value[0].get("text","")` flatten with `ContentUtils.flatten_to_text`. That flatten is a live bug today: Anthropic `thinking` and OpenAI `reasoning_effort` are already whitelisted in `llm_config`, and a thinking-first response currently flattens to `""` (list-of-str content crashes with `AttributeError`).
3. **Capture fix** — `RunContextRunnable.parse_chain_result` + `JournalingCallbackHandler.on_llm_end` (internals) switch to the same projection, ending the divergent-flatten dupe leak between capture and wire.
4. **Peripheral str-safety** — `branch_activation`, `activation_capsule`, `network_copy_middleware`, `chat_history_message_processor` (internals).
5. **Native message preservation** — the journaled/chat-history messages keep block content (`normalize_message`), with `history_safe_text` protecting the assistant-role history down-convert in `OriginatingJournal`.

Then Phase 2 (#1223) adds the additive `content_blocks` wire field and exposes reasoning to clients: `BasicMessageProcessor.get_reasoning()`/`get_content_blocks()` and `AnswerMessageProcessor` (`neuro_san/message/processors/`), `AnswerMessageFilter` (`neuro_san/message/filters/`), `ThinkingFileMessageProcessor` (client). That's also what finally closes old #613. Phase 3 (#1224) is multimodal inbound; Phase 4 (#1225) is tools/knobs/the mcp-adapters unpin.

**Why `neuro_san.message` rather than `internals`:** the first consumer (the converter) and several later ones (processors, filter) live in `neuro_san/message/` — and per the layering your refactor established, which I want to keep, **`message` never imports from `internals`**. An internals-homed ContentUtils would force exactly that import in the very next PR, or a duplicated flatten — and two divergent flatten implementations is the precise disease this class exists to cure. The internals-side consumers can import `message` without violating anything, so `message` is the only home that serves both directions.

On "it's very langchain-y": agreed, but so is this package at its foundation — `message/types` subclasses langchain `BaseMessage`, and the converter's input type is `BaseMessage`. And from Phase 2 onward the langchain v1 block dicts literally become the client-visible `content_blocks` payload, so the block vocabulary ends up part of the wire contract rather than an internals detail.

The "so public" concern is fair though — 434081c0 makes the class docstring say explicitly that this is framework plumbing, not a client API (same character as `SlyDataRedactor` next door, which clients also never touch), and points at #1222 for the consumer map.

neuro_san/message/utils/content_utils.py
Noravee: Addressed the publicness part in 434081c0 — the docstring now states this is framework plumbing, not a client API (same character as SlyDataRedactor in this package). On placement: the first consumer is the wire converter in `message/types`, and `message` never imports `internals` per the layering from your refactor — so an internals home would force either that import or a duplicate flatten in the very next PR. Full phase plan and consumer map in the PR-level comment.

d1donlydfink (approved):

PR #1278, opened 2026-08-21
NS-1246 Unify responses handling + tests refactoring. 

andreidenissov-cog:
## Summary
Two related changes to the record/playback LLM proxy (tests-only tooling):

1. Unify how recorded responses are stored, so a cassette entry always holds a
   "responses" list -- the special-case singular "response" key is gone.
2. Refactor the test suite to the repo convention: one test class per file, moved
   into a dedicated tests/ subfolder separate from the app code.

No production neuro_san code is touched; everything lives under
tests/record_playback_llm_server/.

## 1. Unify response handling (no more "response" key)
Previously a cassette entry stored either a singular "response" (single mode) or
a "responses" list (multi mode), and the handler branched on state.multi_response
to write one or the other. That dual shape complicated both writing and playback.

Now:
- Every entry stores a "responses" list. The singular "response" key is removed.
- Cassette takes a use_multi_mode flag: in single mode it keeps only the first
  recorded response for a key; in multi mode it accumulates each distinct response
  (de-duplicated ignoring per-call timing fields) for round-robin playback.
- ProxyHandler._store collapses to a single cassette.put_response(key, meta,
  response) call -- no more multi/single branching in the handler.
- RecordPlaybackLlmServer builds the cassette as Cassette(args.cassette,
  args.multi_response).

Net effect: one storage shape, simpler record/playback code, same observable
behavior (single = one response per key; multi = round-robin over distinct
responses).

Note: the on-disk cassette format now always uses "responses". Cassettes recorded
before this change (with a singular "response") would need re-recording or a
one-time convert; the cleanup_cassette tool still understands the old shape.

## 2. Test refactor (one class per file, under tests/)
Split the single test module into one test class per file, moved to
tests/record_playback_llm_server/tests/ (separate from app code):
- test_request_canonicalizer.py  -> TestRequestCanonicalizer
- test_cassette.py               -> TestCassette
- test_playback_delay.py         -> TestPlaybackDelay
- test_cassette_cleaner.py       -> TestCassetteCleaner
- test_proxy_integration.py      -> TestProxyIntegration

Supporting scaffolding (also one class per file, no module-level helper functions):
- conftest.py -- the start_app fixture plus the autouse override that lets these
  tests run without an LLM provider key (moved out of the app directory).
- counting_upstream.py / stream_upstream.py / rate_limited_upstream.py -- the fake
  upstream handlers, one class each.
- proxy_test_client.py -- ProxyTestClient, holding the former module-level test
  helpers (post/content/upstream_client) as static methods plus shared constants.

Tests were updated to the unified Cassette API (use_multi_mode; assertions read
"responses" instead of the old "response"); added a single-mode retention test.

## Testing
- 23 tests pass.
- flake8 clean and pylint 10.00/10 across the package, tests directory included.
- The in-process integration tests spin up a fake upstream plus the real proxy on
  ephemeral loopback ports, so they need no real LLM host or neuro-san server and
  run in the default unit suite.

## Related issue
NS-1246


tests/record_playback_llm_server/cassette.py
```diff
@@ -85,52 +86,58 @@ def load(self) -> None:
...
+                if not any(self._same_content(existing, response) for existing in responses):
+                    responses.append(response)
+                    modified = True
+            else:
+                # In single-response mode, use already existing response if we have it:
+                if len(responses) == 0:
+                    responses.append(response)
+                    modified = True
+        if modified:
+            self.save()
```
d1donlydfink, line 127: Sure would be nice to have some comments here now that there is a person understanding what is going on.

d1donlydfink (approved):

Noravee:
Post-merge review notes — the change is solid overall , but a few follow-ups worth tracking:

**1. Single-mode record semantics changed: last-writer-wins → first-writer-wins**
Previously `Cassette.put()` overwrote the entry on every record, so re-recording over an existing cassette refreshed it. Now `put_response()` keeps only the *first* response and drops later ones, so in single mode re-recording an existing cassette is a no-op for existing keys — the file must be deleted to get fresh data. This may be intentional (it matches hybrid semantics), but the README still documents the old behavior ("last-writer-wins on record", in the Multi-response section) and should be updated either way.

**2. Legacy cassettes produce mixed-shape entries when re-recorded**
`put_response()` does `entry.setdefault("responses", [])` without migrating a legacy singular `"response"`, so recording over an old-format entry leaves *both* keys in the file; playback prefers `"responses"` and silently shadows the old response. That also means old entries effectively behave last-writer-wins in single mode — inconsistent with the new first-wins rule. Suggestion: fold a legacy `"response"` into `"responses"` in `Cassette.load()` so the "always a list" invariant is actually true. (Note: playback of old cassettes still works via the `_select_response` fallback, so re-recording isn't strictly required as the PR description suggests.)

**3. `ProxyState.multi_response` is now vestigial**
After this change it is only assigned and logged (`record_playback_llm_server.py`); behavior is controlled by `Cassette.use_multi_mode`. Two flags for one concept can drift — the integration tests already have to pass both. Suggestion: drop the `ProxyState` field and log `state.cassette.use_multi_mode` instead.

**4. Stale docstring in `ProxyHandler._store`**
It still says: *"In single-response mode the entry keeps one \"response\"; in multi-response mode distinct responses accumulate under a \"responses\" list"* — exactly the branching this PR removed. Both modes now use `"responses"`.

**5. `Cassette.VERSION` still 1**
The written format changed (the singular `"response"` key is never written anymore). Readers are tolerant so this is defensible, but if the version field is meant to signal anything, it should probably be bumped.

PR #1280, opened 2026-08-24
Cache dynamically created args_schema models by spec content

Noravee:
Addresses the model-caching item of #1209.

## Why

`args_schema` pydantic models were rebuilt from scratch for every tool of every agent activation on every request, plus once more per agent at registry load (where the validator built and discarded them). pydantic v2's `create_model()` compiles a Rust core schema on every call — **~0.68 ms** for a representative nested spec, multiplied by nesting depth since each nested object is its own `create_model()` call — even though the specs are static registry data.

## What

- `BaseModelDictionaryConverter.from_dict()` now serves models from a class-level, bounded LRU cache keyed on `top_level_field_name` + an order-preserving canonical JSON serialization of the spec.
- `PydanticParametersNetworkValidator` hoists the (stateless) converter out of its per-agent loop; because the cache is shared, registry validation at load time also warms the entries the tool-creation path reads on the first request.

Design decisions, each commented in code:

- **Key preserves key order** (no `sort_keys`): field order flows through to the JSON schema advertised to providers, so specs differing only in key order get separate entries instead of one being served a bytes-different schema.
- **The model name is part of the key** — it becomes the class name pydantic uses in validation error messages.
- **Builds happen outside the lock**: two threads racing on the same spec build two equivalent classes and one wins; nobody blocks on a compile.
- **Spec errors propagate uncached**, so a bad spec is re-reported every time it is seen.
- **Unserializable or too-deeply-nested specs fall back to uncached builds** (`TypeError`/`ValueError`/`RecursionError` from `json.dumps`).
- **Bounded at 256 entries (LRU)** because external agents send their function specs over the network per request — without a bound, a misbehaving remote agent whose spec varies per response would grow the cache without limit. Eviction degrades gracefully to the pre-change rebuild cost.
- **Hand-rolled rather than `functools.lru_cache`**: `lru_cache` can only key on hashables, forcing a `json.loads` round-trip; this cache keys on the serialized form but builds from the original dict, so output is byte-for-byte what an uncached build produces.

## Correctness properties

- **Hot-reload safe by construction**: the server can refresh registries every few seconds; an edited parameters block is different content, hence a guaranteed cache miss — the old model cannot be served for it. Superseded entries just wait for LRU eviction.
- **Shared classes are safe**: audited every consumer of the returned model — neuro-san assigns it to `tool.args_schema` and reads `model_fields`; langchain-core reads it and stamps per-tool name/description onto a *fresh subset class* (`_create_subset_model_v2`), never onto the shared one. `from_dict()`'s docstring now documents the returned class as shared-immutable.

## Testing

- 9 new cache tests: content-keyed sharing across instances, key-order sensitivity, name-in-key, LRU eviction at the bound, unserializable and deep-nesting fallbacks, errors-not-cached, an 8-thread race, and hot-reload freshness.
- Test file restructured to the repo's one-test-class-per-module convention (single `TestBaseModelDictionaryConverter`, one shared `_convert` helper, cache-clearing fixture around every test).
- Full converter + validator suites: 149 passed; flake8 and pylint clean.

### DIFF
--- a/neuro_san/internals/run_context/langchain/core/base_model_dictionary_converter.py
+++ b/neuro_san/internals/run_context/langchain/core/base_model_dictionary_converter.py
@@ -133,11 +133,11 @@ class BaseModelDictionaryConverter(DictionaryConverter):
         # At this point we are not going back to OpenAI functional specs
         raise NotImplementedError
 
-    def from_dict(self, obj_dict: Dict[str, Any]) -> BaseModel:
+    def from_dict(self, obj_dict: Optional[Dict[str, Any]]) -> Optional[Type[BaseModel]]:
         """
         :param obj_dict: The data-only dictionary to be converted into an object
-        :return: An object instance created from the given dictionary.
-                If obj_dict is None, the returned object should also be None.
+        :return: The pydantic model class created from the given dictionary.
+                If obj_dict is None, the returned value should also be None.
                 If obj_dict is not the correct type, it is also reasonable
                 to return None.
 
@@ -163,7 +163,7 @@ class BaseModelDictionaryConverter(DictionaryConverter):
 
         cache: OrderedDict = BaseModelDictionaryConverter._model_cache
         with BaseModelDictionaryConverter._model_cache_lock:
-            base_model: BaseModel = cache.get(cache_key)
+            base_model: Optional[Type[BaseModel]] = cache.get(cache_key)
             if base_model is not None:
                 cache.move_to_end(cache_key)
                 return base_model
@@ -209,15 +209,15 @@ class BaseModelDictionaryConverter(DictionaryConverter):
         # pydantic validation error messages), so it is part of the key.
         return f"{self.top_level_field_name}:{spec_json}"
 
-    def openai_function_to_pydantic(self, name: str, function_dict: Dict[str, Any]) -> BaseModel:
+    def openai_function_to_pydantic(self, name: str, function_dict: Dict[str, Any]) -> Type[BaseModel]:
         """
-        Turns an openai function spec dictionary into a pydantic BaseModel
+        Turns an openai function spec dictionary into a pydantic BaseModel class
         See: https://docs.pydantic.dev/latest/concepts/models/#dynamic-model-creation
 
         :param name: The string name of the object/field undergoing conversion
         :param function_dict: The dictionary describing the OpenAI function to
                     be converted into a pydantic BaseModel
-        :return: The pydantic BaseModel that corresponds to the OpenAI function spec.
+        :return: The pydantic BaseModel class that corresponds to the OpenAI function spec.
         """
 
         # Get stuff from the object-level function dictionary
@@ -315,7 +315,7 @@ class BaseModelDictionaryConverter(DictionaryConverter):
         # crash the run - langchain surfaces them to the calling LLM as a
         # correctable tool-error message, at the cost of a retry round-trip.
         try:
-            model: BaseModel = create_model(name, **fields)
+            model: Type[BaseModel] = create_model(name, **fields)
         except (NameError, TypeError) as exception:
             # Safety net for anything validate_field_name() does not cover:
             # report whatever create_model() rejects as the same spec-error
--- a/neuro_san/internals/run_context/langchain/core/base_model_dictionary_converter.py
+++ b/neuro_san/internals/run_context/langchain/core/base_model_dictionary_converter.py
@@ -101,6 +101,13 @@ class BaseModelDictionaryConverter(DictionaryConverter):
     # specs over the network per request - without a bound, a misbehaving
     # remote agent whose spec varies per response would grow this without
     # limit.  Eviction is least-recently-used.
+    #
+    # Hand-rolled rather than functools.lru_cache: lru_cache can only key
+    # on hashables, so the builder would have to recover the spec with
+    # json.loads(key) and build from the round-tripped dictionary.  This
+    # cache keys on the serialized form but builds from the ORIGINAL
+    # dictionary, so what gets built is byte-for-byte what an uncached
+    # build would have produced.
     MODEL_CACHE_MAX_SIZE: int = 256
     _model_cache: OrderedDict = OrderedDict()
     _model_cache_lock: Lock = Lock()
@@ -133,18 +140,25 @@ class BaseModelDictionaryConverter(DictionaryConverter):
                 If obj_dict is None, the returned object should also be None.
                 If obj_dict is not the correct type, it is also reasonable
                 to return None.
+
+                The returned model class may be shared with every other
+                caller that converted an identical spec (see the cache
+                note on the class attributes above) - treat it as
+                immutable.  Mutating it in place would leak the change
+                into every tool and request sharing the spec.
         """
         # Honor the DictionaryConverter contract stated above.  None reaches
         # here when an external agent reports "parameters": null.
         if obj_dict is None:
             return None
 
-        cache_key: str = self.get_cache_key(obj_dict)
+        cache_key: Optional[str] = self.get_cache_key(obj_dict)
         if cache_key is None:
-            # The spec contains values JSON cannot represent.  Neither hocon
-            # registries nor network specs can produce those (both are born
-            # as JSON), so this is only reachable from in-process callers -
-            # build uncached rather than fail.
+            # The spec cannot be serialized for keying: it contains values
+            # JSON cannot represent, or is nested too deeply to serialize.
+            # Neither hocon registries nor network specs can produce those
+            # (both are born as JSON), so this is only reachable from
+            # in-process callers - build uncached rather than fail.
             return self.openai_function_to_pydantic(self.top_level_field_name, obj_dict)
 
         cache: OrderedDict = BaseModelDictionaryConverter._model_cache
@@ -169,7 +183,7 @@ class BaseModelDictionaryConverter(DictionaryConverter):
 
         return base_model
 
-    def get_cache_key(self, obj_dict: Dict[str, Any]) -> str:
+    def get_cache_key(self, obj_dict: Dict[str, Any]) -> Optional[str]:
         """
         :param obj_dict: The spec dictionary a model is being created from
         :return: A string cache key uniquely identifying the model that
@@ -182,8 +196,13 @@ class BaseModelDictionaryConverter(DictionaryConverter):
             # schema advertised to LLM providers.  Specs that differ only in
             # key order therefore get separate cache entries instead of one
             # of them being served a bytes-different schema.
+            #
+            # RecursionError: dumps() walks the whole spec, including deep
+            # nesting under keys the model builder itself never recurses
+            # into, so it can hit the recursion limit on a spec the
+            # uncached build handles fine.  Fall back to that build.
             spec_json: str = dumps(obj_dict, separators=(",", ":"))
-        except (TypeError, ValueError):
+        except (RecursionError, TypeError, ValueError):
             return None
 
         # The name becomes the created model's class name (visible in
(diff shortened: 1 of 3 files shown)


neuro_san/internals/run_context/langchain/core/base_model_dictionary_converter.py
```diff
@@ -111,15 +140,75 @@ def from_dict(self, obj_dict: Dict[str, Any]) -> BaseModel:
                 If obj_dict is None, the returned object should also be None.
                 If obj_dict is not the correct type, it is also reasonable
                 to return None.
+
+                The returned model class may be shared with every other
+                caller that converted an identical spec (see the cache
+                note on the class attributes above) - treat it as
+                immutable.  Mutating it in place would leak the change
+                into every tool and request sharing the spec.
```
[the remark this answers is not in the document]
Noravee: Agreed — fixed in 5331df99. `from_dict` is now `Optional[Dict[str, Any]] -> Optional[Type[BaseModel]]`, the docstring's "object instance" wording is corrected to the model class it actually returns, and the same class-vs-instance inaccuracy is fixed on `openai_function_to_pydantic` (`-> Type[BaseModel]`) and its `create_model` local for consistency.

tests/neuro_san/internals/run_context/langchain/core/test_base_model_dictionary_converter.py
```diff
@@ -28,22 +31,63 @@
...
+    def clear_model_cache(self):
+        """
+        The model cache is class-level, process-wide state.  Clear it around
+        every test so tests stay order-independent.
+        """
+        BaseModelDictionaryConverter._model_cache.clear()   # pylint: disable=protected-access
+        yield
+        BaseModelDictionaryConverter._model_cache.clear()   # pylint: disable=protected-access
+
+    def _convert(self, parameters: Dict[str, Any], name: str = "parameters") -> BaseModel:
```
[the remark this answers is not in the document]
Noravee: Agreed — fixed in 5331df99. `_convert` is now `Optional[Dict[str, Any]] -> Optional[Type[BaseModel]]`, with the docstring noting that a None argument exercises the documented None -> None contract.

neuro_san/internals/run_context/langchain/core/base_model_dictionary_converter.py
```diff
@@ -83,6 +87,31 @@ class BaseModelDictionaryConverter(DictionaryConverter):
...
+    # remote agent whose spec varies per response would grow this without
+    # limit.  Eviction is least-recently-used.
+    #
+    # Hand-rolled rather than functools.lru_cache: lru_cache can only key
+    # on hashables, so the builder would have to recover the spec with
+    # json.loads(key) and build from the round-tripped dictionary.  This
+    # cache keys on the serialized form but builds from the ORIGINAL
+    # dictionary, so what gets built is byte-for-byte what an uncached
+    # build would have produced.
+    MODEL_CACHE_MAX_SIZE: int = 256
```
Noravee, line 111: **Why class-level shared state, and why hand-rolled**: the whole point is that the load-time validator and every per-request converter instance hit the same entries, so the cache must outlive instances (same pattern as `BaseToolFactory.synthesis_warned`). On `functools.lru_cache`: it was considered and rejected — it can only key on hashables, forcing a `json.loads(key)` round-trip that would build models from a *re-parsed* dict (changing behavior for any non-JSON-native in-process spec). I verified CPython's `lru_cache` also builds outside its lock, so the hand-rolled version loses none of its concurrency behavior — it only gains build-from-original fidelity. The 256 bound exists because external agents send specs over the network per request; eviction degrades gracefully to the pre-change rebuild cost (~0.7 ms), never to incorrectness.

neuro_san/internals/run_context/langchain/core/base_model_dictionary_converter.py
```diff
@@ -104,31 +133,91 @@ def to_dict(self, obj: BaseModel) -> Dict[str, Any]:
...
-                If obj_dict is None, the returned object should also be None.
+        :return: The pydantic model class created from the given dictionary.
+                If obj_dict is None, the returned value should also be None.
                 If obj_dict is not the correct type, it is also reasonable
                 to return None.
+
+                The returned model class may be shared with every other
+                caller that converted an identical spec (see the cache
+                note on the class attributes above) - treat it as
+                immutable.  Mutating it in place would leak the change
```
Noravee, line 147: **The contract change to scrutinize**: `from_dict()` used to return a fresh class per call; identical specs now share one. Audited every consumer before relying on this — neuro-san assigns it to `tool.args_schema` and reads `model_fields`; langchain-core reads it and stamps per-tool name/description onto a *fresh subset class* (`_create_subset_model_v2`), never onto the shared one; the validator discards it. No mutation anywhere today, so the docstring makes immutability an explicit contract for future callers rather than an accident.

neuro_san/internals/run_context/langchain/core/base_model_dictionary_converter.py
```diff
@@ -104,31 +133,91 @@ def to_dict(self, obj: BaseModel) -> Dict[str, Any]:
...
+            return self.openai_function_to_pydantic(self.top_level_field_name, obj_dict)
+
+        cache: OrderedDict = BaseModelDictionaryConverter._model_cache
+        with BaseModelDictionaryConverter._model_cache_lock:
+            base_model: Optional[Type[BaseModel]] = cache.get(cache_key)
+            if base_model is not None:
+                cache.move_to_end(cache_key)
+                return base_model
+
+        # Build outside the lock: create_model() is the slow part, and two
```
Noravee, line 171: **Deliberate thundering-herd choice**: the lock covers only dict bookkeeping, never `create_model()`. Two threads racing on the same spec each build an equivalent class and the second write wins — cheaper than serializing all conversions behind a compile, and the only cost is one wasted build in a rare race (pinned by the 8-thread test). Spec errors raised here propagate *uncached* on purpose: a bad spec is re-reported every time it is seen, so the validator's per-agent error messages never go silent after the first request.

neuro_san/internals/run_context/langchain/core/base_model_dictionary_converter.py
```diff
@@ -104,31 +133,91 @@ def to_dict(self, obj: BaseModel) -> Dict[str, Any]:
...
+            # the spec's own field order, which flows through to the JSON
+            # schema advertised to LLM providers.  Specs that differ only in
+            # key order therefore get separate cache entries instead of one
+            # of them being served a bytes-different schema.
+            #
+            # RecursionError: dumps() walks the whole spec, including deep
+            # nesting under keys the model builder itself never recurses
+            # into, so it can hit the recursion limit on a spec the
+            # uncached build handles fine.  Fall back to that build.
+            spec_json: str = dumps(obj_dict, separators=(",", ":"))
```
Noravee, line 204: **The two keying decisions with wire impact**: (1) no `sort_keys` — field order flows into the JSON schema providers receive, so two specs differing only in key order get separate entries instead of one being served a bytes-different schema; (2) `top_level_field_name` is in the key because it becomes the class name pydantic quotes in validation errors sent back to the calling LLM. Known theoretical limit, accepted: `json.dumps` coerces tuples→lists and int/bool/None dict keys→strings, so a pathological *in-process* spec could alias a JSON-born one — unreachable from hocon registries or network specs, which are both born as JSON. The `RecursionError` in the except tuple is not decorative: `dumps` walks depth the model builder never recurses into, so a deep spec under an ignored key must fall back to the uncached build (regression-tested).

neuro_san/internals/validation/network/pydantic_parameters_network_validator.py
```diff
@@ -62,6 +62,9 @@ def validate_name_to_spec_dict(self, name_to_spec: Dict[str, Any]) -> List[str]:
 
         self.logger.debug("Validating %s parameters via pydantic...", self.network_name)
 
+        # The converter is stateless, so one instance serves every agent.
+        converter = BaseModelDictionaryConverter("parameters")
```
Noravee, line 66: The hoist is the trivial part (stateless converter, one instance per pass — the old per-agent construction was pure churn, and building it outside the `try` keeps the per-agent error message honest about what failed). The quiet win: because the cache is shared, load-time validation now warms the entries the tool-creation path reads on the first request. That warming is an emergent bonus, not a contract — the two paths produce their parameter dicts independently (this one re-runs the filter chain), so the comment here deliberately says nothing about the other class's cache internals.

tests/neuro_san/internals/run_context/langchain/core/test_base_model_dictionary_converter.py
```diff
@@ -226,3 +261,158 @@ def test_array_with_string_items_raises_tool_spec_error(self):
...
+        for thread in threads:
+            thread.join()
+
+        assert len(results) == 8
+        for model in results:
+            assert issubclass(model, BaseModel)
+            assert model.model_validate({"city": "Lisbon"}).city == "Lisbon"
+        assert len(BaseModelDictionaryConverter._model_cache) == 1   # pylint: disable=protected-access
+
+    def test_edited_spec_gets_fresh_model_not_stale_cache(self):
```
Noravee, line 396: This test exists to pin the operational question a cache like this always raises: the server can hot-reload registries every few seconds, so can it serve a stale model? No — the cache never keys on agent name, file, or time, only on what the spec *says*, so an edited parameters block is a guaranteed miss and the old entry just waits for LRU eviction. The membership assertions (`units` present after, absent before) verify the rebuilt model actually reflects the edit, which the key-order test never checks.

Noravee (commented):
**TLDR**

This PR makes `from_dict()` memoize the pydantic model classes it builds, keyed on spec content. Three layers, in increasing order of subtlety:

1. **The cache itself**: `create_model()` compiles a Rust core schema per call (~0.68 ms for a representative nested spec, multiplied by nesting depth), and it ran for every tool of every agent activation on every request — plus once more per agent at registry load, where the validator built models just to throw them away. A class-level, bounded (256, LRU) cache collapses that to one compile per distinct spec: **~4.4 µs per hit, ~155x**.
2. **Wire-format neutrality**: the cache is designed so no provider ever receives different bytes than before — the key preserves spec key order (no `sort_keys`), the model is built from the *original* dict (not a JSON round-trip), and the model name stays part of the key. Hot-reload is safe by construction: an edited parameters block is different content, hence a guaranteed miss.
3. **Fallbacks and contract**: spec errors propagate uncached (re-reported every time); unserializable or too-deeply-nested specs fall back to uncached builds (`TypeError`/`ValueError`/`RecursionError`); the returned class is documented shared-immutable, backed by an audit of every consumer.

**Evidence backing the risky bits**: a max-effort self-review ran 10 finder angles + 13 adversarial verifiers + a gap sweep over this diff before publishing — its confirmed findings (the `RecursionError` gap, annotation mismatches) are already fixed here, and the consumer audit (neuro-san + langchain-core paths, incl. `_create_subset_model_v2`) confirmed nothing mutates the shared class. 149 tests green; benchmark method in the PR description.

Inline notes below point at the decisions most worth scrutinizing.

neuro_san/internals/run_context/langchain/core/base_model_dictionary_converter.py
d1donlydfink: No. Just no.

We do not want to cache anything globally, as we are doing here, as that lets down the barrier of requests being sandboxed from each other.

d1donlydfink (changes requested):
We don't want any request caching anything globally ever at the library level.

What is the problem we are trying to solve here?
If this is a performance problem, I am happy to live with the performance hit to have the firewall between requests remain intact.

If you need something cached on a per-request basis, the proper place is to put something on SessionInvocationContext

Noravee:
Closing per the review: the request-isolation policy stands — no process-global caching of content that flows through request processing, and the performance win here (~ms per tool conversion, amortized against LLM round-trips) isn't worth softening that firewall. Leaving the branch, measurements, and consumer audit in place for reference in case this is ever revisited in a form that respects the boundary (e.g. load-time precomputation scoped to the AgentNetwork, with network-fetched external specs never cached).

PR #1281, opened 2026-08-24
Flatten list content to full text in the wire converter

Noravee:
## What

Fixes the wire converter's projection of list-form message content. `BaseMessageDictionaryConverter.to_dict` previously flattened list content with `value[0].get("text", "")` — the first block only. That has three failure modes, all reachable today:

- **Thinking-first responses flatten to `""`**: a ChatAnthropic response with extended thinking puts the `thinking` block first, so the answer text was dropped entirely.
- **Text after the first block is dropped**: multi-text-block content (e.g. OpenAI Responses API replies) lost everything past block 0.
- **List-of-str content crashes**: `content=["a", "b"]` is legal per the `BaseMessage` annotation, and `.get()` on a `str` raises `AttributeError`.

The fix projects list content through `ContentUtils.flatten_to_text` (#1276), which implements `BaseMessage.text` semantics: concatenate every text block, in order, never strip.

## Backward compatibility

- Plain-string content is untouched — the golden-parity tests locking down exact wire dicts for text-only traffic all pass unchanged.
- Empty-list content keeps **omitting** the `text` key, exactly as before. Emitting `text=""` instead would make such messages newly answer-eligible in `AnswerMessageFilter` and able to displace the real answer.
- Reasoning-only list content still emits `text=""`, same as the old flatten.

## Tests

Four new tests in the converter's golden-parity suite, using the shared provider-shaped fixtures from #1276: thinking-first content yields the answer text (exact wire dict), all text blocks are concatenated (and reasoning-only still emits `""`), list-of-str no longer crashes, and empty-list content still omits the key.

Part of #1222 (Phase 1), second of the PR ladder described there.

### DIFF
--- a/neuro_san/message/types/base_message_dictionary_converter.py
+++ b/neuro_san/message/types/base_message_dictionary_converter.py
@@ -112,6 +112,11 @@ class BaseMessageDictionaryConverter(DictionaryConverter):
                     # displace the real answer.
                     value = None
                 else:
+                    # Deliberately NOT gated on ContentUtils.is_empty_content:
+                    # blank-but-non-empty block content (e.g. reasoning-only)
+                    # already emitted text=""/" " under the old first-block
+                    # flatten, and this projection preserves that wire shape
+                    # exactly.
                     value = ContentUtils.flatten_to_text(value)
 
             if value is not None:
--- a/neuro_san/message/types/base_message_dictionary_converter.py
+++ b/neuro_san/message/types/base_message_dictionary_converter.py
@@ -32,6 +32,7 @@ from neuro_san.message.types.agent_framework_message import AgentFrameworkMessag
 from neuro_san.message.types.agent_progress_message import AgentProgressMessage
 from neuro_san.message.types.agent_tool_result_message import AgentToolResultMessage
 from neuro_san.message.types.chat_message_type import ChatMessageType
+from neuro_san.message.utils.content_utils import ContentUtils
 
 
 class BaseMessageDictionaryConverter(DictionaryConverter):
@@ -97,15 +98,21 @@ class BaseMessageDictionaryConverter(DictionaryConverter):
                 # for, and that is ok.
                 value = None
 
-            # For OpenAI and Ollama, content of AI message is a string but content from
-            # Anthropic AI message can either be a single string or a list of content blocks.
-            # If it is a list, "text" is a key of a dictionary which is the first element of
-            # the list. For more details: https://python.langchain.com/docs/integrations/chat/anthropic/#content-blocks
+            # Content is a plain string for most providers, but can be a list
+            # of content blocks (e.g. Anthropic thinking + text, or list-of-str,
+            # which is also legal per the BaseMessage annotation). Project list
+            # content to its full text: concatenate every text block, not just
+            # the first one - a thinking-first response has no text in its
+            # first block at all.
             if src == "content" and isinstance(value, list):
-                if len(value) > 0:
-                    value = value[0].get("text", "")
-                else:
+                if len(value) == 0:
+                    # Keep omitting the text key for empty-list content:
+                    # emitting text="" instead would make such messages newly
+                    # answer-eligible in AnswerMessageFilter and able to
+                    # displace the real answer.
                     value = None
+                else:
+                    value = ContentUtils.flatten_to_text(value)
 
             if value is not None:
                 chat_message[dest] = value
--- a/tests/neuro_san/message/types/test_base_message_dictionary_converter.py
+++ b/tests/neuro_san/message/types/test_base_message_dictionary_converter.py
@@ -25,20 +25,23 @@ from neuro_san.message.types.agent_tool_result_message import AgentToolResultMes
 from neuro_san.message.types.base_message_dictionary_converter import BaseMessageDictionaryConverter
 from neuro_san.message.types.chat_message_type import ChatMessageType
 
+from tests.neuro_san.message.content_fixtures import ContentFixtures
+
 
 class TestBaseMessageDictionaryConverter:
     """
-    Golden-parity tests for the wire converter.
+    Golden-parity tests for the wire converter, plus the corrected
+    projection of list-form (block) content.
 
-    These lock down the EXACT ChatMessage dictionaries produced for
-    plain-string traffic - the shapes every deployed client sees today.
-    The content-block work (see issue #1222) must keep every test here
+    The plain-string tests lock down the EXACT ChatMessage dictionaries
+    produced for text-only traffic - the shapes every deployed client sees.
+    The content-block work (see issue #1222) must keep every one of them
     green untouched: byte-identical wire output for text-only messages
     is the backward-compatibility guarantee of that whole effort.
 
-    Deliberately absent: list-form (block) content through to_dict. Its
-    current behavior is the bug being fixed (thinking-first content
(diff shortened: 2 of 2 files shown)


tests/neuro_san/message/types/test_base_message_dictionary_converter.py
```diff
@@ -131,6 +134,50 @@ def test_from_dict_unknown_type_yields_none_when_langchain_only(self):
...
+        """
+        Text blocks after the first must not be dropped, and a reasoning-only
+        message still emits text="" exactly as the old flatten did.
+        """
+        converter = BaseMessageDictionaryConverter()
+        result = converter.to_dict(ContentFixtures.openai_responses_reasoning())
+        assert result["text"] == "the answer"
+
+        reasoning_only = AIMessage(content=[{"type": "reasoning", "reasoning": "hidden"}])
+        assert converter.to_dict(reasoning_only)["text"] == ""
```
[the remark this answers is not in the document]
Noravee: Done in 73f501c6 - all three list-content tests (and the suppressed-comment locations) now assert the full wire dict, consistent with the rest of the suite.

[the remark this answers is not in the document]
Noravee: Tightened all three tests to full-dict equality in 73f501c6 so every list-content case now locks the entire wire shape.

neuro_san/message/types/base_message_dictionary_converter.py
```diff
@@ -97,15 +98,21 @@ def to_dict(self, obj: BaseMessage) -> Dict[str, Any]:
...
+            # the first one - a thinking-first response has no text in its
+            # first block at all.
             if src == "content" and isinstance(value, list):
-                if len(value) > 0:
-                    value = value[0].get("text", "")
-                else:
+                if len(value) == 0:
+                    # Keep omitting the text key for empty-list content:
+                    # emitting text="" instead would make such messages newly
+                    # answer-eligible in AnswerMessageFilter and able to
```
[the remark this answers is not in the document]
Noravee: Blank-but-non-empty block content is not newly answer-eligible: the old first-block flatten already emitted text=" " for [{"type":"text","text":" "}] and text="" for reasoning-only content, so this PR preserves those wire shapes exactly. Gating on is_empty_content would newly OMIT the text key for reasoning-only messages - a wire-shape change this parity-focused PR deliberately avoids (is_empty_content's own docstring warns the omission gate must stay list-emptiness only). The one genuinely new emission is blank list-of-str, which previously crashed with AttributeError. Locked this in with a comment and a parity test in 3b8d3cf0.

Noravee:
TLDR: one behavioral hunk, everything else is tests.

- The only production change is in `to_dict`: list-form content now projects through `ContentUtils.flatten_to_text` (all text blocks, in order) instead of `value[0].get("text", "")` (first block only). This fixes thinking-first responses flattening to `""`, later text blocks being dropped, and the `AttributeError` on list-of-str content.
- Everything that worked before produces byte-identical wire output: plain-string content doesn't touch the new code path, empty-list content still omits the `text` key, and blank/reasoning-only block content still emits its blank text. The pre-existing golden-parity tests pass unchanged, and new tests lock the parity edge cases.
- The omission gate is deliberately `len == 0`, not "no visible text" — see the inline note on the converter.

neuro_san/message/types/base_message_dictionary_converter.py
```diff
@@ -97,15 +98,26 @@ def to_dict(self, obj: BaseMessage) -> Dict[str, Any]:
...
+                    # displace the real answer.
                     value = None
+                else:
+                    # Deliberately NOT gated on ContentUtils.is_empty_content:
+                    # blank-but-non-empty block content (e.g. reasoning-only)
+                    # already emitted text=""/" " under the old first-block
+                    # flatten, and this projection preserves that wire shape
+                    # exactly.
+                    value = ContentUtils.flatten_to_text(value)
```
Noravee, line 120: This is the whole production change. The old projection was `value[0].get("text", "")` - first block only - which returned "" for Anthropic thinking-first responses (the answer lives in block 1), dropped any text past block 0, and raised AttributeError on list-of-str content. `flatten_to_text` implements `BaseMessage.text` semantics: every text block, in order, never stripped. Plain-string content never enters this branch, so text-only traffic is byte-identical.

neuro_san/message/types/base_message_dictionary_converter.py
```diff
@@ -97,15 +98,26 @@ def to_dict(self, obj: BaseMessage) -> Dict[str, Any]:
...
+            # of content blocks (e.g. Anthropic thinking + text, or list-of-str,
+            # which is also legal per the BaseMessage annotation). Project list
+            # content to its full text: concatenate every text block, not just
+            # the first one - a thinking-first response has no text in its
+            # first block at all.
             if src == "content" and isinstance(value, list):
-                if len(value) > 0:
-                    value = value[0].get("text", "")
-                else:
+                if len(value) == 0:
```
Noravee, line 108: The omission gate is deliberately list-emptiness, not "no visible text". Three cases to keep straight: empty list [] omitted the text key before and still does; blank-but-non-empty blocks (reasoning-only, whitespace text) emitted text=""/" " before and still do; list-of-str crashed before, so its new text emission is the only new wire behavior. Gating on ContentUtils.is_empty_content instead would newly omit text for reasoning-only messages - a wire-shape change.

tests/neuro_san/message/types/test_base_message_dictionary_converter.py
```diff
@@ -131,6 +134,76 @@ def test_from_dict_unknown_type_yields_none_when_langchain_only(self):
...
         assert converter.from_dict({"type": ChatMessageType.AGENT, "text": "internal"}) is None
 
+    def test_to_dict_thinking_first_content_yields_answer_text(self):
+        """
+        THE headline fix: an Anthropic thinking-first response must produce
+        the answer text on the wire. The old first-block flatten produced ""
+        because the first block is the thinking block, which has no "text".
+        """
+        converter = BaseMessageDictionaryConverter(origin=self.ORIGIN)
+        result = converter.to_dict(ContentFixtures.anthropic_thinking_first())
```
Noravee, line 144: The headline regression test: with extended thinking enabled, ChatAnthropic puts the thinking block first, so the old first-block flatten sent text="" to every client. The fixture is shaped exactly like the pinned langchain-anthropic output.

tests/neuro_san/message/types/test_base_message_dictionary_converter.py
```diff
@@ -131,6 +134,76 @@ def test_from_dict_unknown_type_yields_none_when_langchain_only(self):
...
+    def test_to_dict_blank_text_block_keeps_emitting_text(self):
+        """
+        Blank-but-non-empty block content keeps emitting its (blank) text,
+        exactly as the old first-block flatten did - only the empty LIST
+        omits the text key. Gating omission on "no visible text" instead
+        would newly omit text for reasoning-only messages, changing their
+        wire shape.
+        """
+        converter = BaseMessageDictionaryConverter()
+        result = converter.to_dict(AIMessage(content=[{"type": "text", "text": " "}]))
```
Noravee, line 201: This test exists to defend the gate above against a tempting future refactor: omitting the text key whenever the flattened text is blank looks harmless but would change the wire shape of reasoning-only messages, which have always emitted text="".

d1donlydfink (approved):

PR #1285, opened 2026-08-24
Use the shared full-text projection at both capture points

Noravee:
## What

Fixes how list-form (content-block) LLM output is captured, at the two points where it first enters the pipeline:

- **`RunContextRunnable.parse_chain_result`** flattened list content with an inline loop that took only the first `type=="text"` block. Text after the first block was silently dropped, and list-of-str content (legal per the `BaseMessage` annotation) skipped the non-dict items entirely and became `""`.
- **`JournalingCallbackHandler.on_llm_end`** captured via `ChatGeneration.text` — a second, independent projection of the same content. On the pinned langchain-core its output matches `BaseMessage.text` semantics except for legacy typeless `{"text": ...}` blocks — a shape no supported provider emits, declared out of scope at the call site — its list-content behavior has varied across versions, and any divergence between the two projections defeats `OriginatingJournal`'s dupe suppression: the held AGENT message and the following AI message carry different text, so the comparison fails and clients receive both copies.

Both sites now project through `ContentUtils.flatten_to_text` (#1276) — the same projection the wire converter uses since #1281. One policy, three call sites, no way to diverge.

## Behavior preserved

- Plain-string content is untouched at both sites.
- `on_llm_end` keeps its journaling gate and `.strip()`: tool-call-only steps flatten to `""` and stay unjournaled, so clients still see only the "Invoking:" message for those.
- The error-detector pass at the end of `parse_chain_result` is unchanged.

## Tests

- `parse_chain_result`: thinking-first → answer text, multi-text-block concatenation, list-of-str no longer `""`, and the dict-`messages` chain-result path.
- `on_llm_end`: full text of block content journaled as a held AGENT message, tool-call-only steps stay unjournaled, plain-string strip preserved.
- The dupe-leak regression: a test asserting `parse_chain_result` and `on_llm_end` project the same block content to the identical text that `OriginatingJournal`'s equality check compares.

Verified end-to-end with the keyless mock network (`chat_mock_llm_echo` with the thinking marker): a thinking-first response produces the clean answer text.

## Review follow-ups (second pass)

A deeper review pass on this diff surfaced one gap that is fixed here rather than deferred:

- **`OriginatingJournal`'s pending-dupe comparison is strip-normalized** (`_is_dupe_of_pending`, new `test_originating_journal.py`). This is the journal dedupe item from the #1222 ladder pulled forward: the full-text projection exposes the strip asymmetry between the two capture points (`on_llm_end` strips, `parse_chain_result` doesn't), so without it a trailing whitespace-only text block after the first would flip from dupe-suppressed to leaked. Strip-comparison instead of the originally planned flatten-on-both-sides, which is an identity on plain strings and would not have closed the hole.

Also from that pass: the `on_llm_end` comment no longer claims `ChatGeneration.text` agrees with the shared projection (it differs for legacy typeless text blocks, a shape no supported provider emits — declared out of scope in the comment), the fuller error-detector scan surface is documented as deliberate uniformity with plain-string output, a stale local annotation is widened, and the multi-text-block fixture moved into `ContentFixtures`. Pre-existing issues surfaced by the same pass are tracked in #1222 under "Surfaced by review of #1285".

Part of #1222 (Phase 1), third of the PR ladder described there (after #1276 and #1281).

### DIFF
--- a/neuro_san/internals/journals/originating_journal.py
+++ b/neuro_san/internals/journals/originating_journal.py
@@ -101,11 +101,29 @@ class OriginatingJournal(Journal):
 
         if self.pending is not None:
             # Avoid cases where two different kinds of message hold the same content.
-            if self.pending.content != message.content:
+            if not self._is_dupe_of_pending(message):
                 await self.wrapped_journal.write_message(self.pending, use_origin)
             self.pending = None
 
         await self.wrapped_journal.write_message(message, use_origin)
+
+    def _is_dupe_of_pending(self, message: BaseMessage) -> bool:
+        """
+        :param message: The incoming BaseMessage to compare against the held
+                pending message.
+        :return: True if the pending message carries the same content as the
+                incoming message. String content is compared ignoring leading
+                and trailing whitespace: the pending message's content was
+                stripped when it was captured (JournalingCallbackHandler's
+                on_llm_end), while the incoming message's content is not, so
+                an exact comparison would let a mere trailing newline defeat
+                the suppression and send clients both copies of the same text.
+        """
+        pending_content: Any = self.pending.content
+        incoming_content: Any = message.content
+        if isinstance(pending_content, str) and isinstance(incoming_content, str):
+            return pending_content.strip() == incoming_content.strip()
+        return pending_content == incoming_content
 
     def get_chat_history(self) -> List[BaseMessage]:
         """
--- a/neuro_san/internals/journals/originating_journal.py
+++ b/neuro_san/internals/journals/originating_journal.py
@@ -118,6 +118,14 @@ class OriginatingJournal(Journal):
                 on_llm_end), while the incoming message's content is not, so
                 an exact comparison would let a mere trailing newline defeat
                 the suppression and send clients both copies of the same text.
+
+        Normalizing here, rather than stripping at either capture point, is
+        deliberate: both captured texts are client-visible and locked by
+        backward compatibility (the AGENT copy has always been journaled
+        stripped; the AI text reaches the wire and chat history unstripped),
+        so changing either producer would change client-visible output. The
+        dupe decision is internal, which makes this comparison the one place
+        that can reconcile the two without altering any payload.
         """
         pending_content: Any = self.pending.content
         incoming_content: Any = message.content
--- a/neuro_san/internals/run_context/langchain/core/run_context_runnable.py
+++ b/neuro_san/internals/run_context/langchain/core/run_context_runnable.py
@@ -48,6 +48,7 @@ from neuro_san.internals.run_context.langchain.token_counting.langchain_token_co
 from neuro_san.internals.run_context.langchain.tracing.neuro_san_runnable import NeuroSanRunnable
 from neuro_san.internals.run_context.langchain.util.api_key_error_check import ApiKeyErrorCheck
 from neuro_san.message.types.agent_framework_message import AgentFrameworkMessage
+from neuro_san.message.utils.content_utils import ContentUtils
 
 MINUTES: float = 60.0
 
@@ -375,17 +376,16 @@ class RunContextRunnable(NeuroSanRunnable):
                 # We generally want the content of any single AIMessage we found from above
                 output = ai_message.content
 
-        # In general, output is a string, but it can also be a list of content blocks when there are multiple
-        # message types, such as "thinking", "reasoning", etc.
-        # If it is a list, "text" is a key in the dict containing the actual string output we want to return.
+        # In general, output is a string, but it can also be a list of content blocks when there are
+        # multiple message types, such as "thinking", "reasoning", etc. - or a list of plain strings,
+        # which is also legal per the BaseMessage annotation.
+        # Project list output to its full text - every text block, not just the first one - with the
+        # same projection the wire converter and JournalingCallbackHandler.on_llm_end use. Divergent
+        # projections here made the AI message written below differ from the AGENT message journaled
+        # in on_llm_end, defeating its dupe suppression.
         # For more details: https://docs.langchain.com/oss/python/langchain/messages#standard-content-blocks
         if isinstance(output, list):
-            text_output: str = ""
-            for block in output:
-                if isinstance(block, dict) and block.get("type") == "text":
-                    text_output = block.get("text", "")
-                    break
-            output = text_output
+            output = ContentUtils.flatten_to_text(output)
 
         # See if we had some kind of error and format accordingly, if asked for.
         output = self.error_detector.handle_error(output)
--- a/neuro_san/internals/run_context/langchain/core/run_context_runnable.py
+++ b/neuro_san/internals/run_context/langchain/core/run_context_runnable.py
@@ -339,7 +339,7 @@ class RunContextRunnable(NeuroSanRunnable):
         # Initialize our output.
         # The value here might morph a bit between types, but when we return
         # something we expect it to be a string.
-        output: Union[str, List[Dict[str, Any]]] = None
+        output: Optional[Union[str, List[Any]]] = None
 
         if chain_result is None and exception is not None:
             # We got an exception instead of a proper result. Say so.
@@ -379,10 +379,10 @@ class RunContextRunnable(NeuroSanRunnable):
         # In general, output is a string, but it can also be a list of content blocks when there are
         # multiple message types, such as "thinking", "reasoning", etc. - or a list of plain strings,
         # which is also legal per the BaseMessage annotation.
-        # Project list output to its full text - every text block, not just the first one - with the
-        # same projection the wire converter and JournalingCallbackHandler.on_llm_end use. Divergent
-        # projections here made the AI message written below differ from the AGENT message journaled
-        # in on_llm_end, defeating its dupe suppression.
+        # Project list output to its full text through the single projection policy ContentUtils
+        # defines: OriginatingJournal's dupe suppression relies on the AGENT and AI copies of the
+        # same output projecting to the same text. Full text also means the error detector below
+        # scans everything, exactly as it always has for plain-string output.
         # For more details: https://docs.langchain.com/oss/python/langchain/messages#standard-content-blocks
         if isinstance(output, list):
             output = ContentUtils.flatten_to_text(output)
--- a/neuro_san/internals/run_context/langchain/journaling/journaling_callback_handler.py
+++ b/neuro_san/internals/run_context/langchain/journaling/journaling_callback_handler.py
@@ -37,6 +37,7 @@ from neuro_san.internals.journals.origination import Origination
 from neuro_san.internals.journals.tool_argument_reporting import ToolArgumentReporting
 from neuro_san.message.types.agent_message import AgentMessage
 from neuro_san.message.types.agent_tool_result_message import AgentToolResultMessage
+from neuro_san.message.utils.content_utils import ContentUtils
 
 
 # pylint: disable=too-many-ancestors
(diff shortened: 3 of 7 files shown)


Noravee:
TLDR for reviewers: two one-line production changes, everything else is tests.

- `parse_chain_result` had the real bug: its inline flatten stopped at the first `type=="text"` block (dropping any text after it) and skipped plain strings entirely (list-of-str content became `""`). It now projects through `ContentUtils.flatten_to_text`.
- `on_llm_end` moves from `ChatGeneration.text` onto the same shared projection. On the pinned langchain-core the two agree, so this is not a behavior change — it removes the *possibility* of divergence. When these two projections disagree, `OriginatingJournal`'s dupe suppression fails (it compares content for equality) and clients receive both the AGENT and the AI copy of the same output.
- Preserved exactly: plain-string handling at both sites, the `on_llm_end` journaling gate (tool-call-only steps flatten to `""` and stay unjournaled) and its `.strip()`, and the error-detector pass.
- After the wire converter fix, this makes all three flatten points in the pipeline use one projection policy. Verified end-to-end with the keyless `chat_mock_llm_echo` network using the thinking marker.

neuro_san/internals/run_context/langchain/core/run_context_runnable.py
```diff
@@ -375,17 +376,16 @@ def parse_chain_result(self, chain_result: Optional[Union[Dict[str, Any], AgentF
...
+        # in on_llm_end, defeating its dupe suppression.
         # For more details: https://docs.langchain.com/oss/python/langchain/messages#standard-content-blocks
         if isinstance(output, list):
-            text_output: str = ""
-            for block in output:
-                if isinstance(block, dict) and block.get("type") == "text":
-                    text_output = block.get("text", "")
-                    break
-            output = text_output
+            output = ContentUtils.flatten_to_text(output)
```
Noravee: The substantive fix. The old loop here returned the first type=="text" block and broke - so a response like [text, thinking, text] lost its second text block, and list-of-str content (legal per the BaseMessage annotation) skipped the non-dict items and came out as "". This output becomes the AIMessage written to the journal and the text every client sees for this agent's step.

neuro_san/internals/run_context/langchain/journaling/journaling_callback_handler.py
```diff
@@ -112,8 +113,15 @@ async def on_llm_end(self, response: LLMResult,
...
-            content: str = first_generation.text
-            if content is not None and len(content) > 0:
+            # Project the message content to its full text with the same
+            # projection the wire converter and parse_chain_result use.
+            # ChatGeneration.text happens to agree on the pinned langchain-core,
+            # but its list-content semantics have varied across versions;
+            # capturing through the one shared policy is what keeps the dupe
+            # comparison against the later AI message reliable. Tool-call-only
+            # steps flatten to "" and stay unjournaled, as before.
+            content: str = ContentUtils.flatten_to_text(first_generation.message)
```
Noravee: Not a behavior change on the pinned langchain-core - ChatGeneration.text agrees with flatten_to_text there. The point is convergence: this AGENT message is held by OriginatingJournal and suppressed only if the NEXT message's content is equal, and the next message is built from parse_chain_result's output. Two independent projections of the same content is how the dupe leak happened; now both sites share one.

tests/neuro_san/internals/run_context/langchain/journaling/test_journaling_callback_handler.py
```diff
@@ -42,8 +57,49 @@ def _make_handler():
...
+        message, held for dupe comparison against the AI message that follows.
+        """
+        handler, journal = self._make_handler()
+        await handler.on_llm_end(self._llm_result(ContentFixtures.anthropic_thinking_first()))
+        message = journal.write_message_if_next_not_dupe.call_args.args[0]
+        assert isinstance(message, AgentMessage)
+        assert message.content == "the answer"
+
+    @pytest.mark.asyncio
+    async def test_on_llm_end_tool_call_only_step_stays_unjournaled(self):
```
Noravee: Guards the journaling gate: a tool-call-only step (Anthropic tool_use block, no text) must stay unjournaled, otherwise every tool call would emit an empty AGENT message ahead of the "Invoking:" message.

tests/neuro_san/internals/run_context/langchain/core/test_run_context_runnable.py
```diff
@@ -22,14 +22,110 @@
...
+        """
+        runnable = self._make_runnable()
+        chain_result = {"messages": [
+            HumanMessage(content="question"),
+            ContentFixtures.anthropic_thinking_first(),
+        ]}
+        assert runnable.parse_chain_result(chain_result, exception=None) == "the answer"
+
+    @pytest.mark.asyncio
+    async def test_parse_chain_result_matches_on_llm_end_projection(self):
```
Noravee: The dupe-leak regression test: runs the same block content through both capture points and asserts they produce the identical text that OriginatingJournal's equality check compares. With the old divergent flattens, this pair was ("part one, ", "part one, part two") and both messages reached clients.

neuro_san/internals/journals/originating_journal.py
```diff
@@ -101,12 +101,30 @@ async def write_message(self, message: BaseMessage, origin: List[Dict[str, Any]]
...
         if self.pending is not None:
             # Avoid cases where two different kinds of message hold the same content.
-            if self.pending.content != message.content:
+            if not self._is_dupe_of_pending(message):
                 await self.wrapped_journal.write_message(self.pending, use_origin)
             self.pending = None
 
         await self.wrapped_journal.write_message(message, use_origin)
 
+    def _is_dupe_of_pending(self, message: BaseMessage) -> bool:
```
Noravee: Flagging the one file here that is outside the two capture points: this is the originating_journal dedupe item from the #1222 ladder's str-safety rung, pulled forward. The full-text projection exposes a strip asymmetry between the two capture points (on_llm_end strips its journaled AGENT copy, parse_chain_result does not), so without this a trailing whitespace-only text block after the first would flip from dupe-suppressed to leaked - clients would get both the AGENT and AI copies. Strip-comparison is used rather than the originally planned flatten-on-both-sides because flatten_to_text is an identity on two plain strings and would not have closed the whitespace hole. Covered by the new test_originating_journal.py.

Noravee:
TLDR: three small production changes, everything else is tests.

- `parse_chain_result` had the real bug: its inline flatten stopped at the first `type=="text"` block (dropping any text after it) and skipped plain strings entirely (list-of-str content became `""`). It now projects through `ContentUtils.flatten_to_text`.
- `on_llm_end` moves from `ChatGeneration.text` onto the same shared projection so the two capture points can never diverge. Near-identical on the pinned langchain-core, with one known, deliberate exception: legacy typeless `{"text": ...}` blocks (no `"type"` key) are collected by `ChatGeneration.text` but not by the shared projection — no supported provider emits that shape, and the call-site comment declares it out of scope.
- `OriginatingJournal`'s pending-dupe comparison is strip-normalized (`_is_dupe_of_pending`). Pulled forward from the #1222 str-safety rung: `on_llm_end` strips its journaled AGENT copy while `parse_chain_result` does not, so with the full-text flatten a trailing whitespace-only block after the first would flip from dupe-suppressed to leaked. Strip-comparison rather than the originally planned flatten-on-both-sides, because flatten is an identity on plain strings and would not have closed the whitespace hole.
- Preserved exactly: plain-string handling at both capture points, the `on_llm_end` journaling gate (tool-call-only steps flatten to `""` and stay unjournaled) and its `.strip()`, and the error-detector pass — which now scans the full flattened text, exactly as it always has for plain-string output.
- After the wire converter fix, all three flatten points in the pipeline use one projection policy. Merges cleanly with the descendant-events scope guard from #1259 (its `_is_own_call()` gate sits above this projection in `on_llm_end`). Verified end-to-end with the keyless `chat_mock_llm_echo` network using the thinking marker.

neuro_san/internals/run_context/langchain/core/run_context_runnable.py
```diff
@@ -382,17 +383,16 @@ def parse_chain_result(self, chain_result: Optional[Union[Dict[str, Any], AgentF
...
+        # scans everything, exactly as it always has for plain-string output.
         # For more details: https://docs.langchain.com/oss/python/langchain/messages#standard-content-blocks
         if isinstance(output, list):
-            text_output: str = ""
-            for block in output:
-                if isinstance(block, dict) and block.get("type") == "text":
-                    text_output = block.get("text", "")
-                    break
-            output = text_output
+            output = ContentUtils.flatten_to_text(output)
```
Noravee, line 395: The substantive fix. The old loop here returned the first type=="text" block and broke - so a response like [text, thinking, text] lost its second text block, and list-of-str content (legal per the BaseMessage annotation) skipped the non-dict items and came out as "". This output becomes the AIMessage written to the journal and the text every client sees for this agent's step. Full text also means the error detector below scans everything, exactly as it always has for plain-string output.

neuro_san/internals/run_context/langchain/journaling/journaling_callback_handler.py
```diff
@@ -160,8 +161,16 @@ async def on_llm_end(self, response: LLMResult,
...
-            if content is not None and len(content) > 0:
+            # Project the message content to its full text through the single
+            # projection policy ContentUtils defines: the dupe comparison
+            # against the later AI message relies on both copies projecting to
+            # the same text. ChatGeneration.text is close but not identical -
+            # it also collects legacy text blocks that carry no "type" key, a
+            # shape no supported provider emits and which is deliberately out
+            # of scope here. Tool-call-only steps flatten to "" and are
+            # skipped by the gate below.
+            content: str = ContentUtils.flatten_to_text(first_generation.message)
```
Noravee, line 172: Convergence is the point: this AGENT message is held by OriginatingJournal and suppressed only if the NEXT message carries the same content, and that next message is built from parse_chain_result's output. Two independent projections of the same content is how the dupe leak happened; now both sites share one. One known, deliberate divergence from ChatGeneration.text: legacy typeless {"text": ...} blocks (no "type" key) are collected by ChatGeneration.text but not by the shared projection - no supported provider emits that shape, so the comment here declares it out of scope.

neuro_san/internals/journals/originating_journal.py
Noravee, line 110: Flagging the one file here that is outside the two capture points: this is the originating_journal dedupe item from the #1222 ladder's str-safety rung, pulled forward. The full-text projection exposes a strip asymmetry between the two capture points (on_llm_end strips its journaled AGENT copy, parse_chain_result does not), so without this a trailing whitespace-only text block after the first would flip from dupe-suppressed to leaked - clients would get both the AGENT and AI copies.

Why fix the comparison instead of stripping in parse_chain_result (or unstripping on_llm_end): both captured texts are client-visible and locked by backward compatibility. parse_chain_result's output becomes the AIMessage text on the wire and in chat history, unstripped today - stripping it would change the wire text of every whitespace-padded response from every deployed agent. And on_llm_end's AGENT copy has always been journaled stripped. With both producers pinned, the dupe decision itself - which is purely internal and alters no payload - is the one place that can reconcile them. It is also the single choke point: the pending message is compared against whatever message arrives next (not only the AI answer), and this helper is where the comparison grows a flatten step once native block messages start reaching write_message later in the ladder.

Strip-comparison is used rather than the originally planned flatten-on-both-sides because flatten_to_text is an identity on two plain strings and would not have closed the whitespace hole. Covered by the new test_originating_journal.py.

tests/neuro_san/internals/run_context/langchain/core/test_run_context_runnable.py
```diff
@@ -22,14 +22,104 @@
...
```
Noravee, line 96: The dupe-leak regression test: runs the same block content through both capture points and asserts they project to the identical text. With the old divergent flattens, this pair was ("part one, ", "part one, part two") and both messages reached clients. Whitespace differences between the two projections are absorbed by the journal's strip-normalized dupe comparison, covered separately in test_originating_journal.py.

tests/neuro_san/internals/run_context/langchain/journaling/test_journaling_callback_handler.py
```diff
@@ -47,8 +62,49 @@ def _make_handler() -> Tuple[JournalingCallbackHandler, MagicMock]:
...
+        message, held for dupe comparison against the AI message that follows.
+        """
+        handler, journal = self._make_handler()
+        await handler.on_llm_end(self._llm_result(ContentFixtures.anthropic_thinking_first()))
+        message = journal.write_message_if_next_not_dupe.call_args.args[0]
+        assert isinstance(message, AgentMessage)
+        assert message.content == "the answer"
+
+    @pytest.mark.asyncio
+    async def test_on_llm_end_tool_call_only_step_stays_unjournaled(self) -> None:
```
Noravee, line 83: Guards the journaling gate: a tool-call-only step (Anthropic tool_use block, no text) must stay unjournaled, otherwise every tool call would emit an empty AGENT message ahead of the "Invoking:" message.

d1donlydfink (approved):
Please be sure thinking file output looks more or less the same as (or better than) before.

PR #1289, opened 2026-08-25
NS-1282 Run EventInitiator in only one of neuro-san service instances.

andreidenissov-cog:
## Summary
Run the PeriodicEventInitiator in only one worker of a multi-instance neuro-san
service, instead of in every worker.

When the HTTP server is configured for multiple worker processes
(AGENT_HTTP_SERVER_INSTANCES > 1, where Tornado forks one worker per instance),
each worker previously started its own PeriodicEventInitiator. That means the
periodic events it drives fired once per worker -- duplicated N times -- rather
than once for the service. This change makes the initiator start on exactly one
worker.

## Change
- WatcherThread (the base Startable that runs work on a background thread) gains
  a single_instance flag (default False). When True, start() consults the
  ServerContext for the worker id / worker count and, in a multi-instance
  configuration (num_workers > 1), starts the thread only on worker 0; other
  workers log and skip. A run_single_instance() accessor exposes the flag.
- PeriodicEventInitiator opts in by passing single_instance=True to the base
  constructor.

## Behavior
- Single-instance deployments are unchanged: with num_workers <= 1 the guard is
  a no-op and the thread always starts.
- Other WatcherThread subclasses default to single_instance=False, so their
  behavior is unchanged (they still start in every worker).
- Relies on the worker id / worker count already published on the ServerContext
  during the fork (get_worker_id / get_num_workers).

## Testing
- flake8 clean and pylint 10.00/10 on both changed files.
- Behavior is gated purely by ServerContext worker id / count, so a
  single-instance run is unaffected; in a multi-worker run only worker 0 starts
  the initiator (other workers log "Not starting ... because it is configured as
  a single instance").
- Tested by running 5-instance neuro-san server and verifying that only one PeriodicEventInitiator is started.

## Related issue
NS-1282

### DIFF
--- a/neuro_san/service/watcher/event_initiator/periodic_event_initiator.py
+++ b/neuro_san/service/watcher/event_initiator/periodic_event_initiator.py
@@ -53,7 +53,7 @@ class PeriodicEventInitiator(WatcherThread):
         """
         Constructor
         """
-        super().__init__(server_context)
+        super().__init__(server_context, single_instance=True)
         self.verbose: bool = False
         # NB: do NOT construct the AsyncioExecutor here. When Tornado is
         # configured for multiple worker processes (AGENT_HTTP_SERVER_INSTANCES
--- a/neuro_san/service/watcher/interfaces/watcher_thread.py
+++ b/neuro_san/service/watcher/interfaces/watcher_thread.py
@@ -23,6 +23,7 @@ from threading import Thread
 from time import sleep
 
 from leaf_common.utils.startable import Startable
+from xdist.plugin import worker_id
 
 from neuro_san.service.utils.server_context import ServerContext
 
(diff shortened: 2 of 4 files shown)


neuro_san/service/watcher/interfaces/watcher_thread.py
```diff
@@ -32,26 +32,46 @@ class WatcherThread(Startable):
...
+            if num_workers > 1 and work_id != 0:
+                self.logger.info("Not starting %s in worker %d because it is configured as a single instance",
+                                 self.__class__.__name__, work_id)
+                return
         self.logger.info("Starting %s with %f seconds period",
                          self.__class__.__name__, self.update_period_in_seconds)
         self.update_thread.start()
 
+    def run_single_instance(self) -> bool:
+        """
```
andreidenissov-cog, line 69: Getter for "single instance only" property.

neuro_san/service/watcher/event_initiator/periodic_event_initiator.py
```diff
@@ -53,7 +53,7 @@ def __init__(self, server_context: ServerContext):
         """
         Constructor
         """
-        super().__init__(server_context)
+        super().__init__(server_context, single_instance=True)
         self.verbose: bool = False
```
andreidenissov-cog, line 57: We want PeriodicEventInitiator to only run in one neuro-san instance,
so setup it accordingly.

d1donlydfink (dismissed):
Awesome!

It also would be good to update the comments in the Dockerfiles removing the doubt about HTTP_INSTANCES now.

d1donlydfink (approved):

PR #1290, opened 2026-08-28
Fix mutation of function tool specifications

### LINKED ISSUE #1230: create_function_tool mutates the shared registry spec of a same-server external agent

arnabnandy7:
## Summary

Fixes #1230.

Prevent `BaseToolFactory.create_function_tool()` from mutating the caller-owned function specification when adding the tool lookup name.

## Changes

- Create a shallow copy of `function_json` before adding the `"name"` field.
- Pass the copied specification to `LangChainOpenAIFunctionTool`.
- Add a regression test confirming tool creation leaves the original specification unchanged.

A shallow copy is sufficient because only the top-level `"name"` field is modified.

### DIFF
--- a/neuro_san/internals/run_context/langchain/core/base_tool_factory.py
+++ b/neuro_san/internals/run_context/langchain/core/base_tool_factory.py
@@ -369,5 +369,6 @@ class BaseToolFactory:
             message: str = f"Could not create tool to call external agent '{name}'. Its function_json is None."
             raise ValueError(message)
 
-        function_json["name"] = name
-        return LangChainOpenAIFunctionTool.from_function_json(function_json, self.tool_caller)
+        use_function_json: Dict[str, Any] = dict(function_json)
+        use_function_json["name"] = name
+        return LangChainOpenAIFunctionTool.from_function_json(use_function_json, self.tool_caller)
--- a/neuro_san/internals/run_context/langchain/core/base_tool_factory.py
+++ b/neuro_san/internals/run_context/langchain/core/base_tool_factory.py
@@ -369,6 +369,14 @@ class BaseToolFactory:
             message: str = f"Could not create tool to call external agent '{name}'. Its function_json is None."
             raise ValueError(message)
 
+        # Copy before adding the name. function_json can be a dictionary that
+        # outlives this call: for a same-server external agent it is the
+        # referenced network's live registry spec (AsyncDirectAgentSession
+        # returns it by reference), and internal and toolbox tools funnel
+        # here with their registry/toolbox entries too. Writing the lookup
+        # name into it would leak this caller's reference string into that
+        # shared state (issue #1230). The copy is deliberately shallow: only
+        # the top-level "name" key is written here, so nested dicts stay shared.
         use_function_json: Dict[str, Any] = dict(function_json)
         use_function_json["name"] = name
         return LangChainOpenAIFunctionTool.from_function_json(use_function_json, self.tool_caller)
--- a/tests/neuro_san/internals/run_context/langchain/core/test_base_tool_factory.py
+++ b/tests/neuro_san/internals/run_context/langchain/core/test_base_tool_factory.py
@@ -122,6 +122,31 @@ class TestBaseToolFactory:
         assert list(tool.args_schema.__fields__.keys()) == ["question"]
         factory.journal.write_message.assert_not_awaited()
 
+    def test_create_function_tool_does_not_mutate_function_json(self):
+        """
+        Creating a tool must not add its lookup name to the caller-owned
+        function specification.
+        """
+        function_json: Dict[str, Any] = {
+            "description": "Answers music questions.",
+            "parameters": {
+                "type": "object",
+                "properties": {
+                    "question": {
+                        "type": "string",
+                        "description": "The question to answer."
+                    }
+                },
+                "required": ["question"]
+            }
+        }
+        factory = self.make_factory(function_json)
+
+        tool = factory.create_function_tool(function_json, self.EXTERNAL_AGENT_NAME)
+
+        assert tool is not None
+        assert "name" not in function_json
+
     @pytest.mark.asyncio
     async def test_external_tool_with_empty_properties_gets_default_schema(self):
         """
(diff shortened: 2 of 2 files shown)


neuro_san/internals/run_context/langchain/core/base_tool_factory.py
```diff
@@ -369,5 +369,6 @@ def create_function_tool(self, function_json: Dict[str, Any], name: str) -> Base
             message: str = f"Could not create tool to call external agent '{name}'. Its function_json is None."
             raise ValueError(message)
 
-        function_json["name"] = name
-        return LangChainOpenAIFunctionTool.from_function_json(function_json, self.tool_caller)
+        use_function_json: Dict[str, Any] = dict(function_json)
```
Noravee, line 380: Could you add a short "why" comment here? This file explains the reasoning behind most of its non-obvious lines, and without one the copy looks removable to a future reader. Something like:

```suggestion
        # Copy before adding the name. function_json can be a dictionary that
        # outlives this call: for a same-server external agent it is the
        # referenced network's live registry spec (AsyncDirectAgentSession
        # returns it by reference), and internal and toolbox tools funnel
        # here with their registry/toolbox entries too. Writing the lookup
        # name into it would leak this caller's reference string into that
        # shared state (issue #1230). The copy is deliberately shallow: only
        # the top-level "name" key is written here, so nested dicts stay shared.
        use_function_json: Dict[str, Any] = dict(function_json)
```

tests/neuro_san/internals/run_context/langchain/core/test_base_tool_factory.py
```diff
@@ -122,6 +122,31 @@ async def test_external_tool_with_parameters_is_untouched(self):
...
+                },
+                "required": ["question"]
+            }
+        }
+        factory = self.make_factory(function_json)
+
+        tool = factory.create_function_tool(function_json, self.EXTERNAL_AGENT_NAME)
+
+        assert tool is not None
+        assert "name" not in function_json
```
Noravee: Optional, but two small tightenings would make this pin the whole contract rather than just the `name` key:

- Snapshot the spec before the call and compare after, so any other in-place edit is caught too, and assert that the copy actually carried the name through to the tool (needs `from copy import deepcopy` and the `ExternalAgentParsing` import):

```python
        expected: Dict[str, Any] = deepcopy(function_json)
        ...
        assert function_json == expected
        assert tool.name == ExternalAgentParsing.get_safe_agent_name(self.EXTERNAL_AGENT_NAME)
```

- As written, the mocked session set up by `make_factory(function_json)` is never consulted, because `create_function_tool()` is called directly. An async sibling that does `await factory.create_external_tool(self.EXTERNAL_AGENT_NAME)` with the same assertions would exercise the by-reference path from #1230 itself. I checked that such a variant fails on main and passes on this branch.

Happy to merge without these. The comment above is the only thing I would like in before merging.

Noravee (dismissed):
Thanks for picking this up, and for the regression test. The fix is exactly what #1230 called for, and it covers all three paths that funnel into `create_function_tool()`: internal tools, same-server external agents, and toolbox shared coded tools.

I verified locally that the new test fails on main and passes on this branch, and CI is green on all three Python versions with main merged in.

Approving. One small thing before I merge: please add a short comment above the copy explaining why it is there (inline suggestion below). The test tightening in the other inline comment is optional.

neuro_san/internals/run_context/langchain/core/base_tool_factory.py
arnabnandy7: @Noravee it's done. I added the rationale above the shallow copy, explaining that the function specification may be shared registry or toolbox state, why mutating it leaks the caller’s lookup name, and why a shallow copy is sufficient.

tests/neuro_san/internals/run_context/langchain/core/test_base_tool_factory.py
arnabnandy7: @Noravee it's done as well. The direct test now compares the complete original specification and verifies the sanitized tool name. I also added an async regression test that exercises the external-agent path through the mocked session. The focused suite now passes with 16 tests.

Noravee (approved):
LGTM.

PR #1291, opened 2026-08-28
Validate MCP tools list response

### LINKED ISSUE #1252: Unchecked Input for Loop Condition in McpServiceAgentSession
From upgraded Checkmark scan:

Source
File Name: /neuro_san/session/mcp_service_agent_session.py
Method: function
Element: loads

Destination
File Name: /neuro_san/session/mcp_service_agent_session.py
Method: function
Element: tools_list

Code Snippets
157 response_dict = json.loads(response.text)
162 for tool in tools_list:

arnabnandy7:
## Summary

Fixes #1252.

Validates the MCP `tools/list` response before iterating over tools.

## Changes

- Validate that the JSON-RPC response and `result` are objects.
- Validate that `tools` is an array.
- Reject responses containing more than 1,000 tools.
- Validate that every tool entry is an object.
- Add unit tests for valid, malformed, and oversized responses.

### DIFF
--- a/neuro_san/session/mcp_service_agent_session.py
+++ b/neuro_san/session/mcp_service_agent_session.py
@@ -17,7 +17,6 @@
 
 from typing import Any
 from typing import Dict
-from typing import List
 from typing import Generator
 
 import json
@@ -41,6 +40,7 @@ class McpServiceAgentSession(AbstractHttpServiceAgentSession, AgentSession):
     This is largely only used by command-line tests.
     """
     MCP_PROTOCOL_VERSION: str = "MCP-Protocol-Version"
+    MAX_TOOLS: int = 1000
 
     # pylint: disable=too-many-arguments,too-many-positional-arguments, too-many-locals
     def __init__(self, host: str = None,
@@ -159,8 +159,21 @@ class McpServiceAgentSession(AbstractHttpServiceAgentSession, AgentSession):
         except Exception as exc:  # pylint: disable=broad-exception-caught
             raise ValueError(self.help_message(path)) from exc
 
-        tools_list: List[Dict[str, Any]] = response_dict.get("result", {}).get("tools", [])
+        if not isinstance(response_dict, dict):
+            raise ValueError("Invalid MCP tools/list response: response must be an object")
+        result_dict: Any = response_dict.get("result", {})
+        if not isinstance(result_dict, dict):
+            raise ValueError("Invalid MCP tools/list response: 'result' must be an object")
+
+        tools_list: Any = result_dict.get("tools", [])
+        if not isinstance(tools_list, list):
+            raise ValueError("Invalid MCP tools/list response: 'tools' must be an array")
+        if len(tools_list) > self.MAX_TOOLS:
+            raise ValueError(f"Invalid MCP tools/list response: too many tools (maximum {self.MAX_TOOLS})")
+
         for tool in tools_list:
+            if not isinstance(tool, dict):
+                raise ValueError("Invalid MCP tools/list response: each tool must be an object")
             name: str = tool.get("name", None)
             if name == self.agent_name:
                 tool_description: str = tool.get("description", None)
--- a/neuro_san/session/mcp_service_agent_session.py
+++ b/neuro_san/session/mcp_service_agent_session.py
@@ -21,10 +21,13 @@ from typing import List
 from typing import Generator
 
 import json
+import jsonschema
 import requests
 
+from leaf_common.serialization.util.text_file_reader import TextFileReader
 from leaf_common.time.timeout import Timeout
 
+from neuro_san import TOP_LEVEL_DIR
 from neuro_san.interfaces.agent_session import AgentSession
 from neuro_san.session.abstract_http_service_agent_session import AbstractHttpServiceAgentSession
 from neuro_san.session.mcp_chat_response_dictionary_converter import McpChatResponseDictionaryConverter
@@ -41,6 +44,7 @@ class McpServiceAgentSession(AbstractHttpServiceAgentSession, AgentSession):
     This is largely only used by command-line tests.
     """
     MCP_PROTOCOL_VERSION: str = "MCP-Protocol-Version"
+    _protocol_schema: Dict[str, Any] = None
 
     # pylint: disable=too-many-arguments,too-many-positional-arguments, too-many-locals
     def __init__(self, host: str = None,
@@ -112,7 +116,7 @@ class McpServiceAgentSession(AbstractHttpServiceAgentSession, AgentSession):
             raise ValueError(self.help_message(path)) from exc
 
         # Extract the protocol version from the handshake response
-        result_dict: Dict[str, Any] = self._get_result_dict(response_dict, "initialize")
+        result_dict: Dict[str, Any] = self._get_result_dict(response_dict, "initialize", "InitializeResult")
         self.protocol_version: str = result_dict.get("protocolVersion", None)
 
         # Confirm the protocol version is supported by this client
@@ -165,10 +169,8 @@ class McpServiceAgentSession(AbstractHttpServiceAgentSession, AgentSession):
         # CheckMarx false positive (Unchecked Input for Loop Condition, #1252):
         # tools_list has already been fully parsed by json.loads above, so this
         # O(n) loop with early return cannot cost more than the parse that
-        # preceded it. The isinstance checks above validate the shape.
+        # preceded it. The MCP schema validation above validates the shape.
         for tool in tools_list:
-            if not isinstance(tool, dict):
-                continue
             name: str = tool.get("name", None)
             if name == self.agent_name:
                 tool_description: str = tool.get("description", None)
@@ -179,10 +181,10 @@ class McpServiceAgentSession(AbstractHttpServiceAgentSession, AgentSession):
 
         return None
 
-    @staticmethod
-    def _get_result_dict(response_dict: Any, method: str) -> Dict[str, Any]:
+    @classmethod
+    def _get_result_dict(cls, response_dict: Any, method: str, result_definition: str) -> Dict[str, Any]:
         """
-        Validates an MCP JSON-RPC response and returns its result object.
+        Validates an MCP JSON-RPC response and its result against the protocol schema.
         """
         if not isinstance(response_dict, dict):
             raise ValueError(f"Invalid MCP {method} response: response must be an object")
@@ -194,18 +196,40 @@ class McpServiceAgentSession(AbstractHttpServiceAgentSession, AgentSession):
         result_dict: Any = response_dict.get("result", None)
         if not isinstance(result_dict, dict):
             raise ValueError(f"Invalid MCP {method} response: 'result' must be an object")
+
+        protocol_schema: Dict[str, Any] = cls._get_protocol_schema()
+        validation_schema: Dict[str, Any] = {
+            "$schema": protocol_schema.get("$schema"),
+            "$ref": f"#/definitions/{result_definition}",
+            "definitions": protocol_schema.get("definitions", {}),
+        }
+        try:
+            jsonschema.validate(instance=result_dict, schema=validation_schema)
+        except jsonschema.exceptions.ValidationError as exc:
+            raise ValueError(
+                f"Invalid MCP {method} response: 'result' does not match {result_definition}"
+            ) from exc
         return result_dict
+
+    @classmethod
+    def _get_protocol_schema(cls) -> Dict[str, Any]:
+        """
+        Loads and caches the MCP protocol schema supported by this session.
+        """
+        if cls._protocol_schema is None:
+            schema_name: str = f"service/mcp/validation/mcp-schema-{MCP_VERSION}.json"
+            schema_path = TOP_LEVEL_DIR.get_file_in_basis(schema_name)
+            schema_str: str = TextFileReader.read_text_file(schema_path)
+            cls._protocol_schema = json.loads(schema_str)
+        return cls._protocol_schema
 
     @classmethod
     def _get_tools_list(cls, response_dict: Any) -> List[Any]:
         """
         Validates an MCP tools/list response and returns its tools array.
         """
-        result_dict: Dict[str, Any] = cls._get_result_dict(response_dict, "tools/list")
-        tools_list: Any = result_dict.get("tools", None)
-        if not isinstance(tools_list, list):
-            raise ValueError("Invalid MCP tools/list response: 'tools' must be an array")
-        return tools_list
+        result_dict: Dict[str, Any] = cls._get_result_dict(response_dict, "tools/list", "ListToolsResult")
+        return result_dict["tools"]
 
     def connectivity(self, request_dict: Dict[str, Any]) -> Dict[str, Any]:
         """
--- a/neuro_san/session/mcp_service_agent_session.py
+++ b/neuro_san/session/mcp_service_agent_session.py
@@ -17,6 +17,7 @@
 
 from typing import Any
 from typing import Dict
+from typing import List
 from typing import Generator
 
 import json
@@ -40,7 +41,6 @@ class McpServiceAgentSession(AbstractHttpServiceAgentSession, AgentSession):
     This is largely only used by command-line tests.
     """
     MCP_PROTOCOL_VERSION: str = "MCP-Protocol-Version"
-    MAX_TOOLS: int = 1000
 
     # pylint: disable=too-many-arguments,too-many-positional-arguments, too-many-locals
     def __init__(self, host: str = None,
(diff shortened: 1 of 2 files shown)


neuro_san/session/mcp_service_agent_session.py
```diff
@@ -159,8 +159,21 @@ def function(self, request_dict: Dict[str, Any]) -> Dict[str, Any]:
...
+            raise ValueError("Invalid MCP tools/list response: response must be an object")
+        result_dict: Any = response_dict.get("result", {})
+        if not isinstance(result_dict, dict):
+            raise ValueError("Invalid MCP tools/list response: 'result' must be an object")
+
+        tools_list: Any = result_dict.get("tools", [])
+        if not isinstance(tools_list, list):
+            raise ValueError("Invalid MCP tools/list response: 'tools' must be an array")
+        if len(tools_list) > self.MAX_TOOLS:
+            raise ValueError(f"Invalid MCP tools/list response: too many tools (maximum {self.MAX_TOOLS})")
```
Noravee: Please drop the `MAX_TOOLS` length check (keep the `isinstance` checks above and below, those are a real improvement).

Reasoning:
- By the time this line runs, `json.loads` at line 158 has already downloaded and fully parsed the body, so the cap does not bound any meaningful resource. The loop it guards is O(n) dict lookups with an early return, which is strictly cheaper than the parse that already happened.
- MCP defines no maximum number of tools (the vendored `mcp-schema-2025-06-18.json` `ListToolsResult.tools` has no `maxItems`), so a 1001-tool response is a valid response, and the message wrongly calls it "Invalid".
- Our own server can produce that: `McpToolsProcessor.list_tools()` returns one tool per MCP-exposed agent with no pagination and no bound, and shorthand manifest entries (`"x.hocon": true`) default to `mcp: true` via `ManifestDictConfigFilter.MCP_DEFAULT_MODE`. A large deployment would be locked out of its own server by the client, and there is no constructor or env override for the limit.
- The `ValueError` is raised outside the `try`, so it reaches the CLI user raw via `agent_cli.py:147-150`.

Instead, annotate the Checkmarx finding as a false positive the way the repo does elsewhere (e.g. `abstract_http_service_agent_session.py:88-93`), something like:

```python
        # CheckMarx false positive (Unchecked Input for Loop Condition, #1252):
        # tools_list has already been fully parsed by json.loads above, so this
        # O(n) loop with early return cannot cost more than the parse that
        # preceded it. The isinstance checks above validate the shape.
        for tool in tools_list:
```

If the team decides a numeric bound must stay to clear the scanner, then please: (a) add a comment saying the number is arbitrary and not an MCP limit, (b) raise the default substantially, and (c) `break` instead of `raise` so an agent found before the bound is still returned.

neuro_san/session/mcp_service_agent_session.py
```diff
@@ -41,6 +40,7 @@ class McpServiceAgentSession(AbstractHttpServiceAgentSession, AgentSession):
     This is largely only used by command-line tests.
     """
     MCP_PROTOCOL_VERSION: str = "MCP-Protocol-Version"
+    MAX_TOOLS: int = 1000
```
Noravee: See the comment on lines 171-172. If this constant survives, it needs a comment explaining where 1000 comes from and that it is a scanner-driven sanity bound rather than an MCP or neuro-san limit. Every other class-level `MAX_*` constant in the repo carries that kind of justification (e.g. `MAX_POOL_CONNECTIONS`, `MAX_CONCURRENT_WRITES`).

neuro_san/session/mcp_service_agent_session.py
```diff
@@ -159,8 +159,21 @@ def function(self, request_dict: Dict[str, Any]) -> Dict[str, Any]:
         except Exception as exc:  # pylint: disable=broad-exception-caught
             raise ValueError(self.help_message(path)) from exc
 
-        tools_list: List[Dict[str, Any]] = response_dict.get("result", {}).get("tools", [])
+        if not isinstance(response_dict, dict):
+            raise ValueError("Invalid MCP tools/list response: response must be an object")
+        result_dict: Any = response_dict.get("result", {})
```
Noravee: **Optional (non-blocking):** a JSON-RPC response carries exactly one of `result` or `error`. An error body like `{"jsonrpc": "2.0", "id": 1, "error": {"code": -32602, "message": "..."}}` has no `result`, so this defaults to `{}`, `tools` becomes `[]`, and `function()` returns `None`. The CLI then tells the user the agent "is not implemented on the server", hiding the server's actual error. Since this PR is now the place that decides what a malformed tools/list response means, consider:

```python
        error_dict: Any = response_dict.get("error", None)
        if isinstance(error_dict, dict):
            raise ValueError(f"MCP tools/list error {error_dict.get('code')}: {error_dict.get('message')}")
```

Our own server pairs error objects with 4xx/5xx so `raise_for_status()` catches them today, but third-party servers commonly return 200 plus an error body.

neuro_san/session/mcp_service_agent_session.py
```diff
@@ -159,8 +159,21 @@ def function(self, request_dict: Dict[str, Any]) -> Dict[str, Any]:
...
+        if not isinstance(result_dict, dict):
+            raise ValueError("Invalid MCP tools/list response: 'result' must be an object")
+
+        tools_list: Any = result_dict.get("tools", [])
```
Noravee: **Optional (non-blocking):** `tools` is a required key on `ListToolsResult` in the schema. Right now `"tools": null` is rejected but a missing `tools` key is silently accepted as `[]`. Defaulting to `None` makes the two cases consistent and reuses the check right below:

```python
        tools_list: Any = result_dict.get("tools", None)
```

neuro_san/session/mcp_service_agent_session.py
```diff
@@ -159,8 +159,21 @@ def function(self, request_dict: Dict[str, Any]) -> Dict[str, Any]:
...
+        if not isinstance(tools_list, list):
+            raise ValueError("Invalid MCP tools/list response: 'tools' must be an array")
+        if len(tools_list) > self.MAX_TOOLS:
+            raise ValueError(f"Invalid MCP tools/list response: too many tools (maximum {self.MAX_TOOLS})")
+
         for tool in tools_list:
+            if not isinstance(tool, dict):
+                raise ValueError("Invalid MCP tools/list response: each tool must be an object")
```
Noravee: **Optional (non-blocking):** consider `continue` (optionally with a log line) instead of `raise` here. Two reasons:

- Our own server can put a literal `null` in the list: `McpToolsProcessor.list_tools()` appends the return value of `_get_tool_description()` unconditionally (line 82), and that helper returns `None` when `allow_agent()` disagrees with `list_agents()` (lines 221-222). With a non-default authorizer that can happen, and one bad entry for some *other* agent would then block discovery of the agent the user actually asked for.
- The check is positional: `[match, "garbage"]` returns the match while `["garbage", match]` raises, so whether a response is accepted depends on which agent name was requested.

Skipping bad entries satisfies the scanner just as well and is strictly more useful. (The unconditional append on the server side is a separate bug worth its own issue.)

tests/neuro_san/session/test_mcp_service_agent_session.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def call_function(self, tools: Any) -> Dict[str, Any]:
+        """
+        Calls function() with a tools/list response containing tools.
+        """
+        initialize_response: MagicMock = self.create_response({
+            "result": {"protocolVersion": MCP_VERSION}
+        })
+        initialized_response: MagicMock = self.create_response({})
+        tools_response: MagicMock = self.create_response({"result": {"tools": tools}})
```
Noravee: The envelope is hard-coded here, so two of the five new guards are unreachable from the tests: the top-level `isinstance(response_dict, dict)` check (source 162-163) and the `isinstance(result_dict, dict)` check (source 165-166). Deleting either leaves this file green.

Please let the helper take a full response body, with the current helper delegating to it, e.g.:

```python
    def call_function_with_response(self, response_dict: Any) -> Dict[str, Any]:
        ...
        tools_response: MagicMock = self.create_response(response_dict)
        ...

    def call_function(self, tools: Any) -> Dict[str, Any]:
        return self.call_function_with_response({"result": {"tools": tools}})
```

and add cases for at least: a top-level list (`[]`), `{"result": null}`, `{"result": []}`, and `{"result": {"tools": null}}`. Each raise site in `function()` should have one input that reaches it.

tests/neuro_san/session/test_mcp_service_agent_session.py
```diff
@@ -0,0 +1,82 @@
...
+        """
+        invalid_tools = (
+            "hello_world",
+            ["hello_world"],
+            [{}] * (McpServiceAgentSession.MAX_TOOLS + 1),
+        )
+
+        for tools in invalid_tools:
+            with self.subTest(tools_type=type(tools), tools_length=len(tools)):
+                with self.assertRaisesRegex(ValueError, "Invalid MCP tools/list response"):
```
Noravee: All five new `ValueError`s share this prefix, so the assertion cannot tell which guard fired. This is not theoretical: I removed the `'tools' must be an array` check locally and the whole file still passed, because `"hello_world"` has `len == 11`, iterating it yields `'h'`, and the per-tool check raises with the same prefix.

Please pair each input with the message it must produce, e.g.:

```python
        invalid_cases = (
            ("hello_world", "'tools' must be an array"),
            (["hello_world"], "each tool must be an object"),
        )
        for tools, expected_suffix in invalid_cases:
            with self.subTest(tools=tools):
                with self.assertRaisesRegex(ValueError, re.escape(expected_suffix)):
                    self.call_function(tools)
```

(or `assertRaises` plus `assertEqual` on `str(context.exception)`).

Noravee (changes requested):
Thanks for picking this up. The `isinstance` validation is a genuine improvement over the `AttributeError` crashes on main and I would like to see it land. I am requesting changes for three things, with a few optional notes inline.

**Required before merge**

1. **Remove the `MAX_TOOLS` cap** (or make it non-fatal and documented). See the inline comment on lines 171-172: it runs after the body is already fully parsed, MCP defines no maximum, and our own server can legitimately return more than 1000 tools, so this trades zero protection for a new hard failure.
2. **Tests should cover every guard and assert the specific message.** Two of the five checks are unreachable from the current helper, and the shared message prefix lets a deleted check pass unnoticed (verified by mutation locally). See the inline comments on the test file.
3. **Update the branch.** It is 54 commits behind `main` (no conflicts). The gate now runs unit tests on 3.12, 3.13 and 3.14 and this PR's only CI run predates that.

**Optional / follow-ups (not blocking)**

- Handle a JSON-RPC `error` body and treat a missing `tools` key like `null` (inline on 164 and 168).
- `continue` instead of `raise` on a non-dict tool entry (inline on 175-176).
- `__init__` still reads the initialize response with the same unchecked `.get("result", {}).get(...)` pattern at line 115. A three-line mirror of the new checks would make the class consistent.
- Heads-up: open PR #769 adds a `list()` method to this same file that returns `response.json()` unvalidated for the same `tools/list` call. Whichever lands second should reuse this validation; a small private helper would make that easy.

neuro_san/session/mcp_service_agent_session.py
arnabnandy7: I removed `MAX_TOOLS` and the fatal length check. Since the response has already been downloaded and parsed at this point, the limit did not provide meaningful resource protection and could reject valid MCP responses. I added the Checkmarx false-positive explanation above the loop and included a regression test with more than 1000 tools.

neuro_san/session/mcp_service_agent_session.py
arnabnandy7: The constant has been removed along with the corresponding limit, so there is no longer an arbitrary client-side maximum.

neuro_san/session/mcp_service_agent_session.py
arnabnandy7: added shared response validation that detects JSON-RPC error objects and surfaces their code and message. This is applied to both `tools/list` and `initialize` responses, with focused test coverage for each.

neuro_san/session/mcp_service_agent_session.py
arnabnandy7: Updated. A missing `tools` key and `"tools": null` now follow the same validation path and produce the same error. Both cases are covered by exact-message assertions.

neuro_san/session/mcp_service_agent_session.py
arnabnandy7: Malformed entries are now skipped so they cannot prevent discovery of a valid tool later in the response. I added a regression test with an invalid entry before the requested tool.

tests/neuro_san/session/test_mcp_service_agent_session.py
arnabnandy7: Updated the test helper to accept a complete response body. The tests now independently exercise a non-object response, invalid `result` values, a missing `tools` key, and invalid `tools` values.

tests/neuro_san/session/test_mcp_service_agent_session.py
arnabnandy7: Fixed. Each invalid input is now paired with its complete expected error message, and the tests compare the full exception text. This ensures removing or bypassing an individual guard causes the corresponding test to fail.

arnabnandy7:
Thanks for review and feedback @Noravee 

Following are response for optionals

> * Handle a JSON-RPC `error` body and treat a missing `tools` key like `null` (inline on 164 and 168).

refer the comment in thread.

> * `continue` instead of `raise` on a non-dict tool entry (inline on 175-176).

refer the comment in thread.

> * `__init__` still reads the initialize response with the same unchecked `.get("result", {}).get(...)` pattern at line 115. A three-line mirror of the new checks would make the class consistent.

Addressed as well. Initialization now uses the shared JSON-RPC result validator, including response-shape validation and JSON-RPC error handling.

> * Heads-up: open PR #769 adds a `list()` method to this same file that returns `response.json()` unvalidated for the same `tools/list` call. Whichever lands second should reuse this validation; a small private helper would make that easy.

Added `_get_result_dict()` for shared JSON-RPC validation and `_get_tools_list()` for validating and extracting the tools array. The latter can be reused by the `list()` implementation introduced in PR #769.

andreidenissov-cog:
On the MCP server side, we validate client requests using mcp-schema-2025-06-18.json - which is OpenAPI spec document.
Have you considered using the same validation path in MCP session code also?
Definition is "ListToolsResult".
Using this validation should isolate us from possible (small or not) future changes in response JSON-RPC format.

arnabnandy7:
> On the MCP server side, we validate client requests using mcp-schema-2025-06-18.json - which is OpenAPI spec document. Have you considered using the same validation path in MCP session code also? Definition is "ListToolsResult". Using this validation should isolate us from possible (small or not) future changes in response JSON-RPC format.

@andreidenissov-cog thanks, that makes sense. The existing `McpRequestValidator` validates complete incoming request envelopes, so it cannot be applied directly to a `ListToolsResult`. We can reuse the vendored protocol schema by adding a response validator targeting `#/definitions/ListToolsResult`, while keeping JSON-RPC envelope and `error` handling separate.

One behavioral difference is that full schema validation will reject an entire response when any tool is malformed, whereas the current implementation skips malformed entries so valid tools remain discoverable. I’ll account for that behavior explicitly when updating the validation path.

Noravee:
Thanks for the quick turnaround on both rounds, and for trying the schema-based approach.

Unfortunately the latest commit (12a5c55c) breaks the client against every neuro-san server, including all released 0.6.x versions. I started a server from this branch and pointed both clients at it:

    main client -> {'function': {'description': '\nI can help you to make a terse announcement...
    PR client   -> ValueError: Invalid MCP tools/list response: 'result' does not match ListToolsResult

The cause is on the server side, not in your code: the `inputSchema` our server attaches to every tool is `{"$ref": ..., "components": {...}}` with no top-level `"type": "object"`, which the MCP `Tool` definition requires. I opened #1341 to fix that on the server. But fixing the server does not help this client against servers that are already deployed, and this client never reads `inputSchema` anyway, so validating it here only adds a way to fail.

Could you please:

1. **Go back to the tools/list handling from b8f48b30.** Validate the envelope and that `tools` is a list, then skip entries that are not usable dicts instead of rejecting the whole response. That is what the scanner asked for and it works with every server we have. Keeping the `InitializeResult` schema check is fine; our server's handshake passes it.
2. **Add a test that uses a tool shaped like our server's real output**, i.e. `inputSchema` as a `$ref` plus `components` with no `type`, and assert that `function()` still finds the agent. That guards against a strict check slipping back in.
3. **Update the branch** once more; it is behind main again.

If strict per-tool validation is needed, it should be lenient (skip entries that fail) so the CLI keeps working against older servers. Happy to discuss that as a follow-up once #1341 lands.

arnabnandy7:
> The latest commit breaks the client against every neuro-san server, including all released 0.6.x versions.

@Noravee thanks for catching this and for testing it against a running server. I’ve removed strict `ListToolsResult` validation from the `tools/list` path and restored the more tolerant handling from `b8f48b30` the response envelope and `tools` container are validated, while malformed entries are skipped.

I kept `InitializeResult` schema validation, since the existing handshake response conforms to it.

I also added a regression test using the actual input schema shape returned by neuro-san servers `$ref` with `components` and no top-level `type` and confirmed that the requested agent is still discovered. The targeted test suite passes all seven tests.

The branch is also up to date with `main`.

arnabnandy7:
I've squashed the PR into single commit for better understanding

arnabnandy7:
Fixed conflicts

PR #1295, opened 2026-08-31
Add a flag to allow prompt caching in load test

### LINKED ISSUE #1287: LT tool; add a flag to allow caching
return f"{base_prompt} (request {request_id})"I have this to make sure that I am not load-testing a cache benchmark. LLM, NSS, and Network proxy. This proves that your Boombox is doing caching. :wink:
I am going to add a flag to turn it off.

vince-leaf:
This is needed to allow us to run LT again on our boombox LLM caching. It will be saving tokens.

The load test appends a unique (request N) suffix to every varied-mode prompt so that no cache along the path (LLM prompt cache, agent network, proxy) can serve a response — keeping a run from degenerating into a cache benchmark. This adds --allow-caching to drop that suffix on purpose, so a run can measure cached behavior.

### DIFF
--- a/tests/load_tests/load_test_arguments.py
+++ b/tests/load_tests/load_test_arguments.py
@@ -207,6 +207,17 @@ class LoadTestArguments:
                  "prompts from the agent's prompt pool.",
         )
         parser.add_argument(
+            "--allow-caching",
+            action="store_true",
+            default=False,
+            help="Send prompts verbatim from the agent's prompt "
+                 "pool so caches along the path can serve them. "
+                 "By default a unique '(request N)' suffix is "
+                 "appended to defeat caching; pass this flag to "
+                 "measure cached behavior instead. No effect with "
+                 "--same-prompt, which is already cacheable.",
+        )
+        parser.add_argument(
             "--no-dry-run",
             action="store_true",
             default=False,
--- a/tests/load_tests/load_test_cli.py
+++ b/tests/load_tests/load_test_cli.py
@@ -2270,6 +2270,8 @@ class LoadTestOrchestrator:  # pylint: disable=too-many-instance-attributes
             )
 
         prompt_mode = "same" if self.args.same_prompt else "varied"
+        if self.args.allow_caching:
+            prompt_mode = f"{prompt_mode} (caching allowed)"
         mode = "ramp" if self.args.ramp else "flat"
         logger.info(
             "\nConfig: agent=%s, mode=%s, level=%s, "
@@ -2732,6 +2734,7 @@ class LoadTestOrchestrator:  # pylint: disable=too-many-instance-attributes
                 "num_rounds": self.args.num_rounds,
                 "num_requests": self.args.num_requests,
                 "same_prompt": self.args.same_prompt,
+                "allow_caching": self.args.allow_caching,
                 "chat_filter": self.args.chat_filter,
                 "server_log": self.server_log,
                 "estimated_tokens_per_request": (
--- a/tests/load_tests/prompts/agent_profile.py
+++ b/tests/load_tests/prompts/agent_profile.py
@@ -103,6 +103,9 @@ class AgentProfile:
         base_prompt = prompts[request_id % len(prompts)]
         if allow_caching:
             return base_prompt
+        # The suffix makes every prompt unique, so no cache along the
+        # path (LLM prompt cache, agent network, proxy) can serve the
+        # response and the run measures real work, not cache hits.
         return f"{base_prompt} (request {request_id})"
 
     @classmethod
--- a/tests/load_tests/prompts/agent_profile.py
+++ b/tests/load_tests/prompts/agent_profile.py
@@ -77,11 +77,17 @@ class AgentProfile:
         """
         return self._data.get("failure_patterns", [])
 
-    def get_prompt(self, request_id, same_prompt=False) -> str:
+    def get_prompt(self, request_id, same_prompt=False,
+                   allow_caching=False) -> str:
         """Return the prompt for a given request.
 
         In same_prompt mode, always returns the first prompt.
         In varied mode, cycles through the pool and appends the request_id.
+        The request_id suffix keeps every prompt unique so that no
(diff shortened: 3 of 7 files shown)


tests/load_tests/traffic/runner.py
```diff
@@ -96,6 +96,7 @@ def run_one(self, request_id, global_request_id,
         """
         prompt = self._profile.get_prompt(
             global_request_id, same_prompt=self._args.same_prompt,
+            allow_caching=self._args.allow_caching,
```
vince-leaf, line 99: The profile has no access to the CLI args, so the flag has to be forwarded there, at both call sites (run_one subprocess path).

tests/load_tests/traffic/runner.py
```diff
@@ -171,6 +172,7 @@ def run_one_http(self, request_id, global_request_id,
         prompt = self._profile.get_prompt(
             global_request_id,
             same_prompt=self._args.same_prompt,
+            allow_caching=self._args.allow_caching,
```
vince-leaf, line 175: The profile has no access to the CLI args, so the flag has to be forwarded there at both call sites (the run_one_http HTTP path).

tests/load_tests/load_test_arguments.py
```diff
@@ -206,6 +206,17 @@ def parse_args(epilog) -> argparse.Namespace:
                  "(collision stress test). Default is varied "
                  "prompts from the agent's prompt pool.",
         )
+        parser.add_argument(
+            "--allow-caching",
```
vince-leaf, line 210: This's where the CLI flags are declared — LoadTestArguments builds the argparse parser, so --allow-caching has to be registered

tests/load_tests/load_test_cli.py
```diff
@@ -2270,6 +2270,8 @@ def run(self) -> int:
             )
 
         prompt_mode = "same" if self.args.same_prompt else "varied"
+        if self.args.allow_caching:
+            prompt_mode = f"{prompt_mode} (caching allowed)"
```
vince-leaf, line 2274: Only for reporting. Two lines in load_test_cli.py: the run-header log gains prompt_mode=varied (caching allowed), and the raw JSON export records "allow_caching": self.args.allow_caching next to same_prompt.

Without it, a cached run looks identical to an uncached one in the output/comparison tables, which is the trap the suffix exists to avoid. Drop it if you want the smallest diff.

tests/load_tests/prompts/agent_profile.py
```diff
@@ -77,11 +77,17 @@ def failure_patterns(self) -> List[str]:
         """
         return self._data.get("failure_patterns", [])
 
-    def get_prompt(self, request_id, same_prompt=False) -> str:
+    def get_prompt(self, request_id, same_prompt=False,
```
vince-leaf, line 80: The one file that's genuinely essential — agent_profile.py — holds get_prompt(), the function that builds the prompt and appends (request N). It's the only place the suffix is produced, so it's where "don't append it" has to live.

tests/load_tests/reporting/json_metadata.py
```diff
@@ -50,6 +50,10 @@ def schema() -> Dict[str, str]:
             "config.same_prompt":
                 "If true, all requests use the same prompt "
                 "(collision stress test).",
+            "config.allow_caching":
+                "If true, prompts are sent verbatim and caches may "
+                "serve responses. If false, a unique per-request "
+                "suffix defeats caching.",
```
vince-leaf, line 56: Pure documentation: json_metadata.py contains the field_descriptions map that annotates the raw JSON export, since load_test_cli.py adds a config.allow_caching key: this adds a one-line description so the exported JSON stays self-describing (every other config key has one).

tests/load_tests/unit/test_agent_profile_prompt.py
```diff
@@ -0,0 +1,67 @@
...
+class TestAgentProfilePrompt(TestCase):
+    """
+    Unit tests for AgentProfile.get_prompt().
+
+    The per-request suffix is what keeps a run from measuring cache
+    hits, so a wrong answer here silently changes what is measured.
+    """
+
+    def setUp(self):
+        """Build a profile with a two-prompt pool."""
```
vince-leaf, line 31: Added this test file. It covers get_prompt() in four cases: default varied (suffix present), --allow-caching (verbatim, pool still cycles), --same-prompt taking precedence, and the empty-pool abort.

tests/load_tests/unit/test_raw_json_export.py
```diff
@@ -61,7 +61,7 @@ def _orchestrator(self) -> LoadTestOrchestrator:
             request_timeout=120, idle_timeout=60, stage_timeout=300,
             total_timeout=600, settle_time=5, max_workers=10,
             num_rounds=1, num_requests=3, same_prompt=False,
-            chat_filter=None,
+            allow_caching=False, chat_filter=None,
```
vince-leaf, line 64: That test builds a fake args Namespace with just the fields the export reads, since the export now reads self.args.allow_caching.

tests/load_tests/prompts/agent_profile.py
```diff
@@ -95,6 +101,11 @@ def get_prompt(self, request_id, same_prompt=False) -> str:
         if same_prompt:
             return prompts[0]
         base_prompt = prompts[request_id % len(prompts)]
+        if allow_caching:
+            return base_prompt
+        # The suffix makes every prompt unique, so no cache along the
+        # path (LLM prompt cache, agent network, proxy) can serve the
+        # response and the run measures real work, not cache hits.
```
vince-leaf, line 108: Adding the clarifying comment now — the suffix is deliberate: it makes every prompt unique so no cache (LLM prompt cache, agent network, Boombox/proxy) can serve the response, i.e., the run measures real work instead of cache hits. That's exactly what --allow-caching now lets you turn off.

andreidenissov-cog (approved):
LGTM.

PR #1296, opened 2026-09-01
Accept LLM endpoint base URL without /v1 in record/playback proxy.

andreidenissov-cog:
## Summary
Let the record/playback proxy accept an LLM endpoint base URL with or without
the OpenAI-style `/v1` version segment. Previously the configured
RECORD_PLAYBACK_UPSTREAM_BASE_URL had to already include `/v1`
(e.g. https://api.openai.com/v1); a value without it (https://api.openai.com)
produced wrong upstream paths. Now the `/v1` segment is appended when absent, so
both forms work.

This is tests-only tooling under tests/record_playback_llm_server/; no production
neuro_san code is touched.

## Change
- UpstreamClient normalizes the base URL in a new static _normalize_base_url:
  strip a trailing slash, then append "/v1" only when the URL does not already
  end in "/v1". Applied in the constructor (the request-building code, which
  concatenates base_url + path, is unchanged).
- Examples of the normalization:
    https://api.openai.com      -> https://api.openai.com/v1
    https://api.openai.com/     -> https://api.openai.com/v1
    https://api.openai.com/v1   -> https://api.openai.com/v1   (unchanged)
    https://api.openai.com/v1/  -> https://api.openai.com/v1
    ""                          -> ""   (empty passed through; server still
                                          enforces that record mode requires a URL)

## Tests / docs
- Adds TestUpstreamClient (one class per file, in the tests/ subfolder) covering
  the normalization cases. It asserts through the public constructor's stored
  base_url rather than the private helper.
- Updates the README's RECORD_PLAYBACK_UPSTREAM_BASE_URL note to say the `/v1`
  segment is optional and appended when absent. (The separate note that a
  neuro-san agent's openai_api_base must include /v1 is unrelated -- that is the
  client -> proxy URL, an OpenAI-SDK requirement -- and is left as-is.)

## Verification
- flake8 clean and pylint 10.00/10 across the package (tests included).
- 29 tests pass.

### DIFF
--- a/tests/record_playback_llm_server/README.md
+++ b/tests/record_playback_llm_server/README.md
@@ -80,7 +80,7 @@ required in `record` mode and used for cache-miss fetches in `hybrid` mode:
 
 | Variable | Purpose |
 |---|---|
-| `RECORD_PLAYBACK_UPSTREAM_BASE_URL` | Base URL of the real host, including the version segment. e.g. `https://api.openai.com/v1`. Required in record mode; optional in hybrid mode. |
+| `RECORD_PLAYBACK_UPSTREAM_BASE_URL` | Base URL of the real host, e.g. `https://api.openai.com` or `https://api.openai.com/v1`. The OpenAI-style `/v1` segment is optional and is appended when absent. Required in record mode; optional in hybrid mode. |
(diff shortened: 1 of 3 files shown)


vince-leaf (approved):
LGTM

PR #1310, opened 2026-09-08
Fix openai_proxy config key in OpenAILlmPolicy

### LINKED ISSUE #1308: OpenAILlmPolicy passes the openai_organization config value as openai_proxy
## Summary

`OpenAILlmPolicy.create_llm` reads the wrong config key when building the `openai_proxy` kwarg for `ChatOpenAI`.

`neuro_san/internals/run_context/langchain/llms/openai_llm_policy.py:140-141` (main `0169303b`):

```python
openai_proxy=self.get_value_or_env(config, "openai_organization",
                                   "OPENAI_PROXY", client),
```

The key should be `"openai_proxy"`. The Azure policy uses the correct key at `azure_llm_policy.py:136-137`.

## Impact

Harmless today. Every kwarg in that group is forced to `None` whenever a client object is supplied (see the comment at `openai_llm_policy.py:133`), and neuro-san always creates the client first. The real proxy setting is applied to the httpx `AsyncClient` in `create_http_client` (`openai_llm_policy.py:95-104`). So this is a latent copy-paste slip that would only bite if the client-first ordering ever changed.

## Suggested fix

Change the key to `"openai_proxy"` and add a small unit test that asserts the config key each `get_value_or_env` call reads for this group. Kept separate from #725 so that fix stays focused.

HeaTTap:
Fixes #1308

## Summary

OpenAILlmPolicy.create_llm previously read the wrong config key openai_organization when building the openai_proxy kwarg for ChatOpenAI. This change updates the key to openai_proxy matching AzureLlmPolicy.

## Changes

- Corrected the config key from openai_organization to openai_proxy in neuro_san/internals/run_context/langchain/llms/openai_llm_policy.py
- Added unit tests in tests/neuro_san/internals/run_context/langchain/llms/test_openai_llm_policy.py asserting the config keys passed to get_value_or_env and verifying ChatOpenAI receives the expected proxy setting

## Verification

- Ran pytest on tests/neuro_san/internals/run_context/langchain/llms/test_openai_llm_policy.py and verified all 4 tests pass
- Confirmed test failures on the unpatched code
- Verified flake8 and pylint run cleanly with 10/10 score

### DIFF
--- a/tests/neuro_san/internals/run_context/langchain/llms/test_openai_llm_policy.py
+++ b/tests/neuro_san/internals/run_context/langchain/llms/test_openai_llm_policy.py
@@ -1,0 +1,140 @@
+
+# Copyright © 2023-2026 Cognizant Technology Solutions Corp, www.cognizant.com.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# END COPYRIGHT
+"""
+Unit tests for OpenAILlmPolicy.
+"""
+import asyncio
+from unittest.mock import MagicMock
+
+from neuro_san.internals.run_context.langchain.llms.openai_llm_policy import OpenAILlmPolicy
+
+
+class TestOpenAILlmPolicy:
+    """
+    Test cases for OpenAILlmPolicy.
+    """
+
+    def test_create_llm_config_keys(self):
+        """
+        Test that create_llm reads the correct config keys for the client-supplied kwargs group,
+        specifically ensuring openai_proxy reads 'openai_proxy' and not 'openai_organization'.
+        """
+        captured_kwargs = {}
+
+        def mock_chat_openai(**kwargs):
+            captured_kwargs.update(kwargs)
+            return MagicMock()
+
+        policy = OpenAILlmPolicy()
+        policy.resolver.resolve_class_in_module = MagicMock(return_value=mock_chat_openai)
+
+        config = {
+            "openai_api_key": "sk-test-key",
+            "openai_api_base": "https://custom.api.com",
+            "openai_organization": "org-12345",
+            "openai_proxy": "http://proxy.test:8080",
+            "request_timeout": 60,
+            "max_retries": 3,
+            "temperature": 0.7,
+        }
+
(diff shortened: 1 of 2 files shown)


Noravee:
Thanks @HeaTTap, the fix is correct and the tests are good. I ran your four tests against current `main` with your one-line change applied and they pass alongside the six tests that landed there in #1311; without your change, two of them fail on the `openai_proxy` pairing, so they cover the bug. flake8 and pylint are clean.

The PR shows as conflicting only because #1311 also created `tests/neuro_san/internals/run_context/langchain/llms/test_openai_llm_policy.py`. The change to `openai_llm_policy.py` merges cleanly (your hunk at line 140 does not overlap the #1311 changes further down).

To get this mergeable:

1. Rebase onto `main`:
   ```
   git fetch upstream
   git rebase upstream/main
   ```
   Only `test_openai_llm_policy.py` will conflict (both branches added it).
2. Resolve it by keeping the `main` version of the file and appending your four test methods to the existing `TestOpenAILlmPolicy` class. Add `from unittest.mock import MagicMock` to the imports (`asyncio` is already imported). Drop your module docstring, or place it after the copyright header; `main`'s file has `# pylint: disable=protected-access` right below the header, keep that line.
3. Run the checks and force-push:
   ```
   OPENAI_API_KEY=sk-test PYTHONPATH=. pytest -q tests/neuro_san/internals/run_context/langchain/llms
   flake8 --max-line-length=120 tests/neuro_san/internals/run_context/langchain/llms/test_openai_llm_policy.py
   pylint --load-plugins=pylint_protobuf -rn --rcfile=./.pylintrc tests/neuro_san/internals/run_context/langchain/llms/test_openai_llm_policy.py
   ```
   `tests/conftest.py` skips default-provider tests when `OPENAI_API_KEY` is unset; any dummy value works, nothing in these tests reaches the network. Expected result: 10 passed, 1 pre-existing skip.

Once the rebase is pushed I will approve the CI run and merge.

tests/neuro_san/internals/run_context/langchain/llms/test_openai_llm_policy.py
```diff
@@ -0,0 +1,140 @@
...
+"""
+Unit tests for OpenAILlmPolicy.
+"""
+import asyncio
+from unittest.mock import MagicMock
+
+from neuro_san.internals.run_context.langchain.llms.openai_llm_policy import OpenAILlmPolicy
+
+
+class TestOpenAILlmPolicy:
```
d1donlydfink: Any reason this test does not derive from unittest.TestCase like all the other tests in this repo?

d1donlydfink (changes requested):

Noravee:
Thanks again for reporting this and for the fix. The same one-line correction has now landed on `main` through #1328, together with regression tests added to the `TestOpenAILlmPolicy` module that #1311 introduced in the meantime, which is why this branch could not merge as-is. #1308 is closed by that PR.

Closing this one as superseded. Much appreciated.

Issue #1309, opened 2026-09-08
docs: llm_config section states stale defaults (gpt-4o, temperature 0.7) and omits OPENAI_API_VERSION for Azure


Noravee:
## Summary

`docs/agent_hocon_reference.md` makes several statements about `llm_config` that no longer match `default_llm_info.hocon`, and the Azure row omits an environment variable the Azure client requires. Small, docs-only cleanup; kept separate from the code fixes so it can merge any time.

Verified at main `0169303b`.

## Stale statements

1. **Default model.** `docs/agent_hocon_reference.md:196` says the default "is to use an OpenAI gpt-4o model", and `:201` says "the default model is \"gpt-4o\"". `default_llm_info.hocon` `default_config.model_name` is `gpt-5.2` (`:1746`). Suggest wording that points at `default_config` in `default_llm_info.hocon` rather than naming a model, so it does not go stale again.

2. **Temperature default.** `:309` says "By default this value is 0.7". `classes.openai.args.temperature` is `null` (`default_llm_info.hocon:1454`), meaning the provider default applies and nothing is sent.

3. **Azure environment variables.** The table row at `:220` lists only `AZURE_OPENAI_API_KEY` and `AZURE_OPENAI_ENDPOINT`. `AzureLlmPolicy.create_client` also reads `OPENAI_API_VERSION` (`azure_llm_policy.py:82`), which the openai `AsyncAzureOpenAI` client requires, and the deployment comes from `deployment_name` in `llm_config` or `AZURE_OPENAI_DEPLOYMENT_NAME` (`azure_llm_policy.py:80-81`). See `neuro_san/registries/music_nerd_pro_llm_azure.hocon` for the working shape.

4. **"Unsupported parameters will be ignored"** (`:342-345`). Accurate for the built-in provider classes, where the policy forwards only the kwargs it names. On the user-class path (`class` set to a full Python path) unknown keys raise a `ValueError` from `ArgumentValidator` instead. Worth one clarifying sentence.

## Out of scope here

- Documentation of `reasoning`, `reasoning_effort`, `use_responses_api`, and which Chat-Completions-only parameters the Responses API rejects: part of #725.
- Azure Responses API requirements (`OPENAI_API_VERSION`, `deployment_name` as body `model`): part of #1307.

Issue #1307, opened 2026-09-08
AzureLlmPolicy cannot use the Responses API: no root_async_client, reasoning kwargs dropped, hard-coded n=1; azure-gpt-5.4-pro fails on langchain-openai >=1.2


Noravee:
## Summary

`AzureLlmPolicy.create_llm` builds `AzureChatOpenAI` in a way that makes any Responses API route fail, and it silently drops the reasoning-related `llm_config` keys that the `azure-openai` class inherits from `openai`. Today the Azure class can only ever talk to Chat Completions.

Split out of #725 on purpose: the OpenAI-side fix can be verified offline, while the Azure side needs a live deployment to settle one open question (see item 5).

## Details (main `0169303b`)

1. **`root_async_client` is never passed.** `neuro_san/internals/run_context/langchain/llms/azure_llm_policy.py:125-126` passes `async_client=client` only. langchain-openai creates `root_async_client` itself only when `async_client` is falsy (1.6.0 `azure.py:693-705`), so it stays `None`. Every Responses call goes through `self.root_async_client.responses.create(...)`, which raises `AttributeError: 'NoneType' object has no attribute 'responses'` before any HTTP request. `openai_llm_policy.py:129` does pass `root_async_client=self.async_openai_client` (added in 349b7c4c); it was never mirrored here.

2. **`reasoning`, `reasoning_effort`, `verbosity` are not forwarded.** `azure_llm_policy.py:125-188` has no such kwargs, even though `classes.azure-openai` extends `openai` (`default_llm_info.hocon:1483`) and therefore inherits those args (`default_llm_info.hocon:1476-1478`). Setting them on an Azure model is a silent no-op.

3. **`n=1` is hard-coded** at `azure_llm_policy.py:155`. The Responses API has no `n` parameter, so once `root_async_client` is present this becomes the same `TypeError: AsyncResponses.create() got an unexpected keyword argument 'n'` as #725. Chat Completions defaults `n` to 1, so removing it is wire-neutral there.

[...]

PR #1311, opened 2026-09-08
Stop sending n=1 and expose use_responses_api for OpenAI models

### LINKED ISSUE #725: OpenAI policy hard-codes n=1, breaking every Responses API route; expose ChatOpenAI use_responses_api
Suggestions:
1. Remove reasoning.
# The following parameters are for reasoning models only.
reasoning": null, # Reasoning parameters for reasoning models (Dict[str, Any])

Noravee:
## Summary

`OpenAILlmPolicy` always constructed `ChatOpenAI` with `n=1`. langchain-openai forwards `n` whenever it is not `None`, and the Responses API has no `n` parameter, so every request that reached the Responses API failed client-side before any HTTP call:

```
TypeError: AsyncResponses.create() got an unexpected keyword argument 'n'
```

That path is reached by any `llm_config` with a `reasoning` dict (the original report in #725), and, on langchain-openai >= 1.2, by the shipped `gpt-5.5-pro` / `gpt-5.4-pro` entries with no reasoning config at all, because langchain routes those model names to the Responses API by prefix.

It also matters for the newest models: OpenAI rejects function tools combined with reasoning on Chat Completions. `gpt-5.6-*` only works with tools there if `reasoning_effort` is `"none"`, and `gpt-6-astra` has no Chat Completions tool path at all (no function calling, and `"none"` is rejected). neuro-san had no way to select the Responses API.

## Changes

- `openai_llm_policy.py`: stop passing `n` (Chat Completions defaults it to 1, so this is wire-neutral for every existing config). Forward `use_responses_api` from `llm_config` as a raw tri-state via `config.get`, so `null` keeps langchain's own inference and only an explicit `true`/`false` overrides it. Comments explain why `ConfigUtil.get_bool` is deliberately not used and link the langchain and OpenAI references.
- `default_llm_info.hocon`: `"use_responses_api": null` added to `classes.openai.args` with the tri-state semantics; the stale `n` comment replaced; a note that `reasoning` and `reasoning_effort` should not both be set on the Responses path.
- New `tests/.../test_openai_llm_policy.py` (no network): `n` is gone from both payload shapes, `use_responses_api` defaults to `None` and is forwarded when set, the `reasoning`-dict regression payload binds to the real `AsyncResponses.create` signature, and `reasoning_effort` folds into `reasoning.effort` on the Responses path.
- `docs/agent_hocon_reference.md`: new "OpenAI Reasoning and Responses API Parameters" subsection covering `reasoning` vs `reasoning_effort`, `verbosity`, `use_responses_api`, the gpt-5.6 / gpt-6-astra requirements, the Chat-Completions-only parameters the Responses API rejects, and the interaction with `fallbacks`.

Behavior is unchanged for existing configurations: with `use_responses_api` unset, a plain `gpt-5.2` request differs only by the absent `n` key.

## Out of scope, tracked separately

- Azure side of the same bug and Azure Responses support: #1307
- `openai_proxy` config key typo in the same function: #1308
- Stale `llm_config` docs unrelated to this bug: #1309

## Relationship to #1306

#1306 adds the `gpt-6-astra` entry. This PR should land first so that entry does not advertise tool support with no way to use it; the docs example here uses `gpt-6-astra` on that assumption.

Closes #725

### DIFF
--- a/docs/agent_hocon_reference.md
+++ b/docs/agent_hocon_reference.md
@@ -25,6 +25,7 @@ Items in ***bold*** are essentials. Try to understand these first.
         - [fallbacks](#fallbacks)
         - [temperature](#temperature)
         - [Other LLM-specific Parameters](#other-llm-specific-parameters)
+        - [OpenAI Reasoning and Responses API Parameters](#openai-reasoning-and-responses-api-parameters)
         - [Client-Provided API Keys](#client-provided-api-keys)
     - [***tools*** - list of agent/tool definitions](#tools)
     - [commondefs](#commondefs)
@@ -318,6 +319,75 @@ file, you can set that parameter in any llm_config within its own technical limi
 Note: _We strongly recommend to **not** set secrets as values within any source file, including hocon files._
 These files tend to creep into source control repos, and it is **very** bad practice
 to expose secrets by checking them in.
+
+#### OpenAI Reasoning and Responses API Parameters
+
+The `openai` class accepts a few parameters that only matter for OpenAI reasoning models and for choosing which
+OpenAI endpoint (Chat Completions or the Responses API) requests are sent to. All of them default to `null`.
+
+- `reasoning`: a dictionary passed through as the Responses API `reasoning` object, for example
+  `{"effort": "low", "summary": "auto"}`. Setting it makes langchain route requests to the Responses API.
+- `reasoning_effort`: a string such as `"low"`, `"medium"` or `"high"` that constrains how much reasoning the
+  model does (`"none"` disables reasoning on models that allow it, such as `gpt-5.6-*`; `gpt-6-astra` rejects
+  it). On the Chat Completions path it is sent as-is. On the Responses API path langchain folds it into
+  `reasoning.effort` **only when `reasoning` is absent**, so set either `reasoning` or `reasoning_effort`, not
+  both. Setting `reasoning_effort` on its own does not change the endpoint.
+- `verbosity`: a string (`"low"`, `"medium"` or `"high"`) that controls how long the model's answers are.
+- `use_responses_api`: a tri-state switch for the endpoint:
+    - `null` (the default): let langchain infer the endpoint from the other parameters and the model name, as
+      documented for
+      [`use_responses_api`](https://reference.langchain.com/python/langchain-openai/chat_models/base/BaseChatOpenAI/use_responses_api).
+      langchain switches to the Responses API when any Responses-only setting is present (`reasoning`, `include`,
+      `truncation`, `context_management`, `previous_response_id`, `text`, or a built-in tool such as web search)
+      or when the model is one it knows to be Responses-only (the `gpt-5.x-pro` and `codex` models). Of those,
+      the `openai` class only exposes `reasoning` and the model name, so everything else keeps using Chat
+      Completions and existing configurations behave exactly as before.
+    - `true`: always use the Responses API.
+    - `false`: always use Chat Completions.
+
+**Models that need the Responses API for tool calling.** OpenAI's newest models no longer accept function (tool)
+calls together with reasoning on Chat Completions, and every agent that lists `tools` is a tool-calling agent.
+A Chat Completions request from a `gpt-5.6-*` model that carries tools is rejected with an error such as:
+
+> Function tools with reasoning_effort are not supported for gpt-5.6-sol in /v1/chat/completions.
+> To use function tools, use /v1/responses or set reasoning_effort to 'none'.
+
+The two model families differ in what you can do about it:
+
+- `gpt-5.6-*` (`gpt-5.6-sol`, `gpt-5.6-terra`, `gpt-5.6-luna`): either set `"use_responses_api": true` and keep
+  reasoning, or stay on Chat Completions with `"reasoning_effort": "none"`, which makes tool calls work but
+  without any reasoning.
+- `gpt-6-astra`: the Responses API is the only option. The
+  [OpenAI reasoning guide](https://developers.openai.com/api/docs/guides/reasoning) states that
+  "Chat Completions does not support function calling with GPT-6 Astra" and that GPT-6 Astra
+  "does not support none reasoning effort" (the API answers HTTP 400), so the `"none"` workaround does not exist
+  for it. Any agent that uses tools with gpt-6-astra must set `"use_responses_api": true`.
+
+The minimal configuration is just the endpoint switch. The model's default reasoning effort applies; add
+`reasoning_effort` only when you want a different level.
+
+```hocon
+"llm_config": {
+    "model_name": "gpt-6-astra",
+    "use_responses_api": true
+}
+```
+
+**Parameters the Responses API rejects.** `presence_penalty`, `frequency_penalty`, `seed`, `logprobs` and
+`logit_bias` (plus `stop` on langchain-openai releases before 1.4) exist only on Chat Completions. The Responses
+API rejects requests that carry them, so remove them from any llm_config that reaches the Responses API, whether
+through `"use_responses_api": true` or through auto-routing. OpenAI's
+[migration guide](https://developers.openai.com/api/docs/guides/migrate-to-responses) lists the remaining
+differences between the two endpoints.
+
+**Interaction with fallbacks.** The rejections above surface as exceptions on the client side, but
+[fallbacks](#fallbacks) wrap the model so that any exception moves on to the next llm_config in the list. With
+fallbacks configured, a misconfigured primary model silently fails over instead of reporting the problem, so
+verify a new llm_config without fallbacks first.
+
+**Azure OpenAI.** The `azure-openai` class does not support the Responses API path yet (tracked in
+[#1307](https://github.com/cognizant-ai-lab/neuro-san/issues/1307)), so leave `use_responses_api` unset for
+Azure deployments.
 
 #### class
 
--- a/neuro_san/internals/run_context/langchain/llms/default_llm_info.hocon
+++ b/neuro_san/internals/run_context/langchain/llms/default_llm_info.hocon
@@ -1466,16 +1466,38 @@
                 "logit_bias": null,
                 # streaming defaults to off; set "streaming": true in llm_config to enable.
                 "streaming": false,
-                # n is always 1.  neuro-san will only ever consider a single chat completion.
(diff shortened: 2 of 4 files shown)


neuro_san/internals/run_context/langchain/llms/openai_llm_policy.py
```diff
@@ -156,17 +156,49 @@ def create_llm(self, config: Dict[str, Any], model_name: str, client: Any) -> Ba
             # usage is collected from AIMessage.usage_metadata in LlmTokenCallbackHandler
             # regardless of streaming mode.
             streaming=ConfigUtil.get_bool(config, "streaming"),
-            n=1,  # n is always 1.  neuro-san will only ever consider one chat completion.
+            # n is intentionally not sent. Chat Completions defaults n to 1, which is all neuro-san
```
Noravee, line 159: Root cause of #725. langchain-openai emits `n` whenever `ChatOpenAI.n` is not `None` and never strips it from the Responses payload, so `n=1` reached `AsyncResponses.create()`, which has no `n` parameter. Dropping it is safe: Chat Completions defaults `n` to 1 (openai-openapi `openapi.yaml`, `default: 1`).

Noravee (commented):
**TL;DR**

- **Bug (#725):** `create_llm()` always passed `n=1`. langchain-openai forwards `n` whenever it is set, and `AsyncResponses.create()` has no `n` parameter, so every request that reached the Responses API died client-side with `TypeError: ... unexpected keyword argument 'n'`. Triggered by any `reasoning` dict, and by the shipped `gpt-5.5-pro` / `gpt-5.4-pro` entries with no reasoning config at all (langchain-openai >= 1.2 routes them by name).
- **Fix 1:** stop sending `n`. Chat Completions defaults it to 1, which is all neuro-san consumes.
- **Fix 2:** forward `use_responses_api` from `llm_config` as a raw tri-state. `null` keeps langchain's own inference, `true` forces the Responses API, `false` forces Chat Completions. `true` is what tool calling needs on `gpt-6-astra`, and on `gpt-5.6-*` unless `reasoning_effort` is `"none"`.
- **Unchanged for existing configs:** with the key unset, endpoint selection is exactly what langchain did before; only the `n` key disappears from Chat Completions requests.
- **Next:** #1306 (`gpt-6-astra` entry) lands after this. Azure inherits the hocon key but does not read it yet (#1307). #1308 and #1309 are out of scope.

neuro_san/internals/run_context/langchain/llms/openai_llm_policy.py
```diff
@@ -156,17 +156,49 @@ def create_llm(self, config: Dict[str, Any], model_name: str, client: Any) -> Ba
...
+            # reached the Responses API fail client-side.
+            # See https://github.com/cognizant-ai-lab/neuro-san/issues/725 and the n property of
+            # CreateChatCompletionRequest in https://github.com/openai/openai-openapi/blob/main/openapi.yaml
             top_p=config.get("top_p"),
             max_tokens=config.get("max_tokens"),  # This is always for output
             tiktoken_model_name=config.get("tiktoken_model_name"),
             stop=config.get("stop"),
 
             # The following three parameters are for reasoning models only.
+            # Set either "reasoning" or "reasoning_effort", not both: on the Responses API path
```
Noravee, line 171: Both keys are still forwarded unchanged, so this caveat matters. On the Responses path langchain folds `reasoning_effort` into `reasoning["effort"]` only when `reasoning` is absent; with both set, `reasoning_effort` stays in the payload and `AsyncResponses.create()` raises `TypeError` on it. The hocon and docs carry the same warning.

neuro_san/internals/run_context/langchain/llms/openai_llm_policy.py
```diff
@@ -156,17 +156,49 @@ def create_llm(self, config: Dict[str, Any], model_name: str, client: Any) -> Ba
...
+            #           Chat Completions with reasoning_effort "none" (tool calls work, no reasoning).
+            #           https://developers.openai.com/api/docs/guides/reasoning
+            #           https://developers.openai.com/api/docs/guides/migrate-to-responses
+            #   False - force Chat Completions.
+            # This is deliberately forwarded raw instead of through ConfigUtil.get_bool():
+            # get_bool() turns an absent/null value into False, which would disable langchain's
+            # auto-routing, send a "reasoning" dict to Chat Completions, and pin the *-pro models
+            # to an endpoint OpenAI rejects. Only an explicit true/false in llm_config should
+            # override langchain's choice.
+            use_responses_api=config.get("use_responses_api"),
```
Noravee, line 200: `config.get`, not `ConfigUtil.get_bool`, on purpose. `ChatOpenAI` treats any explicit bool as final, and `get_bool` turns an absent or `null` key into `False`. That would switch off auto-routing for every existing config, send a `reasoning` dict to Chat Completions (`TypeError 'reasoning'`), and pin the `*-pro` models to an endpoint OpenAI rejects.

neuro_san/internals/run_context/langchain/llms/default_llm_info.hocon
```diff
@@ -1466,16 +1466,38 @@
...
+                #   true  - force the Responses API. Needed for tool calling with reasoning on the newest models,
+                #           because Chat Completions rejects function tools combined with reasoning:
+                #             - gpt-6-astra can only call tools through the Responses API. Chat Completions has
+                #               no function calling for it and it rejects reasoning_effort "none".
+                #             - gpt-5.6-* can either use this, or stay on Chat Completions with
+                #               reasoning_effort "none" (tool calls work, but without reasoning).
+                #           https://developers.openai.com/api/docs/guides/reasoning
+                #           https://developers.openai.com/api/docs/guides/migrate-to-responses
+                #   false - force Chat Completions.
+                "use_responses_api": null,
```
Noravee, line 1500: The default must stay `null`, not `false`. `null` lets langchain keep inferring the endpoint, so existing configs keep the same endpoint and the shipped `gpt-5.5-pro` / `gpt-5.4-pro` entries still auto-route on langchain-openai >= 1.2. `azure-openai` extends this class and inherits the key, but `AzureLlmPolicy` does not read it yet (#1307).

tests/neuro_san/internals/run_context/langchain/llms/test_openai_llm_policy.py
```diff
@@ -0,0 +1,212 @@
...
+        payload: Dict[str, Any] = self._request_payload(llm)
+
+        assert llm.n is None
+        assert "n" not in payload
+        # None (not False) is what keeps langchain's auto-routing alive for existing configs.
+        assert llm.use_responses_api is None
+        assert llm._use_responses_api(llm._default_params) is False
+        assert "messages" in payload
+
+    def test_reasoning_dict_auto_routes_to_responses_api_without_n(self, policies: List[OpenAILlmPolicy]) -> None:
```
Noravee, line 134: Regression test for #725. It builds `ChatOpenAI` through the real `create_client()` + `create_llm()`, confirms a `reasoning` dict auto-routes while `use_responses_api` stays `None`, and binds the payload against `inspect.signature(AsyncResponses.create)`. That bind rejects any payload still carrying `n`, with no network call.

tests/neuro_san/internals/run_context/langchain/llms/test_openai_llm_policy.py
```diff
@@ -0,0 +1,212 @@
...
+        :param policies: The fixture list that tracks policies for teardown
+        """
+        llm: ChatOpenAI = self._build_llm(policies, {"use_responses_api": False})
+        payload: Dict[str, Any] = self._request_payload(llm)
+
+        assert llm.use_responses_api is False
+        assert llm._use_responses_api(llm._default_params) is False
+        assert "messages" in payload
+
+    def test_use_responses_api_false_overrides_auto_routing(self, policies: List[OpenAILlmPolicy]) -> None:
```
Noravee, line 184: Guards the `config.get` vs `get_bool` choice at `openai_llm_policy.py:200`. An explicit `false` must reach `ChatOpenAI` as `False` and pin Chat Completions even when a `reasoning` dict would otherwise auto-route. With the `None` and `true` tests above, all three states of the switch are pinned.

docs/agent_hocon_reference.md
```diff
@@ -319,6 +320,75 @@ Note: _We strongly recommend to **not** set secrets as values within any source
...
+
+> Function tools with reasoning_effort are not supported for gpt-5.6-sol in /v1/chat/completions.
+> To use function tools, use /v1/responses or set reasoning_effort to 'none'.
+
+The two model families differ in what you can do about it:
+
+- `gpt-5.6-*` (`gpt-5.6-sol`, `gpt-5.6-terra`, `gpt-5.6-luna`): either set `"use_responses_api": true` and keep
+  reasoning, or stay on Chat Completions with `"reasoning_effort": "none"`, which makes tool calls work but
+  without any reasoning.
+- `gpt-6-astra`: the Responses API is the only option. The
```
Noravee, line 360: This bullet is why the PR adds the switch instead of only dropping `n`. Per the OpenAI reasoning guide, Chat Completions has no function calling for GPT-6 Astra and it rejects reasoning effort `none` (HTTP 400), so the `gpt-5.6-*` workaround does not apply. Any agent with `tools` on gpt-6-astra needs `"use_responses_api": true`, hence #1306 lands after this PR.

docs/agent_hocon_reference.md
```diff
@@ -319,6 +320,75 @@ Note: _We strongly recommend to **not** set secrets as values within any source
...
+```
+
+**Parameters the Responses API rejects.** `presence_penalty`, `frequency_penalty`, `seed`, `logprobs` and
+`logit_bias` (plus `stop` on langchain-openai releases before 1.4) exist only on Chat Completions. The Responses
+API rejects requests that carry them, so remove them from any llm_config that reaches the Responses API, whether
+through `"use_responses_api": true` or through auto-routing. OpenAI's
+[migration guide](https://developers.openai.com/api/docs/guides/migrate-to-responses) lists the remaining
+differences between the two endpoints.
+
+**Interaction with fallbacks.** The rejections above surface as exceptions on the client side, but
```
Noravee, line 383: The `n` TypeError is a client-side exception, and `fallbacks` wraps the model with `with_fallbacks`, which catches `Exception`. With fallbacks configured a misconfigured primary silently fails over instead of reporting the problem, so verify a new llm_config without fallbacks first.

d1donlydfink (approved):

PR #1312, opened 2026-09-09
Flatten the model response in NetworkCopyMiddleware before parsing JSON

Noravee:
## What

`NetworkCopyMiddleware.aafter_agent` reads the provider-native `AIMessage` from the agent state and hands its `.content` straight to `json.loads()`. That content is a **list of blocks**, not a `str`, whenever the model emitted thinking or `tool_use` blocks (Anthropic) or was routed through the OpenAI Responses API (`use_responses_api`, any `reasoning` value, `*-pro` models). `json.loads()` raises `TypeError` on a list, the middleware only caught `JSONDecodeError`, and the request ended as `"Agent stopped due to exception ... TypeError"`.

This is a live bug today, not gated on any later work: the shipped `copy_cat_middleware.hocon` example happens to use a Chat Completions model without reasoning, so it is unaffected by default, but any Anthropic-with-thinking or Responses-API variant hits it.

## Fix

- Project the last message through `ContentUtils.flatten_to_text` before parsing. For plain-string content this returns the same string object, so existing behavior is byte-identical.
- Catch `TypeError` alongside `JSONDecodeError` in `_parse_agent_name`. Unreachable from `aafter_agent` after the projection; kept as a cheap guard for any future caller that bypasses it, with a comment saying so.

## Tests

First unit tests for this middleware (`tests/neuro_san/middleware/test_network_copy_middleware.py`): plain-string JSON still parses; Anthropic thinking-first blocks, OpenAI Responses-shaped blocks, and list-of-str content now parse instead of raising; a `tool_use`-only turn and malformed JSON still produce the "please provide the name" prompt; the `.hocon` suffix is still stripped; `_parse_agent_name` tolerates a non-string.

Part of #1222 (Phase 1), PR 4.1 of the ladder described there. Independent of 4.2 and 4.3.

### DIFF
--- a/neuro_san/middleware/network_copy_middleware.py
+++ b/neuro_san/middleware/network_copy_middleware.py
@@ -145,7 +145,7 @@ class NetworkCopyMiddleware(AgentMiddleware):
         :return: The agent name, or None if parsing/validation fails
         """
         try:
-            args: Dict[str, str] = loads(response)
+            args: Any = loads(response)
         except (JSONDecodeError, TypeError) as json_error:
             # TypeError is what json.loads() raises for non-string input. It
             # cannot happen on the aafter_agent() path, which always projects
@@ -153,6 +153,14 @@ class NetworkCopyMiddleware(AgentMiddleware):
             # any future caller bypassing that projection gets the same
             # "please provide the name" outcome instead of an exception.
             self.logger.error("Cannot parse '%s' into JSON format. Got %s", response, json_error)
+            return None
+
+        if not isinstance(args, dict):
+            # Valid JSON that is not an object (a bare string, list or number)
+            # has no "agent_name" key to read. Treat it like any other
+            # unusable response rather than raising AttributeError on .get(),
+            # which would also end the whole request with an exception.
+            self.logger.error("Expected a JSON object with an agent_name key, got '%s'", response)
             return None
 
         agent_name: str = args.get("agent_name")
--- a/neuro_san/middleware/network_copy_middleware.py
+++ b/neuro_san/middleware/network_copy_middleware.py
@@ -25,12 +25,14 @@ from typing_extensions import override
 from langchain.agents.middleware import AgentMiddleware
 from langchain.agents.middleware import AgentState
 from langchain_core.messages import AIMessage
+from langchain_core.messages import BaseMessage
 
 from neuro_san import REGISTRIES_DIR
 from neuro_san.interfaces.reservationist import Reservationist
 from neuro_san.internals.graph.persistence.agent_network_restorer import AgentNetworkRestorer
 from neuro_san.internals.graph.registry.agent_network import AgentNetwork
 from neuro_san.internals.reservations.reservation_util import ReservationUtil
+from neuro_san.message.utils.content_utils import ContentUtils
 
 
 class NetworkCopyMiddleware(AgentMiddleware):
@@ -81,7 +83,13 @@ class NetworkCopyMiddleware(AgentMiddleware):
         :param runtime: Runtime context
         :return: Dict with error message and jump directive, or None if valid
         """
-        response: str = state.get("messages")[-1].content
+        # The state holds the provider-native message, whose content is a list
+        # of blocks rather than a str whenever the model emitted thinking or
+        # tool_use blocks, or was routed through the OpenAI Responses API.
+        # json.loads() raises TypeError on a list, so project the content to
+        # its text first. For plain-string content this is an identity.
+        last_message: BaseMessage = state.get("messages")[-1]
+        response: str = ContentUtils.flatten_to_text(last_message)
 
         agent_name: str = self._parse_agent_name(response)
         if agent_name is None:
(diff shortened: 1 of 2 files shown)


neuro_san/middleware/network_copy_middleware.py
```diff
@@ -138,7 +146,12 @@ def _parse_agent_name(self, response: str) -> str | None:
         """
         try:
             args: Dict[str, str] = loads(response)
-        except JSONDecodeError as json_error:
+        except (JSONDecodeError, TypeError) as json_error:
+            # TypeError is what json.loads() raises for non-string input. It
+            # cannot happen on the aafter_agent() path, which always projects
```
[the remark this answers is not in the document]
Noravee: Good catch, and it is the same failure class this PR is about. Added an isinstance(args, dict) check in 79f15d79 so non-object JSON yields None and the "please provide the name" prompt; covered by two new tests (a JSON array through aafter_agent, and bare string/number through _parse_agent_name).

Noravee:
TLDR: NetworkCopyMiddleware passed the model's raw `.content` straight to `json.loads()`. For Anthropic thinking/tool_use turns and every OpenAI Responses-API route that content is a list of blocks, so the request died with an uncaught TypeError ("Agent stopped due to exception").

Fix: project the message through `ContentUtils.flatten_to_text` (#1276) before parsing, and make `_parse_agent_name` return None for non-string input and non-object JSON instead of raising. First unit tests for this middleware; 7 of the 10 fail on main.

Backward compat: for str content `flatten_to_text` returns the same str object, so existing Chat Completions traffic (including the shipped copy_cat example) is byte-identical. Part of #1222; no wire-schema change.

neuro_san/middleware/network_copy_middleware.py
```diff
@@ -81,7 +83,13 @@ async def aafter_agent(
         :param runtime: Runtime context
         :return: Dict with error message and jump directive, or None if valid
         """
-        response: str = state.get("messages")[-1].content
+        # The state holds the provider-native message, whose content is a list
+        # of blocks rather than a str whenever the model emitted thinking or
+        # tool_use blocks, or was routed through the OpenAI Responses API.
+        # json.loads() raises TypeError on a list, so project the content to
+        # its text first. For plain-string content this is an identity.
+        last_message: BaseMessage = state.get("messages")[-1]
```
Noravee, line 91: This is the provider-native AIMessage from langchain's agent state, not the flattened copy neuro-san keeps for journaling and the wire, so nothing upstream has normalized it. Anthropic returns list content for thinking/tool_use turns and the OpenAI Responses path always does, which is how `json.loads()` ended up with a list. The shipped copy_cat example stays on Chat Completions (str content), so it never surfaced there.

neuro_san/middleware/network_copy_middleware.py
```diff
@@ -81,7 +83,13 @@ async def aafter_agent(
...
+        response: str = ContentUtils.flatten_to_text(last_message)
```
Noravee, line 92: For str content `flatten_to_text` returns the same string object, so the parsed value and the error log are byte-identical for existing traffic. On list content it concatenates the text blocks the way `BaseMessage.text` does; reasoning and tool_use blocks contribute nothing, which is what a JSON parser wants here.

neuro_san/middleware/network_copy_middleware.py
```diff
@@ -137,11 +145,24 @@ def _parse_agent_name(self, response: str) -> str | None:
         :return: The agent name, or None if parsing/validation fails
         """
         try:
-            args: Dict[str, str] = loads(response)
-        except JSONDecodeError as json_error:
+            args: Any = loads(response)
+        except (JSONDecodeError, TypeError) as json_error:
```
Noravee, line 149: Unreachable from `aafter_agent` now that the response is always projected to str first. Kept as a one-token guard for any future direct caller of `_parse_agent_name`; `test_parse_agent_name_tolerates_non_string` pins it.

neuro_san/middleware/network_copy_middleware.py
```diff
@@ -137,11 +145,24 @@ def _parse_agent_name(self, response: str) -> str | None:
...
+            # TypeError is what json.loads() raises for non-string input. It
+            # cannot happen on the aafter_agent() path, which always projects
+            # the response to str first; it is kept as a cheap guard so that
+            # any future caller bypassing that projection gets the same
+            # "please provide the name" outcome instead of an exception.
             self.logger.error("Cannot parse '%s' into JSON format. Got %s", response, json_error)
             return None
 
+        if not isinstance(args, dict):
```
Noravee, line 158: Pre-existing rather than block-related: valid JSON that is not an object (`"x"`, `["x"]`, `42`) decoded fine and then raised AttributeError on `.get()`, aborting the request the same way the TypeError did. Same failure class, same recovery, so it is folded in as the second commit.

tests/neuro_san/middleware/test_network_copy_middleware.py
```diff
@@ -0,0 +1,196 @@
...
+        """
+        blocks: List[Dict[str, Any]] = [
+            {"type": "thinking", "thinking": "Which network was asked for?", "signature": "sig-abc"},
+            {"type": "text", "text": '{"agent_name": "hello_world"}'},
+        ]
+        middleware: NetworkCopyMiddleware = self.make_middleware()
+        with patch(self.RESTORE_PATH, return_value=None) as restore:
+            response: Dict[str, Any] = await middleware.aafter_agent(self.make_state(blocks), None)
+
+        restore.assert_called_once_with("hello_world")
```
Noravee, line 97: The regression lock: on main this input raises `TypeError: the JSON object must be str, bytes or bytearray, not list` before `_restore_network` is reached. Patching `_restore_network` to return None keeps the registry and Reservationist out of the test while still proving the name was parsed; the "Cannot find agent" text on the next line is just the consequence of that None.

d1donlydfink (approved):

PR #1313, opened 2026-09-09
Flatten both sides of the journal dupe comparison

Noravee:
## What

`OriginatingJournal._is_dupe_of_pending` decides whether the held AGENT copy of an LLM's output (from `on_llm_end`) is the same content as the AI message that follows it, so clients receive only one. Since #1285 it strip-compares when both contents are `str`, and otherwise falls back to a raw `==`.

The held AGENT copy is always a flattened `str`. Once an incoming AI message carries block content (reasoning + text, the next rung of #1222), that raw `==` compares `str == list`, never matches, and both copies leak to clients. This PR makes the comparison text-projected on both sides, ahead of the change that would expose it.

## Fix

Project both `self.pending` and the incoming message through `ContentUtils.flatten_to_text`, then keep the existing strip-compare. `flatten_to_text` is an identity on `str`, so today's string-only traffic compares exactly as before. The method's docstring is restructured to the summary / `:param` / `:return:` template and explains the projection.

## Tests

Two new cases in `test_originating_journal.py`: block content (reasoning + text) with the same visible text as the held AGENT copy is deduped, and block content with different text flushes the AGENT copy first. The three existing tests are unchanged in behavior and now share a small loop-based helper for collecting written messages.

Part of #1222 (Phase 1), PR 4.2 of the ladder described there. A prerequisite for PR 5; independent of 4.1 and 4.3.

### DIFF
--- a/neuro_san/internals/journals/originating_journal.py
+++ b/neuro_san/internals/journals/originating_journal.py
@@ -25,6 +25,7 @@ from langchain_core.messages.system import SystemMessage
 from neuro_san.internals.journals.journal import Journal
 from neuro_san.message.types.agent_tool_result_message import AgentToolResultMessage
 from neuro_san.message.types.base_message_dictionary_converter import BaseMessageDictionaryConverter
+from neuro_san.message.utils.content_utils import ContentUtils
 
 
 class OriginatingJournal(Journal):
@@ -109,15 +110,23 @@ class OriginatingJournal(Journal):
 
     def _is_dupe_of_pending(self, message: BaseMessage) -> bool:
         """
-        :param message: The incoming BaseMessage to compare against the held
-                pending message.
-        :return: True if the pending message carries the same content as the
-                incoming message. String content is compared ignoring leading
-                and trailing whitespace: the pending message's content was
-                stripped when it was captured (JournalingCallbackHandler's
-                on_llm_end), while the incoming message's content is not, so
-                an exact comparison would let a mere trailing newline defeat
-                the suppression and send clients both copies of the same text.
+        Decide whether the held pending message carries the same visible text
+        as the incoming message, so that only one of the two reaches clients.
+
+        Both contents are projected to text with ContentUtils.flatten_to_text
+        and compared ignoring leading and trailing whitespace. The whitespace
+        normalization is needed because the pending message's content was
+        stripped when it was captured (JournalingCallbackHandler's on_llm_end),
+        while the incoming message's content is not, so an exact comparison
+        would let a mere trailing newline defeat the suppression and send
+        clients both copies of the same text.
+
+        Projecting both sides to text matters once an incoming message can
+        carry list content (block content preserved from the provider): the
+        held AGENT copy is always a flattened str, so a raw equality against
+        a block list could never match and clients would receive both copies.
+        flatten_to_text is an identity on str, so string-only traffic compares
+        exactly as it did before.
 
         Normalizing here, rather than stripping at either capture point, is
         deliberate: both captured texts are client-visible and locked by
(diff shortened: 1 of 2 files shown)


Noravee:
TLDR: `_is_dupe_of_pending` now projects both the held AGENT copy and the incoming message through `ContentUtils.flatten_to_text` before the existing strip-compare.

Why: the held copy is always a flattened str (from `on_llm_end`), so once an AI message can carry block content the old `str == list` fallback could never match and clients would receive both copies. Fixed ahead of the change that would expose it (#1222).

Backward compat: `flatten_to_text` is an identity on str, so string-only comparisons behave exactly as they have since #1285. Two new block-content tests; the three existing ones are unchanged in behavior.

neuro_san/internals/journals/originating_journal.py
```diff
@@ -126,12 +135,15 @@ def _is_dupe_of_pending(self, message: BaseMessage) -> bool:
...
+                the incoming message.
         """
-        pending_content: Any = self.pending.content
-        incoming_content: Any = message.content
-        if isinstance(pending_content, str) and isinstance(incoming_content, str):
-            return pending_content.strip() == incoming_content.strip()
-        return pending_content == incoming_content
+        pending_text: str = ContentUtils.flatten_to_text(self.pending)
+        incoming_text: str = ContentUtils.flatten_to_text(message)
+        return pending_text.strip() == incoming_text.strip()
```
Noravee, line 146: The held copy is always a flattened str (`on_llm_end` projects through `ContentUtils` since #1285), so the old raw `==` fallback becomes `str == list` the moment an AI message arrives with block content, and both copies reach clients. Flattening both sides is an identity on str, so today's traffic compares exactly as before; the strip stays because the two capture points still differ in stripping (see the docstring).

tests/neuro_san/internals/journals/test_originating_journal.py
```diff
@@ -87,8 +111,46 @@ async def test_different_content_flushes_pending_first(self):
...
+        """
+        The held AGENT copy is a flattened str while the incoming AI message
+        carries reasoning + text blocks with the same visible text. Both sides
+        are projected to text, so the AGENT copy is still recognized as a dupe
+        and only the AI message reaches the wrapped journal.
+        """
+        journal, wrapped = self._make_journal()
+        await journal.write_message_if_next_not_dupe(AgentMessage(content="the answer"))
+        blocks: List[Dict[str, Any]] = [
+            {"type": "reasoning", "reasoning": "hidden"},
```
Noravee, line 131: The reasoning block plus the trailing newline is deliberate: it exercises both halves of the comparison at once, the text projection (reasoning contributes nothing) and the strip from #1285. Reasoning + text is the shape a block-content answer takes once native block content is preserved through the pipeline (#1222).

tests/neuro_san/internals/journals/test_originating_journal.py
```diff
@@ -87,8 +111,46 @@ async def test_different_content_flushes_pending_first(self):
...
+            {"type": "text", "text": "the answer\n"},
+        ]
+        incoming: AIMessage = AIMessage(content=blocks)
+        await journal.write_message(incoming)
+
+        written: List[BaseMessage] = self._written_messages(wrapped)
+        assert len(written) == 1
+        assert written[0] is incoming
```
Noravee, line 139: Asserting identity rather than `.content == ...` checks that the AI message object itself reaches the wrapped journal untouched: the dedupe must only decide, never rewrite or substitute the payload.

tests/neuro_san/internals/journals/test_originating_journal.py
```diff
@@ -47,8 +58,21 @@ def _make_journal(self):
...
+    @staticmethod
+    def _written_messages(wrapped: MagicMock) -> List[BaseMessage]:
+        """
+        Collect the messages that reached the wrapped journal.
+
+        :param wrapped: The wrapped journal mock whose write_message calls to inspect
+        :return: The messages written to the wrapped journal, in order
+        """
+        written: List[BaseMessage] = []
+        for call in wrapped.write_message.call_args_list:
```
Noravee, line 70: Replaces the inline comprehension the three existing tests carried, per the project's explicit-loop style; the collected list is identical, so their behavior is unchanged.

d1donlydfink (approved):

PR #1314, opened 2026-09-09
Keep use_tool and escape_message str-safe for block content

Noravee:
## What

Three defensive guards at places that assume message content is a `str`. None is reachable with list content today, so all three are identities for current traffic; they exist so that the later rungs of #1222 (which preserve block content through the pipeline) cannot leak a list into a `str` contract.

- **`BranchActivation.use_tool`** and **`ActivationCapsule.use_tool`** are documented to return a `str` but returned the sub-agent's raw `message.content`. They now return `ContentUtils.flatten_to_text(message)`. Their one in-repo caller, the copy_cat `Copyist`, f-strings this value into the calling LLM's tool result, which would otherwise see a Python repr of a block list.
- **`ChatHistoryMessageProcessor.escape_message`** called `str.replace` on the message text. It now passes non-`str` text through untouched rather than crashing or silently dropping the history entry. Every dict it sees today comes from `BaseMessageDictionaryConverter.to_dict`, which always emits `str` text. Its docstring now follows the `:param` / `:return:` template.

## Tests

New modules `test_branch_activation.py`, `test_activation_capsule.py` (with the new `tests/neuro_san/internals/run_context/utils/__init__.py` so CI pylint picks it up), and `test_chat_history_message_processor.py`. Each locks the plain-string behavior unchanged and the block-content projection or pass-through.

Part of #1222 (Phase 1), PR 4.3 of the ladder described there. A prerequisite for PR 5; independent of 4.1 and 4.2.

### DIFF
--- a/neuro_san/internals/chat/chat_history_message_processor.py
+++ b/neuro_san/internals/chat/chat_history_message_processor.py
@@ -18,6 +18,7 @@ from numbers import Number
 from typing import Any
 from typing import Dict
 from typing import List
+from typing import Optional
 
 from copy import copy
 
@@ -113,7 +114,7 @@ class ChatHistoryMessageProcessor(MessageProcessor):
         redacted["text"] = "<redacted>"
         return redacted
 
-    def escape_message(self, chat_message_dict: Dict[str, Any]) -> Dict[str, Any]:
+    def escape_message(self, chat_message_dict: Dict[str, Any]) -> Optional[Dict[str, Any]]:
         """
         Prepare a message such that it can be re-ingested by the system nicely.
         This means properly escaping any text that is sent.
--- a/neuro_san/internals/chat/chat_history_message_processor.py
+++ b/neuro_san/internals/chat/chat_history_message_processor.py
@@ -117,12 +117,24 @@ class ChatHistoryMessageProcessor(MessageProcessor):
         """
         Prepare a message such that it can be re-ingested by the system nicely.
         This means properly escaping any text that is sent.
+
+        :param chat_message_dict: The ChatMessage dictionary to transform
+        :return: A shallow copy of the dictionary with its text escaped,
+                or None if the message carries no text at all
         """
         transformed: Dict[str, Any] = copy(chat_message_dict)
-        text: str = transformed.get("text")
+        text: Any = transformed.get("text")
 
         if text is None:
             return None
+
+        if not isinstance(text, str):
+            # Brace escaping only applies to str text. Every dict reaching this
+            # processor today comes from BaseMessageDictionaryConverter.to_dict,
+            # which always emits str text, so this is a defensive guard for
+            # future non-str text: pass the message through untouched rather
+            # than silently dropping a history entry.
+            return transformed
 
         # Braces are a problem for chat history being read back into the system
         # if they are not properly escaped.
--- a/neuro_san/internals/graph/activations/branch_activation.py
+++ b/neuro_san/internals/graph/activations/branch_activation.py
@@ -34,6 +34,7 @@ from neuro_san.internals.interfaces.callable_activation import CallableActivatio
 from neuro_san.internals.interfaces.invocation_context import InvocationContext
 from neuro_san.internals.run_context.interfaces.run import Run
 from neuro_san.internals.run_context.interfaces.run_context import RunContext
+from neuro_san.message.utils.content_utils import ContentUtils
 
 
 class BranchActivation(CallingActivation, CallableActivation):
@@ -210,8 +211,15 @@ flag to your invocation.
             # load those pools accumulate as leaked sockets.
             await callable_activation.close_of_work(self.run_context)
 
-        # We got a message back, take the content as the return string
-        return message.content
+        # We got a message back. This method's contract is a str. Today the
+        # sub-agent's final message always carries str content (its chain
+        # output is flattened before it is journaled), so this projection is
+        # an identity. Once native block content is preserved through that
+        # message (later rungs of issue #1222), it keeps the contract: callers
+        # hand this value straight to the calling LLM (copyist.py f-strings it
+        # into the tool result) and would otherwise see a Python repr of the
+        # block list.
+        return ContentUtils.flatten_to_text(message)
 
     async def use_reservation(self, reservation_id: str, args: Dict[str, Any], sly_data: Dict[str, Any]) -> str:
         """
--- a/neuro_san/internals/run_context/utils/activation_capsule.py
+++ b/neuro_san/internals/run_context/utils/activation_capsule.py
@@ -31,6 +31,7 @@ from neuro_san.internals.interfaces.context_type_llm_factory import ContextTypeL
 from neuro_san.internals.interfaces.invocation_context import InvocationContext
 from neuro_san.internals.interfaces.lingering_resource import LingeringResource
 from neuro_san.internals.run_context.interfaces.run_context import RunContext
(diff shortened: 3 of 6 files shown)


neuro_san/internals/chat/chat_history_message_processor.py
```diff
@@ -117,13 +117,25 @@ def escape_message(self, chat_message_dict: Dict[str, Any]) -> Dict[str, Any]:
         """
         Prepare a message such that it can be re-ingested by the system nicely.
         This means properly escaping any text that is sent.
+
+        :param chat_message_dict: The ChatMessage dictionary to transform
+        :return: A shallow copy of the dictionary with its text escaped,
```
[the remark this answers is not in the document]
Noravee: Agreed. The None return predates this PR but the docstring now documents it, so the annotation should too. Changed to Optional[Dict[str, Any]] in 3908fb94.

tests/neuro_san/internals/chat/test_chat_history_message_processor.py
```diff
@@ -0,0 +1,71 @@
...
+        Braces in str text are normalized to the doubled, template-safe form,
+        whether or not they were already escaped.
+        """
+        processor: ChatHistoryMessageProcessor = ChatHistoryMessageProcessor()
+        original: Dict[str, Any] = {"type": ChatMessageType.AI, "text": "a {b} and {{c}}"}
+
+        result: Dict[str, Any] = processor.escape_message(original)
+
+        assert result["text"] == "a {{b}} and {{c}}"
+        assert result["type"] == ChatMessageType.AI
```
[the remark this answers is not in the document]
Noravee: Done in 873e3b82: both call sites now annotate the result as Optional and assert it is not None before indexing.

Noravee:
TLDR: Three defensive guards at places that assume message content is a str.

- `BranchActivation.use_tool` and `ActivationCapsule.use_tool` return `ContentUtils.flatten_to_text(message)` instead of raw `message.content`, honoring their documented `-> str` contract.
- `ChatHistoryMessageProcessor.escape_message` passes non-str text through instead of calling `str.replace` on it; its return type is now `Optional[Dict]` to match a `None` return it already had.

Backward compat: every message reaching these points today carries str text, and `flatten_to_text` returns str input as the same object, so this is a no-op for current traffic. The guards start doing work once native block content is preserved through the pipeline (#1222). Three new test modules lock both the unchanged str behavior and the block-content projection.

neuro_san/internals/graph/activations/branch_activation.py
```diff
@@ -210,8 +211,15 @@ async def use_tool(self, tool_name: str, tool_args: Dict[str, Any], sly_data: Di
...
-        return message.content
+        # We got a message back. This method's contract is a str. Today the
+        # sub-agent's final message always carries str content (its chain
+        # output is flattened before it is journaled), so this projection is
+        # an identity. Once native block content is preserved through that
+        # message (later rungs of issue #1222), it keeps the contract: callers
+        # hand this value straight to the calling LLM (copyist.py f-strings it
+        # into the tool result) and would otherwise see a Python repr of the
+        # block list.
+        return ContentUtils.flatten_to_text(message)
```
Noravee, line 222: An identity today: the sub-agent's chain output is flattened to text in `RunContextRunnable.parse_chain_result` before it becomes the AIMessage that `build()` returns, so `message.content` is always a str here. It matters once native block content is preserved through that message (#1222): `Copyist` returns this value as its coded-tool result, which `AbstractClassActivation` f-strings into the calling LLM's tool message, so a block list would render as a Python repr.

neuro_san/internals/run_context/utils/activation_capsule.py
```diff
@@ -107,8 +108,15 @@ async def use_tool(self, tool_name: str, tool_args: Dict[str, Any], sly_data: Di
...
-        return message.content
+        # We got a message back. This method's contract is a str. Today the
+        # sub-agent's final message always carries str content (its chain
+        # output is flattened before it is journaled), so this projection is
+        # an identity. Once native block content is preserved through that
+        # message (later rungs of issue #1222), it keeps the contract: callers
+        # hand this value straight to the calling LLM (copyist.py f-strings it
+        # into the tool result) and would otherwise see a Python repr of the
+        # block list.
+        return ContentUtils.flatten_to_text(message)
```
Noravee, line 119: Kept in lockstep with `BranchActivation.use_tool`: this capsule is what `LangChainRunContext` hands to `MiddlewareFactory`, so it is the `use_tool` entry point middleware reaches, and both must agree on returning text. Same identity for str content.

neuro_san/internals/chat/chat_history_message_processor.py
```diff
@@ -113,17 +114,29 @@ def redact_instructions(self, chat_message_dict: Dict[str, Any]) -> Dict[str, An
...
+                or None if the message carries no text at all
         """
         transformed: Dict[str, Any] = copy(chat_message_dict)
-        text: str = transformed.get("text")
+        text: Any = transformed.get("text")
 
         if text is None:
             return None
 
+        if not isinstance(text, str):
```
Noravee, line 132: Passing non-str text through rather than returning `None` is deliberate: `process_message` treats `None` as "drop this entry", and silently losing a history turn is worse than leaving a block list unescaped. Unreachable today because every dict here comes from `BaseMessageDictionaryConverter.to_dict`, which always emits str text; the `Optional` annotation just documents the pre-existing `return None` above.

tests/neuro_san/internals/graph/activations/test_branch_activation.py
```diff
@@ -0,0 +1,98 @@
...
+    def make_activation(build_result: BaseMessage) -> BranchActivation:
+        """
+        Build a BranchActivation without running its constructor (which
+        creates a real RunContext), wired with a factory whose activation
+        build() returns the given message.
+
+        :param build_result: The BaseMessage the sub-agent's build() will return
+        :return: A BranchActivation ready for use_tool()
+        """
+        activation: BranchActivation = BranchActivation.__new__(BranchActivation)
```
Noravee, line 56: `__new__` plus three attributes because the real constructor goes through `CallingActivation.__init__`, which builds a live `RunContext`; `use_tool` only touches `agent_tool_spec`, `run_context` and `factory`. The third test also locks that `close_of_work` is awaited, since that `finally` is what stops the sub-agent's LLM client pool from leaking on every call.

tests/neuro_san/internals/run_context/utils/test_activation_capsule.py
```diff
@@ -0,0 +1,91 @@
...
+
+from langchain_core.messages.ai import AIMessage
+from langchain_core.messages.base import BaseMessage
+
+from neuro_san.internals.run_context.utils.activation_capsule import ActivationCapsule
+
+from tests.neuro_san.message.content_fixtures import ContentFixtures
+
+
+class TestActivationCapsule:
```
Noravee, line 31: This lives in a new `tests/neuro_san/internals/run_context/utils/` package, hence the empty `__init__.py` next to it: CI pylint only descends into test directories that are packages, so without it this module would never be linted.

neuro_san/internals/run_context/utils/activation_capsule.py
d1donlydfink: FWIW: In the future I'd like all uses of tools which are inherited BranchActivation to use the ActivationCapsule instead. Definitely not for today.

d1donlydfink (approved):

PR #1316, opened 2026-09-09
NS-969 Add parallel data-driven load test driver.

andreidenissov-cog:
## Summary
Adds a parallel, multi-test load driver to the data-driven agent test framework
and refactors the existing single-test driver to sit on top of it. This makes it
possible to fire many agent test cases concurrently (e.g. a 200-request load run)
with per-test timeouts and a success-threshold, while the existing single-test
path keeps working unchanged.

## Changes
- New DataDrivenTestsDriver (data_driven_tests_driver.py): the shared engine that
  runs a sequence of data-driven test-case dicts. run_tests(tests,
  num_need_success) launches one worker thread per test via a ThreadPoolExecutor,
  applies a per-test timeout_in_seconds, and returns a per-test result capture.
  It uses an as_completed_or_timeout iterator so each future is handled exactly
  once -- either as completed or as timed-out -- and counts successes against a
  required threshold (stopping early once enough tests have passed). The per-test
  interaction / response-key checking / thinking-dir setup / sly_data merge logic
  moved here from the single-test driver.
- DataDrivenAgentTestDriver (data_driven_agent_test_driver.py) is now a thin
  wrapper that delegates to DataDrivenTestsDriver for a single test case, so
  existing callers are unaffected (the file shrank from ~425 lines to a small
  delegator).
- New TimedAssertCapture (timed_assert_capture.py): an AssertCapture that also
  records per-test execution time (set/get_execution_time), so load runs can
  report timing per test.
- AssertCapture gains add_assert() to append a captured AssertionError.
- New TestsUtil (tests_util.py) with parse_hocon_test_case(), factoring HOCON
  test-case parsing out so it is shared rather than embedded in the driver.

## Notes
- A single failing/erroring test is captured (with an infinite execution time
  marker) and does not abort the rest of the run; timed-out tests are recorded
  and the executor is shut down without blocking on stragglers.

## Testing
- flake8 clean and pylint 10.00/10 on all changed files.
- Modules import cleanly.

## Related issue
NS-969


neuro_san/test/driver/data_driven_tests_driver.py
```diff
@@ -0,0 +1,520 @@
...
+
+        # Capture asserts for each test execution
+        run_results: List[TimedAssertCapture] = []
+
+        # Loop through each iteration, capturing any asserts.
+        num_successful: int = 0
+
+        # Loop through each test execution in parallel
+        with ThreadPoolExecutor(max_workers=len(tests)) as executor:
+
```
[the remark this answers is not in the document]
andreidenissov-cog: Fixed in next PR update.

neuro_san/test/driver/data_driven_agent_test_driver.py
```diff
@@ -100,381 +71,29 @@ def one_test(self, hocon_file: str):
...
-            for future in as_completed(futures):
-                assert_capture: AssertCapture = future.result()
-                iteration_asserts.append(assert_capture)
-
-                asserts: List[AssertionError] = assert_capture.get_asserts()
-                if len(asserts) > 0:
-                    # Not successful
-                    continue
+        # Capture test results for each iteration
+        test_results: List[TimedAssertCapture] = self.test_driver.run_tests(tests, num_need_success)
```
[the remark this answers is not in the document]
andreidenissov-cog: Fixed in next PR update.

neuro_san/test/driver/data_driven_tests_driver.py
```diff
@@ -0,0 +1,520 @@
...
+                # Don't include an iteration index if there is only one test to do.
+                if len(tests) == 1:
+                    iteration_index = None
+
+                timeout_in_seconds: Optional[float] = test_case.get("timeout_in_seconds", None)
+                future: Future = executor.submit(
+                    self.capture_one_iteration, test_case, timeouts, iteration_index)
+                futures.append(future)
+                future_timeouts[future] = timeout_in_seconds
```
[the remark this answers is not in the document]
andreidenissov-cog: Fixed.

neuro_san/test/driver/data_driven_tests_driver.py
```diff
@@ -0,0 +1,534 @@
...
+
+        while pending:
+            now = monotonic()
+
+            # 1) Emit futures whose deadline has passed but which haven't finished.
+            #    (If one finished right at its deadline, prefer reporting it completed.)
+            expired: List[Future] = []
+            for fut in pending:
+                if deadlines[fut] is not None and now >= deadlines[fut] and not fut.done():
+                    expired.append(fut)
```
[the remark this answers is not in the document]
andreidenissov-cog: That is possible, but: in the test framework we are not that interested in exact timeouts tracking. Timeouts are more protective measure against tests that could hang forever or take VERY long time to finish.

neuro_san/test/driver/data_driven_tests_driver.py
```diff
@@ -0,0 +1,534 @@
...
+
+                num_successful += 1
+                if num_successful == num_need_success:
+                    # Fast path: we have enough successful tests, so we can stop waiting for more.
+                    break
+        finally:
+            # We are done with running tests, so we can shut down the executor and cancel any remaining futures.
+            # Note: this is not a blocking call, but if some timed out tests are still running,
+            # they will continue to run in the background. We are just not waiting for them anymore.
+            executor.shutdown(wait=False, cancel_futures=True)
```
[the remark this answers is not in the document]
andreidenissov-cog: That is also possible, but maybe not that critical: if happens when we have an early test run success, and this outcome is more important than some mis-counting of results.

neuro_san/test/driver/data_driven_tests_driver.py
```diff
@@ -0,0 +1,534 @@
...
+                    timed_capture.add_assert(
+                        AssertionError(f"Test for run {self.test_name} failed: 'agent' not specified in test case."))
+                    run_results.append(timed_capture)
+                    continue
+
+                # Don't include an iteration index if there is only one test to do.
+                if len(tests) == 1:
+                    iteration_index = None
+
+                timeout_in_seconds: Optional[float] = test_case.get("timeout_in_seconds", None)
```
vince-leaf, line 118: When a test hits timeout_in_seconds, as_completed_or_timeout reports it as timed out, and the driver moves on — but nothing tells the worker thread to stop. The old code put this timeout into the timeouts list so the worker itself checked it during streaming and aborted; here, timeouts (line 81) is created but never filled, so the worker never knows it timed out and runs to completion in the background, still consuming a thread, a session, and server load after being reported as failed.

neuro_san/test/driver/data_driven_tests_driver.py
```diff
@@ -0,0 +1,534 @@
...
+                future_timeouts[future] = timeout_in_seconds
+
+            for fut, timed_out in self.as_completed_or_timeout(future_timeouts):
+                timed_capture: TimedAssertCapture = None
+                if timed_out:
+                    # This test iteration has timed out.
+                    # We can't get the result, but we can still record the timeout.
+                    timed_capture = TimedAssertCapture(self.asserts_basis)
+                    timed_capture.set_execution_time(float('inf'))  # Indicate that it timed out
+                    timed_capture.add_assert(AssertionError(f"Test for run {self.test_name} timed out."))
```
vince-leaf, line 133: The timeout message only has test_name, so you can't tell which of the N tests timed out.

neuro_san/test/driver/data_driven_tests_driver.py
vince-leaf, line 163: Since shutdown(wait=False) doesn't join the still-running timed-out threads, the process can hang at exit until they finish.

vince-leaf (changes requested):
See my 3 comments above on data_driven_tests_driver.py. This is not a crash/data-corruption bug — results are still correct. It's a resource/cleanup and diagnosability issue, plus a silent semantic change vs. the documented timeout_in_seconds behavior.

neuro_san/test/driver/data_driven_tests_driver.py
```diff
@@ -0,0 +1,534 @@
...
+                current_sly_data.update(returned_sly_data)
+                return current_sly_data
+            return returned_sly_data.copy()
+        return current_sly_data
+
+    def _setup_thinking_dir(
+        self,
+        fixture_hocon_name: str,
+        iteration_index: int,
+    ) -> str:
```
d1donlydfink, line 428: This comes whole-cloth from DataDrivenAgentTestDriver unchanged.

neuro_san/test/driver/data_driven_tests_driver.py
```diff
@@ -0,0 +1,534 @@
...
+
+    def check_timeouts(self, timeouts: List[Timeout]):
+        """
+        :param timeouts: A list of timeout objects to check
+        """
+        for one_timeout in timeouts:
+            Timeout.check_if_not_none(one_timeout)
+
+    def _merge_sly_data(self, current_sly_data: Dict[str, Any],
+                        returned_sly_data: Dict[str, Any]) -> Dict[str, Any]:
```
d1donlydfink, line 403: This comes whole-cloth from DataDrivenAgentTestDriver unchanged.

neuro_san/test/driver/data_driven_tests_driver.py
```diff
@@ -0,0 +1,534 @@
...
+                                                                                   evaluator_type)
+                if evaluator is not None:
+                    evaluator.evaluate(processor, verify_key, test_key_value)
+                    self.check_timeouts(timeouts)
+
+        # Recurse if there are further dictionary specs to dive into
+        if len(deeper_test_keys) > 0:
+            self.test_response_keys(processor, response_extractor, deeper_test_keys, asserts, timeouts)
+
+    def check_timeouts(self, timeouts: List[Timeout]):
```
d1donlydfink, line 395: This comes whole-cloth from DataDrivenAgentTestDriver unchanged.

neuro_san/test/driver/data_driven_tests_driver.py
```diff
@@ -0,0 +1,534 @@
...
+        else:
+            return_sly_data = current_sly_data
+
+        return return_chat_context, return_sly_data
+
+    def test_response_keys(self, processor: BasicMessageProcessor,
+                           response_extractor: DictionaryExtractor,
+                           keys: List[str],
+                           asserts: AssertForwarder,
+                           timeouts: List[Timeout]):
```
d1donlydfink, line 357: This comes whole-cloth from DataDrivenAgentTestDriver unchanged.

neuro_san/test/driver/data_driven_tests_driver.py
```diff
@@ -0,0 +1,534 @@
...
+            # Although it is unlikely that two tests run at the same time...
+            if os.path.exists(thinking_dir):
+                shutil.rmtree(thinking_dir)
+            # Create the directory anew
+            os.makedirs(thinking_dir)
+
+        return thinking_dir
+
+    def as_completed_or_timeout(self,
+                                future_timeouts: Dict[Future, Optional[float]]) -> Iterator[Tuple[Future, bool]]:
```
d1donlydfink, line 478: This method seems to be new

neuro_san/test/driver/data_driven_tests_driver.py
```diff
@@ -0,0 +1,534 @@
...
+                    fixture_hocon_name,
+                    iteration_index,
+                    carried_sly_data
+                )
+
+    # pylint: disable=too-many-locals,too-many-arguments,too-many-positional-arguments
+    def interact(self, agent: str, session: AgentSession, interaction: Dict[str, Any],
+                 chat_context: Dict[str, Any], asserts: AssertForwarder,
+                 timeouts: List[Timeout], fixture_hocon_name: str, iteration_index: int,
+                 sly_data: Dict[str, Any]) -> tuple:
```
d1donlydfink, line 264: This comes whole-cloth from DataDrivenAgentTestDriver unchanged.

neuro_san/test/driver/data_driven_agent_test_driver.py
```diff
@@ -100,381 +73,33 @@ def one_test(self, hocon_file: str):
...
-                    interaction,
-                    chat_context,
-                    asserts,
-                    timeouts,
-                    fixture_hocon_name,
-                    iteration_index,
-                    carried_sly_data
-                )
-
-    def parse_hocon_test_case(self, hocon_file: str) -> Dict[str, Any]:
```
d1donlydfink, line 248: This moves to TestsUtil

neuro_san/test/driver/data_driven_tests_driver.py
```diff
@@ -0,0 +1,534 @@
...
+        # Perform a single iteration of the test.
+        self.one_iteration(test_case, assert_capture, timeouts, fixture_hocon_name, iteration_index)
+        end_time: float = monotonic()
+        assert_capture.set_execution_time(end_time - start_time)
+
+        return assert_capture
+
+    # pylint: disable=too-many-locals, too-many-arguments, too-many-positional-arguments
+    def one_iteration(self, test_case: Dict[str, Any], asserts: AssertForwarder,
+                      timeouts: List[Timeout], fixture_hocon_name: str, iteration_index: int):
```
d1donlydfink, line 189: This comes whole-cloth from DataDrivenAgentTestDriver unchanged.

neuro_san/test/driver/data_driven_tests_driver.py
```diff
@@ -0,0 +1,534 @@
...
+                    break
+        finally:
+            # We are done with running tests, so we can shut down the executor and cancel any remaining futures.
+            # Note: this is not a blocking call, but if some timed out tests are still running,
+            # they will continue to run in the background. We are just not waiting for them anymore.
+            executor.shutdown(wait=False, cancel_futures=True)
+        return run_results
+
+    def capture_one_iteration(self, test_case: Dict[str, Any], timeouts: List[Timeout],
+                              iteration_index: int) -> TimedAssertCapture:
```
d1donlydfink, line 167: This comes from DataDrivenAgentTestDriver but adds the execution time stuff.

neuro_san/test/driver/data_driven_tests_driver.py
```diff
@@ -0,0 +1,534 @@
...
+        Constructor
+        :param asserts: The AssertForwarder instance to use to integrate failures
+                        back into the test system.
+        :param test_name: Optional name of the test run for logging and traceability.
+        """
+        self.asserts_basis: AssertForwarder = asserts
+        self.test_name: str = test_name
+
+    # pylint: disable=too-many-locals
+    def run_tests(self, tests: Sequence[Dict[str, Any]], num_need_success: int) -> List[TimedAssertCapture]:
```
d1donlydfink, line 73: This is a new method

neuro_san/test/driver/data_driven_tests_driver.py
```diff
@@ -0,0 +1,534 @@
...
+            futures: List[Future] = []
+            iteration_index: int = 0
+            future_timeouts: Dict[Future, Optional[float]] = {}
+
+            for test_case in tests:
+                agent: Optional[str] = test_case.get("agent")
+                if agent is None:
+                    # If the agent is not specified in the test case, we cannot proceed with this test.
+                    # Record an assertion failure and continue to the next test case.
+                    timed_capture = TimedAssertCapture(self.asserts_basis)
```
d1donlydfink, line 107: Do we (at least eventually) want a new/copy-of asserts_basis per test?
That is: Do we want to collect separately multi-run tests failures and then aggregate the results later?

As it stands, I believe the single list in AssertCapture would collect everything, thus muddying which test run iteration was the source of any problem.

neuro_san/test/driver/data_driven_tests_driver.py
```diff
@@ -0,0 +1,534 @@
...
+        run_results: List[TimedAssertCapture] = []
+
+        # Loop through each iteration, capturing any asserts.
+        num_successful: int = 0
+
+        if not tests:
+            return []
+
+        # Loop through each test execution in parallel
+        executor: ThreadPoolExecutor = ThreadPoolExecutor(max_workers=len(tests))
```
d1donlydfink, line 96: Having the executor here could allow the Assessor to go much faster as a bonus.

d1donlydfink (commented):
This is a great start!

Some questions. @vince-leaf + Devin's points are also quite valid.

neuro_san/test/driver/data_driven_tests_driver.py
andreidenissov-cog: Yes. That's a valid point, including general thread-unsafety and racing over shared AssertForwarder instance.
So we do need separate copies per test and then some aggregation API.

andreidenissov-cog: OK now we make full deep copies of underlying AssertForwarder instance for each test.

neuro_san/test/driver/data_driven_tests_driver.py
andreidenissov-cog: OK messages now include "iteration_index" of the test. TimedAssertCapture instances also retain test indices now to help with logging.

neuro_san/test/driver/data_driven_tests_driver.py
andreidenissov-cog: Yes it  can. So here is the tradeoff - for the case when some test(s) get hung or just run for a VERY long time:
1. Our test driver itself will get blocked waiting for all tests to finish: executor.shutdown(wait=True).
    In this case no test results are produced.
2. We detect specific test(s) timeout externally, construct overall test run result and return it to a client. Stray tests continue to run - executor.shutdown(wait=False) possibly blocking test driver process.
  I chose option 2 - getting test run results is a priority.

neuro_san/test/driver/data_driven_tests_driver.py
andreidenissov-cog: Couple of things here:
In Python nothing can tell a thread to stop (big bummer actually) unless thread cooperates;
"worker itself checks timeouts" is generally an illusion. It does check timeouts, but AFTER it gets some response from corresponding session. If session doesn't respond, but hangs, this will not help you.
If you specify timeout of say 30 sec, but session responds after 1000 sec, you will wait for 1000 seconds.
There of course may be other timeouts in execution pipeline, but we don't know or control them here.
In proposed implementation, we will mark test as timed out after approx. 30 sec, give or take, and move on.

neuro_san/test/driver/data_driven_tests_driver.py
andreidenissov-cog: Take that back (bummer) - making deepcopy of AssertForwarder messes up our unit tests.
Because pickle.
Reverted that, maybe we'll have to revisit the issue later. If needed.

neuro_san/test/driver/data_driven_tests_driver.py
```diff
@@ -0,0 +1,550 @@
...
+
+class DataDrivenTestsDriver:
+    """
+    Class which manages the execution of a collection of data-driven test cases
+    represented as Python dictionaries.
+    """
+
+    TEST_KEYS: List[str] = ["text", "structure", "sly_data"]
+
+    def __init__(self, asserts: AssertForwarder, test_name: str = None):
```
d1donlydfink, line 62: A little odd that this class is plural, but the constructor takes something singular.

neuro_san/test/driver/data_driven_tests_driver.py
d1donlydfink: That's fine. Just want this to be in your mind.
Less important for unit test (cuz it already was less important there) but more important for load test.

d1donlydfink (approved):

neuro_san/test/driver/data_driven_tests_driver.py
vince-leaf: Following up on my earlier comment about the test-level timeout, after talking with Andrei about whether there is an API to close the session from the worker side.

I checked the code and verified it locally: there is no explicit close() on AgentSession, but one isn't needed — the existing Timeout mechanism already closes the connection.

Before this PR, in one_test, the test-level timeout_in_seconds was turned into a Timeout and appended to the timeouts list passed to each worker. interact() calls check_timeouts() on every streamed chunk; when the limit is hit, the worker raises, the with requests.post(..., stream=True) block in HttpServiceAgentSession.streaming_chat exits, the HTTP connection closes, and the server sees the disconnect (verified with a local streaming HTTP server: client raised at 2 s, server got BrokenPipeError on its next write).

After this PR, in run_tests (data_driven_tests_driver.py), timeouts: List[Timeout] = [] is created, but nothing is ever appended to it, so check_timeouts() inside the worker has nothing to check. timeout_in_seconds is stored only in future_timeouts and is used by the main thread in as_completed_or_timeout to report the test as timed out. The worker keeps reading the stream until the server finishes: its thread and HTTP connection stay open, the server request stays active, and process exit is blocked until they finish. executor.shutdown(wait=False, cancel_futures=True) doesn't change this — running threads can't be canceled. Verified with a stubbed 6 s stream and timeout_in_seconds = 2: main → 0 threads left after 2 s; PR → 3 threads still streaming, exit blocked for 4 s more.

`main:    one_test returned after 2.0s | live worker threads: 0 | exit blocked: 0.0s
PR:      one_test returned after 2.0s | live worker threads: 3 | exit blocked: 4.0s`


Suggestion: in run_tests, build a Timeout from each test's timeout_in_seconds and pass it in that worker's timeouts list, alongside the existing future_timeouts deadline. The per-interaction timeouts in interact() already work this way, so the plumbing is in place. Fine with doing this as a follow-up ticket.

vince-leaf (approved):
I have approved this with the condition that we look into a timeout on the HTTP request "connect" remaining open from the client to NS.

https://github.com/cognizant-ai-lab/neuro-san/issues/1338
Test-level timeout_in_seconds no longer stops worker thread in DataDrivenTestsDriver

PR #1258, opened 2026-08-16
noisy client on windows

mismadh:
Summary
Fix noisy prompt in agent_cli when running on Windows / PowerShell.

Problem
On Windows, timedinput.timedinput() in the CLI's input loop misbehaves and re-prints the prompt on every poll cycle. Because DEFAULT_PROMPT ends with \n, each redraw lands on a new line — flooding the terminal with:

Please enter your response ('quit' to terminate):
Please enter your response ('quit' to terminate):
Please enter your response ('quit' to terminate):
...
The client is effectively unusable interactively in PowerShell. Users can still submit input via --first_prompt_file, but the interactive mode is broken.

Fix
On Windows only, bypass timedinput and use plain blocking input() in agent_cli.py. POSIX behavior is unchanged.

if sys.platform.startswith("win"):
    try:
        user_input = input(prompt)
    except EOFError:
        ...
else:
    user_input = timedinput(prompt, timeout=timeout, default=self.TIMEOUT_SIGNAL)
    ...
Trade-off
On Windows, the idle-timeout (input_timeout_seconds, default 5000s) no longer applies — the CLI blocks indefinitely for input. Users exit with quit or Ctrl+C. This felt like an acceptable trade for a working interactive prompt. Happy to explore a Windows-native timed-input alternative in a follow-up if reviewers prefer.

Test plan
 macOS/Linux: run python -m neuro_san.client.agent_cli --http --agent hello_world — confirm prompt and idle timeout behavior are unchanged.
 Windows PowerShell: run the same command — confirm the prompt appears once, blocks on input, submits on Enter, and quit exits cleanly.
 Windows PowerShell: run with --first_prompt_file msg.txt — confirm it still submits the file contents first, then prompts once for the next turn.

PR #1318, opened 2026-09-10
Fix noisy windows agent_cli behavior

d1donlydfink:
A rework of the original work from https://github.com/cognizant-ai-lab/neuro-san/pull/1258 from @mismadh
which did not have pylint passing.   Given the way that PR was put together I could not pull the branch and fix it myself in there, so I create this one and obsolete the other.

### DIFF
--- a/neuro_san/client/agent_cli.py
+++ b/neuro_san/client/agent_cli.py
@@ -21,6 +21,7 @@ from typing import List
 
 import os
 import shutil
+import sys
 
 import argparse
 import json
@@ -193,12 +194,22 @@ Some suggestions:
             user_input = state.get("user_input")
             if user_input is None:
                 if prompt is not None and len(prompt) > 0:
-                    user_input = timedinput(prompt, timeout=timeout,
-                                            default=self.TIMEOUT_SIGNAL)
-                    if user_input == self.TIMEOUT_SIGNAL:
-                        # Timeout occurred - nobody is talking to us.
-                        print(f"\nNo input received after {timeout} seconds. Exiting.\n")
-                        break
+                    if not sys.platform.startswith("win"):
+                        user_input = timedinput(prompt, timeout=timeout, default=self.TIMEOUT_SIGNAL)
+                        if user_input == self.TIMEOUT_SIGNAL:
+                            # Timeout occurred - nobody is talking to us.
+                            print(f"\nNo input received after {timeout} seconds. Exiting.\n")
+                            break
+                    else:
+                        # On Windows the timedinput polling loop misbehaves in
+                        # PowerShell (prompt is redrawn on every poll). Use a
+                        # plain blocking input() there — we lose the idle
+                        # timeout but gain a usable CLI.
+                        try:
+                            user_input = input(prompt)
+                        except EOFError:
+                            print("\nStdin closed. Exiting.\n")
+                            break
                 else:
                     user_input = None
 


neuro_san/client/agent_cli.py
```diff
@@ -193,12 +194,22 @@ def loop_until_done(self, user_input: str, sly_data: Dict[str, Any],
...
-                    if user_input == self.TIMEOUT_SIGNAL:
-                        # Timeout occurred - nobody is talking to us.
-                        print(f"\nNo input received after {timeout} seconds. Exiting.\n")
-                        break
+                    if not sys.platform.startswith("win"):
+                        user_input = timedinput(prompt, timeout=timeout, default=self.TIMEOUT_SIGNAL)
+                        if user_input == self.TIMEOUT_SIGNAL:
+                            # Timeout occurred - nobody is talking to us.
+                            print(f"\nNo input received after {timeout} seconds. Exiting.\n")
+                            break
```
d1donlydfink, line 202: Previous behavior  goes into positive  if-block

neuro_san/client/agent_cli.py
```diff
@@ -193,12 +194,22 @@ def loop_until_done(self, user_input: str, sly_data: Dict[str, Any],
...
+                    else:
+                        # On Windows the timedinput polling loop misbehaves in
+                        # PowerShell (prompt is redrawn on every poll). Use a
+                        # plain blocking input() there — we lose the idle
+                        # timeout but gain a usable CLI.
+                        try:
+                            user_input = input(prompt)
+                        except EOFError:
+                            print("\nStdin closed. Exiting.\n")
+                            break
```
d1donlydfink, line 212: @mismadh 's original fix in the windows block with pylint fix in place (no f-string necessary)

andreidenissov-cog (approved):
Good fix.

neuro_san/client/agent_cli.py
```diff
@@ -193,12 +194,22 @@ def loop_until_done(self, user_input: str, sly_data: Dict[str, Any],
...
+                        user_input = timedinput(prompt, timeout=timeout, default=self.TIMEOUT_SIGNAL)
+                        if user_input == self.TIMEOUT_SIGNAL:
+                            # Timeout occurred - nobody is talking to us.
+                            print(f"\nNo input received after {timeout} seconds. Exiting.\n")
+                            break
+                    else:
+                        # On Windows the timedinput polling loop misbehaves in
+                        # PowerShell (prompt is redrawn on every poll). Use a
+                        # plain blocking input() there — we lose the idle
+                        # timeout but gain a usable CLI.
```
Noravee, line 207: The comment attributes the flood to the polling loop redrawing "on every poll", but no released timedinput version writes anything on an idle poll (0.1.1 `win_timedinput`: the idle path is just `time.sleep(INTERVAL)` at timedinput.py:80; same in 2.0.0 at :186). What actually happens: the Windows implementation re-echoes `"\r" + prompt + line`. In 0.1.1/1.0.1 that is only on Backspace (timedinput.py:73-76). In 2.0.0 it is on *every* keystroke via `redraw()` (timedinput.py:136-138, called at :163/:180/:184). Because `DEFAULT_PROMPT` ends in `\n` (line 48), each redraw starts a new line instead of overwriting the old one. It is also not PowerShell-specific: dispatch is `sys.platform.startswith("win")` + msvcrt, any console.

Version matters: requirements.txt:70 is uncapped (`timedinput>=0.1.1`), so CI and fresh installs get 2.0.0 (per-keystroke redraw), while older envs with 0.1.1 only see it on Backspace, which is why it is hard to reproduce locally. #1258 was filed 2026-08-16, three months after 2.0.0 shipped (2026-05-15).

Since the trigger is the newline in *our* prompt, there is a smaller fix that keeps the idle timeout on Windows and needs no platform fork:

```python
print(prompt, end="", flush=True)
user_input = timedinput("", timeout=timeout, default=self.TIMEOUT_SIGNAL)
```

Simulating `win_timedinput` with a fake `msvcrt` (typing `hi`, Backspace, `ello`, Enter): with the prompt passed through, 0.1.1 emits the prompt 2x and 2.0.0 8x. With `""` passed, 0 duplicates on all three versions, and the timeout default still returns. POSIX is unchanged either way.

Trade-off to weigh: plain `input()` does give native console line editing and reads piped stdin on Windows (`_kbhit` is console-only, so the library never saw pipes). If `input()` stays, suggest at least rewording this comment to the actual mechanism so future readers do not go looking for a polling bug.

neuro_san/client/agent_cli.py
```diff
@@ -193,12 +194,22 @@ def loop_until_done(self, user_input: str, sly_data: Dict[str, Any],
...
             if user_input is None:
                 if prompt is not None and len(prompt) > 0:
-                    user_input = timedinput(prompt, timeout=timeout,
-                                            default=self.TIMEOUT_SIGNAL)
-                    if user_input == self.TIMEOUT_SIGNAL:
-                        # Timeout occurred - nobody is talking to us.
-                        print(f"\nNo input received after {timeout} seconds. Exiting.\n")
-                        break
+                    if not sys.platform.startswith("win"):
+                        user_input = timedinput(prompt, timeout=timeout, default=self.TIMEOUT_SIGNAL)
```
Noravee, line 198: With timedinput 2.0.0 (what CI installed for this PR, and what an uncapped `>=0.1.1` resolves to), `posix_timedinput` re-raises `EOFError` when stdin closes (timedinput.py:114-116). This call is unguarded, so on Linux/macOS `echo hello | python -m neuro_san.client.agent_cli --agent X` (without `--one_shot`) prints the answer and then dies with an EOFError traceback, while the new Windows branch below exits cleanly with "Stdin closed. Exiting." (Under 0.1.1 it instead returns `""` and the loop keeps sending empty messages until `--max_input`.)

Pre-existing on main, but since this PR adds EOF handling for one platform, suggest hoisting the `try/except EOFError` so it wraps both calls (or the single call, if the empty-prompt variant above is used).

Verified:

```
printf "" | python -c "from timedinput import timedinput; print(repr(timedinput('P>\n', timeout=2, default='<T>')))"
  2.0.0 -> EOFError: EOF when reading a line
  0.1.1 -> ''
```

Noravee (commented):
Not blocking. The change is safe (POSIX path is behavior-identical, flake8/pylint clean) and the Windows branch does stop the flooding. Leaving notes from digging into *why* it floods, because the stated root cause does not match the library source and a smaller fix that keeps the idle timeout is possible. Details inline.

Two follow-ups outside this diff:

- **Tests.** No unit test covers `loop_until_done` and there is no Windows runner, so CI green here only proves lint. The branch is testable on Linux/macOS by monkeypatching `sys.platform`, builtin `input`, and `timedinput`. Opened #1322 for that.
- **Dependency.** `timedinput>=0.1.1` in requirements.txt is uncapped. 2.0.0 (May 2026) also moved POSIX to SIGALRM + builtin `input()` (main-thread only), added a hard `import readline`, and changed EOF to raise. CLI behavior currently depends on install date; worth an upper bound like the other volatile pins in that file.

Caveat: nothing here was executed on a real Windows console. Windows claims come from the msvcrt-based source of timedinput 0.1.1 / 1.0.1 / 2.0.0 plus a fake-`msvcrt` simulation of `win_timedinput`.

PR #1320, opened 2026-09-10
Journal block-list tool outputs; keep chat history and tool results text

Noravee:
Part of the "preserve native messages" rung of #1222, split into three PRs. This one covers the tool boundaries and is independent of the other two (`find-ai-message-refactor`, and the answer-preservation PR that follows it).

## What changes

- **Journal keeps block-list tool outputs.** `JournalingCallbackHandler.on_tool_end` fires for toolbox and MCP tools. When such a tool returns a list of standard content blocks (text plus image/file data), the journaled `AgentToolResultMessage` now carries that list, with bytes payloads turned into base64 so it serializes, instead of the Python repr `str()` made of it. Every other output shape, including list-of-str, keeps `str(output)` exactly as before.
- **`ContentUtils.to_json_safe` is public.** It was `_to_json_safe`; `on_tool_end` calls it directly. Its two comprehensions are now loops.
- **Chat history stays text.** In `OriginatingJournal.write_message`, the history copy of any tool result, and of any AI message whose content is a block list, is a plain `AIMessage` holding `history_safe_text`: text blocks kept, data blocks replaced by a short `[image attachment: image/png]` reference. Identity for string content. The journaled message keeps its blocks. The history list feeds the next model call, which can be a fallback from a different provider, and is exported as chat_context, so it stays provider-agnostic text.
- **Sub-agent answers reach the calling LLM as text.** `LangChainOpenAIFunctionTool._arun` returns `history_safe_text` of the answer instead of its raw content. An identity today, kept as the guard at that boundary.
- **requirements.txt comment.** The `langchain-mcp-adapters<0.2.0` rationale said `on_tool_end` would log the raw list; that no longer applies. The ceiling itself stays until 0.2.x is verified end to end.

Note on reachability: the pinned adapter (0.1.14) hands `on_tool_end` `str | list[str]` content and puts MCP images in the ToolMessage artifact, so today the block path is reached by toolbox tools that return blocks; the 0.2.x adapter shape is what the fixture documents.

## Parity

Text-only traffic is unchanged at every layer: `str(output)` for non-block tool outputs, the same-instance history append for string AI messages, and the same string returned from `_arun`.

## Tests

12 new tests across `test_content_utils`, `test_journaling_callback_handler`, `test_originating_journal` and `test_langchain_openai_function_tool`.


neuro_san/internals/journals/originating_journal.py
```diff
@@ -90,8 +90,20 @@ async def write_message(self, message: BaseMessage, origin: List[Dict[str, Any]]
...
+            # different provider) and the chat_context returned to clients.
+            # Providers reject data blocks and other providers' native blocks
+            # in assistant-role history (an API 400 mid-conversation), and
+            # neither consumer should change for text-only traffic.
+            # history_safe_text keeps the text, replaces data blocks with a
+            # short reference, and is an identity for plain-string content.
+            # The journaled message keeps the full blocks.
+            if isinstance(message, AgentToolResultMessage) or \
+                    (isinstance(message, AIMessage) and not isinstance(message.content, str)):
+                chat_history_message: BaseMessage = AIMessage(content=ContentUtils.history_safe_text(message))
```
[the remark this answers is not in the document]
Noravee: Deliberate, and now stated at this site (c90d42aa). Nothing breaks for extended-thinking flows today: the session is rebuilt per request from chat_context, and in Phase 1 that is text-only on the wire (`from_dict` reads `text`), so this in-memory list is never replayed with blocks. Each run context also invokes the model once (`LangChainRun.requires_action()` is hard-coded False), and langchain's agent graph carries the signed thinking blocks within a tool loop itself.

Replaying reasoning across turns is provider-specific: Anthropic accepts only its own signed thinking blocks and strips prior-turn thinking anyway, and OpenAI Responses needs its own reasoning item ids. This list is also handed to fallback models that can be a different provider, where a foreign reasoning block produces a rejected payload (verified against `_get_request_payload` for Anthropic, OpenAI Responses and Bedrock Converse). The provider-aware replay path is Phase 2 work, tracked in #1223. Until then the history carries exactly what providers receive today.

Noravee:
**TL;DR** Block content at the tool boundaries: the journal keeps it, everything provider-facing stays text.

- `on_tool_end` (toolbox and MCP tools): a list of standard content blocks is journaled as that list, bytes turned into base64, through the now-public `ContentUtils.to_json_safe`. Any other output shape keeps `str(output)`.
- `OriginatingJournal.write_message`: the chat-history copy of any tool result, and of any AI message with block content, is a plain text `AIMessage` via `history_safe_text` (text kept, data blocks become `[image attachment: image/png]`). Identity for strings. Reasoning is deliberately not replayed in this phase; provider-aware replay is Phase 2 (#1223).
- `LangChainOpenAIFunctionTool._arun`: sub-agent answers reach the calling LLM as that same text. Identity today.
- One deliberate client-visible change: a tool returning a bare text-block list used to be journaled as its Python repr; now the wire text is the text itself.
- 12 new tests; full suite 683 passed; pylint 10.00/10, flake8 clean.

Independent of the other two PRs of this rung (#1321 and the answer-preservation PR that follows it).

neuro_san/internals/run_context/langchain/journaling/journaling_callback_handler.py
```diff
@@ -279,15 +280,36 @@ async def on_tool_end(self, output: Any, run_id: UUID = None, tags: List[str] =
...
+            # image/file data): toolbox tools that return blocks do, and so
+            # does langchain-mcp-adapters>=0.2, while the <0.2 pin in
+            # requirements.txt hands over text only and puts MCP images in the
+            # ToolMessage artifact. Preserve a block list as a JSON-safe list
+            # so the payload reaches the journal intact instead of collapsing
+            # to the Python repr that str() would produce. Every other shape,
+            # str and list-of-str included, keeps exactly the str() form it
+            # has always had.
+            result_content: Union[str, List[Any]]
+            if ContentUtils.looks_like_blocks(output):
```
Noravee, line 293: Only toolbox and MCP tools reach this method (they carry the `langchain_tool` tag). `looks_like_blocks` is strict: every item must be a dict with a known v1 block type, and tool-call blocks or v0 `source_type` shapes disqualify the list, so list-of-str and provider-native lists still take the `str()` branch below.

neuro_san/internals/run_context/langchain/journaling/journaling_callback_handler.py
```diff
@@ -279,15 +280,36 @@ async def on_tool_end(self, output: Any, run_id: UUID = None, tags: List[str] =
...
             )
 
-            # Also log the tool output to the LangChain tool-specific journal
+            # Also log the tool output to the LangChain tool-specific journal.
+            # The structure carries the raw output as before, except that a
+            # block list is the sanitized copy so the structure stays
+            # serializable when blocks carry bytes payloads.
+            tool_output: Any = output
+            if isinstance(result_content, list):
+                tool_output = result_content
```
Noravee, line 309: The tool journal's structure gets the sanitized copy only when the output was a block list, so a bytes payload cannot make the structure unserializable. Every other output is passed through raw, as before.

neuro_san/message/utils/content_utils.py
```diff
@@ -401,7 +401,7 @@ def _data_block(mime_type: str, base64_str: str) -> Dict[str, Any]:
         return {"type": kind, "base64": base64_str, "mime_type": mime_type}
 
     @staticmethod
-    def _to_json_safe(value: Any) -> Any:
+    def to_json_safe(value: Any) -> Any:
```
Noravee, line 404: Was `_to_json_safe`. Public now because `on_tool_end` is a second caller; behavior is unchanged: bytes become base64, non-finite floats and unknown objects become their string form. Deliberately distinct from `ChatMessageConverter.to_json_safe`, which nulls bytes.

neuro_san/internals/journals/originating_journal.py
```diff
@@ -90,8 +90,27 @@ async def write_message(self, message: BaseMessage, origin: List[Dict[str, Any]]
...
+            # short reference, and is an identity for plain-string content.
+            # The journaled message keeps the full blocks.
+            #
+            # Reasoning blocks are deliberately not replayed in this phase.
+            # Replaying them is provider-specific (Anthropic needs its own
+            # signed thinking blocks and strips prior-turn thinking anyway;
+            # OpenAI Responses needs its own reasoning item ids), so a
+            # provider-aware replay path is Phase 2 work (#1223). Until then
+            # the history carries exactly what providers receive today.
+            if isinstance(message, AgentToolResultMessage) or \
```
Noravee, line 111: The one rule for the chat-history copy: tool results, and any AI message whose content is a block list, are appended as plain text. String content is appended as the same instance, unchanged. The second half of the condition is dormant until the answer-preservation PR starts journaling block-content AI messages; it lives here so the two PRs do not edit the same lines. Reasoning replay is deferred to #1223 on purpose, see the comment above.

neuro_san/internals/run_context/langchain/core/langchain_openai_function_tool.py
```diff
@@ -268,9 +270,15 @@ async def _arun(
...
         # any other object - which for a BaseMessage is its pydantic repr
         # ("content='...' additional_kwargs={} ..."), not the answer.
-        return the_answer.content
+        # The answer comes out of the sub-agent's chat history, whose copies
+        # are already text-projected (see OriginatingJournal.write_message),
+        # so this projection is an identity today. It stays as the guard for
+        # this boundary: a reasoning or data block inside a tool result is
+        # rejected by providers, and the blocks are already preserved in the
+        # sub-agent's own journal.
+        return ContentUtils.history_safe_text(the_answer)
```
Noravee, line 284: The answer comes out of the sub-agent's chat history, whose entries are already text after the rule in `OriginatingJournal.write_message`, so this is an identity today. It stays as the guard at the tool boundary: a reasoning or data block inside a tool result is rejected by providers.

requirements.txt
```diff
@@ -84,9 +84,10 @@ objsize>=0.8.0
 
 # Ceiling <0.2.0: 0.2.0 ships langchain-ai/langchain-mcp-adapters#379, which
 # changed ToolMessage.content from a concatenated text string to a list of
-# LangChain content blocks. JournalingCallbackHandler.on_tool_end then calls
-# str() on .content and would log the raw list structure instead of the tool's
-# text output.
+# LangChain content blocks. JournalingCallbackHandler.on_tool_end now journals
+# such block lists intact, so the original str()-repr concern is gone; the
```
Noravee, line 88: Comment-only change. The old rationale described the `str()` repr this PR removes. The ceiling itself stays until 0.2.x is verified end to end (sub-agent tool results, wire shape). Under the pinned 0.1.14 adapter, MCP images arrive in the ToolMessage artifact rather than in content, so today the block branch is reached by toolbox tools that return blocks.

tests/neuro_san/internals/run_context/langchain/journaling/test_journaling_callback_handler.py
```diff
@@ -122,3 +133,97 @@ async def test_on_tool_start_falls_back_to_placeholder_when_name_missing(self) -
...
+        A list that is not standard content blocks (here list-of-str) keeps the
+        str() form it has always been journaled with; the structure keeps the
+        raw output as before.
+        """
+        result, got_result = await self._run_langchain_tool(["a", "b"])
+        assert result.content == "['a', 'b']"
+        assert got_result.structure["tool_output"] == ["a", "b"]
+
+    @pytest.mark.asyncio
+    async def test_on_tool_end_single_text_block_list_is_kept_as_blocks(self) -> None:
```
Noravee, line 221: The one deliberate client-visible change. A tool returning `[{"type": "text", "text": "42"}]` was journaled as the Python repr of that list, and that repr reached the wire text. Now the block list is journaled and the wire text is `42`.

Noravee (commented):
**TL;DR** Block content at the tool boundaries: the journal keeps it, everything provider-facing stays text.

- `on_tool_end` (toolbox and MCP tools): a list of standard content blocks is journaled as that list, bytes turned into base64, through the now-public `ContentUtils.to_json_safe`. Any other output shape keeps `str(output)`.
- `OriginatingJournal.write_message`: the chat-history copy of any tool result, and of any AI message with block content, is a plain text `AIMessage` via `history_safe_text` (text kept, data blocks become `[image attachment: image/png]`). Identity for strings. Reasoning is deliberately not replayed in this phase; provider-aware replay is Phase 2 (#1223).
- `LangChainOpenAIFunctionTool._arun`: sub-agent answers reach the calling LLM as that same text. Identity today.
- One deliberate client-visible change: a tool returning a bare text-block list used to be journaled as its Python repr; now the wire text is the text itself.
- 12 new tests.

Independent of the other two PRs of this rung (#1321 and the answer-preservation PR that follows it).

neuro_san/message/utils/content_utils.py
d1donlydfink: Maybe worth drawing attention to the distinction/differences between similarly named methods in comments in both places.

d1donlydfink (commented):
Worth drawing distinctions and cross-referencing between ChatMessageConverter and ContentUtil's similarly named to_json_safe() methods in their respective comments so they are not mis-used in the future.

d1donlydfink (changes requested):
Generally good.

Worth drawing distinctions and cross-referencing between ChatMessageConverter and ContentUtil's similarly named to_json_safe() methods in their respective comments so they are not mis-used in the future.

neuro_san/message/utils/content_utils.py
Noravee: Done in 01fd3d07. Both docstrings now say which one to use and point at the other: this one is the lossless variant for content (bytes to base64, unknown values to their string form, keys to strings); `ChatMessageConverter.to_json_safe` is the lossy variant for outbound response dictionaries (non-serializable values, bytes included, become None; non-string keys are dropped). The converter's docstring also gained the `:param`/`:return` lines.

d1donlydfink (approved):

PR #1321, opened 2026-09-10
Extract find_ai_message from parse_chain_result (no behavior change)

Noravee:
Part of the "preserve native messages" rung of #1222, split into three PRs. This one is a pure refactor with no behavior change, independent of `tool-result-blocks`. The answer-preservation PR that follows builds on it.

## What changes

`RunContextRunnable.parse_chain_result` did two things in one method: locate the provider-native `AIMessage` inside the chain result, then flatten it to text. The lookup is now the static `find_ai_message`, which handles the same three shapes as before:

- a dict whose `messages` hold the chat history (the last `AIMessage` is the answer),
- a bare `AIMessage`,
- a legacy `AgentFinish`, unwrapped to its `return_values` first.

`parse_chain_result` calls it and keeps its exact behavior, including the `{"output": text}` fallback used when no `AIMessage` is present. That shape is produced inside `invoke_agent_chain` on the API-key error path and the output-parse workaround path.

The `AgentFinish` unwrap is now documented as legacy: it is the return type of langchain's `AgentExecutor`, which neuro-san stopped using in 5c2a0e61. The `create_agent` graph in use returns a state dict and never an `AgentFinish`, so the unwrap is kept for backwards compatibility only.

## Why

The follow-up PR needs the native `AIMessage` object, not just its flattened text, to journal reasoning blocks. Exposing the lookup on its own keeps that PR's diff to the new behavior.

## Tests

6 new tests pin `find_ai_message` on all three shapes plus the no-`AIMessage` cases, and `parse_chain_result` on the `output`-key fallback with and without `AgentFinish`. Those paths had no coverage before.


Noravee:
**TL;DR** Pure refactor, no behavior change. `parse_chain_result` used to both locate the provider-native `AIMessage` inside a chain result and flatten it. The lookup is now the static `find_ai_message`, so the follow-up PR can journal the native message itself.

- Same three shapes as before: a dict whose `messages` hold the history (last `AIMessage` wins), a bare `AIMessage`, and a legacy `AgentFinish` unwrapped to its `return_values`.
- The `{"output": text}` fallback is kept. That shape is built inside `invoke_agent_chain` on the API-key error path and the output-parse workaround path.
- `AgentFinish` is documented as legacy: it is the return type of langchain's `AgentExecutor`, which neuro-san dropped in 5c2a0e61; the pinned langchain and langgraph never produce it.
- 6 new tests pin the lookup and the fallback paths, which had no coverage. Full suite 677 passed; pylint 10.00/10, flake8 clean.

Independent of #1320. The answer-preservation PR is stacked on this one.

neuro_san/internals/run_context/langchain/core/run_context_runnable.py
```diff
@@ -325,6 +325,41 @@ async def journal_retry_reason(self, error: Exception, reason: str):
         if self.sensitive_logger.should_log():
             await self.journal.write_message(AgentFrameworkMessage(content=text))
 
+    @staticmethod
+    def find_ai_message(chain_result: Optional[Union[Dict[str, Any], AgentFinish, AIMessage]]) -> Optional[AIMessage]:
```
Noravee, line 329: Static and called with the raw chain result, so the follow-up PR can use it from `invoke_agent_chain` without re-plumbing. The order of checks is the old one: unwrap `AgentFinish`, then a bare `AIMessage`, then the last `AIMessage` in a dict's `messages`.

neuro_san/internals/run_context/langchain/core/run_context_runnable.py
```diff
@@ -325,6 +325,41 @@ async def journal_retry_reason(self, error: Exception, reason: str):
...
+
+        :param chain_result: The result from invoking the agent chain
+                        (see parse_chain_result() for the possible shapes)
+        :return: The last AIMessage in a dict result's "messages", the result
+                 itself when it is an AIMessage, or None when the result carries
+                 no AIMessage at all (exception, API-key error text, bare "output")
+        """
+        use_result: Any = chain_result
+        if isinstance(use_result, AgentFinish):
+            # AgentFinish is the return type of langchain's legacy AgentExecutor,
```
Noravee, line 341: Kept for backwards compatibility only. `create_agent` returns a `CompiledStateGraph` whose result is the state dict; the pinned langchain 1.3.14 and langgraph contain no reference to `AgentFinish`, and neuro-san stopped using `AgentExecutor` in 5c2a0e61.

neuro_san/internals/run_context/langchain/core/run_context_runnable.py
```diff
@@ -352,36 +387,20 @@ def parse_chain_result(self, chain_result: Optional[Union[Dict[str, Any], AgentF
...
-            elif isinstance(chain_result, AIMessage):
-                # Sometimes we get an AIMessage from a tool call.
-                ai_message = chain_result
-
+            ai_message: Optional[AIMessage] = self.find_ai_message(chain_result)
             if ai_message is not None:
                 # We generally want the content of any single AIMessage we found from above
                 output = ai_message.content
+            else:
+                # We didn't find an AIMessage, so look for a straight-up "output" key:
```
Noravee, line 395: This is where the `output` key built inside `invoke_agent_chain` is read: the API-key error path produces `{"output": message}` and the output-parse workaround produces `{"output": cleaned_text}`. Neither carries an `AIMessage`. The unwrap is repeated here so an `AgentFinish` holding only `output` gives the same result as before.

Noravee (commented):
**TL;DR** Pure refactor, no behavior change. `parse_chain_result` used to both locate the provider-native `AIMessage` inside a chain result and flatten it. The lookup is now the static `find_ai_message`, so the follow-up PR can journal the native message itself.

- Same three shapes as before: a dict whose `messages` hold the history (last `AIMessage` wins), a bare `AIMessage`, and a legacy `AgentFinish` unwrapped to its `return_values`.
- The `{"output": text}` fallback is kept. That shape is built inside `invoke_agent_chain` on the API-key error path and the output-parse workaround path.
- `AgentFinish` is documented as legacy: it is the return type of langchain's `AgentExecutor`, which neuro-san dropped in 5c2a0e61; the pinned langchain and langgraph never produce it.
- 6 new tests pin the lookup and the fallback paths, which had no coverage. Full suite 677 passed; pylint 10.00/10, flake8 clean.

Independent of #1320. The answer-preservation PR is stacked on this one.

tests/neuro_san/internals/run_context/langchain/core/test_run_context_runnable.py
```diff
@@ -92,6 +100,62 @@ def test_parse_chain_result_dict_messages_path_flattens_last_ai_message(self):
...
+
+    def test_find_ai_message_unwraps_agent_finish(self) -> None:
+        """
+        An AgentFinish result is unwrapped to its return_values first.
+        """
+        native: AIMessage = ContentFixtures.anthropic_thinking_first()
+        finish: AgentFinish = AgentFinish(return_values={"messages": [HumanMessage(content="q"), native]}, log="")
+        assert RunContextRunnable.find_ai_message(finish) is native
+
+    def test_find_ai_message_returns_none_without_ai_message(self) -> None:
```
Noravee, line 132: Covers the production `{"output": ...}` shape, a `messages` list without an `AIMessage`, and a `None` result. None of these paths had a test before this PR.

d1donlydfink (dismissed):

neuro_san/internals/run_context/langchain/core/run_context_runnable.py
```diff
@@ -325,6 +325,41 @@ async def journal_retry_reason(self, error: Exception, reason: str):
...
+            # graph used today returns a state dict and never an AgentFinish, so
+            # this unwrap is kept for backwards compatibility only. Its
+            # return_values carry the output in dict form.
+            use_result = use_result.return_values
+
+        if isinstance(use_result, AIMessage):
+            # Sometimes we get an AIMessage from a tool call.
+            return use_result
+
+        if isinstance(use_result, Dict):
```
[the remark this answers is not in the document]
Noravee: Done in dcd4dcc3: both runtime checks now use the built-in `dict`. For the record, the pattern was pre-existing (main had `isinstance(chain_result, Dict)` at the same spot) and works on 3.12 through 3.14, but `dict` is the right spelling for a runtime check; `typing.Dict` stays for the annotations.

d1donlydfink (approved):

PR #1323, opened 2026-09-10
More test tweaking.

d1donlydfink:
* Get rid of some warnings in test running that have accumulated
* Take @vince-leaf 's suggestion to have runtime warnings be errors for tests
* Take @vince-leaf 's suggestion to remove pytest.mark.asyncio from IsolatedAsyncioTestCase
* Ignore a warning about os.fork() from test_hocon_parse_lock.py

PR #1319, opened 2026-09-10
Derive from TestCase. Rename so test classes/files match. Break tests classes out to their own files

d1donlydfink:
A first bunch of transformations on the relative unintelligibility of the test code:

* Break out Test class definitions to their own files when there were multiple in the same file
* Rename files to match their test class
* Make a bunch of classes derive from UnitTest where that is possible.

@vince-leaf and @Noravee : While your vibing of test cases for coverage is commendable, please be sure your code-assist settings are following the rules even for tests:
* One class per file (not entirely addressed here, but at least there is now one _test_ class per file)
* Filename matches test class name

Worth noting that @andreidenissov-cog is generally not an offender.
(My biggest offense is actually not writing tests at all! Worse than all of y'all!)

FWIW: I went here on this PR for 2 reasons:
1) I realized from other external contributors PRs that our tests were setting a bad example.  So we need to raise the bar and this is a start.
2) It's possible we might need to simplify our tests for translatability.

PR #1326, opened 2026-09-10
Fail tests that return an un-awaited coroutine

### LINKED ISSUE #1325: Fail tests that return an un-awaited coroutine
error::RuntimeWarning (added in #1323) doesn't fail a test whose coroutine is never awaited because the warning is emitted during garbage collection, outside the test. Verified on main: changing TestReport from IsolatedAsyncioTestCase to TestCase still gives 5 passed.

The 5 async tests in tests/neuro_san/internals/run_context/langchain/token_counting/test_report.py (TestReport):

test_report_counts_each_agents_own_calls_exactly_once
test_report_flags_unattributed_tokens
test_report_late_external_does_not_restamp_request_latency
test_report_network_message_gating
test_report_separates_main_network_from_total

vince-leaf:
This fix is a follow-up to #1323.

error::RuntimeWarning in pytest.ini does not fail a test whose coroutine is never awaited (the #1319 failure mode). The "coroutine ... was never awaited" RuntimeWarning is emitted from the coroutine's finalizer during garbage collection, after the test has already been marked as passed, so as an error, it only becomes a PytestUnraisableExceptionWarning.

This replaces that line with a filter on the DeprecationWarning unittest raises synchronously when a test method returns a non-None value (which is what happens when an async def test sits on a plain TestCase):

`error:It is deprecated to return a value that is not None from a test case:DeprecationWarning`

I have verified:
1. Full unit suite: 708 passed, 1 skipped — same as main.
2. The 12 IsolatedAsyncioTestCase files from #1319: 83 passed, identical to main.
3. Sanity check: switching TestReport from IsolatedAsyncioTestCase to TestCase now gives 5 failures (was 5 passed with error::RuntimeWarning).

Confirmed from my local branch:

##### 1. Bug re-created (TestReport on plain TestCase), pytest.ini as on main:
15:    error::RuntimeWarning
5 passed, 10 warnings in 0.11s
 
##### 2. Same broken file, pytest.ini with suggested line:
15:    error:It is deprecated to return a value that is not None from a test case:DeprecationWarning
FAILED ...test_report.py::TestReport::test_report_counts_each_agents_own_calls_exactly_once
FAILED ...test_report.py::TestReport::test_report_flags_unattributed_tokens
FAILED ...test_report.py::TestReport::test_report_late_external_does_not_restamp_request_latency
FAILED ...test_report.py::TestReport::test_report_network_message_gating
FAILED ...test_report.py::TestReport::test_report_separates_main_network_from_total
5 failed, 5 warnings in 0.22s
 
##### 3. File restored (IsolatedAsyncioTestCase), suggested line kept:
5 passed in 0.15s
 
##### 4. Full unit suite with suggested line:
708 passed, 1 skipped, 12 subtests passed in 18.34s
 
##### restored: 0 files modified
So: current line lets the bug through (1); suggested line catches it (2), doesn't touch correct tests (3), and the whole suite stays green (4). Working tree left clean.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -12,6 +12,7 @@ markers =
     smoke_non_default_llm_provider_needs_server: Tests that use a non-default llm provider and need server.
     ollama: Tests that specifically use ollama as the llm provider
 filterwarnings =
+    # Fail async tests on a plain TestCase (method returns an un-awaited coroutine)
     error:It is deprecated to return a value that is not None from a test case:DeprecationWarning
     # Ignore warnings about protobuf 4's PyType_Spec metaclass usage
     ignore:Type google\._upb\._message.* uses PyType_Spec with a metaclass that has custom tp_new:DeprecationWarning
--- a/pytest.ini
+++ b/pytest.ini
@@ -12,7 +12,7 @@ markers =
     smoke_non_default_llm_provider_needs_server: Tests that use a non-default llm provider and need server.
     ollama: Tests that specifically use ollama as the llm provider
 filterwarnings =
-    error::RuntimeWarning
+    error:It is deprecated to return a value that is not None from a test case:DeprecationWarning
     # Ignore warnings about protobuf 4's PyType_Spec metaclass usage
     ignore:Type google\._upb\._message.* uses PyType_Spec with a metaclass that has custom tp_new:DeprecationWarning
     # Ignore warnings about fork from test_hocon_parse_lock.py


pytest.ini
```diff
@@ -12,7 +12,7 @@ markers =
     smoke_non_default_llm_provider_needs_server: Tests that use a non-default llm provider and need server.
     ollama: Tests that specifically use ollama as the llm provider
 filterwarnings =
-    error::RuntimeWarning
+    error:It is deprecated to return a value that is not None from a test case:DeprecationWarning
```
vince-leaf, line 16: **Why error::RuntimeWarning didn't catch it:** when an async def test runs on a plain TestCase, calling it just returns a coroutine object; the body never runs, and the test is marked as passed. The "coroutine ... was never awaited" RuntimeWarning is only emitted later, when the garbage collector destroys that coroutine — after the test has already finished. Turning it into an error at that point can't change the result; pytest just logs it as an unraisable-exception warning.

**Why the new line catches it**: unittest (3.11+) checks each test method's return value while the test is still running. A coroutine is a non-None return, so unittest emits a DeprecationWarning: It is deprecated to return None from a test case inside the test. Making that specific warning an error raises an error inside the test, so the test fails. Correct tests return None and never trigger it.

andreidenissov-cog (approved):
Looks OK, AFAICT.

PR #1327, opened 2026-09-11
Journal the provider-native answer when it carries block content

Noravee:
Last of the three PRs for the "preserve native messages" rung of #1222, after #1320 (tool boundaries, merged) and #1321 (`find_ai_message` refactor, merged).

## What changes

`RunContextRunnable.invoke_agent_chain` journaled a fresh `AIMessage(text)` for every answer, so Anthropic thinking and OpenAI Responses reasoning blocks were lost before they reached the journal. The new `build_return_message` decides between the model's own message and that plain copy. It defaults to the plain copy and keeps the model's message only when four checks pass, in this order:

1. `find_ai_message` found a native `AIMessage`. Otherwise (exception, API-key error, bare `{"output": ...}`) there is nothing to preserve.
2. The ErrorDetector did not rewrite the text. If it did, the client gets the formatted error, not the model's content.
3. After `ContentUtils.normalize_message` (standard v1 blocks plus the `output_version` stamp) the content is still a block list. Text-only content, including a single text block, list-of-str, and text plus tool_use, keeps today's exact `AIMessage(text)` shape with no provider ids or metadata.
4. The normalized blocks still flatten to the parsed text. Langchain's translators skip items they do not recognize, so this fails closed to the plain copy if any text would be lost.

An `AIMessageChunk` is converted to a complete `AIMessage` before normalizing, so a custom middleware returning a chunk cannot give the journaled answer an unknown wire type.

## What does not change

- The wire: the converter still emits only the flattened `text` in this phase, and `parse_chain_result` produces that text exactly as before.
- The chat history: the copy appended for a block-content answer is plain text under the rule merged in #1320, so what is handed to providers, fallbacks included, is unchanged.
- Text-only traffic at every layer: the journaled object for a plain-string answer is byte-for-byte what it was.

## Tests

10 new invoke-level tests: block answer preserved with `output_version` and `usage_metadata`; text-only shape kept; error rewrite wins; tool_use blocks excluded from content; tool_use turn and list-of-str collapse to text; bare output dict; AgentFinish; mixed-content fallback; chunk converted. The test class carries a `too-many-public-methods` disable with its reason, as four other test modules do. Full suite: 703 passed, 1 skipped (main alone: 693). pylint 10.00/10, flake8 clean.

### DIFF
--- a/neuro_san/internals/run_context/langchain/core/run_context_runnable.py
+++ b/neuro_san/internals/run_context/langchain/core/run_context_runnable.py
@@ -30,7 +30,9 @@ from pydantic import ConfigDict
 from langchain_core.agents import AgentFinish
 from langchain_core.callbacks.base import BaseCallbackHandler
 from langchain_core.messages.ai import AIMessage
+from langchain_core.messages.ai import AIMessageChunk
 from langchain_core.messages.base import BaseMessage
+from langchain_core.messages.utils import message_chunk_to_message
 from langchain_core.runnables.base import Runnable
 from langchain_core.runnables.config import ensure_config
 from langchain_core.runnables.config import merge_configs
@@ -298,7 +300,7 @@ class RunContextRunnable(NeuroSanRunnable):
             self.sensitive_logger.error("%s", backtrace)
 
         output: str = self.parse_chain_result(chain_result, exception)
-        return_message: BaseMessage = AIMessage(output)
+        return_message: BaseMessage = self.build_return_message(chain_result, output)
 
         # Chat history is updated in write_message
         await self.journal.write_message(return_message)
@@ -416,3 +418,62 @@ class RunContextRunnable(NeuroSanRunnable):
         # See if we had some kind of error and format accordingly, if asked for.
         output = self.error_detector.handle_error(output)
         return output
+
+    def build_return_message(self, chain_result: Optional[Union[Dict[str, Any], AgentFinish, AIMessage]],
+                             output: str) -> BaseMessage:
+        """
+        Build the AIMessage that is journaled as this agent's answer for the turn.
+
+        Plain-string answers, error paths, and anything the ErrorDetector
+        rewrote keep the exact shape they have always had: a fresh AIMessage
+        carrying only the parsed text. When the provider-native message
+        carries block content (Anthropic thinking, OpenAI Responses reasoning,
+        ...), that message is preserved instead, with its content normalized
+        to standard v1 blocks (see ContentUtils.normalize_message) so the
+        blocks and its usage_metadata reach the journal. The wire is the
+        same either way: the converter projects block content to the same
+        text and emits nothing else for it in this phase.
+
+        The chat history does not depend on which shape is chosen:
+        OriginatingJournal.write_message appends a text-only copy of any
+        block-content AI message, so what is handed to a provider (or a
+        fallback provider) stays provider-agnostic.
+
+        :param chain_result: The result from invoking the agent chain
+        :param output: The parsed, error-checked text from parse_chain_result()
+        :return: The BaseMessage to journal as the answer
+        """
+        ai_message: Optional[AIMessage] = self.find_ai_message(chain_result)
+        if ai_message is None:
+            # Exception, API-key error text or a bare "output": nothing native to preserve.
+            return AIMessage(output)
+
+        if output != ContentUtils.flatten_to_text(ai_message):
+            # The ErrorDetector rewrote the text. The client-visible answer is
+            # the formatted error, not the native content.
+            return AIMessage(output)
+
+        if isinstance(ai_message, AIMessageChunk):
+            # normalize_message is for complete messages only: chunk-only
+            # shapes are not content. Chains return complete messages today,
+            # but a custom middleware could hand back a chunk, and preserving
+            # its class would give the journaled answer an unknown wire type.
+            ai_message = message_chunk_to_message(ai_message)
+
+        preserved: BaseMessage = ContentUtils.normalize_message(ai_message)
+        if isinstance(preserved.content, str):
+            # Trivial (text-only) content: keep the exact message shape that has
+            # always been journaled for it. Preserving the native object would
+            # change nothing a client can see and would only drag provider
+            # bookkeeping (ids, metadata) along with it.
+            return AIMessage(output)
+
+        if ContentUtils.flatten_to_text(preserved) != output:
+            # Fail closed. The provider translators behind normalize_message
+            # skip content items they do not recognize (e.g. a bare string
+            # inside a block list), which can drop text. The client-visible
+            # answer must stay the parsed text, so blocks are preserved only
+            # while they still project to it.
+            return AIMessage(output)
+
+        return preserved
--- a/tests/neuro_san/internals/run_context/langchain/core/test_run_context_runnable.py
+++ b/tests/neuro_san/internals/run_context/langchain/core/test_run_context_runnable.py
@@ -386,7 +386,6 @@ class TestRunContextRunnable(IsolatedAsyncioTestCase):  # pylint: disable=too-ma
             block_types.append(block["type"])
         return block_types
 
-    @pytest.mark.asyncio
     async def test_invoke_agent_chain_preserves_block_content_answer(self) -> None:
         """
         A thinking-first native answer is journaled as the native message with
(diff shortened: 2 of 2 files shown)


neuro_san/internals/run_context/langchain/core/run_context_runnable.py
```diff
@@ -298,7 +300,7 @@ async def invoke_agent_chain(self, inputs: Dict[str, Any], runnable_config: Dict
             self.sensitive_logger.error("%s", backtrace)
 
         output: str = self.parse_chain_result(chain_result, exception)
-        return_message: BaseMessage = AIMessage(output)
+        return_message: BaseMessage = self.build_return_message(chain_result, output)
```
Noravee, line 303: The only call-site change. `output` is still the text `parse_chain_result` produced and ran through the ErrorDetector, so what the client sees is decided before this line; `build_return_message` only decides which object carries it into the journal.

neuro_san/internals/run_context/langchain/core/run_context_runnable.py
```diff
@@ -416,3 +418,62 @@ def parse_chain_result(self, chain_result: Optional[Union[Dict[str, Any], AgentF
         # See if we had some kind of error and format accordingly, if asked for.
         output = self.error_detector.handle_error(output)
         return output
+
+    def build_return_message(self, chain_result: Optional[Union[Dict[str, Any], AgentFinish, AIMessage]],
```
Noravee, line 422: Read this as a funnel: five exits return the plain `AIMessage(output)` that was always journaled, one exit returns the model's message. Anything unusual falls back to the old behavior.

neuro_san/internals/run_context/langchain/core/run_context_runnable.py
```diff
@@ -416,3 +418,62 @@ def parse_chain_result(self, chain_result: Optional[Union[Dict[str, Any], AgentF
...
+        :param chain_result: The result from invoking the agent chain
+        :param output: The parsed, error-checked text from parse_chain_result()
+        :return: The BaseMessage to journal as the answer
+        """
+        ai_message: Optional[AIMessage] = self.find_ai_message(chain_result)
+        if ai_message is None:
+            # Exception, API-key error text or a bare "output": nothing native to preserve.
+            return AIMessage(output)
+
+        if output != ContentUtils.flatten_to_text(ai_message):
```
Noravee, line 451: The ErrorDetector returns the same text when there is no error, so a difference here means it rewrote the answer into a formatted error. The client-visible answer is that error, so the model's content must not be journaled.

neuro_san/internals/run_context/langchain/core/run_context_runnable.py
```diff
@@ -416,3 +418,62 @@ def parse_chain_result(self, chain_result: Optional[Union[Dict[str, Any], AgentF
...
+            # The ErrorDetector rewrote the text. The client-visible answer is
+            # the formatted error, not the native content.
+            return AIMessage(output)
+
+        if isinstance(ai_message, AIMessageChunk):
```
Noravee, line 456: `normalize_message` is documented for complete messages only. The shipped chains never return a chunk, but a custom middleware could, and `ChatMessageType.from_message` does an exact type lookup, so a preserved chunk would get the unknown wire type. Converting first closes that.

neuro_san/internals/run_context/langchain/core/run_context_runnable.py
```diff
@@ -416,3 +418,62 @@ def parse_chain_result(self, chain_result: Optional[Union[Dict[str, Any], AgentF
...
+            # normalize_message is for complete messages only: chunk-only
+            # shapes are not content. Chains return complete messages today,
+            # but a custom middleware could hand back a chunk, and preserving
+            # its class would give the journaled answer an unknown wire type.
+            ai_message = message_chunk_to_message(ai_message)
+
+        preserved: BaseMessage = ContentUtils.normalize_message(ai_message)
+        if isinstance(preserved.content, str):
```
Noravee, line 464: This is the text-only identity. A plain string, a single text block, list-of-str, and text plus tool_use all normalize to a string, and they keep exactly today's shape: fresh `AIMessage(output)`, no provider id, no `response_metadata`, no `usage_metadata`. Two or more plain text blocks are not trivial by the `is_trivial` rule from the converter PR, so they are preserved as blocks; the wire text is unchanged either way.

neuro_san/internals/run_context/langchain/core/run_context_runnable.py
```diff
@@ -416,3 +418,62 @@ def parse_chain_result(self, chain_result: Optional[Union[Dict[str, Any], AgentF
...
+            # Trivial (text-only) content: keep the exact message shape that has
+            # always been journaled for it. Preserving the native object would
+            # change nothing a client can see and would only drag provider
+            # bookkeeping (ids, metadata) along with it.
+            return AIMessage(output)
+
+        if ContentUtils.flatten_to_text(preserved) != output:
```
Noravee, line 471: Fail-closed guard. langchain's anthropic and openai translators skip non-dict items, so legal content like `["hello ", {thinking}, {text: "world"}]` would normalize to blocks that flatten to `world` while the client got `hello world`. Preserving those blocks would also defeat the journal's dupe suppression, which compares flattened text. So blocks are kept only while they still say what `output` says.

neuro_san/internals/run_context/langchain/core/run_context_runnable.py
```diff
@@ -416,3 +418,62 @@ def parse_chain_result(self, chain_result: Optional[Union[Dict[str, Any], AgentF
...
+            # Fail closed. The provider translators behind normalize_message
+            # skip content items they do not recognize (e.g. a bare string
+            # inside a block list), which can drop text. The client-visible
+            # answer must stay the parsed text, so blocks are preserved only
+            # while they still project to it.
+            return AIMessage(output)
+
+        return preserved
```
Noravee, line 479: Only here does the native message reach the journal: standard v1 blocks with the `output_version` stamp, `usage_metadata` intact. Its chat-history copy is still text because of the rule in `OriginatingJournal.write_message`, and the converter still emits only the flattened text on the wire.

Noravee (commented):
**TL;DR** The journal now keeps the model's own answer when it carries block content; everything else keeps today's plain `AIMessage(text)`.

- `invoke_agent_chain` calls the new `build_return_message` instead of always building `AIMessage(output)`. That method defaults to the plain copy and returns the model's message only when four checks pass in order: a native `AIMessage` exists, the ErrorDetector did not rewrite the text, the normalized content is still a block list (text-only shapes collapse back to the plain copy, with no provider ids or metadata), and the normalized blocks still flatten to the parsed text.
- The preserved message is normalized to standard v1 blocks with the `output_version` stamp, so its reasoning blocks and `usage_metadata` reach the journal. An `AIMessageChunk` is converted to a complete message first.
- Nothing changes on the wire (the converter still emits only `text` in this phase) or in the chat history (the copy appended for a block answer is text, per #1320).
- 10 new invoke-level tests, one per branch of the decision.

Closes the "preserve native messages" rung of #1222 after #1320 and #1321.

neuro_san/internals/run_context/langchain/core/run_context_runnable.py
d1donlydfink: Very thorough!

neuro_san/internals/run_context/langchain/core/run_context_runnable.py
d1donlydfink: Nit: Perhaps leave a comment in the code as to what kind of native messages you expect to get to this point.

tests/neuro_san/internals/run_context/langchain/core/test_run_context_runnable.py
```diff
@@ -332,3 +344,216 @@ async def test_invoke_agent_chain_does_not_log_stale_backtrace(self):
...
+            ],
+            response_metadata={"model_provider": "anthropic"},
+        )
+        runnable, written = self._make_invoking_runnable({"messages": [native]})
+        await runnable.invoke_agent_chain(inputs={}, runnable_config={}, max_attempts=1)
+
+        final: BaseMessage = written[-1]
+        assert type(final) is AIMessage  # pylint: disable=unidiomatic-typecheck
+        assert self._block_types(final) == ["reasoning", "text"]
+        assert ContentUtils.flatten_to_text(final) == "the answer"
```
d1donlydfink, line 559: Please strongly consider using the TestCase assert methods as opposed to bare assert in the future whenever possible. I believe TestCase assert methods actually give us slightly more information upon failure.

Our test style is all over the place with these vibed tests, we need to start reining it in.

d1donlydfink (approved):
Very thorough.

2 nits:

1) Add a code comment as to the kinds of blocks we expect to get through the gauntlet in build_return_message()

2) Please strongly consider using TestCase assert methods which I believe give slightly more information on failure.  This will help in solidifying test style for external contributors, which is currently all over the map.

Issue #487, opened 2025-10-20
Be sure framework level errors are logged


d1donlydfink:
Some of these errors are retried before they fail, but still these need to be logged so administrators know there might be an issue.

In the end, though, error messages may or may not come back to the user and/or over the message stream to the client. These are always agent implementation decisions.

Issue #486, opened 2025-10-20
Add reasoning fields for anthropic, bedrock, and gemini models in default llm info file


Noravee:
Currently, only OpenAI and Ollama models have fields for changing reasoning levels.

PR #1332, opened 2026-09-14
Make the Responses API the default OpenAI endpoint and forward store and include

### LINKED ISSUE #1335: Make the Responses API the default OpenAI endpoint and expose store and include in llm_config
## Problem

OpenAI's newest models only accept function tools together with reasoning on the Responses API. gpt-5.6-* reject the combination on Chat Completions ("Function tools with reasoning_effort are not supported ... use /v1/responses or set reasoning_effort to 'none'"), and gpt-6-astra has no Chat Completions function calling at all and also rejects `reasoning_effort: "none"`. With the `openai` class default `use_responses_api: null`, every such model needs a per-hocon `"use_responses_api": true`, and a `model_name`-only `llm_config` fails for any tool-calling agent.

Two related gaps in `OpenAILlmPolicy`:

- `store` is not forwarded, so every Responses API request uses OpenAI's stored-by-default behavior (responses kept server-side for 30 days) with no way to opt out from `llm_config`.
- `include` is not forwarded, so the legacy `reasoning.encrypted_content` request cannot be made even when a user wants it.

## Proposal

- `default_llm_info.hocon`, `classes.openai.args`: `use_responses_api: true`, `store: false`, `include: null`, with comments covering the three values of each and how to restore Chat Completions per agent or server-wide through an `AGENT_LLM_INFO_FILE` override.
- `OpenAILlmPolicy.create_llm()` forwards `store` and `include` raw (a null must reach langchain as None so the field is omitted).
- `include` must stay `null` by default: a non-null value is a Responses routing trigger in langchain-openai and the Chat Completions SDK call rejects the keyword.
- Update the gpt-6-astra / gpt-5.6-* entry comments, the docs section "OpenAI Reasoning and Responses API Parameters", and add an upgrading note (endpoint change, no server-side storage, `stop` dropped, Chat-Completions-only args rejected, gateways need `false`).
- Tests for the forwarding in `TestOpenAILlmPolicy`.

## Evidence

Live runs against gpt-5.2-2025-12-11 and gpt-6-astra with `use_responses_api: true`, `store: false` and no `include` showed reasoning items carrying `encrypted_content` unprompted and the tool-calling loop replaying them correctly, matching the OpenAI reasoning guide ("include an `encrypted_content` property by default"; the legacy `include` value "is accepted but not required"). Offline: langchain-openai 1.6.0 treats `include` as a routing trigger (`_use_responses_api`) and omits `store`/`include` from the payload when None.

## Out of scope

Azure: `azure-openai` inherits the class defaults through `extends` but `AzureLlmPolicy` does not forward them, so Azure stays on Chat Completions (#1307).

Related: #725, #1222.

Noravee:
## What

Core change (three files):

- `default_llm_info.hocon`, `classes.openai.args`: `use_responses_api` defaults to `true`, and two new args are declared, `store: false` and `include: null`, each with a comment explaining the three values and how to restore Chat Completions per agent (`"use_responses_api": false` in `llm_config`) or server-wide (a sparse `AGENT_LLM_INFO_FILE` override). The gpt-6-astra and gpt-5.6-* entry comments now say tool calling works with the class defaults. `stop` is marked Chat Completions only. A note on `azure-openai` records that the inherited values are not forwarded by `AzureLlmPolicy` and that Responses-only model names still auto-route there (#1307).
- `OpenAILlmPolicy.create_llm()` forwards `store` and `include` raw, next to `use_responses_api`, with the rationale in comments.
- `requirements.txt`: `langchain-openai>=1.4.0` and `langchain-core>=1.5.0`, with the reasons as comments. 1.4.0 is the first langchain-openai release that drops `stop` on the Responses path and has the store-aware reasoning replay the `store: false` default relies on; 1.4.0 itself requires langchain-core 1.5.0. After merging `main` this PR also lifts the temporary `langchain-openai<1.6.3` ceiling that #1377 added: that cap only protected the previous `null` default against 1.6.3's changed endpoint inference (gpt-6* with tools and gpt-5.6-sol auto-routing, and the inferred endpoint being written onto the instance), and with an explicit `true` default the explicit value wins. The `openai>=2.45.0,<3` pin from #1377 stays until the httpx2 client migration (#1379).

Documentation and tests of the core change:

- Documentation: the provider-specific llm_config docs (the rewritten OpenAI section with `store`, `include`, OpenAI-compatible gateways such as LiteLLM, and an "Upgrading existing OpenAI configurations" note) are consolidated in #1334, which moves them from `docs/agent_hocon_reference.md` into a new "Provider-Specific Arguments" section of `docs/llm_info_hocon_reference.md`. This PR no longer touches the agent reference; the five `default_llm_info.hocon` comment URLs point at the new location.
- `TestOpenAILlmPolicy` gains three store/include tests and is converted to the `unittest.TestCase` conventions (no pytest fixtures, `self.assert*` only, `@override` on `setUp`/`tearDown`).

Bundled OpenAI-compatible test servers (a direct consequence of the flip):

- Both `tests/mock_llm_server` and `tests/record_playback_llm_server` register only `/v1/chat/completions`. With the new default, a model-name-only agent such as `math_guy` posts to `/v1/responses` and gets a 404 from them.
- New `tests/mock_llm_server/llm_info_chat_completions.hocon`: a sparse llm info overlay that pins the `openai` class to Chat Completions. Point `AGENT_LLM_INFO_FILE` at it when starting a neuro-san server against either test server.
- `tests/load_tests/load_test_mock_llm_service.py`: `--auto-start` passes the overlay to the server it launches; the manual-start check warns when the running server has no `AGENT_LLM_INFO_FILE` (an empty value counts as unset, because the shipped Dockerfiles define `AGENT_LLM_INFO_FILE=""`); the docstring and error recipes include the export. Without the overlay every request 404s, yet the load test still reports success, because the CLI exits 0 with the exception text as the answer. The wiring is what keeps the numbers meaningful.
- README and module docstring of both servers: the example `llm_config` adds `use_responses_api = false`, and the "single `openai_api_base` change" wording is corrected.
- `RequestCanonicalizer.VOLATILE_BODY_KEYS` gains `store`: langchain-openai sends `store` on Chat Completions bodies too, so cassettes recorded before this default would stop matching. `test_request_canonicalizer.py` gets a regression test for it and is converted to the `unittest.TestCase` conventions in the process.

## Why

OpenAI's newest models only accept function tools together with reasoning on the Responses API, and gpt-6-astra has no Chat Completions function calling at all and rejects `reasoning_effort: "none"`. With the previous `null` default every such model needed a per-hocon `use_responses_api: true`. Flipping the class default makes a `model_name`-only `llm_config` work.

`store` defaults to `false` so requests stay stateless. Live runs against gpt-5.2-2025-12-11 and gpt-6-astra confirmed that reasoning items carry `encrypted_content` even with `store: false` and no `include`, and that the tool-calling loop replays them and answers correctly, which matches the OpenAI reasoning guide ("include an `encrypted_content` property by default"). `include` stays `null` because a non-null value is a Responses routing trigger in langchain-openai and the Chat Completions SDK call rejects the keyword.

## Behavior changes for existing OpenAI llm_configs

- Requests move from `/chat/completions` to `/responses`. OpenAI-compatible gateways without a `/responses` route need `"use_responses_api": false`, per agent or server-wide. The bundled mock and record/playback test servers are such gateways; the new overlay file covers them.
- Responses are no longer stored server-side (`store: false`) any more.  This is in alignment with our "Security by Default" ethos.
- `presence_penalty`, `frequency_penalty`, `seed`, `logprobs` and `logit_bias` are rejected client-side unless `use_responses_api` is `false`; `stop` is silently dropped on the Responses API and only applies with `use_responses_api: false`.
- Dependency floors rise to langchain-openai 1.4.0 and langchain-core 1.5.0, and the temporary `langchain-openai<1.6.3` ceiling from #1377 is removed (a fresh install now resolves to 1.6.3). The `openai<3` pin is unchanged.
- Azure OpenAI: `AzureLlmPolicy` does not forward these keys, so Azure requests keep following langchain's own inference (#1307).

## Verification

Offline: `pytest tests/neuro_san/internals/run_context/langchain/llms tests/neuro_san/internals/run_context/langchain/core tests/record_playback_llm_server tests/load_tests/unit` passes; the new policy tests fail with the two forwarding lines removed, and the canonicalizer regression test fails with `store` removed from `VOLATILE_BODY_KEYS`. flake8, pylint (`.pylintrc`) and pymarkdown are clean. A stubbed end-to-end run of math_guy, hello_world and copy_cat_middleware against a fake `/v1/responses` with the flipped default completed normally with correct answers, tool execution and token accounting. The load test was run against the bundled mock on localhost both without the overlay (every request a 404 on `/v1/responses`, reported as success) and with it (real Chat Completions round trips). The live gpt-5.2 and gpt-6-astra runs above were done separately with real keys. CI is green on Python 3.12, 3.13 and 3.14. After the merge of `main` (post-#1377) the same suites pass with the environment's langchain-openai 1.6.0 and with a scratch install of langchain-openai 1.6.3, and a dry-run resolve of `requirements.txt` picks langchain-openai 1.6.3, langchain-core 1.6.4 and openai 2.54.0.

Fixes #1335

## Related

#725 (the `n=1` fix that made any Responses route possible), #1222 (Phase 1 arg plumbing), #1307 (Azure), #1337 (factory-level test of the resolved defaults, deliberately separate). The documentation for this PR and for #1333 lives in #1334, which documents the new defaults and must merge after both. #1377 (temporary langchain-openai ceiling and openai pin on `main`; the ceiling is lifted here) and #1379 (httpx2 client migration that lifts the openai pin). Follow-ups from review: #1362 (LLM API errors come back as a normal answer, pre-existing) and cognizant-ai-lab/neuro-san-studio#1432 (raise the studio langchain-openai floor when it bumps neuro-san).

## Reviewer guide

The TL;DR comment on this PR summarizes the change in a few lines. Inline comments mark the parts worth a close look: the three class defaults in the hocon, the raw forwarding in the policy, the dependency floors, the overlay and load-test wiring, and the canonicalizer key. Everything else follows from those.


docs/agent_hocon_reference.md
```diff
@@ -385,9 +418,19 @@ differences between the two endpoints.
 fallbacks configured, a misconfigured primary model silently fails over instead of reporting the problem, so
 verify a new llm_config without fallbacks first.
 
-**Azure OpenAI.** The `azure-openai` class does not support the Responses API path yet (tracked in
-[#1307](https://github.com/cognizant-ai-lab/neuro-san/issues/1307)), so leave `use_responses_api` unset for
-Azure deployments.
+**Azure OpenAI.** The `azure-openai` class extends `openai` and therefore inherits the `use_responses_api`,
+`store` and `include` defaults, but `AzureLlmPolicy` does not forward any of them to `AzureChatOpenAI`, so Azure
+deployments stay on Chat Completions regardless of what those keys say (tracked in
+[#1307](https://github.com/cognizant-ai-lab/neuro-san/issues/1307)).
```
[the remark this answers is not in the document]
Noravee: Agreed. Not forwarding the switch leaves `AzureChatOpenAI.use_responses_api` at None, and `azure-gpt-5.4-pro` resolves to a `gpt-5.4-pro` snapshot that langchain-openai auto-routes. Reworded in 0eeabcbd: the inherited keys have no effect on Azure, Azure follows langchain's inference (Chat Completions for most models, Responses for Responses-only names, which `AzureLlmPolicy` does not support yet), and #1307 covers pinning the endpoint.

neuro_san/internals/run_context/langchain/llms/default_llm_info.hocon
```diff
@@ -1512,23 +1516,47 @@
...
+                "include": null,
 
                 # If you really need more parameters, you will likely have to create your own LlmFactory.
             }
         },
         "azure-openai": {
             "extends": "openai",
+            # The use_responses_api, store and include values inherited from "openai" are not forwarded by
+            # AzureLlmPolicy, so Azure stays on Chat Completions regardless of what they say. Tracked in
+            # https://github.com/cognizant-ai-lab/neuro-san/issues/1307
```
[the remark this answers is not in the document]
Noravee: Reworded in 0eeabcbd the same way as the docs paragraph: the inherited values are ignored rather than forcing Chat Completions, and Responses-only model names such as the gpt-5.4-pro snapshot behind azure-gpt-5.4-pro still auto-route, which is unsupported until #1307.

Noravee:
Two of the review's suppressed comments, addressed here since they had no inline thread:

- **`stop` on langchain-openai before 1.4.** Correct that earlier releases still send it. Commit 0eeabcbd qualifies the upgrading note: 1.4 and later drop the parameter, earlier releases send it and fail client-side, so remove `stop` or set `use_responses_api` to `false`. The dependency floor stays where it is; raising it is a separate decision.
- **Factory-level coverage of the new hocon defaults.** A `DefaultLlmFactory`-level test module that loaded the shipped hocon and asserted `use_responses_api`/`store`/`include` on the resolved config and on the built model was written and then dropped at the maintainer's request: the class defaults are configuration data, and the repo's test convention is one test module per class. `TestOpenAILlmPolicy` covers the forwarding of every value the policy can receive; the overlay itself is exercised by the existing factory tests.

tests/neuro_san/internals/run_context/langchain/llms/test_openai_llm_policy.py
```diff
@@ -104,161 +109,191 @@ def _request_payload(llm: ChatOpenAI) -> Dict[str, Any]:
...
 
-    def test_default_config_keeps_chat_completions_without_n(self, policies: List[OpenAILlmPolicy]) -> None:
+    def test_default_config_keeps_chat_completions_without_n(self) -> None:
         """
-        A plain gpt-5.2 config stays on Chat Completions, does not force an endpoint, and no longer sends n.
-
-        :param policies: The fixture list that tracks policies for teardown
+        With use_responses_api absent from llm_config the policy itself does not force an endpoint: the
+        llm stays on Chat Completions and no longer sends n. The hocon class default of true is applied
+        by DefaultLlmFactory and is not exercised by this policy-level test.
```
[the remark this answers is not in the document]
Noravee: Agreed that the overlay path deserves its own coverage, and it is tracked as a separate change in #1337: a new `TestDefaultLlmFactory` module (there is none for `DefaultLlmFactory` today) asserting that a model-only OpenAI config resolves `use_responses_api=True`, `store=False`, `include=None` after loading the hocon. Keeping it out of this PR so the flip and the plumbing stay reviewable on their own.

Noravee:
Follow-up in 8376615b after discussion with the maintainer: the langchain-openai floor moves to 1.4.0 (with langchain-core 1.5.0, which it requires). Wheel inspection shows 1.4.0 is the first release that drops `stop` on the Responses path and the first with the store-aware reasoning replay (`store is not False or encrypted_content`) that the new `store: false` default relies on, so the flip and the floor belong together. `stop` stays for Chat Completions users and is now marked Chat-Completions-only in the policy, the hocon and the docs; the pre-1.4 caveats are gone. The factory-level default test is tracked separately in #1337.

neuro_san/internals/run_context/langchain/llms/default_llm_info.hocon
```diff
@@ -1512,23 +1518,49 @@
...
+                #           /chat/completions. For gpt-5.6-* combine it with "reasoning_effort": "none" so that tool
+                #           calls still work (without reasoning).
+                #   null  - let langchain infer the endpoint from the other parameters and the model name: a
+                #           "reasoning" dict, an "include" list or a Responses-only model (gpt-5.x-pro, codex) selects
+                #           the Responses API; otherwise requests stay on Chat Completions (the previous default).
+                # To restore Chat Completions for one agent, set "use_responses_api": false in its llm_config.
+                # To restore it server-wide, point AGENT_LLM_INFO_FILE at a sparse hocon file that deep-merges over
+                # this one, for example:
+                #   { "classes": { "openai": { "args": { "use_responses_api": false } } } }
+                "use_responses_api": true,
```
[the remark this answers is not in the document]
Noravee: Good catch. Both servers only serve /v1/chat/completions, so their examples now pin Chat Completions: c87c4e61 adds `use_responses_api = false` to the four example snippets (tests/mock_llm_server/README.md and mock_llm_server.py, tests/record_playback_llm_server/README.md and record_playback_llm_server.py) with a note explaining why. Adding a /v1/responses route to the servers is a larger change and out of scope here.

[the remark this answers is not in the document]
Noravee: Already addressed in c87c4e61 (pushed before this review was generated): all four bundled examples, tests/mock_llm_server/README.md, mock_llm_server.py, tests/record_playback_llm_server/README.md and record_playback_llm_server.py, now include `use_responses_api = false` with a note that those servers only implement /v1/chat/completions.

Noravee:
On the suppressed comment about the server-wide override: correct, `DefaultLlmFactory` takes an agent network's own `llm_info_file` first and only falls back to `AGENT_LLM_INFO_FILE` (default_llm_factory.py:98-107). 0646f739 says so next to the override recipe in both the docs and the hocon comment, and links the existing `llm_info_file` section. The mock and record/playback examples were pinned to Chat Completions in c87c4e61.

neuro_san/internals/run_context/langchain/llms/default_llm_info.hocon
```diff
@@ -1512,23 +1518,51 @@
...
+                #   null  - let langchain infer the endpoint from the other parameters and the model name: a
+                #           "reasoning" dict, an "include" list or a Responses-only model (gpt-5.x-pro, codex) selects
+                #           the Responses API; otherwise requests stay on Chat Completions (the previous default).
+                # To restore Chat Completions for one agent, set "use_responses_api": false in its llm_config.
+                # To restore it server-wide, point AGENT_LLM_INFO_FILE at a sparse hocon file that deep-merges over
+                # this one, for example:
+                #   { "classes": { "openai": { "args": { "use_responses_api": false } } } }
+                # An agent network that sets its own llm_info_file ignores AGENT_LLM_INFO_FILE and needs the same
+                # override in that file.
+                "use_responses_api": true,
```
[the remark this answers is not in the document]
Noravee: Checked against OpenAI's model pages rather than assumed: the endpoints table for GPT-4 (snapshots gpt-4-0613 and gpt-4-0314, https://developers.openai.com/api/docs/models/gpt-4) and for GPT-3.5 Turbo (https://developers.openai.com/api/docs/models/gpt-3.5-turbo) both mark `v1/responses` as Supported alongside `v1/chat/completions`. Neither legacy entry is Chat-Completions-only, so no per-model endpoint override is needed for them. For the record, `create_full_llm_config()` does not overlay model-entry keys into the config today, so a per-model `use_responses_api` would be a new factory feature; if a Chat-Completions-only OpenAI model ever appears in the table, that is the change to make.

tests/mock_llm_server/README.md
```diff
@@ -86,13 +86,17 @@ llm_config {
...
 }
 ```
 
 Notes:
 
 - `openai_api_base` must include the `/v1` path segment — the OpenAI client SDK
   is strict about this.
+- `use_responses_api = false` is required: the `openai` class defaults to the
+  Responses API, and the mock only implements `/v1/chat/completions`, so
+  without it every request is a 404 on `/v1/responses`.
```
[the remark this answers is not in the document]
Noravee: Fixed in c42239f8. A sparse overlay now ships at tests/mock_llm_server/llm_info_chat_completions.hocon; the load test's --auto-start passes it to the neuro-san server as AGENT_LLM_INFO_FILE, the manual prerequisites and every error hint export it, and _check_server_api_base warns when a running server lacks it, explaining the /v1/responses 404s.

tests/mock_llm_server/mock_llm_server.py
```diff
@@ -23,13 +23,16 @@
     GET  /v1/models             (lists the configured mock model)
     GET  /healthz               (liveness probe)
 
-Wire it into a `.hocon` agent network exactly like a real OpenAI endpoint:
+Wire it into a `.hocon` agent network exactly like a real OpenAI endpoint. The
+`openai` class defaults to the Responses API and this server only implements
+`/v1/chat/completions`, so pin Chat Completions with use_responses_api = false:
```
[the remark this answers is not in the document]
Noravee: Fixed in c42239f8: the load test's --auto-start path now sets AGENT_LLM_INFO_FILE=tests/mock_llm_server/llm_info_chat_completions.hocon for the neuro-san server, and the manual path documents the same export, so the model-name-only math_guy and hello_world configs stay on Chat Completions against the mock.

tests/record_playback_llm_server/README.md
```diff
@@ -183,10 +183,14 @@ llm_config {
...
     openai_api_base = "http://localhost:8899/v1"
     openai_api_key = "not-needed"
+    use_responses_api = false
 }
 ```
 
 - `openai_api_base` must include the `/v1` path segment.
+- `use_responses_api = false` is required: the `openai` class defaults to the
+  Responses API, and the proxy only implements `/v1/chat/completions`, so
+  without it every request is a 404 on `/v1/responses`.
```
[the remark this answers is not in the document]
Noravee: Fixed in c42239f8: both READMEs' opening paragraphs now say an openai_api_base change plus use_responses_api = false is needed, explain why, and link the section with the full instructions, which also mention the server-wide overlay.

Noravee:
On the suppressed comment about record/playback cassette keys: correct, the class default puts `store: false` into Chat Completions payloads too, and `RequestCanonicalizer` hashed every body field, so cassettes recorded before this change would miss. c42239f8 adds `store` to `VOLATILE_BODY_KEYS` (it only controls server-side retention and never changes the answer) and documents that in the proxy README, so existing cassettes keep matching without re-recording. The same commit ships the Chat Completions overlay for the bundled servers and wires it into the mock-based load test.

tests/load_tests/load_test_mock_llm_service.py
```diff
@@ -334,19 +340,40 @@ def _check_server_api_base(server_proc, mock_port):
...
             )
             sys.exit(1)
 
         logger.info("  OPENAI_API_BASE=%s", api_base)
+
+        # Without the overlay, model-name-only openai configs (math_guy, hello_world) use the
+        # Responses API, which the mock does not serve, and every request is a 404. Warn rather
+        # than exit: a deployment may pin Chat Completions through a per-network llm_info_file.
+        llm_info_file: str = server_env.get("AGENT_LLM_INFO_FILE")
+        if llm_info_file is None:
```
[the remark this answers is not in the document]
Noravee: Right, neuro_san/deploy/Dockerfile:166 and Dockerfile.py314t:319 set `ENV AGENT_LLM_INFO_FILE=""` and DefaultLlmFactory ignores an empty path. 59e3fce3 switches the check to a falsy test with a comment saying why.

Noravee:
On the suppressed canonicalizer comment: 59e3fce3 adds `test_store_is_volatile_but_other_fields_are_not` to `TestRequestCanonicalizer`, asserting that bodies differing only by `store` (absent, false, true) hash identically while a changed message does not, and that `store` is in `VOLATILE_BODY_KEYS`. Verified the test fails when the volatile key is removed. The module was converted to the repo's TestCase conventions in the same commit.

neuro_san/internals/run_context/langchain/llms/default_llm_info.hocon
Noravee, line 1541: The flip itself: `true` replaces the previous `null` (langchain inference). gpt-6-astra and gpt-5.6-* only allow function tools together with reasoning on the Responses API, and gpt-6-astra has no Chat Completions function calling at all, so with `null` every such model needed a per-hocon `use_responses_api: true`. The comment above documents the three values and both ways to get Chat Completions back (per agent, or server-wide via `AGENT_LLM_INFO_FILE`). Note the last two lines: a network that sets its own `llm_info_file` ignores the environment variable.

neuro_san/internals/run_context/langchain/llms/default_llm_info.hocon
```diff
@@ -1512,23 +1518,51 @@
...
+                # Whether OpenAI stores the response server-side (bool or null). OpenAI stores by default (for 30
+                # days); false keeps every request stateless. Stateless reasoning replay in the tool-calling loop
+                # still works, because the Responses API includes encrypted_content on reasoning items by default:
+                # https://developers.openai.com/api/docs/guides/reasoning
+                #   false - (default) do not store responses server-side.
+                #   true  - let OpenAI store responses so they can be retrieved later.
+                #   null  - omit the field entirely, for gateways that reject fields they do not know.
+                "store": false,
```
Noravee, line 1549: `store: false` keeps every request stateless, which is how the other providers already behave (Anthropic and Gemini carry their thinking signatures inline). Live runs against gpt-5.2-2025-12-11 and gpt-6-astra confirmed that reasoning items carry `encrypted_content` even with `store: false` and no `include`, and that the tool-calling loop replays them correctly. `null` is documented for gateways that reject unknown fields: langchain leaves the key out entirely when the value is `None`.

Noravee (commented):
**TL;DR**

- The `openai` class now defaults to the Responses API (`use_responses_api: true`) with stateless requests (`store: false`) and no `include`. A `model_name`-only `llm_config` now works for gpt-6-astra and gpt-5.6-*, which only allow tool calling with reasoning on that endpoint.
- `OpenAILlmPolicy` forwards `store` and `include` raw, like `use_responses_api`, so a `null` reaches langchain as `None` and the field stays out of the request.
- Floors move to langchain-openai 1.4.0 and langchain-core 1.5.0. 1.4.0 is the first release that drops `stop` on the Responses path and replays reasoning items by `encrypted_content` when `store` is false.
- The bundled mock and record/playback servers only serve `/v1/chat/completions`, so the PR adds a sparse overlay (`tests/mock_llm_server/llm_info_chat_completions.hocon`), wires it into the load test's `--auto-start`, and corrects both servers' examples. `store` becomes a volatile key in the cassette canonicalizer so recordings made before this default keep matching.
- Opting back into Chat Completions: `"use_responses_api": false` per agent, or `AGENT_LLM_INFO_FILE` pointing at the overlay server-wide.
- Where to start reviewing: `default_llm_info.hocon` lines 1541-1554 and `openai_llm_policy.py` lines 200-226. The inline comments below mark the important parts of the rest.

neuro_san/internals/run_context/langchain/llms/default_llm_info.hocon
```diff
@@ -1512,23 +1518,51 @@
...
+                # Extra output fields to return (list of strings or null), e.g. [ "reasoning.encrypted_content" ].
+                # OpenAI still accepts that legacy value but no longer requires it: encrypted_content is included
+                # by default. Any non-null value selects the Responses API in langchain when use_responses_api is
+                # null, and is rejected by Chat Completions, so this must never be given a non-null default here.
+                "include": null,
```
Noravee, line 1554: `include` must stay `null` here. In langchain-openai any non-`None` `include` is a Responses routing trigger when `use_responses_api` is `null`, and `chat.completions.create()` rejects the keyword, so a non-null class default would break anyone who opts back into Chat Completions. The legacy `["reasoning.encrypted_content"]` can still be set per agent; OpenAI no longer requires it.

neuro_san/internals/run_context/langchain/llms/openai_llm_policy.py
```diff
@@ -178,27 +180,51 @@ def create_llm(self, config: Dict[str, Any], model_name: str, client: Any) -> Ba
...
+            #           the default endpoint.
             # This is deliberately forwarded raw instead of through ConfigUtil.get_bool():
-            # get_bool() turns an absent/null value into False, which would disable langchain's
-            # auto-routing, send a "reasoning" dict to Chat Completions, and pin the *-pro models
-            # to an endpoint OpenAI rejects. Only an explicit true/false in llm_config should
-            # override langchain's choice.
+            # get_bool() turns an absent/null value into False. An explicit false must reach langchain
+            # as False (pin Chat Completions), and a null must reach it as None so that inference still
+            # works for users who opt back into it; a null collapsed to False would send a "reasoning"
+            # dict to Chat Completions and pin the *-pro models to an endpoint OpenAI rejects.
```
Noravee, line 204: The endpoint-related keys are forwarded raw rather than through `ConfigUtil.get_bool()`. `get_bool()` collapses an absent or null value to `False`, which would pin everyone to Chat Completions and send a `reasoning` dict to an endpoint that rejects it. With raw forwarding an explicit `false` reaches langchain as `False` and a `null` as `None`, so inference still works for users who opt back into it.

neuro_san/internals/run_context/langchain/llms/openai_llm_policy.py
```diff
@@ -178,27 +180,51 @@ def create_llm(self, config: Dict[str, Any], model_name: str, client: Any) -> Ba
...
+            # (for 30 days); the class default in default_llm_info.hocon is false, which keeps every
+            # request stateless. Reasoning replay in the tool-calling loop still works statelessly,
+            # because the Responses API includes encrypted_content on reasoning items by default and
+            # langchain-openai replays any reasoning item that carries it. Chat Completions accepts
+            # store as well, so forwarding it is safe on either endpoint. Forwarded raw for the same
+            # reason as use_responses_api: a null must reach langchain as None so the field is left
+            # out of the request entirely (some gateways reject fields they do not know).
+            # https://developers.openai.com/api/docs/guides/reasoning
+            # https://docs.langchain.com/oss/python/integrations/chat/openai#responses-api
+            store=config.get("store"),
```
Noravee, line 217: Chat Completions accepts `store` as well, so forwarding it is safe on both endpoints. That also means a `use_responses_api: false` gateway config now sends `store: false` in its Chat Completions bodies, which is why `RequestCanonicalizer` treats `store` as volatile (see the comment there).

neuro_san/internals/run_context/langchain/llms/openai_llm_policy.py
```diff
@@ -178,27 +180,51 @@ def create_llm(self, config: Dict[str, Any], model_name: str, client: Any) -> Ba
...
+
+            # include is the list of extra output fields to return, e.g. ["reasoning.encrypted_content"].
+            # OpenAI still accepts that legacy value but no longer requires it, because encrypted_content
+            # is included by default. Any non-None include is a Responses API routing trigger in
+            # langchain-openai, and the Chat Completions SDK call rejects the keyword outright, so it is
+            # forwarded raw and must never be given a non-null class default.
+            # https://developers.openai.com/api/docs/guides/reasoning
+            # https://reference.langchain.com/python/langchain-openai/chat_models/base/BaseChatOpenAI/use_responses_api
+            include=config.get("include"),
```
Noravee, line 226: Forwarded raw for the same reason as `use_responses_api`. See the hocon comment on why the class default must stay `null`.

neuro_san/internals/run_context/langchain/llms/openai_llm_policy.py
```diff
@@ -165,6 +165,8 @@ def create_llm(self, config: Dict[str, Any], model_name: str, client: Any) -> Ba
             top_p=config.get("top_p"),
             max_tokens=config.get("max_tokens"),  # This is always for output
             tiktoken_model_name=config.get("tiktoken_model_name"),
+            # Only takes effect on Chat Completions: langchain-openai drops stop from Responses API
+            # requests (the API has no such parameter) rather than failing.
```
Noravee, line 169: `stop` stays. langchain-openai 1.4.0 and later drop it from Responses API requests instead of sending it (older releases raised a client-side `TypeError`), so it only takes effect when `use_responses_api` is `false`. The requirements floor guarantees the dropping behavior; the hocon comment and the docs say the same.

requirements.txt
```diff
@@ -48,7 +49,14 @@ pydantic>=2.9.2,<3.0
...
 # NOTE: This version should be aligned with the langchain and langchain-core versions above.
-langchain-openai>=1.0.0,<2.0
+# Floor 1.4.0: the Responses API is the default OpenAI endpoint (default_llm_info.hocon,
+# classes.openai.args.use_responses_api) with store=false. 1.4.0 is the first release that
+# drops the Chat-Completions-only "stop" parameter from Responses API requests instead of
+# sending it (a client-side TypeError before), and the first whose request builder replays
+# reasoning items by their encrypted_content when store is false, which the stateless
+# default relies on for tool-calling loops. langchain-openai 1.4.0 itself requires
+# langchain-core>=1.5.0 and openai>=2.45.0.
+langchain-openai>=1.4.0,<2.0
```
Noravee, line 59: Why 1.4.0 and not the previous floor: inspecting the released wheels, 1.4.0 is the first langchain-openai that (a) pops `stop` from Responses payloads and (b) passes `store` into `_construct_responses_api_input()` so reasoning items are replayed by `encrypted_content` when `store` is false. Both are needed for the new defaults. 1.4.0 itself requires langchain-core>=1.5.0 and openai>=2.45.0, hence the langchain-core floor above.

tests/mock_llm_server/llm_info_chat_completions.hocon
```diff
@@ -0,0 +1,22 @@
...
+#     export AGENT_LLM_INFO_FILE=tests/mock_llm_server/llm_info_chat_completions.hocon
+#
+# The file deep-merges over the default llm info, so nothing else changes. An agent network that
+# sets its own "llm_info_file" ignores the environment variable and needs the same
+# "use_responses_api": false in that file instead.
+{
+    "classes": {
+        "openai": {
+            "args": {
+                "use_responses_api": false
```
Noravee, line 18: Why this file exists: both bundled test servers register only `/v1/chat/completions` (`mock_llm_server.py:67`, `record_playback_llm_server.py:99-100`). After the flip a model-name-only agent such as `math_guy` posts to `/v1/responses` and gets a 404. This sparse file deep-merges over `default_llm_info.hocon` through `AGENT_LLM_INFO_FILE`, so nothing else changes. The alternatives were larger or wrong: a Responses handler in both servers, or pinning the shipped registries to Chat Completions against real OpenAI.

tests/load_tests/load_test_mock_llm_service.py
```diff
@@ -404,7 +435,12 @@ def _auto_start_servers(self):
             stderr=self._mock_log_fh,
         )
 
-        server_env = {**os.environ, "OPENAI_API_BASE": api_base}
+        # The overlay keeps openai-class models on Chat Completions, the only endpoint the mock serves.
+        server_env = {
+            **os.environ,
+            "OPENAI_API_BASE": api_base,
+            "AGENT_LLM_INFO_FILE": CHAT_COMPLETIONS_LLM_INFO_PATH,
```
Noravee, line 442: `--auto-start` now launches the server with the overlay. Without it every request 404s on `/v1/responses`, but the load test would still report 100% success: the 404 is caught and returned as an "Agent stopped due to exception" answer, `agent_cli --one_shot` exits 0, and `_run_one` counts exit code 0 as OK. The latency and leak numbers would then measure the error path. Verified both ways against the bundled mock on localhost.

tests/load_tests/load_test_mock_llm_service.py
```diff
@@ -334,19 +340,42 @@ def _check_server_api_base(server_proc, mock_port):
...
 
         logger.info("  OPENAI_API_BASE=%s", api_base)
+
+        # Without the overlay, model-name-only openai configs (math_guy, hello_world) use the
+        # Responses API, which the mock does not serve, and every request is a 404. Warn rather
+        # than exit: a deployment may pin Chat Completions through a per-network llm_info_file.
+        # An empty value counts as unset: the shipped Dockerfiles define AGENT_LLM_INFO_FILE=""
+        # and DefaultLlmFactory ignores an empty path just like a missing variable.
+        llm_info_file: str = server_env.get("AGENT_LLM_INFO_FILE")
+        if not llm_info_file:
```
Noravee, line 358: For a manually started server the test cannot know how the LLM is pinned (a network may use its own `llm_info_file`), so it warns instead of exiting. `if not llm_info_file` rather than an `is None` check because the shipped Dockerfiles define `ENV AGENT_LLM_INFO_FILE=""` and `DefaultLlmFactory` ignores an empty path exactly like a missing variable.

tests/record_playback_llm_server/request_canonicalizer.py
```diff
@@ -48,10 +48,15 @@ class RequestCanonicalizer:
...
-    VOLATILE_BODY_KEYS: Tuple[str, ...] = ()
+    # every caller in-process). Extend this tuple in source if a client is
+    # found to inject a per-run volatile value (a random request id, a
+    # timestamp, etc.) into the request body.
+    #
+    # "store" only tells OpenAI whether to keep the response server-side and
+    # never changes the answer. neuro-san's openai class now sends store=false
+    # on every request, so dropping it keeps cassettes recorded before that
+    # default existed matching the requests neuro-san sends today.
+    VOLATILE_BODY_KEYS: Tuple[str, ...] = ("store",)
```
Noravee, line 59: langchain-openai emits `store` on Chat Completions bodies too (it is a valid parameter there), so with the new default every request carries `store: false` and cassettes recorded before this PR would stop matching. `store` never changes the answer, so dropping it from the hash is safe. No cassettes are committed in the repo; this keeps developers' local recordings usable.

tests/record_playback_llm_server/tests/test_request_canonicalizer.py
```diff
@@ -19,28 +18,78 @@
...
+    def test_path_and_method_participate(self) -> None:
+        """
+        Method and path are part of the key, not just the body.
+        """
+        body: bytes = b"{}"
+
+        self.assertNotEqual(RequestCanonicalizer.key("POST", "/chat/completions", body),
+                            RequestCanonicalizer.key("GET", "/chat/completions", body))
+
+    def test_store_is_volatile_but_other_fields_are_not(self) -> None:
```
Noravee, line 75: Regression test for the volatile key; it fails if `store` is removed from `VOLATILE_BODY_KEYS` (checked locally). The module is converted to `unittest.TestCase` per the repo test conventions since it was being touched. The three existing tests assert exactly what they did before.

tests/neuro_san/internals/run_context/langchain/llms/test_openai_llm_policy.py
```diff
@@ -104,161 +109,191 @@ def _request_payload(llm: ChatOpenAI) -> Dict[str, Any]:
...
-        assert llm.openai_api_key.get_secret_value() == "sk-test"
-        assert llm.openai_api_base == "http://127.0.0.1:9"
-        assert llm.openai_organization == "org-test"
-        assert llm.openai_proxy == "http://127.0.0.1:9"
+        self.assertEqual(llm.openai_api_key.get_secret_value(), "sk-test")
+        self.assertEqual(llm.openai_api_base, "http://127.0.0.1:9")
+        self.assertEqual(llm.openai_organization, "org-test")
+        self.assertEqual(llm.openai_proxy, "http://127.0.0.1:9")
+
+    def test_store_and_include_from_llm_config_reach_the_payload(self) -> None:
```
Noravee, line 258: The three new tests cover the forwarding matrix: both keys present, `store` alone (no invented `include`), and both absent (both keys left out of the request). Each payload is also bound against the real `responses.create()` signature, so an unknown keyword fails here rather than at runtime. The rest of the module is the conversion to `unittest.TestCase` (`setUp`/`tearDown` with `@override`, `self.assert*` only, no pytest fixtures).

docs/agent_hocon_reference.md
```diff
@@ -385,9 +421,27 @@ differences between the two endpoints.
...
-Azure deployments.
+**Azure OpenAI.** The `azure-openai` class extends `openai` and therefore inherits the `use_responses_api`,
+`store` and `include` defaults, but `AzureLlmPolicy` does not forward any of them to `AzureChatOpenAI`, so those
+keys have no effect on Azure. Azure requests instead follow langchain's own inference: Chat Completions for most
+models, but a model name langchain knows to be Responses-only (such as the `gpt-5.4-pro` snapshot behind
+`azure-gpt-5.4-pro`) auto-routes to the Responses API, which `AzureLlmPolicy` does not support yet. Responses API
+support for Azure, including pinning the endpoint, is tracked in
+[#1307](https://github.com/cognizant-ai-lab/neuro-san/issues/1307).
+
+**Upgrading existing OpenAI configurations.** Making the Responses API the default changes a few observable
```
Noravee: The upgrade note is the user-facing list of everything that changes for llm_configs that never set `use_responses_api`: no server-side storage, `stop` ignored, the two floors, and the gateway case with a ready-made overlay. If you read only one part of the docs change, read this one.

docs/agent_hocon_reference.md
```diff
@@ -323,27 +323,62 @@ to expose secrets by checking them in.
...
+      {
+          "classes": {
+              "openai": {
+                  "args": {
+                      "use_responses_api": false
+                  }
+              }
+          }
+      }
+      ```
```
d1donlydfink: Why is this information about llm_info in the agent_hocon_reference?
Shouldn't this be in the llm_info_hocon_reference instead?

neuro_san/internals/run_context/langchain/llms/default_llm_info.hocon
d1donlydfink: Is there any indication what will happen to LiteLLM instances which co-opt the OpenAI LLM Interface?
(We use these in production)

docs/agent_hocon_reference.md
Noravee: Agreed. The provider-specific llm_config documentation for #1332, #1333 and #1334 is now consolidated in #1334, which moves it, including the pre-existing OpenAI section, into a new "Provider-Specific Arguments" section of `docs/llm_info_hocon_reference.md` and leaves one pointer paragraph in the agent reference. 34ceca22 removes the agent reference changes from this PR; the `default_llm_info.hocon` comment URLs now point at the llm_info reference.

neuro_san/internals/run_context/langchain/llms/default_llm_info.hocon
Noravee: Good question. LiteLLM translates as well as routes, so it depends on what sits behind it:

- **LiteLLM in front of OpenAI**: nothing to do. LiteLLM serves `/v1/responses` and its OpenAI Responses path accepts every Responses parameter, `store` included, so the only visible change is the path and the `store: false` field. The one prerequisite is a LiteLLM version recent enough to serve `/v1/responses`; an older proxy would answer 404 there, in which case pin Chat Completions as below.
- **LiteLLM translating to another provider** (an Anthropic, Bedrock or Gemini model behind an OpenAI-style alias): LiteLLM checks each OpenAI parameter against that provider's supported list and raises `UnsupportedParamsError` for `store` unless the proxy has `drop_params: true` (its Responses-to-Chat bridge does not list `store` either). Setting `"store": null` makes langchain omit the field entirely, and `"use_responses_api": false` keeps the request on `/chat/completions` if the proxy does not bridge the Responses API.

Both keys can be set per agent in `llm_config`.

The documentation paragraph on gateways such as LiteLLM is in #1334. 

In short, for OpenAI behind a current LiteLLM I would expect no change to be needed.

vince-leaf:
Regarding the 404-as-success behavior noted in load_test_mock_llm_service.py:

When the server posts to /v1/responses against a backend that only serves /chat/completions, the 404 is caught and returned as an "Agent stopped due to exception" answer; agent_cli --one_shot exits with 0, and the load test reports 100% success. The overlay and the warning route around this for the load test, but the underlying failure mode is still silent, and it now applies to any deployment whose gateway lacks /responses.

vince-leaf:
What remains to verify (can't do without a key): whether LiteLLM's Responses support works for the specific model names configured in staging (it needs each model's provider to be Responses-capable; for OpenAI, it forwards natively), and whether LiteLLM cost/usage tracking in the gateway logs Responses calls the same way — the Datadog/Langfuse dashboards mention LiteLLM cost metrics. 
But we will need to deploy this to test or stage to confirm.
A quick smoke test of one agent in staging after the neuro-san image bump would settle it.

vince-leaf:
Merge order note: the five default_llm_info.hocon comment URLs point to docs/llm_info_hocon_reference.md#openai-reasoning-and-responses-api-parameters, which only exists once #1334 lands. Until then, main has dead links, and docs/agent_hocon_reference.md still documents use_responses_api: null as the default. Can we merge #1333 → #1332 → #1334 back-to-back, or fold the doc move into this PR so main is never inconsistent?

vince-leaf:
@Noravee 
Dependency floors look fine. Checked:

- langchain-openai>=1.4.0 requires langchain-core>=1.5.0 and openai>=2.45,<3; current PyPI is langchain-openai 1.6.2 / langchain-core 1.6.3, so a fresh resolve already lands well above both floors.
- The other provider packages in requirements-build.txt (langchain-anthropic, -google-genai, -aws, -ollama, -nvidia-ai-endpoints, -openrouter) all have current releases compatible with langchain-core 1.6.
- No lockfile in this repo or neuro-san-studio; both Dockerfiles run pip install -r requirements.txt from scratch, so images pick up the new versions automatically.

One follow-up outside this PR: neuro-san-studio pins neuro-san==0.7.0 and langchain-openai>=1.1.9. When studio bumps to the neuro-san release containing this change, its langchain-openai floor should be raised to >=1.4.0 to match. No ticket exists for that yet.

vince-leaf:
Verified locally on this branch (langchain-openai 1.6.2, real OpenAI key): a tool-call loop with {"model_name": "gpt-5.2", "reasoning_effort": "medium"} sends both requests to /v1/responses with store: false, and the second request replays the reasoning item with encrypted_content. Works as described.

One note on the store comment in default_llm_info.hocon: with the class defaults alone (no reasoning/reasoning_effort), gpt-5.2 runs at reasoning_effort: none and returns no reasoning items, so encrypted_content replay only comes into play for agents that turn reasoning on. Might be worth a short clarification in the comment. Not a blocker.

vince-leaf:
@Noravee, as you requested. I ran a smoke test again on this PR, and it passed. Note: This is not with the LiteLLM config.

Noravee:
@vince-leaf thanks for the thorough pass. Point by point:

1. **404 reported as success.** Agreed, and it is pre-existing: `parse_chain_result()` has turned the final exception into an `Agent stopped due to exception ...` answer since the first import (`run_context_runnable.py:390`), `agent_cli --one_shot` exits 0 on it, and the load test judges by exit code. The endpoint default adds a common trigger, so I opened #1362 to surface API errors as failures (message marked as an error, non-zero exit in one-shot mode, load test judging by the response). Kept out of this PR.
2. **Staging.** Agreed that a one-agent smoke test in staging after the image bump is what settles LiteLLM's Responses support for the staging model names and the cost tracking. For OpenAI-backed aliases LiteLLM forwards natively. If anything is off, the two-line override (`use_responses_api: false`, `store: null`) through `AGENT_LLM_INFO_FILE` restores the previous traffic without a code change.
3. **Merge order.** The plan is back-to-back: #1333, then this PR, then #1334. #1334 is already approved, so once this PR and #1333 are approved the three can go in within minutes and main is inconsistent only for that window. Folding the doc move back into this PR would undo the consolidation Dan asked for. If the window is a concern, the alternative is to point the five hocon URLs back at the agent reference here and let #1334 move them, which trades the dead links for a rebase conflict in #1334.
4. **Dependency floors.** Thanks for checking the resolution. Opened cognizant-ai-lab/neuro-san-studio#1432 to raise the studio `langchain-openai` floor when it bumps neuro-san.
5. **Live verification and the `store` comment.** Thanks, and good point: `none` is gpt-5.2's default reasoning effort, so the replay only matters once an agent turns reasoning on. c50392b7 adds that clarification to the `store` comment in `default_llm_info.hocon`.
6. **Smoke test.** Thanks for running it.

tests/record_playback_llm_server/tests/test_request_canonicalizer.py
```diff
@@ -19,28 +18,78 @@
...
-            RequestCanonicalizer.key("GET", "/chat/completions", body)
+class TestRequestCanonicalizer(TestCase):
+    """
+    Canonical-key stability guarantees the record/playback matching relies on.
+    """
+
+    PATH: str = "/v1/chat/completions"
+
+    @staticmethod
+    def _key(body: Dict[str, Any], method: str = "POST", path: str = "/v1/chat/completions") -> str:
```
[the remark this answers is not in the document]
Noravee: Fixed in f54e9112: `_key()` now takes `PATH` as its default, so the string lives in one place.

tests/load_tests/load_test_mock_llm_service.py
```diff
@@ -334,19 +341,42 @@ def _check_server_api_base(server_proc, mock_port):
...
+            logger.warning(
+                "neuro-san server does not have AGENT_LLM_INFO_FILE set.\n"
+                "  The openai class defaults to the Responses API, which the mock LLM server does not\n"
+                "  serve, so model-name-only agent configs will get HTTP 404 from /v1/responses.\n"
+                "  Restart the server with:\n"
+                "    export AGENT_LLM_INFO_FILE=%s\n"
+                "    export OPENAI_API_BASE=%s\n"
+                "    python -m neuro_san.service.main_loop.server_main_loop",
+                CHAT_COMPLETIONS_LLM_INFO_PATH, expected_url,
+            )
```
d1donlydfink, line 369: This is not a warning-level message.

The default is to run with no AGENT_LLM_INFO_FILE set at all, even in the Dockerfiles.
You will be polluting everyone's test output with this message with no path to turn it off
and the instructions to get rid of it increase out-of-box complexity for running the tests.

Strongly consider demoting this to debug or getting rid of it altogether.
Alternatively, maybe we get rid of MockLlmServer. (See question below)

tests/load_tests/load_test_mock_llm_service.py
```diff
@@ -334,19 +341,42 @@ def _check_server_api_base(server_proc, mock_port):
...
+        else:
+            logger.info("  AGENT_LLM_INFO_FILE=%s", llm_info_file)
```
d1donlydfink, line 371: This else should not be necessary.

tests/load_tests/load_test_mock_llm_service.py
```diff
@@ -334,19 +341,42 @@ def _check_server_api_base(server_proc, mock_port):
...
+
         if mock_port not in api_base:
             logger.error(
                 "OPENAI_API_BASE does not reference port %s.\n"
                 "  Mock LLM server is running on port %s,\n"
                 "  but OPENAI_API_BASE=%s\n"
                 "  Restart the server with:\n"
+                "    export AGENT_LLM_INFO_FILE=tests/mock_llm_server/llm_info_chat_completions.hocon\n"
```
d1donlydfink, line 379: What does AGENT_LLM_INFO_FILE have to do with this error message?

tests/load_tests/load_test_mock_llm_service.py
```diff
@@ -366,7 +396,8 @@ def _find_local_processes(self):
         if self.server_proc is None:
             logger.error(
                 "neuro-san server process not found.\n"
-                "Start it with OPENAI_API_BASE pointing to the mock LLM server:\n"
+                "Start it with the Chat Completions overlay and OPENAI_API_BASE pointing to the mock LLM server:\n"
+                "  export AGENT_LLM_INFO_FILE=tests/mock_llm_server/llm_info_chat_completions.hocon\n"
```
d1donlydfink, line 400: Overkill.

tests/load_tests/load_test_mock_llm_service.py
```diff
@@ -379,6 +410,7 @@ def _find_local_processes(self):
                 "Start the mock LLM server first, then the neuro-san server:\n"
                 "  python -m tests.mock_llm_server.mock_llm_server --port 8888\n"
                 "Then:\n"
+                "  export AGENT_LLM_INFO_FILE=tests/mock_llm_server/llm_info_chat_completions.hocon\n"
```
d1donlydfink, line 413: Overkill

tests/mock_llm_server/llm_info_chat_completions.hocon
d1donlydfink: > The alternatives were larger or wrong: a Responses handler in both servers, or pinning the shipped registries to Chat Completions against real OpenAI.

This MockLlmServer seems to be severely complicating things, but it is only associated with the load_tests framework, which we are in the process of making generic.    

@andreidenissov-cog , @vince-leaf : 
* Why do we need the MockLlmServer?
* What do we lose if we simply get rid of it?

andreidenissov-cog: 1. That's a question to @vince-leaf 
2. I'm sure we can simplify things in LT area a lot, but again - @vince-leaf would know better.

andreidenissov-cog: Ah. Take that back: MockLlmServer (if under tests/mock_llm_server) has nothing to do with load test (tests/load_tests). It is pretty basic LLM requests responder when we need something to quickly test neuro-san against. No tokens, no external endpoint.
Do we need it? Maybe from time to time.
Get we get rid of it? Yes we can. Small testing - real LLM with cheap model; bigger runs - Boombox.

vince-leaf: Here is my 2 cents, but @andreidenissov-cog would make a better call here as he created both of these features.

> `mock_llm_server` (Andrei's formalization of the mock approach Nor first hacked together) is the zero-cost, zero-rate-limit backend for the leak/load tests in `tests/load_tests/load_test_mock_llm_service.py`. It's not used by CI — only the manual load framework.
>
> **Different tool, different job:**
> - Mock: synthetic, no setup, handles any prompt/agent, tunable latency, stateless — good for "hammer the server with N concurrent requests and watch for leaks".
> - Record/playback: needs a cassette recorded first against real OpenAI (tokens + API key); an unmatched request in `playback` mode returns 504. Load tests that vary prompts/sly_data or hit new agents miss unless run in `hybrid` mode (which spends tokens on misses). Great for deterministic regression runs, less so for generic load.
>
> **If we drop `mock_llm_server` and use record/playback for load tests:**
> - Same Responses-API 404 — `record_playback_llm_server` also only serves `/v1/chat/completions` (lines 99-100), so the override hocon is still required. Nothing is simplified.
> - Every load-test agent/prompt needs a cassette recorded first; any request that doesn't match fails with 504.
> - Latency becomes whatever was recorded, not tunable.
>
> **Net:** we'd trade a zero-setup load backend for one that needs recording upkeep, and still carry the override. Keep the mock and the sparse `AGENT_LLM_INFO_FILE` override for now; if we want one server, add a `/v1/responses` handler as a follow-up.

neuro_san/internals/run_context/langchain/llms/default_llm_info.hocon
```diff
@@ -78,9 +78,10 @@
             "output": [ "text" ],
         },
         "capabilities": [ "tools" ],
-        # Tool calling needs "use_responses_api": true in llm_config: Chat Completions rejects tools combined
-        # with reasoning, and this model rejects reasoning_effort "none", so there is no Chat Completions path. See
-        # https://github.com/cognizant-ai-lab/neuro-san/blob/main/docs/agent_hocon_reference.md#openai-reasoning-and-responses-api-parameters
+        # Tool calling works with the class defaults (the Responses API is the default endpoint). Do not set
+        # "use_responses_api": false for this model: Chat Completions rejects tools combined with reasoning and the
+        # model rejects reasoning_effort "none", so there is no Chat Completions path. See
+        # https://github.com/cognizant-ai-lab/neuro-san/blob/main/docs/llm_info_hocon_reference.md#openai-reasoning-and-responses-api-parameters
```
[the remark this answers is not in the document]
Noravee: Known and intentional: the anchor `#openai-reasoning-and-responses-api-parameters` is added by #1334 (approved), which consolidates the provider-specific llm_config docs in `docs/llm_info_hocon_reference.md`. The PR body's Related section records that #1334 merges right after this PR; until then the link is dead for the few minutes between the two merges. Pointing the comments at the old agent-reference section instead would leave five stale URLs behind once #1334 lands.

neuro_san/internals/run_context/langchain/llms/default_llm_info.hocon
```diff
@@ -117,9 +119,10 @@
             "output": [ "text" ],
         },
         "capabilities": [ "tools" ],
-        # Tool calling needs "use_responses_api": true in llm_config, or "reasoning_effort": "none" to stay on
-        # Chat Completions without reasoning (Chat Completions rejects tools combined with reasoning). See
-        # https://github.com/cognizant-ai-lab/neuro-san/blob/main/docs/agent_hocon_reference.md#openai-reasoning-and-responses-api-parameters
+        # Tool calling works with the class defaults (the Responses API is the default endpoint). To call tools on
+        # Chat Completions instead, set "use_responses_api": false together with "reasoning_effort": "none" (Chat
+        # Completions rejects tools combined with reasoning). See
+        # https://github.com/cognizant-ai-lab/neuro-san/blob/main/docs/llm_info_hocon_reference.md#openai-reasoning-and-responses-api-parameters
```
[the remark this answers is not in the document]
Noravee: Known and intentional: the anchor `#openai-reasoning-and-responses-api-parameters` is added by #1334 (approved), which consolidates the provider-specific llm_config docs in `docs/llm_info_hocon_reference.md`. The PR body's Related section records that #1334 merges right after this PR; until then the link is dead for the few minutes between the two merges. Pointing the comments at the old agent-reference section instead would leave five stale URLs behind once #1334 lands.

Noravee:
## LiteLLM: tested, fixed on the gateway side, re-verified

This closes out the staging smoke-test question from review. The short version: the flip did break our LiteLLM gateway, the fix was on the gateway and is now live, and this PR needs no change for it.

**Correction to my earlier reply.** In [my answer to Dan's LiteLLM question](https://github.com/cognizant-ai-lab/neuro-san/pull/1332#discussion_r4008165079) I described the LiteLLM-in-front-of-OpenAI case. Ours is not that. `litellm.evolution.ml` fronts **Azure OpenAI**: three `gpt-5.2` deployments behind one model name, with the Azure api-version taken from a shared secret that carried the dated GA `2024-10-21`. No dated Azure api-version has a `/responses` route, so LiteLLM forwarded every Responses call to `{azure}/openai/responses?api-version=2024-10-21` and Azure answered `404 Resource not found`.

**What we tested.** Against `litellm.evolution.ml` with the staging key, using the exact request bodies this PR's code produces for the staging `llm_config` (`{"class": "openai", "model_name": "gpt-5.2"}`, tools bound, non-streaming), captured from both trees through the real factory and policy:

| Request | api-version 2024-10-21 | api-version v1 |
| --- | --- | --- |
| `POST /chat/completions` (main today) | 200 | 200 |
| `POST /responses` with `store: false` (this PR) | 404 | 200 |

The `v1` column was first measured with LiteLLM's per-request `?api-version=v1` override, before anything was changed.

**What we fixed**, in cognizant-ai-lab/manifest:

- #487 sets `openai_api_version` to `v1` in `infra/azure-openai`, so the secret itself carries it. It also updates `USAGE_FROM_AWS.md`, since on `v1` chat completions move to `/openai/v1/chat/completions` and the `AzureOpenAI` SDK client no longer builds a valid URL.
- #486 documents the value in the LiteLLM chart README, including that a secret change needs a pod restart before it reaches the proxy.

Applied on 2026-09-24. All 8 LiteLLM replicas are on the new revision with `OPENAI_API_VERSION=v1` confirmed inside a running pod.

**Re-verified after the apply**, with no query string: both calls above return **200**. Chat Completions keeps working on `v1`, so current traffic was unaffected by the gateway change.

**Scope.** Only staging and test route through LiteLLM (`OPENAI_API_BASE` is set only in `values-staging.yaml` and `values-test.yaml` in neuro-san-deploy). Dev, prod and byok call `api.openai.com` directly, where the Responses API needs no gateway support.

**What is left.** Nothing on this PR. The deployments install neuro-san from PyPI, so the Responses API reaches staging and test only after a release containing this change and the matching bump in neuro-san-studio (cognizant-ai-lab/neuro-san-studio#1432 covers raising its langchain-openai floor at the same time). The gateway is already on `v1`, so that rollout needs no coordination with the proxy.

tests/load_tests/load_test_mock_llm_service.py
Noravee: Keeping this one, and I have taken the other four comments in 18a1f38f: the else, and the overlay line in the recipes it had nothing to do with (including the OPENAI_API_BASE-missing one you did not flag, same pattern).

The reach is narrower than "everyone's test output". The check runs only in this script's manual mode: `--auto-start` sets the overlay itself and never gets here, remote mode does no env check, pytest does not collect the file, and no workflow runs it. A server started without `OPENAI_API_BASE` (the default, including both Dockerfiles) hits the hard error above before this line. So it fires once, for someone who pointed a local server at the mock and skipped the overlay.

In that case it is the only signal in the load test's own output that the run is meaningless. Reproduced end to end: every request POSTs `/v1/responses`, the mock 404s, the exception comes back as the answer, `agent_cli --one_shot` exits 0, `_run_one` counts it as OK, the summary reports N passed / 0 failed, and `_finalize_test_log` deletes the log file because nothing failed. The 404s do appear in the mock and server terminals, but not in the report anyone reads.

The real fix is #1362, judging success from the response rather than the exit code, which would make this check redundant. And #1352 already lists a `/v1/responses` route on the mock, which removes the overlay requirement entirely. Both are out of scope for a PR that flips the endpoint default.

tests/load_tests/load_test_mock_llm_service.py
Noravee: Dropped in 18a1f38f.

tests/load_tests/load_test_mock_llm_service.py
Noravee: Agreed, nothing. Removed in 18a1f38f, along with the same line in the OPENAI_API_BASE-missing message above it.

tests/load_tests/load_test_mock_llm_service.py
Noravee: Removed in 18a1f38f; the message is back to its previous wording.

tests/load_tests/load_test_mock_llm_service.py
Noravee: Removed in 18a1f38f.

tests/mock_llm_server/llm_info_chat_completions.hocon
Noravee: Closing the loop on this thread: the mock stays for now, per Vince's reasoning above (dropping it would not remove the overlay, since record/playback also serves only chat completions). The `/v1/responses` handler he suggests is already listed in #1352, and once it lands the overlay and the check below go away for the mock path. In the meantime 18a1f38f trims the overlay mentions in this script down to the docstring, the warning, and `--auto-start`.

d1donlydfink (approved):

PR #1333, opened 2026-09-14
Forward include_thoughts to ChatGoogleGenerativeAI

### LINKED ISSUE #1336: Forward include_thoughts to ChatGoogleGenerativeAI so llm_config can request Gemini thought summaries
## Problem

`GeminiLlmPolicy.create_llm()` forwards `thinking_level` and `thinking_budget` but not `include_thoughts`, the `ChatGoogleGenerativeAI` switch that makes Gemini return its thought summaries as thinking blocks in the message content. An `llm_config` that sets `"include_thoughts": true` is silently ignored. The only way to get Gemini thoughts today is the full-class-path route (`"class": "langchain_google_genai.chat_models.ChatGoogleGenerativeAI"`), which bypasses the model alias table, the llm_info-derived settings and the provider policy.

Also, the `thinking_level` comment in `default_llm_info.hocon` lists only Gemini 3 Pro and 3 Flash. gemini-3.7-flash and gemini-3.8-flash reject `"minimal"` (supported: `"low"`, `"medium"` (default), `"high"`), which is not recorded anywhere.

## Proposal

- Forward `include_thoughts=config.get("include_thoughts")` in `GeminiLlmPolicy.create_llm()`.
- Declare `"include_thoughts": null` in `classes.gemini.args` and add the Gemini 3.7 / 3.8 Flash rows to the `thinking_level` comment.
- Document the Gemini thinking parameters in `docs/agent_hocon_reference.md`.
- Tests in `test_gemini_llm_policy.py` for the forwarding and the class default. While there, fold the existing AFC-disable tests into the same class so `GeminiLlmPolicy` has one test module.

## Evidence

Live runs of gemini-3.8-flash through the class pass-through route with `include_thoughts: true` returned `[thinking, text]` content with the thought signature on the text block; `thinking_level: "minimal"` returned an error on 3.7 and 3.8 Flash. langchain-google-genai 4.3.1 declares the field as `include_thoughts: bool | None` and forwards it into the thinking config only when not None.

Related: #1222, #1096.

Noravee:
## What

- `GeminiLlmPolicy.create_llm()` forwards `include_thoughts` from `llm_config` to `ChatGoogleGenerativeAI`.
- `default_llm_info.hocon`, `classes.gemini.args`: declares `"include_thoughts": null` with a comment, and extends the `thinking_level` model rows with Gemini 3.7 Flash and 3.8 Flash (`"low"`, `"medium"` (default), `"high"`; `"minimal"` returns an error).
- Documentation: the "Gemini Thinking Parameters" section (`thinking_level`, `thinking_budget`, `include_thoughts`, temperature advice) lands in #1334, which consolidates the provider-specific llm_config docs in a new "Provider-Specific Arguments" section of `docs/llm_info_hocon_reference.md`. This PR no longer touches `docs/agent_hocon_reference.md`.
- Tests: new `test_gemini_llm_policy.py` with the single class `TestGeminiLlmPolicy`. It covers the forwarding and the class default, and absorbs the former `test_gemini_llm_policy_afc_disabled.py` (deleted): the four stub classes became `MagicMock` stand-ins built by two helpers, and the optional Gemini packages are resolved lazily through the leaf-common `Resolver`, the same way `LlmPolicy` does, so a missing package skips the affected tests instead of failing the import.

## Why

`include_thoughts` is the switch that makes Gemini return its thought summaries as thinking blocks in the message content. Until now an `llm_config` could set it but the policy dropped it silently, so the only way to see Gemini's reasoning was the full-class-path route, which bypasses the alias table and the shared client setup. Live runs of gemini-3.8-flash through that route confirmed the shapes and that `thinking_level: "minimal"` is rejected on 3.7 and 3.8 Flash, which the hocon comment and docs now record.

## Verification

`pytest tests/neuro_san/internals/run_context/langchain/llms` passes (the one skip is the "google-genai SDK not installed" case, which cannot run when it is installed). The forwarding test fails with the policy line removed. With `langchain_google_genai.chat_models` and `google.genai.types` poisoned in `sys.modules`, the module loads and the dependent tests skip cleanly. flake8, pylint (`.pylintrc`) and pymarkdown are clean.

Fixes #1336

## Related

#1222 (Phase 1 arg plumbing), #1096 (the AFC banner the merged tests cover). The documentation for this PR lives in #1334, which documents the new default and must merge after this PR.


neuro_san/internals/run_context/langchain/llms/default_llm_info.hocon
```diff
@@ -1748,6 +1754,15 @@
                 # Set to -1 for dynamic thinking (model decides)
                 # Set to a positive integer to constrain token usage
                 # See https://docs.langchain.com/oss/python/integrations/chat/google_generative_ai#gemini-2-5-models%3A-thinking_budget
+
+                "include_thoughts": null, # Whether to return thought summaries (bool or null).
```
Noravee, line 1758: Add `include_thoughts` argument so that gemini models can return thinking block.

tests/neuro_san/internals/run_context/langchain/llms/test_gemini_llm_policy.py
```diff
@@ -0,0 +1,319 @@
...
+
+from typing_extensions import override
+
+from leaf_common.resolution.resolver import Resolver
+
+from neuro_san.internals.run_context.langchain.llms.default_llm_factory import DefaultLlmFactory
+from neuro_san.internals.run_context.langchain.llms.gemini_llm_policy import GeminiLlmPolicy
+
+
+class TestGeminiLlmPolicy(TestCase):
```
Noravee, line 39: Fix the unit test to follow the repo template.

docs/agent_hocon_reference.md
```diff
@@ -389,6 +390,42 @@ verify a new llm_config without fallbacks first.
 [#1307](https://github.com/cognizant-ai-lab/neuro-san/issues/1307)), so leave `use_responses_api` unset for
 Azure deployments.
 
+#### Gemini Thinking Parameters
```
d1donlydfink: Again, why is this in agent_hocon_reference and not in llm_info_hocon_reference?

What should be happening is that agent_hocon_reference should be referring to llm_info_hocon_reference for the nitty-gritty details on llm_config. Otherwise the main doc is bloated with non-essential information.

It's quite possible a whole section of this doc needs to move to llm_info_hocon_reference.

Noravee: Agreed, and the same for the pre-existing OpenAI section. The whole provider-specific block moves to a new "Provider-Specific Arguments" section of `docs/llm_info_hocon_reference.md` in #1334, with the agent reference keeping one pointer paragraph. ea842225 removes the agent reference changes from this PR; the Gemini section is now part of #1334.

d1donlydfink (approved):

PR #1340, opened 2026-09-15
NS-1338 Add close() to agent sessions to abort in-flight requests.

andreidenissov-cog:
## Summary
Adds a close() method to the agent session API so a client can drop its
connection to abort a streaming_chat() it no longer wants -- for example a
load-test driver cancelling a request whose client-side timeout has fired, so
the neuro-san service can stop the corresponding work.

This is the reusable session capability. Its first consumer -- per-test
cancellation in the data-driven load driver -- lives on a separate branch and
depends on this.

## What close() does, per session type
- AgentSession / AsyncAgentSession (interfaces): new close() with a no-op
  default -- in-process direct sessions hold no connection, so there is nothing
  to close. close() is synchronous on both interfaces so it can be called from a
  supervising sync context or another thread (this matches the existing
  synchronous AsyncDirectAgentSession.close()).
- HttpServiceAgentSession (sync requests): tracks the active streaming response
  and closes it to interrupt the read. close() is non-blocking: it closes the
  response on a background daemon thread, because closing a `requests` streaming
  response while another thread is blocked reading it can itself block, and
  close() must never hang its caller. Thread-safe and idempotent.
- AsyncHttpServiceAgentSession (aiohttp): aborting needs different handles in
  different phases -- closing the response unblocks an in-flight read once
  headers have arrived, while closing the ClientSession aborts a request still
  awaiting headers (there may be no total timeout). Both are tracked and, on the
  request's owning loop (aiohttp is not thread-safe), the response is closed via
  call_soon_threadsafe and the session via run_coroutine_threadsafe. An early
  closed-check + capturing the loop before issuing the request cover the
  cancel-before-first-iteration and pre-headers windows.
- McpServiceAgentSession: close() is a documented no-op. The MCP transport is
  request/response -- requests.post() does not return until the server sends
  headers, and the MCP handler writes only after the full tool run completes --
  so there is no abortable in-flight connection to release mid-run. This session
  deliberately does not claim to support cancellation.

## Single-use contract (AgentSessionClosedError)
Once close() is called, a remote HTTP session is single-use:
- New AgentSessionClosedError is raised by streaming_chat(), function() and
  connectivity() on a closed session, so a deliberate close is distinguishable
  from a genuine connectivity failure (still a ValueError with debug help text)
  and from a legitimately empty stream.
- A close() that interrupts an in-flight streaming read raises
  AgentSessionClosedError (not the generic connectivity ValueError).
- is_closed() is added to the interfaces (default False; True after close() on
  the HTTP sessions) so callers can check state.

## Scope of the guarantee (documented in close())
- Async client: close() reliably aborts the request in both the pre-header and
  post-header windows.
- Sync client: close() reliably interrupts the streaming read WHILE DATA IS
  FLOWING (results streaming, or a keep-alive heartbeat enabled). A read blocked
  on a fully SILENT stream (no results and heartbeat disabled) cannot be
  interrupted through the synchronous `requests` library from another thread and
  is not reliably cancellable; this is documented and close() stays non-blocking
  so it cannot deadlock the caller.
- Server side: close() releases the client connection; it does NOT by itself
  cancel the server-side work. StreamingChatHandler flushes headers before
  entering service.streaming_chat() and has no on_connection_close hook, so it
  ends the request only when it next detects the dropped connection (next result
  or heartbeat flush). Prompt while results/heartbeat flow, not guaranteed
  otherwise. True server-side disconnect cancellation is tracked as a follow-up
  (issue #1344).

## Tests
tests/neuro_san/session/close_contract/ (one class per file):
- Contract unit tests (no server, no LLM key): is_closed() False initially and
  True after close(); close() idempotent; streaming_chat()/function()/
  connectivity() raise AgentSessionClosedError after close(); interface default
  is_closed() is False.
- Transport-level integration tests using a real in-process server
  (StreamingCancelServer + CancelTestHandler on their own thread; state
  synchronized with threading.Events, no sleeps):
    - sync: a data-flowing post-header read is interrupted by close() from another
      thread -> streaming_chat() raises AgentSessionClosedError.
    - async: post-header (response closed via call_soon_threadsafe) and pre-header
      (ClientSession closed via run_coroutine_threadsafe) are both aborted by
      close() from another thread -> AgentSessionClosedError.
- A scoped conftest overrides the project-wide OPENAI_API_KEY autouse skip for
  this directory only, leaving sibling session tests unaffected.
- 17 tests pass; every wait/join is timeout-bounded so a regression fails fast
  rather than hanging. flake8 clean and pylint 10.00/10 across the changed files.

Note: writing the transport tests uncovered that the sync `requests` client
cannot interrupt a read blocked on a fully silent stream from another thread (and
that response.close() could itself block there). That is the reason sync close()
is now non-blocking and its contract is narrowed as described above.

## Also in this branch
- Enable streaming for the MCP service agent request (and a follow-on syntax fix).

## Related issue
NS-1338 (follow-up: #1344 -- server-side disconnect cancellation)


neuro_san/session/http_service_agent_session.py
```diff
@@ -93,6 +114,14 @@ def streaming_chat(self, request_dict: Dict[str, Any]) -> Generator[Dict[str, An
                                timeout=self.streaming_timeout_in_seconds) as response:
                 response.raise_for_status()
 
+                # Register this response so close() (possibly from another thread)
+                # can drop the connection and unblock the read loop below.
+                # If the session was already closed, abort before consuming any stream.
+                with self._stream_lock:
+                    if self._closed:
+                        return
+                    self._active_response = response
```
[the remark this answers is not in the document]
andreidenissov-cog: Fixed.

neuro_san/interfaces/agent_session.py
```diff
@@ -67,3 +67,18 @@ def streaming_chat(self, request_dict: Dict[str, Any]) -> Generator[Dict[str, An
...
         """
         raise NotImplementedError
+
+    def close(self):
+        """
+        Close the session, releasing any underlying client connection.
+
+        For remote sessions this drops the connection to the service so that an
+        in-flight streaming_chat() request is aborted (the service detects the
+        client disconnect and terminates the corresponding server-side work).
```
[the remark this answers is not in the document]
andreidenissov-cog: Addressed.

neuro_san/session/mcp_service_agent_session.py
```diff
@@ -211,11 +222,25 @@ def streaming_chat(self, request_dict: Dict[str, Any]) -> Generator[Dict[str, An
...
+        # Check if the session has been closed already before starting the streaming request
+        with self._stream_lock:
+            if self._closed:
+                return
+
         path: str = self.get_request_path("streaming_chat")
         try:
             with requests.post(path, json=mcp_payload, headers=headers,
-                               timeout=self.streaming_timeout_in_seconds) as response:
+                               stream=True, timeout=self.streaming_timeout_in_seconds) as response:
```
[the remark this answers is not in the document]
andreidenissov-cog: Addressed.

neuro_san/session/http_service_agent_session.py
```diff
@@ -135,3 +171,44 @@ def streaming_chat(self, request_dict: Dict[str, Any]) -> Generator[Dict[str, An
 
         except Exception as exc:  # pylint: disable=broad-exception-caught
```
vince-leaf, line 181: After close() is called from another thread, the read fails with AttributeError: 'NoneType' object has no attribute 'read', which this except turns into ValueError(help_message) — the generic connectivity-debug text. So a deliberate close() looks exactly like a lost connection to the caller. Verified with a local streaming HTTP server.

Maybe checking self._closed here and returning quietly (or raising a dedicated exception) so callers can tell the two apart.

neuro_san/session/http_service_agent_session.py
```diff
@@ -87,12 +108,27 @@ def streaming_chat(self, request_dict: Dict[str, Any]) -> Generator[Dict[str, An
         separator: bytes = b"\n"
         max_chunk_size: int = 64 * 1024
         path: str = self.get_request_path("streaming_chat")
+
+        # Abort before issuing the request if close() has already been called
+        # (e.g. the caller cancelled before first iterating this generator).
+        with self._stream_lock:
+            if self._closed:
+                return
```
vince-leaf: Once close() has been called, _closed is never reset, so every later streaming_chat() hits this return and yields nothing — no error, just an empty stream. A caller reusing the session would see it as "agent returned no messages." Meanwhile, function() and connectivity() still work on a closed session, which is inconsistent.

Suggest raising here (or providing is_closed()) and documenting that a session is single-use after close().

neuro_san/session/http_service_agent_session.py
andreidenissov-cog: Addressed.

neuro_san/session/http_service_agent_session.py
andreidenissov-cog: Should be addressed.

neuro_san/session/async_http_service_agent_session.py
```diff
@@ -18,24 +18,69 @@
...
 
+    def is_closed(self) -> bool:
+        """:return: True if close() has been called on this session."""
+        with self._stream_lock:
+            return self._closed
+
+    def _raise_if_closed(self):
+        """Raise AgentSessionClosedError if this session has been closed."""
+        if self.is_closed():
+            raise AgentSessionClosedError("async HTTP agent session has been closed")
```
d1donlydfink: Bad style to have new methods before the constructor.
The first thing anyone looks at when they look at a source file is the constructor to satisfy the question "What does this thing need?"

neuro_san/session/async_http_service_agent_session.py
```diff
@@ -167,5 +249,72 @@ async def streaming_chat(self, request_dict: Dict[str, Any]) -> Generator[Dict[s
...
             raise ValueError(self.help_message(path)) from exc
+        finally:
+            # The request is over (normally, by error, or because close() aborted
+            # it); stop tracking it.
+            with self._stream_lock:
+                self._active_response = None
+                self._active_session = None
+                self._active_loop = None
+
+    def close(self):
```
d1donlydfink, line 270: Should the neuro-san internals that use sessions (like ExternalActivation via AsyncAgentSessionFactory) now be using close()?  Doesn't have to be this PR, but soon, probably.

neuro_san/session/async_http_service_agent_session.py
```diff
@@ -167,5 +249,72 @@ async def streaming_chat(self, request_dict: Dict[str, Any]) -> Generator[Dict[s
...
+        on its next result or heartbeat flush. That is prompt while results stream
+        (or a keep-alive heartbeat is enabled), but a request producing no output
+        with heartbeat and request-timeout disabled may keep running server-side
+        until it does.
+        """
+        with self._stream_lock:
+            self._closed = True
+            response: Optional[ClientResponse] = self._active_response
+            session: Optional[ClientSession] = self._active_session
+            loop: Optional[asyncio.AbstractEventLoop] = self._active_loop
```
d1donlydfink: Please type the variables outside the with- block. This facilitates transpiling.

neuro_san/session/http_service_agent_session.py
```diff
@@ -18,20 +18,57 @@
...
 
+    def is_closed(self) -> bool:
+        """:return: True if close() has been called on this session."""
+        with self._stream_lock:
+            return self._closed
+
+    def _raise_if_closed(self):
+        """Raise AgentSessionClosedError if this session has been closed."""
+        if self.is_closed():
+            raise AgentSessionClosedError("HTTP agent session has been closed")
```
d1donlydfink: Similar deal as above: These are not the most important methods for a first view. Take them to the bottom of the file.

neuro_san/session/http_service_agent_session.py
```diff
@@ -134,4 +186,51 @@ def streaming_chat(self, request_dict: Dict[str, Any]) -> Generator[Dict[str, An
...
+        directly cancel the server-side work. The neuro-san service ends the
+        corresponding request only when it next detects the dropped connection --
+        on its next result or heartbeat flush. That is prompt while results stream
+        (or a keep-alive heartbeat is enabled), but a request producing no output
+        with heartbeat and request-timeout disabled may keep running server-side
+        until it does.
+        """
+        with self._stream_lock:
+            self._closed = True
+            response: Optional[requests.Response] = self._active_response
```
d1donlydfink: Similar deal with typing before the with- block.

d1donlydfink (changes requested):

neuro_san/session/async_http_service_agent_session.py
d1donlydfink: Also: Does close() need to be asynchronous for asynchronous sessions?
(Maybe aclose() or async_close() if intended to mesh together with sync API?)

vince-leaf:
@andreidenissov-cog, 
Ok, close() was called from a different thread at t=2 s, while the server streams had been running for 6 s.
Tested:
- sync  HttpServiceAgentSession
- async AsyncHttpServiceAgentSession

vince-leaf:
@andreidenissov-cog, 
We still need to test the data_driven_tests_driver read timeout from HOCON and enforce close().
No. There is no branch or PR for it on the remote — I checked all remote branches, and none touch neuro_san/test/driver with a close() call.

neuro_san/session/async_http_service_agent_session.py
andreidenissov-cog: Moved.

neuro_san/session/async_http_service_agent_session.py
andreidenissov-cog: Yes - we'll need to look over sessions usage inside neuro-san. Some close()s would be probably needed.

neuro_san/session/http_service_agent_session.py
andreidenissov-cog: OK. Instructed Claude to do this refactoring and save the rule for later use.

tests/neuro_san/session/close_contract/conftest.py
```diff
@@ -0,0 +1,35 @@
...
+network call), so the project-wide skip does not apply. This conftest overrides
+that fixture with a no-op for this directory only, leaving sibling session tests
+unaffected.
+"""
+import pytest
+
+
+# pylint: disable=unused-argument
+@pytest.fixture(autouse=True)
+def configure_llm_provider_keys(request, monkeypatch):
```
d1donlydfink, line 48: Filename doesn't exactly say what's going on in this file.

neuro_san/session/http_service_agent_session.py
andreidenissov-cog: Moved.

tests/neuro_san/session/close_contract/conftest.py
andreidenissov-cog: conftest.py is a mandatory fixed name for pytest framework, the same as in other places with our tests.

neuro_san/session/async_http_service_agent_session.py
```diff
@@ -101,43 +138,76 @@ async def streaming_chat(self, request_dict: Dict[str, Any]) -> Generator[Dict[s
...
+        session: ClientSession = ClientSession(headers=self.get_headers(), timeout=timeout)
+        # Track the session BEFORE issuing the request so close() can abort it even
+        # while it is still awaiting response headers. If close() raced in between
+        # the check above and here, tear the session down and bail.
+        abort: bool = False
+        with self._stream_lock:
+            if self._closed:
+                abort = True
+            else:
+                self._active_session = session
```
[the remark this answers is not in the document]
andreidenissov-cog: Test added.

neuro_san/session/http_service_agent_session.py
```diff
@@ -87,12 +116,28 @@ def streaming_chat(self, request_dict: Dict[str, Any]) -> Generator[Dict[str, An
...
                                timeout=self.streaming_timeout_in_seconds) as response:
                 response.raise_for_status()
 
+                # Register this response so close() (possibly from another thread)
+                # can drop the connection and unblock the read loop below.
+                # If the session was already closed, abort before consuming any stream.
+                with self._stream_lock:
+                    if self._closed:
+                        raise AgentSessionClosedError("HTTP agent session has been closed")
+                    self._active_response = response
```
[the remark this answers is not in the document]
andreidenissov-cog: Test added, limitation documented.

tests/neuro_san/session/close_contract/conftest.py
d1donlydfink: ahh. Leave it to Python to exude clarity...

d1donlydfink (approved):

PR #1346, opened 2026-09-16
Fix multi-instance service shutdown

### LINKED ISSUE #1343: Fix Ctrl-C behavior for neuro-san multi-instance service.
When we are running neuro-san service in multi-instance mode, Ctrl-C (process SIGINT) does not terminate all running server processes. "Main" orchestrator process keeps running.
Also: when we directly kill "main" process, worker processes seem to keep running.

arnabnandy7:
## Summary

Fixes #1343 by making multi-instance service shutdown reliable when the supervisor receives `SIGINT` or `SIGTERM`, or exits unexpectedly.

## Changes

- Register termination handlers before spawning workers.
- Start workers in separate sessions so terminal signals are handled by the supervisor.
- Forward shutdown signals to all active workers.
- Guarantee worker termination from the supervisor’s cleanup path.
- Preserve conventional process exit codes for `SIGINT` and `SIGTERM`.
- Pass the supervisor PID to workers and monitor it to prevent orphaned processes.
- Handle workers that exit between polling and termination.
- Add focused tests for worker supervision and parent-process monitoring.

## Root Cause

The existing implementation registered signal handlers only after all workers had started. It also relied on `Popen` context-manager cleanup, which waits for child processes but does not terminate them.

Additionally, the supervisor and workers shared the terminal process group, allowing Ctrl-C to reach them concurrently. Workers had no way to detect an unexpectedly terminated supervisor.

## Verification

```
python -m unittest -v tests/neuro_san/service/main_loop/test_worker_supervisor.py
```


arnabnandy7:
@andreidenissov-cog FYI the fix

andreidenissov-cog (approved):
Looks OK.

PR #1354, opened 2026-09-16
Resolve reservation specs through NetworkConfigFilterChain

### LINKED ISSUE #1353: Reservation (temporary) networks bypass NetworkConfigFilterChain, so the global sly_data_schema never reaches the front man
## Summary

Agent networks deployed through the Reservations API (temporary networks) never go through `NetworkConfigFilterChain`. As a result the top-level `sly_data_schema` that a deployment's `llm_config.hocon` defines is never merged into the front man's `function.sly_data_schema`, and the `/function` endpoint for such a network does not advertise that the client must supply BYOK keys in `sly_data.llm_config`. Generic clients (MAUI, nsflow) therefore never send the key, and the first chat fails with:

```
No fully-specified LLM found in llm_config or fallbacks.
LLM operation for this agent requires at least one of the following set in sly_data.llm_config:
    anthropic_api_key
    google_api_key
    openai_api_key
```

Networks loaded from hocon files in the manifest are unaffected because `AgentNetworkRestorer` runs the filter chain.

Downstream report, with the schema-less `/function` output and screenshot: cognizant-ai-lab/neuro-san-studio#1424

## Reproduction

1. Run a server whose `llm_config.hocon` defines a global `sly_data_schema` with `"required": ["llm_config"]` (for example neuro-san-studio's `config/byok_llm_config.hocon`).
2. Deploy any network through a `Reservationist` (Agent Network Designer with `AGENT_NETWORK_DESIGNER_USE_RESERVATIONS=true` is the easiest way).
3. Call `GET /api/v1/<reservation_id>/function`.

Expected: `function.sly_data_schema.required` contains `"llm_config"`, as it does for every static network under the same `llm_config.hocon`.
Actual: `function` contains only `description` and `parameters`.

Minimal in-process check (spec shaped like the one the reservationist receives):

```python
from neuro_san.internals.graph.filters.network_config_filter_chain import NetworkConfigFilterChain

spec = {
    "llm_config": {"fallbacks": [{"class": "openai", "model_name": "gpt-5.2", "openai_api_key": "sly_data"}]},
    "sly_data_schema": {"type": "object",
                        "properties": {"llm_config": {"type": "object",
                                                      "properties": {"openai_api_key": {"type": "string"}}}},
                        "required": ["llm_config"]},
    "tools": [{"name": "front", "function": {"description": "front", "parameters": {"type": "object", "properties": {}}},
               "instructions": "x"}],
}
print("before:", sorted(spec["tools"][0]["function"]))
filtered = NetworkConfigFilterChain().filter_config(spec)
print("after: ", sorted(filtered["tools"][0]["function"]),
      filtered["tools"][0]["function"]["sly_data_schema"]["required"])
print("idempotent:", NetworkConfigFilterChain().filter_config(filtered) == filtered)
```

```
before: ['description', 'parameters']
after:  ['description', 'parameters', 'sly_data_schema'] ['llm_config']
idempotent: True
```

`ExpiringAgentNetworkStorage` stores and serves the "before" shape.

## Root cause

The merge lives in `DefaultsConfigFilter` ([defaults_config_filter.py#L56-L60](https://github.com/cognizant-ai-lab/neuro-san/blob/9383d5dfc57231e34dd45b8157a781e32bd55657/neuro_san/internals/graph/filters/defaults_config_filter.py#L56-L60)): copy top-level `sly_data_schema` to `function.sly_data_schema`, front man only, union `required`. This is the documented contract for the global key ([agent_hocon_reference.md#L513-L526](https://github.com/cognizant-ai-lab/neuro-san/blob/9383d5dfc57231e34dd45b8157a781e32bd55657/docs/agent_hocon_reference.md#L513-L526)).

The chain is only invoked from:

- [agent_network_restorer.py#L125](https://github.com/cognizant-ai-lab/neuro-san/blob/9383d5dfc57231e34dd45b8157a781e32bd55657/neuro_san/internals/graph/persistence/agent_network_restorer.py#L125), the hocon-file path.
- [parameters_schema_network_validator.py#L75](https://github.com/cognizant-ai-lab/neuro-san/blob/9383d5dfc57231e34dd45b8157a781e32bd55657/neuro_san/internals/validation/network/parameters_schema_network_validator.py#L75), which filters a copy for validation and discards it.

Every reservation path builds `AgentNetwork` from the raw spec:

- In-memory add: [expiring_agent_network_storage.py#L158](https://github.com/cognizant-ai-lab/neuro-san/blob/9383d5dfc57231e34dd45b8157a781e32bd55657/neuro_san/internals/network_providers/expiring_agent_network_storage.py#L158)
- Base-storage write persists the raw spec: [s3_reservations_writer.py#L184](https://github.com/cognizant-ai-lab/neuro-san/blob/9383d5dfc57231e34dd45b8157a781e32bd55657/neuro_san/service/watcher/temp_networks/s3/s3_reservations_writer.py#L184), [local_reservations_storage.py#L203-L217](https://github.com/cognizant-ai-lab/neuro-san/blob/9383d5dfc57231e34dd45b8157a781e32bd55657/neuro_san/service/watcher/temp_networks/local/local_reservations_storage.py#L203-L217)
- Base-storage read on another pod: [s3_reservations_reader.py#L122](https://github.com/cognizant-ai-lab/neuro-san/blob/9383d5dfc57231e34dd45b8157a781e32bd55657/neuro_san/service/watcher/temp_networks/s3/s3_reservations_reader.py#L122), [local_reservations_storage.py#L253](https://github.com/cognizant-ai-lab/neuro-san/blob/9383d5dfc57231e34dd45b8157a781e32bd55657/neuro_san/service/watcher/temp_networks/local/local_reservations_storage.py#L253)

`/function` returns the front man's `function` dict verbatim ([async_direct_agent_session.py#L125-L129](https://github.com/cognizant-ai-lab/neuro-san/blob/9383d5dfc57231e34dd45b8157a781e32bd55657/neuro_san/session/async_direct_agent_session.py#L125-L129)), so whatever was stored is what the client sees.

The network is otherwise usable because `llm_config` has its own runtime merge: [calling_activation.py#L96-L97](https://github.com/cognizant-ai-lab/neuro-san/blob/9383d5dfc57231e34dd45b8157a781e32bd55657/neuro_san/internals/graph/activations/calling_activation.py#L96-L97) overlays the network-level `llm_config` onto each agent at activation time. There is no equivalent runtime merge for `sly_data_schema`; the only place it happens is the filter chain.

## Proposed fix

Run `NetworkConfigFilterChain` on each reservation spec in `ExpiringAgentNetworkStorage.add_reservations`, **before** the `self.base_storage.add_reservations(...)` call at [expiring_agent_network_storage.py#L141](https://github.com/cognizant-ai-lab/neuro-san/blob/9383d5dfc57231e34dd45b8157a781e32bd55657/neuro_san/internals/network_providers/expiring_agent_network_storage.py#L141), not only before the in-memory `AgentNetwork(...)` at line 158. Filtering at that point means the JSON written to S3/local/Azure storage is already resolved, so other pods that hydrate through `get_agent_network_provider` and the base-storage readers serve the same schema. Filtering only the in-memory copy would leave multi-instance BYOK deployments broken.

Notes:

- The chain is idempotent on these specs (see output above), so also filtering on the read side is harmless if wanted as belt-and-braces.
- `DefaultsConfigFilter` copies `llm_config`, `max_execution_seconds`, etc. into every tool, so stored JSON and `AgentNetwork.get_size_in_bytes()` grow somewhat.
- Commondefs filters are no-ops for reservation specs produced by neuro-san-studio (the assembler strips `commondefs`), and `NameCorrectionConfigFilter` has already run on a copy during `validate_with`.
- `tests/neuro_san/internals/network_providers/test_expiring_agent_network_storage.py` is the natural home for a test asserting `function.sly_data_schema` on the front man after `add_reservations`.

## Versions

- Reported on neuro-san 0.6.94 (neuro-san-studio 0.3.19). That tag already has `NetworkConfigFilterChain`, the `sly_data_schema` default mapping, and the unfiltered construction at line 158.
- Reproduced on `main` at 9383d5dfc57231e34dd45b8157a781e32bd55657 (0.7.2).

Noravee:
## Summary

Agent networks deployed through the Reservations API (temporary networks) were stored and served as raw dictionaries, so the top-level defaults that hocon-file networks receive from `AgentNetworkRestorer.filter_config()` were never applied to them. Most visibly, a deployment's global `sly_data_schema` never reached the front man's `function.sly_data_schema`, so the `/function` endpoint of a temporary network did not tell clients that it needs BYOK keys in `sly_data.llm_config`. Generic clients therefore never sent a key and the first chat failed with `No fully-specified LLM found in llm_config or fallbacks.`

Fixes #1353
Downstream report: cognizant-ai-lab/neuro-san-studio#1424

## Changes

- `ResolvedNetworkConfigFilter` (new, `neuro_san/internals/graph/filters/`): a `ConfigFilter` that runs a spec through `NetworkConfigFilterChain`, strips the consumed `commondefs` block, and always hands back a copy (the chain returns the caller's own dict when a spec has no tools, and the S3 writer injects metadata into whatever it is given). Extracted from `ExpiringAgentNetworkStorage` after review so it is testable on its own and shared by the write side and both readers.
- `ExpiringAgentNetworkStorage.add_reservations()`: applies that filter to every spec in the batch **before** the base-storage write and builds the in-memory `AgentNetwork` from the same resolved dictionaries. Filtering ahead of the base-storage write is what lets other server instances that hydrate from S3/local storage serve the same resolved spec; filtering only the in-memory copy would leave multi-instance BYOK deployments broken. Stripping `commondefs` makes the persisted spec a fixed point of the chain: the commondefs filters resolve one substitution level per pass and never remove the block, so a spec that kept it would come out different on the pod that reads it back.
- `S3ReservationsReader` and `LocalReservationsStorage`: `get_one_reservation()` runs the same `ResolvedNetworkConfigFilter` before constructing `AgentNetwork`. This is a no-op for specs written by this code and covers objects written by older instances during a rolling upgrade or by other tooling (a legacy spec that still carries `commondefs` gets the substitution applied and the block dropped, exactly like the write side). The S3 reader now catches failures from the chain or from `AgentNetwork` construction and reports the reservation as absent, matching the local storage's existing policy, instead of letting a malformed foreign object raise onto the request path.
- `docs/agent_hocon_reference.md`: the global `sly_data_schema` section notes that the merge also applies to Reservations API networks.
- `NameCorrectionConfigFilter`: stored the corrected agent name under a key named after the old name (`tool[name] = new_name`) instead of updating `tool["name"]`, so an agent named `a/b` stayed registered as `a/b` while its callers referenced `a_b`. Pre-existing for hocon networks; surfaced by Copilot because this branch now routes reservation specs (write and read side) through the same filter. One-line fix.

## Behaviour notes

- Every `DefaultsConfigFilter` key now reaches the agents of a temporary network, not just `sly_data_schema`: `llm_config`, `verbose`, `max_steps`/`max_iterations`, `max_execution_seconds`, `max_attempts`, `error_formatter`, `error_fragments`. The runtime reads these per agent, so a top-level value in a reservation spec used to be silently ignored (only `llm_config` had a runtime overlay). For example, neuro-san-studio's deployable template sets a top-level `max_execution_seconds` of 600, which now applies to each agent instead of the 300s runtime default. This matches what the identical hocon file has always done.
- Persisted reservation JSON grows accordingly (per-agent `llm_config` copies, front-man `sly_data_schema`).

## Tests

- New `test_resolved_network_config_filter.py` (`TestResolvedNetworkConfigFilter`): seven direct tests for the filter: defaults merge (front-man-only `sly_data_schema`, per-agent `llm_config`/`max_execution_seconds`), `required`-list union with a front man's own schema, `commondefs` substitution and strip, whole-dict non-mutation of the input (including the in-place slash-name correction), a copy returned even with no tools, output is a fixed point (chained/order-dependent `commondefs` spec filtered twice), and `None` passthrough.
- `test_expiring_agent_network_storage.py`: three new integration tests covering the front-man merge as seen through the stored `AgentNetwork`, the resolved spec handed to the base storage (and shared with the in-memory network), and whole-dict non-mutation of the caller's spec through a metadata-injecting base storage. Two new helper modules alongside: `ByokAgentSpecBuilder` and a `RecordingReservationsStorage` test double.
- `test_s3_reservations_reader.py` (`TestS3ReservationsReader`): reading a raw spec directly from the bucket yields the merged schema; a legacy spec with `commondefs` comes back substituted and stripped; chain-breaking specs yield `(None, None)` instead of raising.
- `test_local_reservations_storage.py` (`TestLocalReservationsStorage`): a raw file read back resolves in memory and the file is left exactly as written; a legacy file with `commondefs` comes back substituted and stripped.
- `test_round_trip.py`: the `tools` assertion now compares against the resolved form of the written spec, since the reader resolves on read.
- New `test_name_correction_config_filter.py` (`TestNameCorrectionConfigFilter`): first coverage for the filter; the slash-correction test fails on main's implementation and passes here.
- Scoped `conftest.py` in `tests/neuro_san/internals/network_providers/` and `tests/neuro_san/service/watcher/temp_networks/local/`, mirroring the existing s3 one, so these LLM-free suites run instead of skipping when `OPENAI_API_KEY` is unset.

```
flake8 / pylint --rcfile=.pylintrc on touched files: clean / 10.00
pymarkdown --config .pymarkdownlint.yaml on the changed doc: clean
pytest graph/filters + network_providers + temp_networks: 93 passed, 10 subtests passed
pytest temp_networks with OPENAI_API_KEY unset: 46 passed, 0 skipped
pytest tests/neuro_san --ignore=tests/neuro_san/zzz_hocons (dummy key): 646 passed, 2 skipped
```


neuro_san/internals/network_providers/expiring_agent_network_storage.py
```diff
@@ -128,6 +131,49 @@ def is_expired(reservation: Reservation) -> bool:
         """
         return time.time() > reservation.get_expiration_time_in_seconds()
 
+    @staticmethod
+    def filter_reservations(reservations_dict: Dict[Reservation, Dict[str, Any]]) \
```
Noravee: This is the core of the fix. Reservation specs are plain dicts. Unlike hocon files they were stored unresolved: validation ran `NetworkConfigFilterChain` on them but dropped the result. Running it here applies commondefs substitution, copies top-level defaults onto each agent, merges the global `sly_data_schema` into the front man, and corrects names.

neuro_san/internals/network_providers/expiring_agent_network_storage.py
```diff
@@ -140,17 +186,26 @@ async def add_reservations(self, reservations_dict: Dict[Reservation, Dict[str,
...
             return
 
+        # Resolve every spec through the standard filter chain *before* anything is stored.
+        # Doing this ahead of the base storage write matters for multi-instance deployments:
+        # other instances hydrate temporary networks from the base storage (S3, local files, ...),
+        # so the persisted spec has to be the resolved one or those instances would serve a
+        # network that never got its defaults (e.g. the front man's sly_data_schema).
+        # The same resolved dictionaries are then used for the in-memory AgentNetworks below,
+        # so any metadata a writer adds to them stays consistent with what is served here.
+        use_reservations: Dict[Reservation, Dict[str, Any]] = self.filter_reservations(reservations_dict)
```
Noravee: Filtering happens *before* the base-storage write on purpose. Other pods load the spec from base storage, not from this pod's memory. Storing the resolved spec makes the read-side filter pass a no-op, and the in-memory network shares the dict the S3 writer stamps metadata into.

neuro_san/internals/network_providers/expiring_agent_network_storage.py
```diff
@@ -128,6 +131,49 @@ def is_expired(reservation: Reservation) -> bool:
...
+        """
+        filter_chain: ConfigFilter = NetworkConfigFilterChain()
+        filtered: Dict[Reservation, Dict[str, Any]] = {}
+        for reservation, agent_spec in reservations_dict.items():
+            resolved: Dict[str, Any] = filter_chain.filter_config(agent_spec)
+            if resolved is agent_spec:
+                # The chain only copies when there are tools to work on; for anything else it
+                # hands back the caller's own dict.  Copy it ourselves so that neither the
+                # commondefs removal below nor a metadata-injecting writer touches the caller's object.
+                resolved = deepcopy(agent_spec)
```
Noravee: The chain does not always copy. A spec with no tools and no `commondefs` comes back as the caller's own dict, and `S3ReservationsWriter` then writes `metadata` straight into it. Copying here keeps the docstring's promise that the caller's dict is never touched.

neuro_san/internals/network_providers/expiring_agent_network_storage.py
```diff
@@ -128,6 +131,49 @@ def is_expired(reservation: Reservation) -> bool:
...
+            if isinstance(resolved, dict):
+                resolved.pop("commondefs", None)
```
Noravee: The commondefs filters resolve one substitution level per pass and never remove the `commondefs` block. If the stored spec kept it, the S3 and local readers would change it again when they re-run the chain on read: `hello world {punct}` becomes `hello world !`. Dropping the consumed block makes that second pass a no-op. `test_filter_reservations_is_idempotent` pins this.

neuro_san/service/watcher/temp_networks/s3/s3_reservations_reader.py
```diff
@@ -118,8 +121,28 @@ def get_one_reservation(self, reservation_id: str) -> Tuple[Optional[Reservation
...
+            # writers, which would otherwise be served without their top-level defaults
+            # applied (e.g. the front man's sly_data_schema).
+            #
+            # Those foreign objects never went through the reservationist's validators, so
+            # the chain (or AgentNetwork itself) can raise on a malformed shape.  Treat that
+            # like any other unreadable object - log it and report the reservation as absent -
+            # rather than letting the exception escape onto the request path.  This mirrors
+            # LocalReservationsStorage.get_one_reservation().
+            try:
+                resolved_spec: Dict[str, Any] = NetworkConfigFilterChain().filter_config(agent_spec)
```
Noravee: Read-side safety net. Normally a no-op, because `filter_reservations` already resolved the spec before it was written. It matters during a rolling upgrade (an older pod writes a raw spec, a newer pod reads it) and for objects written by other tooling.

Objects from other tooling never went through the reservationist's validators, so the chain can raise on a malformed shape. The `except` below reports the reservation as absent, as `LocalReservationsStorage.get_one_reservation` already does. Without it the exception would escape `get_agent_network_provider` and surface as a 500 on the request path. Pinned by `test_s3_reservations_reader.py`.

neuro_san/service/watcher/temp_networks/local/local_reservations_storage.py
```diff
@@ -250,7 +251,14 @@ def get_one_reservation(self, obj_key: str) -> Tuple[Reservation, AgentNetwork]:
...
                 self.logger.debug("%s: reservation %s is expired", self._name, reservation_id)
                 return None, None
-            agent_network = AgentNetwork(agent_spec, reservation.get_reservation_id())
+            # Specs written through ExpiringAgentNetworkStorage.filter_reservations() are already
+            # resolved with their commondefs stripped, so for those this second pass is a no-op;
+            # it only changes files written by older instances or by other tooling.  See
+            # filter_reservations() for why an unresolved spec is a problem.  Any exception the
+            # chain raises on a malformed foreign spec lands in the broad except below and is
+            # reported as "not present", like every other reconstruction failure.
+            resolved_spec: Dict[str, Any] = NetworkConfigFilterChain().filter_config(agent_spec)
```
Noravee: Same filter-chain pass as the S3 reader. The existing `except Exception` below already returns `(None, None)` on any failure, so no new try/except was needed.

neuro_san/internals/graph/filters/name_correction_config_filter.py
```diff
@@ -69,8 +69,10 @@ def filter_config(self, basis_config: Dict[str, Any]) \
                 # Add to the errors
                 errors.append(new_name)
             elif new_name != name:
-                # Make a correction
-                tool[name] = new_name
+                # Make a correction to the tool's own "name" field.  The references other
+                # tools hold to it are rewritten below from the corrections table, so both
+                # sides of every edge end up using the corrected name.
+                tool["name"] = new_name
```
Noravee, line 75: Pre-existing bug, one line. This used to be `tool[name] = new_name`, which stored the corrected name under a key named after the *old* name and left `tool["name"]` alone. The references below were still rewritten, so callers pointed at `a_b` while the registry still had `a/b`. Fixed here because reservation specs now pass through this filter too. The reservationist's validators already reject `/` in names, so in practice this mattered for hocon networks.

tests/neuro_san/service/watcher/temp_networks/s3/test_round_trip.py
```diff
@@ -106,10 +110,16 @@ async def test_add_then_get_returns_equivalent_reservation(self):
...
         )
+        # The reader resolves the spec through NetworkConfigFilterChain before building the
+        # AgentNetwork (see S3ReservationsReader.get_one_reservation), so the tools come back
+        # with the top-level defaults applied - here the network-level llm_config copied onto
+        # the front man. Compare against the resolved form of what was written, which is what
+        # any client of the storage is meant to receive.
+        resolved_spec: Dict[str, Any] = NetworkConfigFilterChain().filter_config(agent_spec)
         self.assertEqual(
-            agent_spec.get("tools"),
+            resolved_spec.get("tools"),
```
Noravee, line 122: Only existing test whose assertions change. `S3ReservationsReader` now resolves what it reads, so the tools come back with the network-level `llm_config` copied onto each tool, here just the front man. Comparing against the resolved input keeps the check meaningful. Top-level `name` and `llm_config` are still compared raw.

This module, like most of `s3/`, is named after a behaviour rather than the class it tests. Folding the legacy modules into one test module per source class is tracked in #1356 and deliberately left out of this PR.

tests/neuro_san/internals/network_providers/conftest.py
```diff
@@ -0,0 +1,40 @@
...
+project-wide skip condition does not apply. This conftest overrides that
+fixture with a no-op for tests in this subtree only, mirroring
+tests/neuro_san/service/watcher/temp_networks/s3/conftest.py.
+"""
+import pytest
+
+
+# pylint: disable=unused-argument
+@pytest.fixture(autouse=True)
+def configure_llm_provider_keys(request, monkeypatch):
```
Noravee, line 34: Same override the s3/ tests already had. The root `tests/conftest.py` skips every unmarked test when `OPENAI_API_KEY` is unset; nothing in this directory calls an LLM. Without this file, all 28 tests here and all 13 in `local/` skip in keyless runs, including fork PRs in CI. With it: 41 passed, 0 skipped.

Noravee (commented):
**TL;DR**

**Problem.** Networks deployed through the Reservations API ("temporary networks", e.g. what Agent Network Designer builds in reservations mode) were stored and served as the deployer handed them over. Hocon networks go through `NetworkConfigFilterChain` when loaded from disk. Reservation specs were only validated against it; the raw spec was kept. So the deployment-wide `sly_data_schema` from the studio's `llm_config.hocon` never landed on the front man, `/function` never told the client to send its BYOK key in `sly_data.llm_config`, and the first chat failed with *No fully-specified LLM found*. Reported as cognizant-ai-lab/neuro-san-studio#1424.

**Fix.** Run the same filter chain on reservation specs:
1. In `ExpiringAgentNetworkStorage.add_reservations`, before the spec is written to the base storage (S3/local), so every instance sees the resolved spec.
2. Again when the S3 and local readers load a spec, so objects written by older instances still get their defaults.

**What else changes.** Other defaults only the filter chain applies (`max_execution_seconds`, `verbose`, ...) now reach temporary networks too. Stored JSON grows a bit. Also fixes a latent bug in `NameCorrectionConfigFilter` (`tool[name]` should have been `tool["name"]`), which reservation specs now pass through.

**Reading order.** `expiring_agent_network_storage.py`, then `s3_reservations_reader.py`, `local_reservations_storage.py`, `name_correction_config_filter.py`, then the tests.

neuro_san/internals/network_providers/expiring_agent_network_storage.py
```diff
@@ -128,6 +131,49 @@ def is_expired(reservation: Reservation) -> bool:
...
+        :param reservations_dict: A mapping of Reservation -> agent network spec as deployed
+        :return: A new mapping with the same Reservation keys whose values are the fully resolved
+                 agent network specs.  Every value is a copy: the caller's dictionaries are never
+                 returned or mutated, so a base-storage writer that injects metadata into what it
+                 is given (S3ReservationsWriter does) cannot reach them either.
+        """
+        filter_chain: ConfigFilter = NetworkConfigFilterChain()
+        filtered: Dict[Reservation, Dict[str, Any]] = {}
+        for reservation, agent_spec in reservations_dict.items():
+            resolved: Dict[str, Any] = filter_chain.filter_config(agent_spec)
```
d1donlydfink: This process itself should be its own new FilterConfig, separately and singly testable from ExpiringAgentNetworkStorage

d1donlydfink (changes requested):
Make the new process the tis part of ExpiringAgentNetworkStorage its own FilterConfig. Then it can be separately and singly testable and re-shared wherever we need it.

neuro_san/internals/graph/filters/name_correction_config_filter.py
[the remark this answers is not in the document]
Noravee: Agreed that `args.tools` is not rewritten, but that gap predates this PR: the filter has never touched `args`, so hocon networks on `main` have the same behaviour, and the deploy path for reservations rejects slash names before the filter runs. It is unrelated to the fix here, so it is tracked separately in #1358 rather than folded into this PR.

Noravee:
**What changed after the review**

- New `ResolvedNetworkConfigFilter` in `neuro_san/internals/graph/filters/`. It runs `NetworkConfigFilterChain`, copies when the chain hands back the input object, and drops the consumed `commondefs` block. The rationale that used to live on the static method is now its class docstring.
- `ExpiringAgentNetworkStorage.filter_reservations()` is gone. `add_reservations()` applies the filter to each spec in the batch before the base-storage write, same behaviour as before.
- `S3ReservationsReader` and `LocalReservationsStorage` now use the same filter on the read side instead of the bare chain, so a legacy spec read back also loses its `commondefs` block and every instance serves the same network.

**Tests**

- New `test_resolved_network_config_filter.py` with seven direct tests: defaults merge, `required`-list union, `commondefs` substitution and strip, input never mutated, copy returned without tools, output is a fixed point, `None` passthrough. It builds its own spec and imports nothing from other test packages.
- The two `filter_reservations` tests moved there; the storage-level union test that had become an exact duplicate was dropped. The base-storage write and caller non-mutation tests stay as integration coverage.
- One legacy-`commondefs` read test per reader. Both fail if the readers use the bare chain.

neuro_san/internals/network_providers/expiring_agent_network_storage.py
Noravee: This code moved into a new `ResolvedNetworkConfigFilter` (`neuro_san/internals/graph/filters/resolved_network_config_filter.py`), a `ConfigFilter` with its own tests in `test_resolved_network_config_filter.py`. `add_reservations()` now just applies that filter to each spec, and the S3 and local readers reuse the same filter on the read side.

neuro_san/service/watcher/temp_networks/s3/s3_reservations_reader.py
```diff
@@ -118,8 +121,27 @@ def get_one_reservation(self, reservation_id: str) -> Tuple[Optional[Reservation
...
+            # rather than letting the exception escape onto the request path.  This mirrors
+            # LocalReservationsStorage.get_one_reservation().
+            try:
+                resolved_spec: Dict[str, Any] = ResolvedNetworkConfigFilter().filter_config(agent_spec)
+                agent_network = AgentNetwork(resolved_spec, reservation.get_reservation_id())
+            except Exception as exception:  # pylint: disable=broad-exception-caught
+                sensitive_logger = SensitiveLogger(self.logger)
+                sensitive_logger.error("%s: Failed to reconstruct agent network for reservation %s: %s",
+                                       self.name, reservation_id, exception)
+                return None, None
```
d1donlydfink, line 144: Several million points out of 10 for using SenstiveLogger right off the bat for the exception!

d1donlydfink (approved):

Issue #1344, opened 2026-09-15
Server-side cancellation of streaming_chat on client disconnect


andreidenissov-cog:
## Summary
Add server-side cancellation of an in-flight `streaming_chat` request when the
client disconnects, so that dropping the client connection actually terminates
the corresponding server-side agent work.

## Background
PR #1340 added `AgentSession.close()` (and concrete session implementations) so a
client can drop its connection to abort a request it no longer wants (e.g. a
load-test driver cancelling a timed-out test). During review it was established
that `close()` reliably releases the *client* connection but does **not**
guarantee the *server* stops working. The `close()` docstrings were narrowed in
#1340 to state this; this issue tracks the real server-side fix.

## The gap
In `neuro_san/service/http/handlers/streaming_chat_handler.py`:
- Response headers are flushed immediately, before entering
  `service.streaming_chat()` (~lines 140–146).
- There is **no `on_connection_close` hook**, so a client disconnect is only
  detected on a *later* result or heartbeat flush (~lines 160–164).
- The keep-alive heartbeat is **off by default**
  (`DEFAULT_KEEP_ALIVE_INTERVAL_SECONDS = 0`), and a per-request timeout is not
  always set.

Consequently, if the service is awaiting a result indefinitely and produces no
output (heartbeat and request-timeout disabled), a client `close()` frees the
client socket but the server-side agent run continues to completion.

## Proposed fix
Give `StreamingChatHandler` a Tornado `on_connection_close()` hook (or equivalent
disconnect detection) that cancels the in-flight request work:
- Hold a reference to the task driving `post()` / the `service.streaming_chat()`
  async generator.
- On disconnect, cancel that task and `aclose()` the result generator so the
  underlying agent run is torn down promptly rather than only on the next flush.
- Ensure clean resource release (journals, executor tasks, `finish_client_request`
  accounting) on this cancellation path.

## Acceptance criteria
- Dropping the client connection mid-request terminates the server-side agent
  run promptly, even with the heartbeat disabled and no request timeout and no
  intermediate results being produced.
- No regression to the normal streaming path or to the existing
  results/heartbeat-based disconnect detection.
- Test coverage for the disconnect-during-indefinite-wait case.

## References
- PR #1340 (client-side `close()` API + narrowed contract)
- `neuro_san/service/http/handlers/streaming_chat_handler.py`
- `neuro_san/interfaces/agent_session.py` (`close()` contract)

PR #1359, opened 2026-09-17
NS-1338 Cancel server-side work when a load test times out (drop client connection)

andreidenissov-cog:
## Summary
Cancel a data-driven load test's server-side work when the test times out on the
client side. When a test exceeds its timeout, the driver now closes that test's
client connection(s), so the neuro-san service observes the disconnect and
terminates the still-running request instead of continuing work whose result is
discarded.

This is the consumer of the session close() API (added in #1340, now on main).
It closes the gap where a client-side timeout abandoned the future but the
server kept running the request.

## Background
DataDrivenTestsDriver runs test cases in parallel, one worker thread per test,
and uses as_completed_or_timeout() to give each future its own timeout. On a
timeout the driver recorded the result and moved on -- but the worker thread was
still blocked reading the streaming response, and nothing dropped the connection,
so the server kept computing. The session close() API now makes that connection
droppable from another thread; this change wires it in.

## Changes
- New SessionCanceller (neuro_san/test/driver/session_canceller.py): a thread-safe
  handle for aborting one test's in-flight work from another thread. A worker
  registers the AgentSession(s) it creates; the driver's main thread calls
  cancel() to close them all, dropping their connections. There is one canceller
  per test (per future), but a test can open more than one session -- one per
  entry in its "connections" list -- so register() accumulates a list and cancel()
  closes them all. It also handles the race where cancel() fires before a session
  is registered (that session is closed immediately on registration).
- DataDrivenTestsDriver: create a SessionCanceller per test and map each future to
  it; thread the canceller through capture_one_iteration() -> one_iteration(),
  where each created session registers itself; and on a timed-out future call
  its canceller.cancel() so the connection is dropped and the server terminates
  the request.
- Early-exit cancellation: once num_need_success is reached the loop breaks with
  other tests possibly still running. executor.shutdown(cancel_futures=True) only
  drops futures that never started, so the finally block now also calls cancel()
  on every still-running future's canceller, dropping those connections too.
  SessionCanceller.cancel() is idempotent, so cancelling once on timeout and again
  on shutdown closes each session only once.

## Only cancellable sessions are registered
Only HTTP sessions (sync HttpServiceAgentSession and async
AsyncHttpServiceAgentSession) have a close() that actually drops an in-flight
connection. Direct sessions run in-process and MCP is a request/response transport
whose close() is a no-op, so registering them would be a false promise -- their
server-side work runs to completion regardless. The driver now registers a session
only when it is an AbstractHttpServiceAgentSession. A timed-out direct/MCP test is
simply recorded as a timeout; nothing pretends its connection was dropped, and the
connect_timeout_in_seconds passed to those sessions is unaffected.

## Notes
- Cancellation is only as strong as the underlying session.close() (see #1340):
  reliable for the async client and for the sync client while data/heartbeat is
  flowing; server-side termination then follows once the service detects the
  disconnect. Server-side disconnect cancellation independent of a flush is
  tracked separately (issue #1344).

## Testing
- New deterministic unit tests for the SessionCanceller synchronization boundary
  (tests/neuro_san/test/driver/test_session_canceller.py), using a thread-safe
  mock session that counts close() calls:
  - normal multi-session path: sessions registered before cancel() are each
    closed exactly once;
  - cancel-before-register race: a session registered after cancel() has fired is
    closed immediately on registration, not left running;
  - mixed ordering (some registered before cancel, one after) closes every
    session exactly once;
  - thread-safety invariant: many threads calling register() while the main
    thread calls cancel() still closes each session exactly once;
  - idempotency: cancel() called more than once closes each session only once.
  A directory conftest overrides the project-wide provider-key skip so these
  LLM-free tests run under the default unit filter.
- flake8 clean and pylint 10.00/10 on the changed files.

## Related issue
NS-1338


neuro_san/test/driver/data_driven_tests_driver.py
```diff
@@ -251,6 +267,10 @@ def one_iteration(self, test_case: Dict[str, Any], asserts: AssertForwarder,
                     use_direct=use_direct,
                     metadata=metadata,
                     connect_timeout_in_seconds=timeout_in_seconds)
+            # Register the session so a timeout on the main thread can close it
+            # (dropping the connection) even while this worker thread is blocked
+            # reading the streaming response.
+            canceller.register(session)
```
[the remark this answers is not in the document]
andreidenissov-cog: Addressed.

neuro_san/test/driver/data_driven_tests_driver.py
```diff
@@ -251,6 +267,24 @@ def one_iteration(self, test_case: Dict[str, Any], asserts: AssertForwarder,
...
+                # connection (McpServiceAgentSession.close() is a no-op). A timed-out
+                # MCP test cannot drop the connection, so the server-side tool runs
+                # to completion -- cancellation-on-timeout is unsupported for MCP.
+                # Make that explicit instead of registering an un-cancellable session
+                # and pretending cancel() drops its connection.
+                asserts.assertIsNone(
+                    timeout_in_seconds,
+                    "connection 'mcp' does not support cancellation on timeout: a "
+                    "timed-out MCP request runs the server-side tool to completion. "
+                    "Remove timeout_in_seconds or use a cancellable connection.")
```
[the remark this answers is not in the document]
andreidenissov-cog: Should be addressed.

neuro_san/test/driver/session_canceller.py
```diff
@@ -0,0 +1,79 @@
...
+        Register a session created by the worker thread so a later cancel() can
+        close it. If cancel() already fired (a race where the timeout beat the
+        session's creation), close this session immediately instead.
+
+        :param session: The AgentSession to track for cancellation.
+        """
+        with self._lock:
+            if not self._cancelled:
+                self._sessions.append(session)
+                return
```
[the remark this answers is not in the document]
andreidenissov-cog: Tests added.

d1donlydfink (dismissed):
Cool

vince-leaf (changes requested):
The new MCP guard asserts that timeout_in_seconds is None for connections: ["mcp"], but tests/fixtures/music_nerd_pro/combination_responses_with_history_mcp.hocon still sets timeout_in_seconds: 60 for connections: ["mcp"] and is active in test_smoke_test_hocons.py. That test will now fail unconditionally when run with an OpenAI key (CI is green only because it skips when no key is present). Please update the fixture(s) or the guard. Note that timeout_in_seconds is also passed as connect_timeout_in_seconds, so simply removing it changes connect behavior.

Confirmed. Ran the PR branch's driver against tests/fixtures/music_nerd_pro/combination_responses_with_history_mcp.hocon with the MCP session and LLM interaction mocked to succeed. Result: 1 assert failure —

60 is None: connection 'mcp' does not support cancellation on timeout: a timed-out MCP request runs the server-side tool to completion. Remove timeout_in_seconds or use a cancellable connection.

neuro_san/test/driver/data_driven_tests_driver.py
```diff
@@ -251,6 +267,24 @@ def one_iteration(self, test_case: Dict[str, Any], asserts: AssertForwarder,
...
+            else:
+                # Register the session so a timeout on the main thread can close it
+                # (dropping the connection) even while this worker thread is blocked
+                # reading the streaming response.
+                canceller.register(session)
```
vince-leaf, line 302: This also registers DirectAgentSession (the default use_direct / "direct" path). DirectAgentSession.close() calls finish_request() and sets invocation_context = None while streaming_chat() on the worker thread may still be using it, and it doesn't cancel the background task — so a timeout can tear down the context under a live in-process request. Suggest only registering http/async sessions (or add a synchronized direct-session cancel first).

neuro_san/test/driver/data_driven_tests_driver.py
```diff
@@ -133,6 +139,10 @@ def run_tests(self, tests: Sequence[Dict[str, Any]], num_need_success: int) -> L
                 test_index: Optional[int] = future_indices.get(fut)
                 if timed_out:
                     # This test iteration has timed out.
+                    # Close its client connection(s) so the neuro-san service sees
+                    # the disconnect and terminates the still-running server-side
+                    # request instead of finishing work whose result we discard.
+                    future_cancellers[fut].cancel()
```
vince-leaf, line 145: Cancellers are only called for timed-out futures. When num_need_success is reached, the loop breaks, and executor.shutdown(wait=False, cancel_futures=True) leaves already-running futures and their server requests alive. Please also call cancel() on the remaining future_cancellers in that early-exit path (e.g., in the finally block).

neuro_san/test/driver/data_driven_tests_driver.py
```diff
@@ -251,6 +267,24 @@ def one_iteration(self, test_case: Dict[str, Any], asserts: AssertForwarder,
...
                     connect_timeout_in_seconds=timeout_in_seconds)
+
+            if connection.lower() == "mcp":
+                # MCP is a request/response transport with no abortable in-flight
+                # connection (McpServiceAgentSession.close() is a no-op). A timed-out
+                # MCP test cannot drop the connection, so the server-side tool runs
+                # to completion -- cancellation-on-timeout is unsupported for MCP.
+                # Make that explicit instead of registering an un-cancellable session
+                # and pretending cancel() drops its connection.
+                asserts.assertIsNone(
```
vince-leaf: Nit (same MCP guard)
Confirm that the assertIsNone failure recorded on the worker's TimedAssertCapture actually surfaces in the final report and isn't overwritten by the timeout/exception capture created in run_tests().

neuro_san/test/driver/data_driven_tests_driver.py
andreidenissov-cog: That's a good catch.

neuro_san/test/driver/data_driven_tests_driver.py
andreidenissov-cog: Should be addressed now.

neuro_san/test/driver/data_driven_tests_driver.py
andreidenissov-cog: Should be addressed now.

neuro_san/test/driver/data_driven_tests_driver.py
andreidenissov-cog: This assertIsNone is removed now.

vince-leaf (approved):
The PR is fine to approve as-is; the two leftovers are optional nits which I posted.

PR #1363, opened 2026-09-18
Consolidate temp_networks s3/ and local/ tests into one module per source class

### LINKED ISSUE #1356: Consolidate temp_networks s3/ and local/ tests into one test module per source class
## Summary

The reservations storage tests under `tests/neuro_san/service/watcher/temp_networks/s3/` and `.../local/` are organized one module per behaviour (`test_retry_on_throttling.py`, `test_local_reservations_storage_start.py`, ...) rather than one module per source class. PR #1319 finished the one-test-class-per-file step here (it split `TestCredentialRejectionGate` out of the reader recovery module and renamed two reader modules to `test_reader_*`); the `local/` modules were already one class per file since commit e95e08e2. This ticket takes the next step so the file and class are named after the source module and class they exercise, matching how most of `tests/neuro_san/` is laid out (`foo_bar.py` / `FooBar` -> `test_foo_bar.py` / `TestFooBar`).

PR #1354 (open at time of filing) adds the first modules in these directories whose file and class both mirror a source class (`test_s3_reservations_reader.py::TestS3ReservationsReader`, `test_local_reservations_storage.py::TestLocalReservationsStorage`) and, apart from an assertion update in `test_round_trip.py`, leaves the behaviour-named modules untouched so its bug-fix diff stays focused. This ticket is the follow-up and should land after #1354 merges; "(exists)" below refers to the post-#1354 tree.

## Current state

- `s3/`: 22 test modules. 21 are behaviour-named; 20 of those subclass `S3ReservationsStorageTestBase` and one (`test_credential_rejection_gate.py`) subclasses `TestCase`.
- `local/`: 5 test modules. 4 are behaviour-named bare pytest classes (no `TestCase` base). Between them they use the `tmp_path` fixture (all four), `monkeypatch` (constructor only), `pytest.raises` (constructor only) and `@pytest.mark.asyncio` (expiration and write_read).
- Two `s3/` modules define a second, private helper class inside the test module: `_TokenBoundS3Client` in `test_proactive_credential_refresh.py` and `_InvalidTokenClient` in `test_reader_invalid_token_recovery.py`. Each also builds that class in a nested factory closure (`create_token_bound_client`, `create_client_with_recovery`).
- A module-level function `_make_client_error()` is duplicated verbatim in `test_credential_rejection_gate.py:29` and `test_reader_invalid_token_recovery.py:29`.
- 27 test modules, 3,361 lines in total. No `test_*` method names collide within any of the merge groups below.

## Proposed layout

Group by the class that owns the code under test. `S3ReservationsStorage.add_reservations` / `get_one_reservation` / `expire_reservations` are one-line delegations to `self.writer` / `self.reader` / `self.expiration` (`s3_reservations_storage.py:121,134,141`), so a test that goes through the facade is really testing the component behind it. Grouping by "class whose method is called" instead would put 13 modules (about 1,300 lines after header dedup) into one `test_s3_reservations_storage.py`, over pylint's default `max-module-lines=1000`, which `.pylintrc` does not override. Do not add a `too-many-lines` disable to get around that.

| Target module | Legacy modules folded in | Raw lines today |
|---|---|---|
| `s3/test_s3_reservations_writer.py` (new) | write-only tests: `test_batch_partial_failure`, `test_custom_prefix`, `test_empty_batch_is_no_op`, `test_json_body_format`, `test_object_key_format`, `test_retry_on_throttling`, `test_writer_credential_recovery`, `test_no_retry_on_access_denied` (its one `get_one_reservation` call only confirms nothing was written) | 962 |
| `s3/test_s3_reservations_storage.py` (new) | write-then-read round trips through the facade: `test_round_trip`, `test_multiple_reservations`, `test_overwrite_duplicate_id`, `test_spec_without_metadata`, `test_user_authored_metadata` | 604 |
| `s3/test_s3_reservations_reader.py` (exists) | reader-lifecycle tests: `test_reader_client_reuse`, `test_reader_empty_chain_recovery`, `test_reader_invalid_token_recovery`, `test_proactive_credential_refresh`. They call the facade today; switch them to `self.storage.reader.get_one_reservation` like the existing tests in that module (behaviour-preserving, the facade only forwards) | 641 incl. existing |
| `s3/test_s3_reservations_expiration.py` (new) | `test_expiration_malformed_object_policy`, `test_expiration_credential_expiry_mid_sweep`, `test_retry_on_listing_pagination` (all call `self.storage.expiration.*` directly) | 640 |
| `s3/test_s3_util.py` (new) | `test_credential_rejection_gate` | 125 |
| `local/test_local_reservations_storage.py` (exists) | `test_local_reservations_storage_constructor`, `..._expiration`, `..._start`, `..._write_read` | 389 incl. existing |

Raw line counts include the 16-line license header of every module, so the merged files come out smaller than the sums.

`s3_reservations_retriever.py`, `aws_sync_client_worker.py` and `aws_async_client_worker.py` keep having no module of their own. The retry and credential-recovery tests pin worker behaviour (`retry_with_new_client`, `do_with_retries`) but reach it through the writer/reader/expiration entry points and the base class's `create_client` patches, so they stay with the component they drive. Testing the workers directly is a separate ticket if wanted.

Non-test support modules stay where they are: `s3_reservations_storage_test_base.py`, `fake_s3_client.py`, `fake_async_s3_client.py`, `local_reservations_test_helpers.py`, and both `conftest.py` files.

## Work items

1. Move `_TokenBoundS3Client` and `_InvalidTokenClient` into their own non-test modules in `s3/` (one class per module, public names matching `fake_s3_client.py` / `FakeS3Client`, e.g. `token_bound_s3_client.py` / `TokenBoundS3Client`). The token-bound client imports `S3ReservationsStorageTestBase` for `make_expired_token_error`. Hoist one copy of `_make_client_error()` onto `S3ReservationsStorageTestBase` as `@staticmethod make_client_error(...)` next to `make_expired_token_error`, and have the invalid-token client and `TestS3Util` use it (`TestS3Util` derives from `TestCase`, so it can call the static method through the class or get its own `@staticmethod`). The nested factory closures become methods or `@staticmethod`s on the receiving test class, bound with `functools.partial` where the patch target needs arguments. No module-level or nested functions in test code.
2. Create the four new `s3/` target modules and fold the legacy test methods in verbatim, keeping their method docstrings. Eight legacy modules also carry module-level docstrings (`test_expiration_credential_expiry_mid_sweep`, `test_expiration_malformed_object_policy`, `test_proactive_credential_refresh`, `test_reader_client_reuse`, `test_reader_empty_chain_recovery`, `test_retry_on_listing_pagination`, `test_local_reservations_storage_constructor`, `test_local_reservations_storage_start`): move that rationale into the receiving class docstring or the relevant method docstring rather than dropping it. Start each new module with `git mv` of the largest legacy module folded into it (`test_batch_partial_failure.py` -> writer, `test_round_trip.py` -> storage, `test_expiration_malformed_object_policy.py` -> expiration, `test_credential_rejection_gate.py` -> util), committed on its own before the fold so rename detection pairs the files. Blame survives for that module only; the other folded lines are attributed to the fold commit and stay reachable via `git log -- <old path>`.
3. Fold the four reader-lifecycle modules into `test_s3_reservations_reader.py`, switching their call to `self.storage.reader.get_one_reservation`.
4. Convert the four `local/` bare pytest classes into `TestLocalReservationsStorage(IsolatedAsyncioTestCase)`:
   - `tmp_path` -> in `setUp`, create a `tempfile.TemporaryDirectory`, register its `cleanup` with `self.addCleanup`, and keep `self.tmp_path = Path(tmp.name).resolve()`. The `Path` wrapper is needed because the tests use `/`, `.iterdir()`, `.is_file()`, `.read_text()`, `.mkdir()`, `.exists()`, `.is_dir()` and `.name`. The `.resolve()` is needed because `LocalReservationsStorage` stores `os.path.abspath(...)` (`local_reservations_storage.py:119`) and the two constructor tests compare against `str(tmp_path.resolve())`; on macOS the temp dir is a symlink under `/var/folders`, so an unresolved path fails them (pytest's `tmp_path` is already resolved, which is why they pass today). Either `setUp` or the per-test `with tempfile.TemporaryDirectory()` block that the existing `test_get_one_reservation_resolves_raw_spec` uses is fine; pick one and use it for the whole class.
   - `monkeypatch.setenv` / `monkeypatch.delenv` -> `with patch.dict(os.environ): os.environ[...] = ...` / `os.environ.pop(..., None)` (precedent: `tests/neuro_san/internals/chat/test_connectivity_reporter.py`). Note `patch.dict(os.environ, {...})` alone cannot delete a key.
   - `pytest.raises(ValueError, match=...)` -> `self.assertRaisesRegex(ValueError, ...)`.
   - bare `assert` -> `self.assert*`; drop `@pytest.mark.asyncio` and `import pytest`.
   - None of these classes use `@pytest.mark.parametrize` or `@pytest.fixture`, so the reason commit 9f1e3d46 (in PR #1319) dropped `TestCase` elsewhere (pytest-injected method arguments) does not apply once the fixtures above are replaced.
5. Update the prose that names the old layout: the `TestLocalReservationsStorage` class docstring ("covered by the sibling test_local_reservations_storage_*.py modules"), the `local_reservations_test_helpers.py` module docstring, the `test_retry_on_listing_pagination` header (`iter_reservation_keys` lives on `S3ReservationsExpiration`, not the storage), the `test_writer_credential_recovery` reference to `test_invalid_token_recovery.py`, and the `test_custom_prefix` mention of "tests T1-T9".
6. Delete the emptied legacy modules.

## Done when

- `python -m pytest -q tests/neuro_san/service/watcher/temp_networks` still collects 46 tests (33 `s3/`, 13 `local/`) and reports `46 passed, 8 subtests passed`. No `OPENAI_API_KEY` is needed: the `s3/` and `local/` `conftest.py` files override the project-wide `configure_llm_provider_keys` skip (the `local/` override arrives with #1354). Diff the `--collect-only -q` node lists before and after with the module prefix stripped: the set of test method names must be identical, not just the count.
- No `test_*.py` in `s3/` or `local/` without a same-stem module under `neuro_san/service/watcher/temp_networks/`.
- Every test class derives from `TestCase` or `IsolatedAsyncioTestCase`; no module-level functions, nested functions or second classes in any `test_*.py`.
- flake8 and pylint clean (baseline today is 10.00/10 on both directories).

## Out of scope

The same behaviour-per-module layout exists elsewhere and can get its own ticket if wanted. The two largest clusters are `tests/neuro_san/session/close_contract/` (4 modules, bare pytest classes named `<source class>_close[_integration]`) and `tests/neuro_san/internals/run_context/langchain/token_counting/` (4 of 6 modules have no same-name source module; `test_nested_token_accounting.py` carries three `Fake*` helper classes). About a dozen single modules are scattered across `tests/neuro_san/`; a census of `tests/neuro_san/**/test_X.py` without a matching `neuro_san/**/X.py` lists them all.

## References

- PR #1319: one test class per file, file named after the class.
- PR #1354: added `test_s3_reservations_reader.py` and `test_local_reservations_storage.py`; left legacy modules for this ticket.
- Commit e95e08e2 (2026-07-14) split the then-single `test_local_reservations_storage.py`, which already held four behaviour-named bare pytest classes, into one module per class plus `local_reservations_test_helpers.py`; commit 400a8df1 later moved them under `local/`. Commit 748e0735 (PR #1319) split `test_credential_rejection_gate.py` out of `test_invalid_token_recovery.py` and renamed the remainder to `test_reader_invalid_token_recovery.py`.

Noravee:
## Summary

Reorganizes the reservations storage tests under `tests/neuro_san/service/watcher/temp_networks/` from one module per behaviour to one module per source class, each holding a single `Test<Class>` deriving from `TestCase`/`IsolatedAsyncioTestCase`. Follow-up to #1354, which added the first two source-mirroring modules and left the legacy ones in place.

Fixes #1356


Noravee (commented):
## Layout

| Module | Class | Folded legacy modules |
|---|---|---|
| `s3/test_s3_reservations_writer.py` | `TestS3ReservationsWriter` (9 tests) | batch_partial_failure, custom_prefix, empty_batch_is_no_op, json_body_format, object_key_format, retry_on_throttling, writer_credential_recovery, no_retry_on_access_denied |
| `s3/test_s3_reservations_storage.py` | `TestS3ReservationsStorage` (5) | round_trip, multiple_reservations, overwrite_duplicate_id, spec_without_metadata, user_authored_metadata |
| `s3/test_s3_reservations_reader.py` | `TestS3ReservationsReader` (8) | reader_client_reuse, reader_empty_chain_recovery, reader_invalid_token_recovery, proactive_credential_refresh, now calling `self.storage.reader.get_one_reservation` like the existing tests |
| `s3/test_s3_reservations_expiration.py` | `TestS3ReservationsExpiration` (8) | expiration_malformed_object_policy, expiration_credential_expiry_mid_sweep, retry_on_listing_pagination |
| `s3/test_s3_util.py` | `TestS3Util` (4) | credential_rejection_gate |
| `local/test_local_reservations_storage.py` | `TestLocalReservationsStorage` (14) | the four bare pytest `test_local_reservations_storage_*` modules |

Grouping is by the class that owns the code under test rather than the facade called, which keeps every module under pylint's 1,000-line default (largest is the writer at 892 lines).

tests/neuro_san/service/watcher/temp_networks/s3/s3_reservations_storage_test_base.py
```diff
@@ -206,12 +225,42 @@ def make_expired_token_error(operation_name: str) -> ClientError:
             operation_name,
         )
 
+    @staticmethod
+    def make_client_error(code: str, operation_name: str = "GetObject") -> ClientError:
```
Noravee: One copy of the `_make_client_error()` that two legacy modules each defined at module level. `TestS3Util` calls it through the class because it derives from `TestCase`, not from this base.

tests/neuro_san/service/watcher/temp_networks/s3/test_s3_reservations_reader.py
```diff
@@ -138,3 +175,377 @@ def test_get_one_reservation_reports_defaults_merge_error_as_absent(self) -> Non
...
+        ClientError-only retry), it propagates out of
+        get_one_reservation() after a single chain resolution and this
+        test errors instead of passing.
+        """
+        reservation_id: str = self._put_live_reservation("copy_cat-empty-chain")
+
+        resolutions = {"count": 0}
+        # See the class docstring for why the partial is wrapped in a MagicMock
+        # rather than patched onto Session.get_credentials bare.
+        get_credentials = MagicMock(side_effect=partial(self._resolve_chain_mid_rewrite, resolutions))
```
Noravee: Patched onto a class attribute (`Session.get_credentials`). A bare `functools.partial` there trips Python 3.13's FutureWarning about partial becoming a method descriptor. The MagicMock forwards the call unchanged. Patches on instance attributes use bare partials.

tests/neuro_san/service/watcher/temp_networks/s3/test_s3_reservations_writer.py
```diff
@@ -0,0 +1,892 @@
...
+        # iterations 1 and 2 fall through to the real in-memory store.
+        # Status 403 lands the error on the non-retryable branch of
+        # S3Util.is_retryable_client_error so we know the failure is final.
+        real_put = self.fake_s3.put_object
+        # Defensively restore put_object on the fake at end-of-test. The
+        # base class hands each test a fresh FakeS3Client in setUp, so
+        # this is a no-op today; the cleanup is here to document intent
+        # and to keep the test correct if a future refactor turns
+        # self.fake_async_s3 into a shared object.
+        self.addCleanup(setattr, self.fake_s3, "put_object", real_put)
```
Noravee: Pre-existing slip fixed while moving this test: `real_put` was taken from `fake_s3` but restored onto `fake_async_s3`.

tests/neuro_san/service/watcher/temp_networks/s3/test_s3_util.py
```diff
@@ -14,37 +14,27 @@
...
-                "Message": f"{code} (test)",
-            },
-            "ResponseMetadata": {"HTTPStatusCode": 400},
-        },
-        operation_name,
-    )
 
-
-class TestCredentialRejectionGate(TestCase):
+class TestS3Util(TestCase):
```
Noravee: Derives from `TestCase`, not the S3 base: these tests never touch a storage or a fake client.

tests/neuro_san/service/watcher/temp_networks/s3/token_bound_s3_client.py
```diff
@@ -0,0 +1,93 @@
...
+from typing import List
+from typing import Optional
+from typing import Tuple
+
+from tests.neuro_san.service.watcher.temp_networks.s3.fake_s3_client import FakeS3Client
+from tests.neuro_san.service.watcher.temp_networks.s3.s3_reservations_storage_test_base \
+    import S3ReservationsStorageTestBase
+
+
+class TokenBoundS3Client:
```
Noravee: Moved out of the test module so each module holds one class, with a public name like `FakeS3Client`. Same for `InvalidTokenS3Client`.

tests/neuro_san/service/watcher/temp_networks/local/test_local_reservations_storage.py
```diff
@@ -36,75 +46,309 @@ class TestLocalReservationsStorage(IsolatedAsyncioTestCase):
...
+        # addCleanup runs even if the test fails or raises.
+        tmp_dir: str = tempfile.mkdtemp()
+        self.addCleanup(shutil.rmtree, tmp_dir)
+        # LocalReservationsStorage stores os.path.abspath(base_path), which does
+        # not resolve symlinks, and the constructor tests compare against the
+        # resolved path. On macOS the temp dir lives under a symlink
+        # (/var -> /private/var), so resolve() here keeps the expected and
+        # actual paths identical; pytest's tmp_path fixture, which these tests
+        # used before, hands out an already-resolved path.
+        self.tmp_path: Path = Path(tmp_dir).resolve()
```
Noravee: Replaces pytest's `tmp_path`. The `resolve()` matters: `LocalReservationsStorage` stores `os.path.abspath()` and two constructor tests compare against it, and on macOS the temp dir is a symlink. pytest's `tmp_path` is already resolved, which is why those tests passed before.

tests/neuro_san/service/watcher/temp_networks/local/test_local_reservations_storage.py
```diff
@@ -36,75 +46,309 @@ class TestLocalReservationsStorage(IsolatedAsyncioTestCase):
...
+            names.append(entry.name)
+        return sorted(names)
+
+    def test_missing_path_raises(self) -> None:
+        """No base_path arg and no env var -> ValueError."""
+        # patch.dict snapshots os.environ and restores it on exit, which also
+        # puts back the key popped here; patch.dict(os.environ, {...}) alone
+        # can only add or override keys, not delete one, hence the explicit pop.
+        with patch.dict(os.environ):
+            os.environ.pop("AGENT_RESERVATIONS_LOCAL_PATH", None)
```
Noravee: Replaces `monkeypatch.delenv`. `patch.dict(os.environ)` puts the key back on exit; `patch.dict(os.environ, {...})` on its own cannot delete one.

tests/neuro_san/service/watcher/temp_networks/s3/test_s3_reservations_reader.py
```diff
@@ -138,3 +175,377 @@ def test_get_one_reservation_reports_defaults_merge_error_as_absent(self) -> Non
...
+        # create_client seam with this mock and discards the client built in
+        # setUp, so the constructions counted are exactly the ones the reads
+        # below cause.
+        create_client_mock = MagicMock(return_value=self.fake_s3)
+
+        with self._fresh_reader_client(create_client_mock):
+            for _ in range(self.NUM_READS):
+                reservation: Optional[Reservation]
+                agent_network: Optional[AgentNetwork]
+                reservation, agent_network = self.storage.reader.get_one_reservation(reservation_id)
```
Noravee: The four folded reader tests call the reader directly instead of the storage facade. The facade is a one-line delegate, so the path exercised is the same, and it matches the tests that were already here.

d1donlydfink (approved):
Nice consolidation. I was wondering about this organization when I was last fixing these

PR #1364, opened 2026-09-18
Make NameCorrectionConfigFilter rewrite args.tools references

### LINKED ISSUE #1358: NameCorrectionConfigFilter does not rewrite args.tools references when it corrects an agent name
## Summary

`NameCorrectionConfigFilter` replaces `/` in agent names with `_` and rewrites the string entries of every tool's `tools` list so both ends of each edge agree (`neuro_san/internals/graph/filters/name_correction_config_filter.py:78-95`). It does not touch `args.tools`, the documented convention for CodedTools deriving from `BranchActivation` to declare the agents they may call (`docs/agent_hocon_reference.md`, "tools (args)"). After a correction, `args.tools` still holds the old name while the agent is registered under the new one.

## Who reads `args.tools`

- `ToolsShapeValidator.validate_args_tools` (`tools_shape_validator.py:99`) accepts a dict of label -> agent name or a list of agent names.
- `ConnectivityReporter` merges `args.tools` with `tools` through `AbstractNetworkValidator.coerce_args_tools` (`connectivity_reporter.py:205-211`, `abstract_network_validator.py:162`), so connectivity reports an edge to a name that no longer exists in the network.
- Coded tools that look up their children by the names in their own `args["tools"]` miss the renamed agent.

## Scope

Pre-existing on `main` for hocon-file networks. Reservation specs also pass through the filter since PR #1354, but the deploy-time `ToolNameNetworkValidator` (`^[a-zA-Z0-9_-]+$`, `tool_name_network_validator.py:36`) rejects slash names, so on that path only specs already sitting in S3/local storage from before that PR can hit it. Raised by Copilot on PR #1354 (https://github.com/cognizant-ai-lab/neuro-san/pull/1354#discussion_r4039731621) and deliberately left out of that PR as unrelated to its fix.

## Proposed fix

In the second loop of `filter_config`, after rewriting `tool["tools"]`, also rewrite `tool["args"]["tools"]` when present, using the same `corrections` table: every string element of a list, every string value of a dict (keys untouched). Leave non-string entries and other shapes alone; the shape validator is the place that reports those. Do not create `args` or `args.tools` where they are absent.

## Tests

Extend `tests/neuro_san/internals/graph/filters/test_name_correction_config_filter.py::TestNameCorrectionConfigFilter` (introduced in PR #1354):

- list form: `args.tools: ["help/er"]` becomes `["help_er"]`
- dict form: `args.tools: {"helper": "help/er"}` becomes `{"helper": "help_er"}`, key unchanged
- non-string entries and a tool without `args` are left exactly as they were
- a corrected name referenced from both `tools` and `args.tools` is rewritten in both

## References

- PR #1354: the `tool["name"]` assignment fix (was `tool[name]`) and the filter's first unit tests.
- `docs/agent_hocon_reference.md`, section "tools (args)".

Noravee:
## Summary

`NameCorrectionConfigFilter` replaces `/` in agent names with `_` and rewrites the references other tools hold in their `tools` lists. It did not rewrite `args.tools`, the coded-tool convention for declaring downstream agents (docs "tools (args)"), which `ToolsShapeValidator.validate_args_tools` and `ConnectivityReporter` (through `AbstractNetworkValidator.coerce_args_tools`) both read as edges. After a correction the child was registered under the new name while `args.tools` still held the old one, so connectivity reported a dangling edge and a coded tool looking its children up by those names missed.

Fixes #1358
Raised by Copilot on #1354 and split out as pre-existing behaviour.

## Changes

- `neuro_san/internals/graph/filters/name_correction_config_filter.py`: new `correct_args_tools(tool, corrections)` static method, called from the second loop after the existing `tools`-list rewrite. It rewrites string entries of `args.tools` in place, in both accepted shapes: every element of a list, every value of a dict (labels untouched, they are the coded tool's own lookup keys). Non-string entries and other shapes are left alone (the shape validator is what reports those), and neither `args` nor `args.tools` is created where absent.
- The loop's `continue` on a missing or empty `tools` list became an `if` around the unchanged list rewrite, because a coded tool typically has no `tools` list at all and still needs its `args.tools` rewritten.
- Class docstring updated to say both kinds of reference are rewritten.

## Tests

Six added to `TestNameCorrectionConfigFilter`; the four existing tests are untouched. List form, dict form with keys preserved, non-string entries and an `args`-less tool left alone, a name referenced from both `tools` and `args.tools` rewritten in both, `args` without a `tools` key does not gain one, and a malformed (string) `args.tools` left as-is. Four of the six fail against `main`'s filter.


neuro_san/internals/graph/filters/name_correction_config_filter.py
```diff
@@ -79,20 +83,27 @@ def filter_config(self, basis_config: Dict[str, Any]) \
...
-                # leave complex tool objects (e.g. MCP dictionaries) as-is
-                if isinstance(agent_tool, str):
-                    new_agent_tools.append(corrections.get(agent_tool, agent_tool))
-                else:
-                    new_agent_tools.append(agent_tool)
-
-            tool["tools"] = new_agent_tools
+            # This is an if rather than a continue on purpose: a coded tool typically has
+            # no "tools" list at all, yet still needs its args.tools rewritten below.
+            if isinstance(agent_tools, list) and len(agent_tools) > 0:
```
Noravee, line 88: Was a `continue`. A coded tool usually has no `tools` list at all but still needs its `args.tools` rewritten below, so the list rewrite is now just guarded and the loop body keeps going.

neuro_san/internals/graph/filters/name_correction_config_filter.py
```diff
@@ -79,20 +83,27 @@ def filter_config(self, basis_config: Dict[str, Any]) \
...
+
+                tool["tools"] = new_agent_tools
+
+            # CodedTools deriving from BranchActivation declare the agents they might call
+            # in args.tools rather than in tools (see "tools (args)" in agent_hocon_reference.md).
+            # ToolsShapeValidator.validate_args_tools and ConnectivityReporter (through
+            # AbstractNetworkValidator.coerce_args_tools) both read that key, so a reference
+            # left under the old name would make connectivity report an edge to an agent that
+            # no longer exists, and the coded tool itself would fail to find its renamed child.
+            self.correct_args_tools(tool, corrections)
```
Noravee, line 106: Uses the same `corrections` table as the `tools` rewrite above, so both kinds of reference end up on the corrected name.

neuro_san/internals/graph/filters/name_correction_config_filter.py
```diff
@@ -105,6 +116,45 @@ def filter_config(self, basis_config: Dict[str, Any]) \
...
+        strings, or a dict of label -> agent-name string. Only string entries (list elements,
+        dict values) are rewritten. Dict keys, non-string entries and any other shape are left
+        untouched, since ToolsShapeValidator is the place that reports malformed shapes.
+        Neither "args" nor "args.tools" is created where absent.
+
+        :param tool: The agent/tool spec dictionary whose args.tools is corrected in place
+        :param corrections: A dictionary of original agent name -> corrected agent name
+        """
+        args: Any = tool.get("args")
+        if not isinstance(args, dict) or "tools" not in args:
```
Noravee, line 136: Absent `args` or `args.tools` stays absent. The filter only renames; it never adds keys the author did not write.

neuro_san/internals/graph/filters/name_correction_config_filter.py
```diff
@@ -105,6 +116,45 @@ def filter_config(self, basis_config: Dict[str, Any]) \
...
+            # List form: every element is an agent name. Rewrite by index so that the list
+            # object and any non-string entries stay exactly as they were given.
+            for index, agent_tool in enumerate(args_tools):
+                if isinstance(agent_tool, str):
+                    args_tools[index] = corrections.get(agent_tool, agent_tool)
+        elif isinstance(args_tools, dict):
+            # Dict form: label -> agent name.  Only the values are names; the labels are the
+            # coded tool's own lookup keys and must stay stable.  Assigning to existing keys
+            # while iterating is safe; only adding or removing keys would not be.
+            for label, agent_tool in args_tools.items():
```
Noravee, line 153: Dict form: the values are agent names, the keys are the coded tool's own labels and stay as they are. Non-string values and any other shape are left alone; `ToolsShapeValidator` is what reports those.

tests/neuro_san/internals/graph/filters/test_name_correction_config_filter.py
```diff
@@ -101,3 +109,88 @@ def test_uncorrectable_name_is_logged_and_left_as_is(self) -> None:
...
+
+        # Only the string reference changes; the int and the dict are value-equal to the copy.
+        expected_front_args["tools"][0] = "help_er"
+        self.assertEqual(expected_front_args, filtered["tools"][0]["args"])
+        # The helper never had "args" and must not have acquired one.
+        expected_helper["name"] = "help_er"
+        self.assertEqual(expected_helper, filtered["tools"][1])
+        self.assertNotIn("args", filtered["tools"][1])
+
+    def test_corrects_name_referenced_from_both_tools_and_args_tools(self) -> None:
```
Noravee: Four of the six new tests fail on main's filter: this one plus the list, dict and mixed-entries cases. The other two only pin that nothing is created or touched.

Noravee (commented):
Inline notes on the parts worth a look.

d1donlydfink (commented):
Nice catch!

d1donlydfink (approved):

PR #1361, opened 2026-09-17
Add more typing

d1donlydfink:

PR #1366, opened 2026-09-18
Honor the https scheme in external agent URL references

### LINKED ISSUE #1365: External agent references written with https:// are sent as plain http to port 8080 (URL scheme dropped)
## Summary

An external-agent tool reference written with an `https://` URL is not called over https. `ExternalAgentParsing.parse_external_agent()` discards the URL scheme, `ExternalAgentSessionFactory` never passes a `security_cfg`, and `AbstractHttpServiceAgentSession` then defaults to `http` and port 8080. A reference such as

```
"tools": ["https://agents.example.com/deep/math_guy"]
```

is actually requested as

```
http://agents.example.com:8080/api/v1/deep/math_guy/function
```

Against a TLS-terminated deployment that port is closed, so the call times out and the calling agent drops the tool as "unreachable". There is no hocon key or environment variable that can make an external-agent reference use https.

## Reproduction

```python
from unittest.mock import MagicMock
from neuro_san.internals.utils.external_agent_parsing import ExternalAgentParsing
from neuro_san.session.external_agent_session_factory import ExternalAgentSessionFactory

url = "https://agents.example.com/deep/math_guy"
location = ExternalAgentParsing.parse_external_agent(url)
print(location)
# {'host': 'agents.example.com', 'port': None, 'agent_name': 'deep/math_guy'}   <- no scheme

ic = MagicMock()
ic.get_metadata.return_value = {}
session = ExternalAgentSessionFactory(use_direct=False).create_session_from_location_dict(location, ic)
print(session.get_request_path("function"))
# http://agents.example.com:8080/api/v1/deep/math_guy/function                 <- http + 8080
```

Writing `https://host:443/deep/math_guy` does not help: the port is honored but the scheme is still `http`, so plain HTTP is sent to the TLS port (nginx answers `400 The plain HTTP request was sent to HTTPS port`).

Writing `http://host:80/deep/math_guy` against a host that redirects 80 to 443 looks like it works but does not: the `function()` GET follows the 301 so the tool registers, but `streaming_chat()` is a POST and aiohttp turns a redirected POST into a GET, which the streaming handler rejects.

## Root cause (main @ 5e13d134)

- `neuro_san/internals/utils/external_agent_parsing.py` L93-98: the returned dict has only `host`, `port`, `agent_name`. `parse_result.scheme` is never read. For a non-localhost host without an explicit port, `port` stays `None` (the localhost rule at L76-80 only substitutes `server_port` for `localhost`).
- `neuro_san/session/external_agent_session_factory.py` L156: `AsyncHttpServiceAgentSession(host, port, agent_name=..., metadata=..., streaming_timeout_in_seconds=None)` with no `security_cfg`.
- `neuro_san/session/abstract_http_service_agent_session.py` L73-75: `port is None` becomes `DEFAULT_HTTP_PORT` (8080, `neuro_san/interfaces/agent_session_constants.py` L26); L94-96: scheme is `https` only when `security_cfg is not None`.
- The server does hold a `security_cfg` (`neuro_san/service/generic/async_agent_service.py` L88) but only forwards it to `AsyncDirectAgentSession`, never to the `ExternalAgentSessionFactory` it constructs at L278.
- `AGENT_SESSION_REQUIRE_HTTPS=true` (L100-104) only raises `ValueError` for `http`; it never switches the scheme. With the shipped Dockerfile default of `true`, every cross-server external-agent reference therefore fails: `http://` references are rejected and `https://` references are mis-built.

The only place in the repo that ever gets an https session is client code that passes `security_cfg={}` explicitly (`neuro_san/client/agent_session_factory.py` L76-79, `tests/load_tests/traffic/http_client.py` L86-91).

## History

- #571 (closed) covered the localhost case: a `/name` reference on a non-8080 server. Its fix is the `server_port` substitution at L76-80, which applies to `localhost` only; remote hosts were left with the session default.
- #84 introduced the "https only if security_cfg is non-None" skeleton.
- #1053 added `AGENT_SESSION_REQUIRE_HTTPS`; its description assumed "client code eventually calling this class specifies https in their urls and security configuration", which the external-agent path never does.
- #1072 made the Dockerfile default `AGENT_SESSION_REQUIRE_HTTPS="true"`.
- `docs/agent_hocon_reference.md` L778-795 documents URL references with only an `http://localhost:8080/...` example; `docs/clients.md` L25-30 also states the server uses `use_direct=False`, while `async_agent_service.py` L278 constructs the factory with `use_direct=True`.

## Proposed fix

1. `parse_external_agent()` returns a `scheme` key (`""` for scheme-less `/name` references) and, only for `https` without an explicit port and after the existing localhost rule, defaults the port to a new `AgentSessionConstants.DEFAULT_HTTPS_PORT = 443`. Bare `http://host/agent` keeps today's 8080 default; explicit ports are kept as written. The session class always emits an explicit port, so leaving `None` would produce `https://host:8080`.
2. `ExternalAgentSessionFactory.create_session_from_location_dict()` passes `security_cfg={}` when the scheme is `https` (same convention as the CLI client), `None` otherwise. Dicts without a `scheme` key keep plain http.
3. Docs: https example and port rules in the External Agents section; correct the `use_direct` statement in `docs/clients.md`; note in the Dockerfile that `AGENT_SESSION_REQUIRE_HTTPS` also governs external-agent references.
4. Unit tests for the parser (scheme/port matrix, localhost precedence, unparseable URLs) and the factory (https -> `https://host:443/...`, http -> `http://host:8080/...`, legacy dict without scheme, behavior under `AGENT_SESSION_REQUIRE_HTTPS=true`).

Implementation is ready on branch `https-external-agents`. It overlaps files touched by open PR #1361 (typing edits in `external_agent_session_factory.py` and `abstract_http_service_agent_session.py`).

## Related

- cognizant-ai-lab/neuro-san-studio#600 is a different bug (MCP tool names containing `/`) that is often hit in the same setup; it is tracked separately.

Noravee:
## Summary

External-agent tool references written with an `https://` URL were never called over https. `ExternalAgentParsing.parse_external_agent()` discarded the URL scheme, `ExternalAgentSessionFactory` never passed a `security_cfg`, and `AbstractHttpServiceAgentSession` therefore fell back to `http` and port 8080. A reference such as `https://agents.example.com/deep/math_guy` was requested as `http://agents.example.com:8080/api/v1/deep/math_guy/function`, which times out against a TLS-terminated deployment, so the calling agent dropped the tool as unreachable. Writing `:443` explicitly sent plain http to the TLS port instead. With the shipped Dockerfile default `AGENT_SESSION_REQUIRE_HTTPS=true`, every cross-server external-agent reference failed: `http://` was rejected and `https://` was mis-built.

Fixes #1365

## Changes

- `ExternalAgentParsing.parse_external_agent()` returns a new `scheme` key (`""` for scheme-less `/name` references). For `https` references without an explicit port it defaults the port to the new `AgentSessionConstants.DEFAULT_HTTPS_PORT` (443), after the existing localhost/`server_port` rule. Bare `http://host/agent` keeps today's 8080 session default; explicit ports are kept as written. The session class always emits an explicit port, so leaving it `None` would have produced `https://host:8080`.
- `ExternalAgentSessionFactory.create_session_from_location_dict()` passes `security_cfg={}` when the scheme is `https` (the same convention `neuro_san/client/agent_session_factory.py` uses), `None` otherwise. Location dicts without a `scheme` key keep plain http.
- `parse_external_agent()` reads the host and port from `urlparse`'s bracket-aware `hostname` and `port` fields instead of `netloc.split(":")`, which turned a bracketed IPv6 reference like `https://[2001:db8::1]/agent` into host `[2001` and port `db8` (so the 443 default was skipped and the session emitted a malformed address). IPv6 literals keep their brackets so the session layer can build `https://[v6]:443/...`. An explicit port is returned exactly as written (`:0443` stays `0443`); a non-numeric or empty port, or an authority without a host (`http://user@/agent`, `http://:8080/agent`), is treated as unparseable and returns `None` rather than falling back to localhost. An empty authority (`http:///agent`) keeps its same-server behavior.
- `parse_external_agent()` catches the `ValueError` that `urlparse` raises on malformed urls (mismatched brackets read as an IPv6 netloc) and returns `None` per its contract, instead of crashing connectivity reporting and external-tool resolution.
- Docs: https example and port rules in the External Agents section of `docs/agent_hocon_reference.md`; `docs/clients.md` corrected (the server constructs the factory with `use_direct=True`, `async_agent_service.py:278`); both Dockerfiles note that `AGENT_SESSION_REQUIRE_HTTPS` also governs external-agent references.

## Tests

- New `tests/neuro_san/internals/utils/test_external_agent_parsing.py::TestExternalAgentParsing`: scheme/port matrix (https without port -> `"443"`, explicit port kept, http without port -> `None`, `/name` with and without `server_port`, `https://localhost` where the localhost rule wins, bracketed IPv6 hosts with and without a port, explicit port text preserved, non-numeric and empty ports, hostless authorities, empty authority), plus the malformed-url cases.
- New `tests/neuro_san/session/test_external_agent_session_factory.py::TestExternalAgentSessionFactory`: https -> `https://host:443/...`, http -> `http://host:8080/...`, legacy dict without scheme -> http, end-to-end parse -> factory for an https reference (including a bracketed IPv6 host), and behavior under `AGENT_SESSION_REQUIRE_HTTPS=true` (https succeeds, http raises).
- `tests/neuro_san/internals/chat/test_connectivity_reporter.py`: a malformed string-form url no longer crashes the connectivity report.
- Local: `flake8` clean, `build_scripts/run_pylint.sh` 10.00/10, 295 passed across `tests/neuro_san/{session,internals/utils,internals/chat,internals/validation,internals/run_context/langchain/core,internals/graph,client}` with `OPENAI_API_KEY=dummy`; the two new modules also pass with `AGENT_SESSION_REQUIRE_HTTPS=true`.

## Notes

- The behavior change is limited to `https://` references: the scheme is now honored, and a missing port means 443 instead of 8080. Anyone who relied on an `https://` reference silently being served over plain http will now get a real TLS handshake.
- Overlaps files touched by #1361 (typing edits in `external_agent_session_factory.py` and `abstract_http_service_agent_session.py`); no functional conflict expected.
- Live check: the URL now built for a nested network on a TLS-terminated neuro-san 0.7.0 deployment returned 200 on its `/function` endpoint.


neuro_san/internals/utils/external_agent_parsing.py
```diff
@@ -60,14 +80,26 @@ def parse_external_agent(agent_url: str, server_port: int = None) -> Dict[str, s
...
+            # which turned "[2001:db8::1]" into host "[2001" and port "db8".
+            # hostname also drops any userinfo and lower-cases the name.
+            host = parse_result.hostname
+            if ":" in host:
+                # hostname strips the brackets from an IPv6 literal; put them
+                # back so the session layer can build "https://[v6]:443/...".
+                host = f"[{host}]"
+            if port_number is not None:
+                # Keep the historical str type for the port.
+                port = str(port_number)
```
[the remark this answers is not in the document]
Noravee: Fixed in f654d50d: .port is now used only to validate the number; the returned port is the raw text after the last ':' of the netloc, so ':0443' stays '0443'. Test added.

neuro_san/internals/utils/external_agent_parsing.py
```diff
@@ -60,14 +80,29 @@ def parse_external_agent(agent_url: str, server_port: int = None) -> Dict[str, s
...
-            # We have a host specified
-            split: List[str] = parse_result.netloc.split(":")
-            host = split[0]
-            if len(split) > 1:
-                port = split[1]
+        if parse_result.hostname:
+            # hostname/port are bracket-aware, unlike a naive netloc.split(":"),
+            # which turned "[2001:db8::1]" into host "[2001" and port "db8".
+            # hostname also drops any userinfo and lower-cases the name.
+            host = parse_result.hostname
```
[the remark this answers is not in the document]
Noravee: Fixed in 14d90ade: a present-but-hostless authority ("user@", ":8080") now returns None instead of falling back to localhost; an empty authority ("http:///agent") keeps its same-server behavior. Tests added for all three.

Noravee:
**TL;DR**

Problem (#1365): an `https://host/agent` tool reference was silently called as `http://host:8080/...`. The parser dropped the scheme and the factory never passed `security_cfg`, so the http session always built an http URL.

Fix:
- Parser returns the scheme. An https reference with no explicit port defaults to 443 (a TLS proxy, not the 8080 dev server). `https://localhost/...` keeps the server's own port, like `/name`.
- Factory turns scheme `https` into a non-None `security_cfg`, which is how the http session picks `https://`.

Unchanged: `/name` and well-formed `http://host[:port]/name` references keep their host, port and plain http.

Also fixed along the way: IPv6 hosts were mangled by `netloc.split(":")`, and malformed references (no host, empty or non-numeric port, bad IPv6 bracket) now return None instead of a bogus host or a ValueError out of `is_external_agent`.

Where to look: `external_agent_parsing.py` (scheme and port rules) and `ExternalAgentSessionFactory.get_security_cfg` (the https switch). User-facing rules are in the External Agents section of `docs/agent_hocon_reference.md`.

neuro_san/internals/utils/external_agent_parsing.py
```diff
@@ -47,7 +57,17 @@ def parse_external_agent(agent_url: str, server_port: int = None) -> Dict[str, s
...
             return None
 
-        parse_result: ParseResult = urlparse(agent_url)
+        port_number: Optional[int] = None
+        try:
+            parse_result: ParseResult = urlparse(agent_url)
+            # .port raises ValueError for a non-numeric or out-of-range port,
+            # which is as unparseable as a malformed netloc.
+            port_number = parse_result.port
+        except ValueError:
```
Noravee, line 66: urlparse and `.port` raise ValueError for a mismatched IPv6 bracket or a non-numeric port. Before, that escaped `is_external_agent` and could crash connectivity reporting and tool setup. Now such a reference just returns None, meaning "not an external agent".

neuro_san/internals/utils/external_agent_parsing.py
```diff
@@ -56,18 +76,43 @@ def parse_external_agent(agent_url: str, server_port: int = None) -> Dict[str, s
...
+        # "http://:8080/agent") is malformed. It must not fall through to the
+        # localhost default below, which would silently turn a broken remote
+        # reference into a call to a local agent of the same name. An explicit
+        # but empty port ("https://host:/agent") is malformed too: .port reports
+        # it as None, so without this check it would receive a default port.
+        # The host-info part is what follows any userinfo "@".
+        hostinfo: str = parse_result.netloc.rsplit("@", 1)[-1]
+        malformed_authority: bool = bool(parse_result.netloc) and \
+            (not parse_result.hostname or hostinfo.endswith(":"))
+        if not parse_result.path.startswith("/") or malformed_authority:
```
Noravee, line 89: A URL with `//` but no host (`http://user@/x`, `http://:8080/x`) or with an empty port (`https://host:/x`) now returns None. Without this guard, a hostless reference would fall back to localhost and call a local agent of the same name, and an empty port would silently get a default.

neuro_san/internals/utils/external_agent_parsing.py
```diff
@@ -56,18 +76,43 @@ def parse_external_agent(agent_url: str, server_port: int = None) -> Dict[str, s
...
+            # Either not an external agent specification, or unparseable.
             return None
 
+        # No normalization is done on the scheme: the network validators
+        # (AbstractNetworkValidator.is_url_or_path) only recognize the lower-case
+        # "http://" and "https://" spellings in hocon tool references, so that is
+        # the only form that reaches this parser in normal operation.
+        scheme: str = parse_result.scheme or ""
```
Noravee, line 97: The scheme is now kept and returned. Before, it was dropped here, so the location dict never told the factory a reference was https.

neuro_san/internals/utils/external_agent_parsing.py
```diff
@@ -56,18 +76,43 @@ def parse_external_agent(agent_url: str, server_port: int = None) -> Dict[str, s
...
+            if ":" in host:
+                # hostname strips the brackets from an IPv6 literal; put them
+                # back so the session layer can build "https://[v6]:443/...".
+                host = f"[{host}]"
+            if port_number is not None:
+                # .port only validated the number. Return the port as written in
+                # the reference ("0443" stays "0443"), as this parser always has.
+                # When present, the port is what follows the last ":" of the
+                # netloc, which is bracket-safe because an IPv6 literal ends in "]".
+                port = parse_result.netloc.rsplit(":", 1)[1]
```
Noravee, line 115: Host now comes from urlparse's bracket-aware `hostname`; the old `netloc.split(":")` broke IPv6 hosts. `.port` only validates the number; the returned port is the text after the last `:`, so `0443` stays `0443`.

neuro_san/internals/utils/external_agent_parsing.py
```diff
@@ -79,12 +124,24 @@ def parse_external_agent(agent_url: str, server_port: int = None) -> Dict[str, s
...
 
-        # Get the agent name from the URL by looking at the path
+        if port is None and scheme == "https":
+            # An https reference without an explicit port is expected to reach
+            # a TLS-terminating proxy on the well-known https port, not the bare
+            # 8080 dev server the session layer defaults to. Only https gets this
+            # treatment: http and scheme-less references keep port None so the
+            # session layer's own http default still applies. This runs after
+            # the localhost rule so a configured server_port wins for localhost.
+            port = str(AgentSessionConstants.DEFAULT_HTTPS_PORT)
```
Noravee, line 134: The main behavior change: an https reference with no port now defaults to 443 instead of the session's 8080. http and `/name` references keep port None here, as before. The localhost rule above runs first, so `https://localhost/...` keeps the server port when one is configured.

neuro_san/session/external_agent_session_factory.py
```diff
@@ -153,11 +158,38 @@ def create_session_from_location_dict(self, agent_location: Dict[str, str],
             #   b)  We figure that the regular connection aspects to the server in question
             #       have already been sorted out in the obligitory call to function() that
             #       precedes any streaming_chat() call.
+            security_cfg: Dict[str, Any] = self.get_security_cfg(agent_location)
```
Noravee, line 161: This is the actual https switch. The http session builds an `https://` URL only when `security_cfg` is not None (abstract_http_service_agent_session.py L94-96); nothing was passed here before, so every external agent reached over the network got http. Empty dict means https, None means http, same convention as neuro_san/client/agent_session_factory.py.

neuro_san/deploy/Dockerfile
```diff
@@ -423,6 +423,10 @@ ENV AGENT_MCP_ONLY="false"
 # When set to "true", this parameter requires https for internal agent session communication.
 # The default is "true". Developers should set this to false to work locally without acquiring certificates.
 # Strongly consider keeping this on for production deployments.
+# This also governs external-agent references in the tools lists of agent network hocons:
```
Noravee, line 426: Same note added to Dockerfile.py314t. With `AGENT_SESSION_REQUIRE_HTTPS=true` (the Dockerfile default; the code default is false), a remote `http://` external agent is dropped from the calling agent's tools as unreachable (base_tool_factory.py L141-144). `/name` and `http://localhost/...` references to networks this server serves are resolved in-process and are not affected.

docs/clients.md
```diff
@@ -22,12 +22,15 @@ There are two ways to connect to an agent network:
...
+client → Server → http / https → external network on another server (`http://...` / `https://...`)
 
 The client sends requests to a running server. The server runs the agent network and resolves
-any external networks (tools prefixed with `/`) via HTTP back to itself.
+same-server external networks (tools prefixed with `/`) in-process: it constructs its
+`ExternalAgentSessionFactory` with `use_direct=True`, so no HTTP round trip back to itself is made.
+External networks referenced by URL are called over http or https, matching the URL's scheme.
 
-`use_direct` is hardcoded to `False` on the server — the client's setting is ignored.
+`use_direct` is hardcoded to `True` on the server — the client's setting is ignored.
```
Noravee, line 33: Doc correction. The server constructs the factory with `use_direct=True` (async_agent_service.py L278), so same-server references are resolved in-process, not via an HTTP call back to itself.

neuro_san/internals/utils/external_agent_parsing.py
d1donlydfink: Great catch!

d1donlydfink (approved):
Nice!

PR #1368, opened 2026-09-18
Add provider-side tool runtime support

### LINKED ISSUE #1348: provider_tools: runtime plumbing into create_agent, user-class strip, same-provider fallback guard, docs
Part of #1347 (1 of 5).

## Scope

Make `llm_config.provider_tools` work end to end for OpenAI, Anthropic and Gemini.

- `LangChainRunContext.create_agent` (`langchain_run_context.py:333-339`): pass `tools=self.tools + (self.llm_config.get("provider_tools") or [])`. `self.llm_config` at `:103` is already the network+agent overlay (`calling_activation.py:96-97`), so a top-level key covers both plain and fallback configs. `self.tools` stays `List[BaseTool]`.
- `default_llm_factory.py:47`: add `"provider_tools"` to `KEYS_TO_REMOVE_FOR_USER_CLASS` so `"class": "pkg.mod.ChatX"` configs no longer raise in `ArgumentValidator`.
- `create_llm_with_fallbacks` (`default_llm_factory.py:569-695`): when the top-level `provider_tools` is non-empty and more than one entry was built, compare `type(resources.get_llm_policy())` across main and fallbacks and raise `ValueError` if they differ. `RunnableWithFallbacks.bind_tools` broadcasts one list to every member, and cross-provider dicts either raise at bind time (`KeyError: 'function'`, `ValueError: Unsupported function`) or degrade silently. Nested peer groups: pass the key down or document same-provider.
- `default_llm_info.hocon`: `"provider_tools": null` at the top of the `openai`, `anthropic` and `gemini` args blocks (above the hunks #1332/#1333 rewrite), with a comment that the key is consumed by the run context, not the chat-model constructor.
- Docs: `#### provider_tools` under `llm_config` in `docs/agent_hocon_reference.md` (semantics: replace/clear, same-provider fallbacks, model decides, text-only wire and history, provider billing outside token accounting, no secrets in dicts); provider dict shapes and rejections in the "Provider-Specific Arguments" section of `docs/llm_info_hocon_reference.md` once #1334 lands; carve-outs at `:315` ("scalar") and `:412-415` ("Unsupported parameters will be ignored").

## Tests

`unittest.TestCase`, keyless (`pytestmark = [non_default_llm_provider, ollama]` as in `test_langchain_run_context_llm_fallbacks.py:33`):

- run context: `create_agent` receives `[base_tool, dict]` and `run_context.tools` is unchanged; no key means `tools=self.tools`; canary with the real `create_agent` and a `FakeMessagesListChatModel` recording `bind_tools`.
- factory: user-class path accepts the key; mixed policy types with a non-empty key raise; same-provider chain builds a `RunnableWithFallbacks`.

## Notes

Land after #1332, #1333 and #1334. Do not edit `test_openai_llm_policy.py` (#1332 converts it) or create `test_gemini_llm_policy.py` (#1333 does).

arnabnandy7:
## Summary

Adds end-to-end runtime plumbing for provider-side tools declared through `llm_config.provider_tools`.

- Passes provider-native tool dictionaries to LangChain during agent creation
- Preserves the existing client-side tool collection
- Removes `provider_tools` before constructing user-defined chat model classes
- Requires fallback models to use the same provider when provider tools are configured
- Propagates provider tools through nested fallback peer groups
- Declares the option for OpenAI, Anthropic, and Gemini configurations
- Documents inheritance, clearing, fallback, billing, security, and provider-specific behavior
- Adds keyless unit coverage and a real LangChain binding canary

Closes #1348.


neuro_san/internals/run_context/langchain/core/langchain_run_context.py
```diff
@@ -332,7 +332,7 @@ def create_agent(self, instructions: str, llm: BaseLanguageModel) -> Runnable:
 
         return create_agent(
             model=llm,
-            tools=self.tools,
+            tools=self.tools + (self.llm_config.get("provider_tools") or []),
```
d1donlydfink, line 335: Simplify the number of operations on one line by having the tools value be computed outside the constructor call

arnabnandy7: @d1donlydfink done

PR #1369, opened 2026-09-18
Break out ChatMessageType methods to ChatMessageTypeUtil

d1donlydfink:
This decomplexifies what wants to be a simpler Enum.

Also:
* Some de-comprehending
* More type hints

Please double-check my unravelling of comprehensions.


neuro_san/client/thinking_file_message_processor.py
```diff
@@ -110,8 +111,8 @@ def write_message(self, response: Dict[str, Any], origin_str: str):
 
         timestamp: float = time()
         response_type: str = response.get("type")
-        message_type: ChatMessageType = ChatMessageType.from_response_type(response_type)
-        message_type_str: str = ChatMessageType.to_string(message_type)
+        message_type: ChatMessageType = ChatMessageTypeUtil.from_response_type(response_type)
+        message_type_str: str = ChatMessageTypeUtil.to_string(message_type)
```
d1donlydfink, line 115: These methods now come from ChatMessageTypeUtil

neuro_san/message/types/chat_message_type.py
```diff
@@ -37,66 +36,18 @@ class ChatMessageType(IntEnum):
...
     These all need to match what is defined in chat.proto
     """
-    UNKNOWN_MESSAGE_TYPE = 0
-    SYSTEM = 1
-    HUMAN = 2
-    AI = 4
+    UNKNOWN_MESSAGE_TYPE: int = 0
+    SYSTEM: int = 1
+    HUMAN: int = 2
+    AI: int = 4
```
d1donlydfink: Add int typing.

neuro_san/message/types/chat_message_type.py
```diff
@@ -37,66 +36,18 @@ class ChatMessageType(IntEnum):
...
-    @classmethod
-    def to_string(cls, chat_message_type: ChatMessageType) -> str:
-        """
-        :param chat_message_type: A ChatMessageType instance
-        :return: A string corresponding to the chat_message_type
-        """
-        message_type_str: str = _CHAT_MESSAGE_TYPE_TO_STRING.get(chat_message_type)
-        if message_type_str is None:
-            message_type_str = _CHAT_MESSAGE_TYPE_TO_STRING.get(cls.UNKNOWN_MESSAGE_TYPE)
-        return message_type_str
```
d1donlydfink, line 98: All these methods move to ChatMessageTypeUtil

neuro_san/message/utils/content_utils.py
```diff
@@ -87,21 +87,27 @@ def flatten_to_text(source: Union[BaseMessage, str, List[Any], None]) -> str:
         if content is None:
             return ""
         if isinstance(content, str):
-            return content
+            content_string: str = content
+            return content_string
```
d1donlydfink, line 94: A few extra typings to make transpile easier

neuro_san/message/utils/content_utils.py
```diff
@@ -87,21 +87,27 @@ def flatten_to_text(source: Union[BaseMessage, str, List[Any], None]) -> str:
...
         if not isinstance(content, list):
             # Not a legal langchain content shape, but do not crash over it.
             return str(content)
 
+        content_list: List[Any] = content
         parts: List[str] = []
-        for item in content:
+        for item in content_list:
```
d1donlydfink, line 101: Parts below best viewed with Hide Whitespace on.

neuro_san/message/utils/content_utils.py
```diff
@@ -121,13 +127,24 @@ def is_empty_content(source: Union[BaseMessage, str, List[Any], None]) -> bool:
         :param source: A BaseMessage, or raw message content
         :return: True if the content is empty in the sense above
         """
-        content: Any = source.content if isinstance(source, BaseMessage) else source
+        content: Any = source
+        if isinstance(source, BaseMessage):
+            content = source.content
```
d1donlydfink, line 135: Simplify

neuro_san/message/utils/content_utils.py
```diff
@@ -121,13 +127,24 @@ def is_empty_content(source: Union[BaseMessage, str, List[Any], None]) -> bool:
...
+            string_content: str = content
+            return string_content.strip() == ""
+
         if isinstance(content, list):
-            return all(ContentUtils._is_blank_block(item) for item in content)
+            list_content: List[Any] = content
+            for item in list_content:
+                if not ContentUtils._is_blank_block(item):
+                    return False
+            return True
```
d1donlydfink, line 149: Comprehensions are incomprehensible

neuro_san/message/utils/content_utils.py
```diff
@@ -171,11 +188,19 @@ def standard_blocks(message: BaseMessage) -> List[Dict[str, Any]]:
...
+        block: Dict[str, Any] = None
+        for block in content_blocks:
+            use_block: Dict[str, Any] = block
+            block_type: str = use_block.get("type")
+            if block_type in ContentUtils.TOOL_CALL_BLOCK_TYPES:
+                continue
+            blocks.append(use_block)
+
+        json_safe_blocks: List[Dict[str, Any]] = ContentUtils.to_json_safe(blocks)
+        return json_safe_blocks
```
d1donlydfink, line 206: Comprehensions are incomprehensible

neuro_san/message/utils/content_utils.py
```diff
@@ -243,18 +274,35 @@ def normalize_message(message: BaseMessage) -> BaseMessage:
         :return: The same message instance if nothing changed,
                  otherwise a model_copy with normalized content
         """
+        update_dict: Dict[str, Any] = {}
```
d1donlydfink, line 286: Introduce a variable separate to pass into the model_copy() call

neuro_san/message/utils/content_utils.py
```diff
@@ -243,18 +274,35 @@ def normalize_message(message: BaseMessage) -> BaseMessage:
...
+    @staticmethod
+    def message_model_copy(message: BaseMessage, update_dict: Dict[str, Any]) -> BaseMessage:
+        """
+        Funnel method to return a model_copy() of the message with the given update_dict applied.
+
+        :param message: The BaseMessage to copy
+        :param update_dict: The dictionary to update the message with
+        :return: A model_copy() of the message with the update_dict applied
+        """
+        return message.model_copy(update=update_dict)
```
d1donlydfink, line 316: Funnel calls to model_copy() here.

neuro_san/message/utils/content_utils.py
```diff
@@ -273,14 +321,22 @@ def looks_like_blocks(value: Any) -> bool:
         """
         if not isinstance(value, list) or len(value) == 0:
             return False
-        for item in value:
+
+        list_value: List[Any] = value
+        item: Any = None
+        for item in list_value:
```
d1donlydfink, line 338: More typing.

neuro_san/message/utils/content_utils.py
```diff
@@ -424,7 +501,11 @@ def to_json_safe(value: Any) -> Any:
         if isinstance(value, float):
             return value if math.isfinite(value) else str(value)
         if isinstance(value, (bytes, bytearray)):
-            return base64.b64encode(bytes(value)).decode("ascii")
+            # Convert bytes to base64
+            bytes_value: bytes = bytes(value)
+            encoded: bytes = base64.b64encode(bytes_value)
+            encoded_bytearray: bytearray = bytearray(encoded)
+            return ContentUtils.decode(encoded_bytearray, "ascii")
```
d1donlydfink: Funnel all calls to common decode()

neuro_san/message/utils/content_utils.py
```diff
@@ -436,3 +517,16 @@ def to_json_safe(value: Any) -> Any:
...
+        """
+        :param instring: The bytes to decode
+        :param encoding: The encoding to use
+        :return: The decoded string
+        """
+        if not instring:
+            return ""
+
+        decoded: str = instring.decode(encoding)
+        return decoded
```
d1donlydfink: Common decode()

neuro_san/session/abstract_http_service_agent_session.py
```diff
@@ -149,12 +149,3 @@ def help_message(self, path: str) -> str:
...
         return message
-
-    @staticmethod
-    def decode_utf8(instring: bytearray) -> str:
-        """
-        :param instring: The bytes to decode
-        :return: The decoded string
-        """
-        decoded: str = instring.decode("utf-8")
-        return decoded
```
d1donlydfink, line 160: This moves to ContentUtils.

neuro_san/message/utils/content_utils.py
```diff
@@ -243,18 +274,35 @@ def normalize_message(message: BaseMessage) -> BaseMessage:
...
                 return message
-            return message.model_copy(update={"content": normalized})
+            update_dict["content"] = normalized
+            return ContentUtils.message_model_copy(message, update_dict)
 
-        response_metadata: Dict[str, Any] = dict(message.response_metadata or {})
+        response_metadata: Dict[str, Any] = message.response_metadata
+        if response_metadata is None:
+            response_metadata = {}
         response_metadata["output_version"] = ContentUtils.OUTPUT_VERSION_V1
```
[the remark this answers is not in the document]
d1donlydfink: Taking your advice a different way.

neuro_san/message/utils/content_utils.py
```diff
@@ -313,28 +369,38 @@ def blocks_from_chat_message(chat_message: Optional[Dict[str, Any]]) -> Optional
             return None
 
         content_blocks: Any = chat_message.get("content_blocks")
-        if content_blocks is not None and content_blocks != []:
+        content_blocks_list: List[Any] = content_blocks
+        if content_blocks is not None and len(content_blocks_list) != 0:
```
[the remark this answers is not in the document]
d1donlydfink: Addressed a different way

neuro_san/session/mcp_chat_response_dictionary_converter.py
```diff
@@ -55,7 +56,7 @@ def to_dict(self, obj: object) -> Dict[str, object]:
         response: Dict[str, Any] = content_seq[0]
         final_response: Dict[str, Any] = {
             "response": {
-                "type": ChatMessageType.AGENT_FRAMEWORK.name,
+                "type": ChatMessageTypeUtil.to_string(ChatMessageType.AGENT_FRAMEWORK),
```
d1donlydfink, line 59: Use the to_string() instead.

neuro_san/message/utils/content_utils.py
```diff
@@ -243,18 +275,37 @@ def normalize_message(message: BaseMessage) -> BaseMessage:
...
         if isinstance(normalized, str):
             if normalized == message.content:
                 return message
-            return message.model_copy(update={"content": normalized})
-
-        response_metadata: Dict[str, Any] = dict(message.response_metadata or {})
+            update_dict["content"] = normalized
+            return ContentUtils.message_model_copy(message, update_dict)
+
+        response_metadata: Dict[str, Any] = None
```
Noravee: This pre-declare-then-reassign shape transpiles badly. python2ts 1.4.3 (the version the ts-package pins) emits a new block-scoped `let response_metadata` inside each branch, so the outer one is still `null` when line 291 indexes it. `update_dict` also gets a second `let` in the same scope (lines 278 and 293). And `copy()` produces a duplicate `copy` import that substitutions.txt only strips for ten named files, not this one.

The base one-liner avoids all of that and is still a shallow copy:

```python
response_metadata: Dict[str, Any] = dict(message.response_metadata or {})
```

Suggest the same for `update_dict`: build it once in each branch instead of pre-declaring it. This also clears the mypy error for assigning `None` to a `Dict[str, Any]`.

neuro_san/message/utils/content_utils.py
```diff
@@ -121,13 +128,24 @@ def is_empty_content(source: Union[BaseMessage, str, List[Any], None]) -> bool:
         :param source: A BaseMessage, or raw message content
         :return: True if the content is empty in the sense above
         """
-        content: Any = source.content if isinstance(source, BaseMessage) else source
+        content: Any = source
```
Noravee, line 133: Same transpile issue as in normalize_message: `content = source` followed by `content = source.content` inside the `if` becomes a shadowed `let content` in the TS output, so `content` stays the BaseMessage. The ternary form still used at line 87 transpiles cleanly. Suggest keeping that form here and at lines 424-426.

neuro_san/message/utils/content_utils.py
```diff
@@ -171,11 +189,19 @@ def standard_blocks(message: BaseMessage) -> List[Dict[str, Any]]:
...
         """
-        blocks: List[Dict[str, Any]] = [
-            block for block in message.content_blocks
-            if block.get("type") not in ContentUtils.TOOL_CALL_BLOCK_TYPES
-        ]
-        return ContentUtils.to_json_safe(blocks)
+        blocks: List[Dict[str, Any]] = []
+
+        content_blocks: List[Dict[str, Any]] = message.content_blocks
+        block: Dict[str, Any] = None
```
Noravee, line 197: `block: Dict[str, Any] = None` before `for block in ...` has no effect in Python (the loop creates the name) and it assigns `None` to a non-Optional hint, which mypy flags. Same at lines 329, 394 and 435. python2ts emits a dead `let block = null` for each. Suggest deleting the four pre-declarations.

neuro_san/message/utils/content_utils.py
```diff
@@ -171,11 +189,19 @@ def standard_blocks(message: BaseMessage) -> List[Dict[str, Any]]:
...
+        for block in content_blocks:
+            use_block: Dict[str, Any] = block
+            block_type: str = use_block.get("type")
```
Noravee, line 200: `dict.get` can return `None`, so `block_type: str` is not accurate and mypy flags it (same at line 336). Line 448 hints the identical expression as `Any`; suggest the same here.

Related: `content_blocks_list: List[Any] = content_blocks` at line 375 runs before the `isinstance(..., list)` check on 376, so that hint is false for non-list values.

neuro_san/message/utils/content_utils.py
```diff
@@ -215,14 +241,20 @@ def normalize_content(message: BaseMessage) -> Union[str, List[Dict[str, Any]]]:
...
-        if isinstance(content, list) and all(isinstance(item, str) for item in content):
-            # A list of plain strings carries no block structure at all -
-            # collapse to text rather than promoting it to a block list.
-            return ContentUtils.flatten_to_text(message)
+            content_string: str = content
+            return content_string
+
+        if isinstance(content, list):
+            content_list: List[Any] = content
+            if all(isinstance(item, str) for item in content_list):
```
Noravee: One comprehension-style expression survived the unravelling: `all(isinstance(item, str) for item in content_list)`. `is_empty_content` got unrolled at lines 144-147; suggest the same here for consistency.

neuro_san/message/types/chat_message_type.py
```diff
@@ -37,66 +36,18 @@ class ChatMessageType(IntEnum):
     Python enum to mimic protobufs for chat.ChatMessageType without dragging in all of gRPC.
     These all need to match what is defined in chat.proto
     """
-    UNKNOWN_MESSAGE_TYPE = 0
-    SYSTEM = 1
-    HUMAN = 2
-    AI = 4
+    UNKNOWN_MESSAGE_TYPE: int = 0
```
Noravee: Enum members should stay unannotated. `AI: int = 4` says the member is an `int`, but `type(ChatMessageType.AI)` is `ChatMessageType`. The typing spec asks checkers to reject annotated enum members; current mypy reports all eight lines (mypy 1.10 is silent). python2ts emits identical TypeScript with or without the annotation, so it does not help the transpile either. Suggest reverting to bare `NAME = value`.

neuro_san/message/types/chat_message_type_util.py
```diff
@@ -0,0 +1,80 @@
...
+# END COPYRIGHT
+from __future__ import annotations
+
+from typing import Type
+from typing import Union
+
+from langchain_core.messages.base import BaseMessage
+
+from neuro_san.message.types.chat_message_type import ChatMessageType
+from neuro_san.message.types.chat_message_type import _CHAT_MESSAGE_TYPE_TO_STRING
```
Noravee: These two imports reach into another module's underscore names, the only place in the repo that does this. Transpiled, they become non-exported consts imported by this file, and substitutions.txt only has an export rule for `const ChatMessageType`.

Suggest making both maps class attributes of `ChatMessageTypeUtil`. Then chat_message_type.py drops its langchain and Agent*Message imports and is just the IntEnum, which is what the title is after. `from __future__ import annotations` can go in both files too.

neuro_san/session/http_service_agent_session.py
```diff
@@ -30,6 +30,7 @@
 from requests import post
 from requests import Response
 
+from neuro_san.message.utils.content_utils import ContentUtils
```
Noravee: This import makes both HTTP client sessions load langchain_core and pydantic at import time; on main they were langchain-free (443 vs 565 modules). ContentUtils' own docstring calls it framework plumbing, not client-facing. The commit also removes public `AbstractHttpServiceAgentSession.decode_utf8` with no shim (I found no external callers, so minor). In the ts-package, substitutions.txt:541 that rewrites `this.decode_utf8(` goes dead, and the new `instring.decode(encoding)` has no mapping there.

Simplest is to keep `decode_utf8` on the session base and call it as `AbstractHttpServiceAgentSession.decode_utf8(...)`. Otherwise the PR description should mention the API removal and the dependency change.

Noravee:
Summary:

- `normalize_message` and the `content = source` pattern transpile to shadowed `let`s under python2ts 1.4.3; the base one-liners did not.
- Enum member annotations and several `= None` / `str` hints are wrong. mypy on the touched modules goes from 2 errors on main to 7.
- The private maps are imported across modules; moving them onto `ChatMessageTypeUtil` fixes that and makes the enum file a plain IntEnum.
- HTTP client sessions now import langchain via ContentUtils; `decode_utf8` removed without a shim.

Outside this diff: `streaming_input_processor.py` lines 149 and 194 still use `ChatMessageType.X.name`, which is undefined on the transpiled enum object. Same fix as 766c8ad8.

neuro_san/message/utils/content_utils.py
d1donlydfink, line 294: Nope, Claude:  The suggested fix also transpires badly, that's why I went there.

neuro_san/message/utils/content_utils.py
d1donlydfink: Fine. We are not using mypy as a checker though.

neuro_san/message/utils/content_utils.py
d1donlydfink: This is fixable in the substitutions.

neuro_san/message/utils/content_utils.py
d1donlydfink: We are not using mypy as a tool in this stack.

Noravee: I think this one should be Optional[str].

neuro_san/message/utils/content_utils.py
d1donlydfink: Reasonable. This particular one translated ok.

neuro_san/message/types/chat_message_type.py
d1donlydfink: I can revert this.

neuro_san/message/types/chat_message_type_util.py
d1donlydfink: nice find.

neuro_san/session/http_service_agent_session.py
d1donlydfink: The shim was only put there in the previous PR. No need for compatibility, claude.
Dependencies need time and releases to accumulate.

neuro_san/message/utils/content_utils.py
d1donlydfink: Done.

neuro_san/message/types/chat_message_type.py
d1donlydfink: Reverted.

neuro_san/message/types/chat_message_type_util.py
d1donlydfink: Fixed.

neuro_san/session/http_service_agent_session.py
d1donlydfink: It's an interesting observation about the import dependency though.
ContentUtils itself is under message, which we expect clients to use as a package, but the only thing within messages that uses it is BaseMessageDictionaryConverter.  This makes me think that ContentUtils and BaseMessageDictionaryConverter as a unit aren't quite in the right place either.

d1donlydfink: Upon inspection, the thing to do is to break out flatten_to_text() into a separate utility as well.
This way BaseMessageDictionaryConverter can just use that and only that.
The rest of ContentUtil would then be free to move back into the internals area.
Not for this PR.

Noravee (approved):
LGTM.

Issue #1371, opened 2026-09-20
MCP tool names for nested registry networks contain "/", which OpenAI and Anthropic reject


Noravee:
### Problem

Agent networks kept in a registry sub-directory get a network name that includes the directory, for example `deep/math_guy` for `registries/deep/math_guy.hocon`. When such a network is served as an MCP tool (`mcp: true` in the manifest), the neuro-san MCP server advertises it in `tools/list` under that same name, "/" included, and `tools/call` expects that name back.

LLM providers reject tool names containing "/": OpenAI requires `^[a-zA-Z0-9_-]{1,64}$` and Anthropic `^[a-zA-Z0-9_-]{1,128}$`. Any agent network that consumes such a tool through `LangChainMcpAdapter` fails the whole LLM request with a 400 as soon as the tool is in its tool list. This is what cognizant-ai-lab/neuro-san-studio#600 reports for `basic/music_nerd_pro`:

```
Invalid 'tools[0].function.name': string does not match pattern.
```

The same happens with any third-party MCP server whose tool names contain "/", since the adapter passes names through unchanged.

### Where

- `neuro_san/service/mcp/processors/mcp_tools_processor.py`: `_get_tool_description()` uses the network name as the MCP tool name, and `tools/call` looks the tool up by that name.
- `neuro_san/internals/run_context/langchain/mcp/langchain_mcp_adapter.py`: hands MCP tool names to LangChain unchanged.
- `neuro_san/session/mcp_service_agent_session.py`: neuro-san's own MCP client matches `tools/list` entries by the exact network name.

### Workarounds today

- Add a slash-free alias network at the registry root that references the nested network, and expose the alias over MCP instead.
- Call the nested network over the native HTTP API rather than MCP.

### Proposed fix, as stacked PRs

1. `McpToolNamePolicy` (derive `deep__math_guy` from `deep/math_guy`, validate against the provider pattern) and an advertised tool name on `AgentNetwork`. No behavior change on its own.
2. `McpServiceAgentSession` accepts both spellings in `tools/list` and calls the tool by whichever spelling the server advertised, so it works with servers from before and after the change.
3. Server: advertise the provider-safe name, carry the original network name in the MCP `title` field, accept `tools/call` under either spelling, support an optional `mcp_name` manifest key, and resolve collisions such as `a/b` vs `a__b`.
[...]

PR #1372, opened 2026-09-20
Add McpToolNamePolicy and an advertised MCP tool name on AgentNetwork (1 of 4)

### LINKED ISSUE #600: Invalid MCP tool name causes 400 error
When attempting to call the MCP tool named basic/music_nerd_pro, the request fails with the following error:

```
openai.BadRequestError: Error code: 400 - {
  'error': {
    'message': "Invalid 'tools[0].function.name': string does not match pattern. Expected a string that matches the pattern '^[a-zA-Z0-9-]+$'.",
    'type': 'invalid_request_error',
    'param': 'tools[0].function.name',
    'code': 'invalid_value'
  }
}
```

**Root cause**
The MCP tool name contains a forward slash (`/`), which is not allowed for the tool name.

Noravee:
Part of #1371 (MCP tool names for nested registry networks contain "/", which OpenAI and Anthropic reject).

First of four PRs, opened one at a time. This one adds the building blocks and changes no behavior.

- `McpToolNamePolicy` (new, `neuro_san/internals/utils/mcp_tool_name_policy.py`): derives a provider-safe tool name from a network name (`deep/math_guy` becomes `deep__math_guy`; any other character outside `[A-Za-z0-9_-]` becomes `_`), validates names against the pattern OpenAI and Anthropic require, flags names over OpenAI's 64-character cap, and matches allow-list entries written in either spelling.
- `AgentNetwork`: new `mcp_tool_name` field with `set_as_mcp_tool(tool_name=None)`, `get_mcp_tool_name()` and `clear_mcp_tool()`. Calling `set_as_mcp_tool()` with no name keeps using the network name, so existing callers behave exactly as before.

Tests: `tests/neuro_san/internals/utils/test_mcp_tool_name_policy.py` and `tests/neuro_san/internals/graph/registry/test_agent_network.py` (new test package).

Follow-ups, each opened once the previous one has merged (they are closed until then):
1. #1373 client session: `McpServiceAgentSession` matches either spelling in `tools/list` and calls the tool by the spelling the server advertised.
2. #1374 server side: advertise `deep__math_guy`, carry the network name in the MCP `title`, accept `tools/call` under either spelling, manifest `mcp_name`, collision handling.
3. #1375 LangChain adapter: rename unsafe tool names from any MCP server before the LLM sees them. This is what fixes cognizant-ai-lab/neuro-san-studio#600 for adapter users.


Noravee:

Noravee:
Reopening: the four PRs for #1371 are now going one at a time, starting here.

neuro_san/internals/utils/mcp_tool_name_policy.py
```diff
@@ -0,0 +1,260 @@
...
+        # Finally compare both sides in mangled form so an entry written with
+        # the original "/" still matches a server that has already renamed its
+        # tools (the server-side fix). Only entries carrying a NETWORK_SEPARATOR
+        # get this treatment: the server-side rename applies to network names,
+        # which are registry paths, and without the gate the lossy
+        # UNSAFE_REPLACEMENT mapping would let an entry "a.b" claim an unrelated
+        # tool "a_b".
+        if McpToolNamePolicy.NETWORK_SEPARATOR not in entry:
+            return False
+        return McpToolNamePolicy.to_tool_name(entry) == mangled_original
```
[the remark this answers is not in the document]
Noravee: Fixed in 8be38ff7: the "/" gate is gone, so matches() compares the mangled forms for every unsafe character (an entry "a.b" now selects a tool advertised as "a_b"). resolve() still prefers an exact spelling when the server offers one, which is what prevents the lossy mapping from picking an unrelated tool. Tests updated accordingly.

Noravee:
**TL;DR**

Foundation for #1371: nested registry networks get MCP tool names like `deep/math_guy`, and OpenAI and Anthropic reject the "/". This PR adds the pieces the later PRs use; nothing observable changes yet.

- `McpToolNamePolicy` (new): `to_tool_name()` maps `deep/math_guy` to `deep__math_guy` and any other unsafe character to `_`. `is_valid_tool_name()` and `is_over_soft_limit()` check the provider pattern and OpenAI's 64-character cap. `select_allowed()` and `find_unmatched()` match allow-list entries written in either spelling.
- `AgentNetwork`: new `mcp_tool_name` field with `set_as_mcp_tool(tool_name=None)`, `get_mcp_tool_name()` and `clear_mcp_tool()`. Today's only caller passes no name, so the tool name stays the network name.

Not in this PR: nothing calls `to_tool_name()` yet. The client session (#1373), server (#1374) and LangChain adapter (#1375) do.

Where to look: `mcp_tool_name_policy.py` for the mapping and matching rules, `agent_network.py` for the field and accessors. Each rule has a unit test.

neuro_san/internals/utils/mcp_tool_name_policy.py
```diff
@@ -0,0 +1,261 @@
...
+    NETWORK_SEPARATOR: str = "/"
+
+    # What NETWORK_SEPARATOR becomes in a tool name. "__" is used rather than
+    # "." because "." fails both the OpenAI and Anthropic regexes, and rather
+    # than "-" because a single hyphen is ambiguous with hyphenated stems
+    # ("a-b" could be the network "a/b" or a network literally named "a-b").
+    # "__" is also the separator ExternalAgentParsing.get_safe_agent_name uses
+    # for "/" (that helper additionally prefixes a leading "__"), so external
+    # agents and MCP tools follow the same separator convention.
+    SEPARATOR: str = "__"
```
Noravee: Why `__` rather than `-` or `.`: `.` fails both provider regexes, and a single `-` would be ambiguous with names that already contain a hyphen (`a-b` could be the network `a/b` or a network literally named `a-b`).

neuro_san/internals/utils/mcp_tool_name_policy.py
```diff
@@ -0,0 +1,261 @@
...
+
+    # Strictest tool-name shape common to OpenAI and Anthropic: ASCII letters,
+    # digits, underscore and hyphen. The length cap is Anthropic's 128;
+    # OpenAI's tighter 64 is handled as a soft limit below.
+    TOOL_NAME_PATTERN: str = r"^[a-zA-Z0-9_-]{1,128}$"
+
+    # OpenAI caps tool names at 64 characters. This is warn-only because the
+    # same name is still valid for Anthropic and for lenient local providers,
+    # so callers log rather than refuse the tool.
+    SOFT_MAX_LENGTH: int = 64
```
Noravee, line 55: Two limits on purpose. The pattern is the strictest shape both providers accept, with Anthropic's 128-character cap. OpenAI's 64 is warn-only because the same name still works for Anthropic and for local models.

neuro_san/internals/utils/mcp_tool_name_policy.py
```diff
@@ -0,0 +1,261 @@
...
+
+    @staticmethod
+    def to_tool_name(network_name: str) -> str:
```
Noravee: The core mapping every consumer calls: `/` becomes `__`, any other character outside `[A-Za-z0-9_-]` becomes `_`, and a name that is already safe comes back unchanged, so applying it twice changes nothing.

neuro_san/internals/utils/mcp_tool_name_policy.py
```diff
@@ -0,0 +1,261 @@
...
+        # the original spelling still matches a server that has already renamed
+        # its tools. The server-side rename applies to_tool_name() to the whole
+        # network name, so this has to cover every unsafe character, not only
+        # the "/" of nested registry paths: a network "a.b" is advertised as
+        # "a_b" and the entry "a.b" must still select it. The mapping is lossy
+        # ("a.b" and "a b" both become "a_b"), so an entry could in principle
+        # claim an unrelated tool that happens to carry the mangled spelling;
+        # resolve() keeps that in check by preferring an exact spelling whenever
+        # the server offers one.
+        return McpToolNamePolicy.to_tool_name(entry) == mangled_original
```
Noravee: Both sides are compared in mangled form for every unsafe character, not only `/` (Copilot round 1). The mapping is lossy, so `resolve()` below tries the exact spelling first; that is what stops an entry `a.b` from taking a literal `a_b` when the server offers both.

neuro_san/internals/utils/mcp_tool_name_policy.py
```diff
@@ -0,0 +1,261 @@
...
+        :return: The advertised name equal to entry when there is one; otherwise
+                 the first advertised name that matches() the entry; None when
+                 nothing matches or original_names is None or empty.
+        """
+        # A None original list means the server advertised nothing, so there is
+        # nothing to resolve to rather than an error.
+        if not original_names:
+            return None
+
+        if entry in original_names:
```
Noravee, line 155: Exact spelling wins before any equivalence. With a server offering both `a/b` and `a__b`, the entry `a/b` means `a/b` only and `a__b` means `a__b` only.

neuro_san/internals/graph/registry/agent_network.py
```diff
@@ -47,6 +47,33 @@ def __init__(self, config: Dict[str, Any], name: str):
...
+        #
+        # The network name stays the internal key everywhere else (network storage,
+        # agent authorization, the /api/v1/{agent_name} http path). This field is
+        # only the spelling shown to MCP clients; when the two differ, the MCP tool
+        # description can still carry self.name in its "title" field so clients can
+        # see which network a renamed tool stands for.
+        #
+        # None until set_as_mcp_tool() is called, i.e. for networks that are not
+        # served as MCP tools at all.
+        self.mcp_tool_name: str = None
```
Noravee, line 75: New field, kept apart from `self.name`. The network name stays the key for storage, authorization and the `/api/v1/{name}` path; this is only the spelling MCP clients see. The table above lists the three cases.

neuro_san/internals/graph/registry/agent_network.py
```diff
@@ -62,11 +89,43 @@ def get_config(self) -> Dict[str, Any]:
...
+                          a nested network advertised that way keeps its "/", which
+                          OpenAI and Anthropic reject in tool names.
         """
         self.is_mcp_network = True
+        # An empty string is treated like None: an MCP tool must have some name,
+        # and the network name is the only sensible fallback.
+        if tool_name:
+            self.mcp_tool_name = tool_name
+        else:
+            self.mcp_tool_name = self.name
```
Noravee, line 128: This branch is what keeps the PR behavior-neutral: the only caller today, `RegistryManifestRestorer`, passes no name, so the tool name is still the network name until the server-side PR.

neuro_san/internals/graph/registry/agent_network.py
```diff
@@ -75,6 +134,37 @@ def is_mcp_tool(self) -> bool:
...
+        (network "a__b") both derive the MCP tool name "a__b". The network whose
+        own name already is that tool name keeps it, so the other one is cleared:
+
+            a_slash_b.clear_mcp_tool()
+                a_slash_b.is_mcp_tool() -> False, a_slash_b.get_mcp_tool_name() -> None
+                tools/list then shows a single "a__b" tool, backed by network "a__b"
+
+        Calling set_as_mcp_tool() again makes the network an MCP tool once more.
+        """
+        self.is_mcp_network = False
```
Noravee, line 165: Only used by the server-side PR's collision handling (`a/b` vs `a__b`). It withdraws the network from MCP alone; the network stays served over the http API and as an external agent.

neuro_san/internals/graph/registry/agent_network.py
d1donlydfink: I take it the above documentation will make it to manifest_hocon_reference when everything is in place?

neuro_san/internals/utils/mcp_tool_name_policy.py
d1donlydfink: Maybe call this constant NEUR_SAN_MCP_SEPARATOR to make it clearer?

neuro_san/internals/utils/mcp_tool_name_policy.py
d1donlydfink: You talk about the length cap in TOOL_NAME_PATTERN, but SOFT_MAX_LENGTH comment should probably contain the info about the OpenAI limit as it's closer to the actual usage of the limit.

neuro_san/internals/utils/mcp_tool_name_policy.py
```diff
@@ -0,0 +1,261 @@
...
+#
+# END COPYRIGHT
+import re
+
+from typing import List
+from typing import Optional
+from typing import Set
+
+
+class McpToolNamePolicy:
```
d1donlydfink: Strongly consider establishing the basis of this class as a StringFilter instead of all static methods:

```
class StringFilter:
    """
    Interface whose filter() method takes one string and returns another string
    which is a filtered version of the input.

    These kinds of things are necessary for the following reasons:
    * Tensorflow only likes column names with certain characters in them
    * HTML UI sometimes only likes certain characters in strings.
    """

    def filter(self, in_string: str) -> str:
        """
        :param in_string: an input string to filter
        :return: a filtered version of the in_string, according to implementation policy
        """
        raise NotImplementedError
```

In this case to_tool_name() becomes filter().

Using such an interface allows complex composition later on down the line.
Static methods do not allow this.

The StringFilter interface used to be in leaf-common, and we can reinstate it again if needed.
Easier to restart here and push down.

ofrancon (dismissed):
works for me

neuro_san/internals/graph/registry/agent_network.py
Noravee: Yes. The server-side PR (#1374, closed until this one and #1373 merge) adds an `mcp_name` section to `docs/manifest_hocon_reference.md` covering the derivation (`deep/math_guy` is advertised as `deep__math_guy`, a top-level name is unchanged, an explicit `mcp_name` wins), the startup warnings for names outside the provider pattern or over 64 characters, and the collision rule. `docs/mcp_service.md` gets the tools/list naming and the `title` field. I am adding the same three-row example table from this comment to that section so the doc and the code comment match.

neuro_san/internals/utils/mcp_tool_name_policy.py
Noravee: Done, spelled `NEURO_SAN_MCP_SEPARATOR`. The constant now lives in the new `McpToolNameFilter` (`neuro_san/internals/utils/mcp_tool_name_filter.py`), and its comment says it is neuro-san's own convention.

neuro_san/internals/utils/mcp_tool_name_policy.py
Noravee: Done. The `TOOL_NAME_PATTERN` comment now only describes the shape and Anthropic's 128-character cap; the OpenAI 64-character limit and the warn-only rationale live on `SOFT_MAX_LENGTH`.

neuro_san/internals/utils/mcp_tool_name_policy.py
Noravee: Done:

- `StringFilter` reinstated at `neuro_san/internals/interfaces/string_filter.py` (your interface, docstrings in the repo template). leaf-common 1.4.x no longer ships it, so it starts here and can be pushed down later.
- The mapping is now `McpToolNameFilter(StringFilter)` in `neuro_san/internals/utils/mcp_tool_name_filter.py`; `filter()` is the former `to_tool_name()`.
- `McpToolNamePolicy` is instance-based: the constructor takes a `StringFilter` (default `McpToolNameFilter`), the policy is itself a `StringFilter` that delegates `filter()` to it, and the allow-list matching goes through the same filter, so an injected or composed mapping drives matching as well. `is_valid_tool_name()` and `is_over_soft_limit()` stay static since they only read the class constants.
- Tests: new `test_string_filter.py` and `test_mcp_tool_name_filter.py`; the policy tests are instance-based and include an injected-filter test.

The three follow-up branches already hold one policy or filter instance each (client session, restorer, adapter) and pass their suites.

Noravee:
**Changes since the review**

  - `StringFilter` interface reinstated under `neuro_san/internals/interfaces/`.
  - New `McpToolNameFilter(StringFilter)` holds the network-name to tool-name mapping as `filter()`; its separator constant is now `NEURO_SAN_MCP_SEPARATOR`.
  - `McpToolNamePolicy` is instance-based and itself a `StringFilter`: it takes the mapping filter at construction (default `McpToolNameFilter`), delegates `filter()` to it, and its matching methods use that filter. The two validators stay static.
  - `TOOL_NAME_PATTERN` and `SOFT_MAX_LENGTH` comments each describe their own limit.
  - Tests: new modules for the interface and the filter; policy tests instance-based, plus an injected-filter test.

Still no behavior change: the only caller of `set_as_mcp_tool()` passes no name. 
The example table for the `mcp_name` section of `docs/manifest_hocon_reference.md` is queued in #1374.

d1donlydfink (approved):

PR #1373, opened 2026-09-20
Make McpServiceAgentSession use the provider-safe MCP tool name (2 of 4)

Noravee:
Part of #1371. Second of four PRs, opened one at a time. The foundation (#1372) has merged; this one must merge before #1374.

neuro-san's own MCP client session (`agent_cli --connection mcp`, `AgentSessionFactory.create_session("mcp", ...)`) is created with the network name, for example `--agent deep/math_guy`. Once a server advertises `deep__math_guy` instead, the current client no longer finds its tool in `tools/list`, `function()` returns None, and `agent_cli` turns that into an error.

Changes in `McpServiceAgentSession`:
- Derives the provider-safe name from the network name at construction.
- `function()` matches a `tools/list` entry under either spelling and remembers which spelling the server used.
- `streaming_chat()` sends `tools/call` with that remembered spelling, falling back to the provider-safe name if `function()` was not called first.

So the new client works against servers from before this change, which only know `deep/math_guy`, and after it. For a top-level network the two spellings are identical and nothing changes.

Why this must merge before the server-side PR: a client without this change only matches the exact network name, so servers must not start advertising the new name until clients can follow it.

Tests: `tests/neuro_san/session/test_mcp_service_agent_session.py` (new), covering both server generations.


Noravee:

Noravee:
Reopening: the foundation PR #1372 has merged, so this one is next.

Noravee:
Superseded by #1376. This PR could not be reopened: its base branch was deleted when #1372 merged, and GitHub does not allow changing the base of a closed pull request.

PR #1374, opened 2026-09-20
Advertise MCP tools under provider-safe names and add manifest mcp_name (3 of 4)

Noravee:
Part of #1371. Third of four stacked PRs: based on #1373, which must merge first (see there for why).

Server side of the fix. A nested network such as `deep/math_guy` is now advertised in `tools/list` as `deep__math_guy`, and `tools/call` accepts either spelling.

- `RegistryManifestRestorer.derive_mcp_tool_name()`: the tool name is the manifest's `mcp_name` when present, otherwise `McpToolNamePolicy.to_tool_name(network_name)`. Names that do not match the provider pattern, or exceed 64 characters, are still exposed but warned about at startup.
- `RegistryManifestRestorer.resolve_mcp_tool_name_collisions()`: two public networks resolving to the same tool name (`a/b` and `a__b`, or duplicate `mcp_name` values) are detected after all manifests are loaded. The network whose own name is the tool name keeps it; otherwise the first in sorted order does. The other network stays served over the regular APIs but is withdrawn from MCP, with an error log naming both.
- `McpManifestDictConfigFilter`: new optional `mcp_name` manifest key. A non-empty string implies `mcp: true` (and therefore `public: true`); any other value is dropped with a warning. `ManifestDictFilterChain` now passes the manifest file and network name to it for logging.
- `McpToolsProcessor`: `tools/list` uses the advertised name and, when it differs from the network name, carries the network name in the MCP `title` field (MCP 2025-06-18). `tools/call` maps the advertised name back to the network name before authorization and lookup; the legacy slash spelling and top-level names pass through unchanged.

Docs: `docs/manifest_hocon_reference.md` (`mcp_name` section) and `docs/mcp_service.md`.

Tests: new `test_registry_manifest_restorer.py`, `test_mcp_manifest_dict_config_filter.py`, and `tests/neuro_san/service/mcp/processors/test_mcp_tools_processor.py` (new test packages with `__init__.py`).

Compatibility note: an `agent_cli` or client library without #1373 matches `tools/list` by the exact network name and will not find nested networks on a server with this change. Upgrade clients first.

PR #1375, opened 2026-09-20
Rename provider-unsafe MCP tool names in LangChainMcpAdapter (4 of 4)

Noravee:
Part of #1371, and the part that fixes cognizant-ai-lab/neuro-san-studio#600 as adapter users see it. Fourth of four stacked PRs: based on #1372, independent of #1373 and #1374.

OpenAI and Anthropic reject the whole request when any tool name falls outside `^[a-zA-Z0-9_-]+$`. `LangChainMcpAdapter` passed MCP tool names through unchanged, so a neuro-san server's `basic/music_nerd_pro`, or any third-party MCP tool with a "/" in its name, produced the 400 in studio#600.

Changes in `LangChainMcpAdapter.get_mcp_tools()`:
- Allow-list filtering happens first, against the names the server actually advertised. Entries may be written in either spelling; when a server offers both spellings of one name, an entry selects the one it spells exactly. Entries that match nothing are logged together with the server's tool list instead of being dropped silently.
- Tools are then renamed with `McpToolNamePolicy.to_tool_name()` (`/` becomes `__`, other unsafe characters become `_`). Only the LangChain tool's name changes: langchain-mcp-adapters builds each tool's coroutine as a closure over the original `mcp.types.Tool`, so the server is still called with the original name.
- Two tools that would collide on one safe name: the tool that needed no rename wins, otherwise the first one listed; the other is skipped with a warning. A renamed name longer than OpenAI's 64-character cap is kept but warned about.
- Renames are logged at INFO, since the adapter is rebuilt on every agent run.

This applies to any MCP server, not only neuro-san, because the providers fail regardless of where the tool came from. Once the server-side change (#1374) is deployed, neuro-san servers already advertise safe names and nothing is renamed for them.

Docs: `docs/agent_hocon_reference.md` (MCP tools section).

Tests: `tests/neuro_san/internals/run_context/langchain/mcp/test_langchain_mcp_adapter.py` extended.


Noravee:

Noravee:
This PR cannot be reopened for the same reason as #1374 (head branch force-pushed, base branch deleted). The adapter change will be opened as a new PR against main when its turn comes. The branch `mcp-adapter-tool-name` stays current.

PR #1376, opened 2026-09-21
Let McpServiceAgentSession find its network under any advertised MCP tool name (2 of 4)

Noravee:
Part of #1371. Second of four PRs, opened one at a time. The foundation (#1372) has merged; this one must merge before the server-side PR. Supersedes #1373, which could not be reopened after its base branch was deleted when #1372 merged.

neuro-san's own MCP client session (`agent_cli --mcp`, `AgentSessionFactory.create_session("mcp", ...)`, `SimpleOneShot`) is created with the network name, for example `--agent deep/math_guy`. Once a server advertises `deep__math_guy` instead, or a custom manifest `mcp_name` such as `calculator`, the current client no longer finds its tool in `tools/list`, `function()` returns None, and `agent_cli` turns that into an error.

Changes in `McpServiceAgentSession`:
- `function()` picks the `tools/list` entry that stands for the network. A server that renames a tool keeps the network name in the entry's MCP `title`, so when a title is present it decides; an untitled entry matches only if its `name` is the network name, as servers from before the rename advertise it. The client does not guess from the spelling: an unrelated network can legitimately be named `deep__math_guy`, and an entry named `math_guy` but titled `x/y` belongs to `x/y`.
- `function()` records the advertised `name`, and `streaming_chat()` sends `tools/call` with it. A caller that skips `function()` sends the network name, which every neuro-san server accepts: older servers key on it directly and the coming server-side change passes an unadvertised name through to the same network-keyed lookup.

About `mcp_name`: the client does not depend on the manifest key itself, only on the `title` contract above. The key is described in `AgentNetwork.set_as_mcp_tool()` from #1372; the server-side PR (3 of 4) wires it into the manifest and documents it in `docs/manifest_hocon_reference.md`. Servers on main today advertise every network under its own name and send no `title`, so against them this client behaves exactly as before.

So the new client works against servers from before this change, servers that advertise the provider-safe name, and servers that advertise a custom `mcp_name`. For a top-level network nothing changes.

Why this must merge before the server-side PR: a client without this change only matches the exact network name in `tools/list`, so servers must not start advertising other names until clients can follow them.

Tests: `tests/neuro_san/session/test_mcp_service_agent_session.py` (new), one fake server per naming scheme plus the lookalike network, a title overriding a matching name, an entry without description, and a session without an agent name.


Voyagerroc-Lab (commented):
The PR’s current revision is not merge-ready: all three required unit-test matrix jobs are red. The failure log shows `tests/neuro_san/internals/run_context/langchain/llms/test_openai_llm_policy.py::TestOpenAILlmPolicy::test_reasoning_dict_auto_routes_to_responses_api_without_n` failing because `use_responses_api` is `True` instead of `None`; that test is outside this diff, so please either rebase onto a revision where the baseline is green or document/fix the interaction before merge. The new MCP-name tests cannot be treated as sufficient while the required suite is failing.

neuro_san/session/mcp_service_agent_session.py
```diff
@@ -176,7 +187,12 @@ def function(self, request_dict: Dict[str, Any]) -> Dict[str, Any]:
         for tool in tools_list:
             use_tool: Dict[str, Any] = tool
             name = use_tool.get("name", None)
-            if name == self.agent_name:
+            # Accept both spellings: servers before the rename advertise the network
+            # name itself, servers after it advertise the provider-safe name.
+            if name in (self.agent_name, self.mcp_tool_name):
```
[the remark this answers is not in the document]
Noravee: Done in a1b69c04. function() now matches the entry by title when present, falling back to name only for untitled entries, and records the advertised name for tools/call.

While doing this the derived-spelling guess went away entirely: an untitled entry named deep__math_guy can legitimately be an unrelated network of that name (that is exactly what a server lists after a name collision withdraws the nested one), so guessing could route the chat to the wrong network. Before function() has run, tools/call now sends the network name, which older servers key on and newer ones pass through to the same lookup. Tests added for the custom-name, lookalike, and title-overrides-name cases.

Noravee:
**TL;DR**

`McpServiceAgentSession` is neuro-san's own MCP client (`agent_cli --mcp`, `SimpleOneShot`). It is created with a network name such as `deep/math_guy`, but a future server may list that network under another name (`deep__math_guy`, or an alias like `calculator`) and put the network name in the entry's `title`.

- `function()` now finds the network's entry in `tools/list` by `title`, or by `name` when there is no title, and remembers the advertised `name`.
- `streaming_chat()` calls the tool by that remembered name. If `function()` was never called, it sends the network name, which every neuro-san server accepts.
- No guessing from the spelling: an untitled `deep__math_guy` entry is a network of that name, not a renamed `deep/math_guy`.
- Against today's servers, which send no `title`, behaviour is unchanged.

The client does not read the `mcp_name` manifest key; that is wired in by the server PR (3 of 4).

neuro_san/session/mcp_service_agent_session.py
```diff
@@ -81,6 +81,18 @@ def __init__(self, host: str = None,
...
+        # The name the server advertises for this network, learned by function()
+        # from tools/list. A server may advertise a network under a name other
+        # than the network name (see AgentNetwork.set_as_mcp_tool(): nested
+        # "deep/math_guy" becomes the provider-safe "deep__math_guy", or an
+        # alias such as "calculator"), while users keep passing the network name
+        # (e.g. --agent deep/math_guy). Until function() has looked, tools/call
+        # uses the network name, which every neuro-san server accepts: older
+        # servers key on it directly and newer ones pass an unadvertised name
+        # through to the same network-keyed lookup.
+        self.advertised_tool_name: str = None
```
Noravee, line 94: Learned from `tools/list` in `function()`. Stays None for callers that never call `function()`, such as `SimpleOneShot`.

neuro_san/session/mcp_service_agent_session.py
```diff
@@ -171,18 +183,57 @@ def function(self, request_dict: Dict[str, Any]) -> Dict[str, Any]:
...
         result_dict: Dict[str, Any] = response_dict.get("result", empty_dict)
         tools_list: List[Dict[str, Any]] = result_dict.get("tools", empty_list)
-        name: str = None
-        tool_description: str = None
+        use_tool: Dict[str, Any] = self.find_tool_for_network(tools_list)
+        if use_tool is None:
+            # Nothing stands for the network any more, so forget whatever an
+            # earlier listing advertised; tools/call then falls back to the
+            # network name rather than a stale one.
+            self.advertised_tool_name = None
```
Noravee, line 191: If a later listing no longer has the network, forget the old name so `tools/call` falls back to the network name.

neuro_san/session/mcp_service_agent_session.py
```diff
@@ -171,18 +183,57 @@ def function(self, request_dict: Dict[str, Any]) -> Dict[str, Any]:
...
+            return None
+
+        # Remember the name this server uses so tools/call sends a name the
+        # server can resolve. When the network is advertised under an alias,
+        # this is a name we could not have derived from the network name.
+        self.advertised_tool_name = use_tool.get("name", None)
```
Noravee, line 197: Recorded before the description check, so `tools/call` can use the advertised name even when the entry has no description.

neuro_san/session/mcp_service_agent_session.py
```diff
@@ -171,18 +183,57 @@ def function(self, request_dict: Dict[str, Any]) -> Dict[str, Any]:
...
-                    return {
-                        "function": {"description": tool_description}
-                    }
-
+            name: str = use_tool.get("name", None)
+            if name is None:
+                # Not a callable tool entry: tools/call needs a name.
+                continue
+            # An empty title carries no information, so it falls back to the name.
+            network_name: str = use_tool.get("title", None) or name
```
Noravee, line 234: The one rule: a `title` says which network the entry stands for; without one, the `name` must be the network name. An empty title counts as absent.

neuro_san/session/mcp_service_agent_session.py
```diff
@@ -208,13 +259,20 @@ def streaming_chat(self, request_dict: Dict[str, Any]) -> Generator[Dict[str, An
             Note that responses to the chat input might be numerous and will come as they
             are produced until the system decides there are no more messages to be sent.
         """
+        # Call the tool by the name function() saw the server advertise. A caller
+        # that skips function() (e.g. SimpleOneShot) has not learned it, so it
+        # sends the network name instead, which every neuro-san server accepts.
+        tool_name: str = self.advertised_tool_name
+        if tool_name is None:
+            tool_name = self.agent_name
```
Noravee, line 267: Advertised name when known, otherwise the network name. Main's server keys `tools/call` by network name, and the coming server passes an unadvertised name through to the same lookup, so the network name works against both.

ofrancon (approved):
lgtm

d1donlydfink (approved):
LGTM

PR #1377, opened 2026-09-21
Fix the OpenAI policy test for langchain-openai 1.6.3 and pin langchain-openai and openai for now

Noravee:
## What

1. **Test fix.** In `tests/neuro_san/internals/run_context/langchain/llms/test_openai_llm_policy.py`, `test_reasoning_dict_auto_routes_to_responses_api_without_n` no longer asserts that `llm.use_responses_api` is still `None` after building a `ChatOpenAI` with a `reasoning` dict; it asserts that the policy did not pin Chat Completions (`is not False`). The outcome assertions stay: langchain routes the request to the Responses API and the payload carries no `n`.
2. **Two temporary dependency pins in `requirements.txt`**, each with its reason as a comment next to the line:
    - `langchain-openai>=1.0.0,<1.6.3` (was `<2.0`).
    - `openai>=2.45.0,<3` (new explicit pin; it was only a transitive dependency of langchain-openai).

## Why

**The test.** langchain-openai 1.6.3, released 2026-09-21, adds a pydantic model validator `_infer_use_responses_api` (`langchain_openai/chat_models/base.py:1329-1333` at the `langchain-openai==1.6.3` tag, absent in 1.6.2). At construction, when `use_responses_api` is `None` and instance settings such as `reasoning`, `include`, `truncation` or a Responses-only model already imply Responses routing, it sets `use_responses_api = True`. CI installs the newest release inside our cap, so this test failed on Python 3.12, 3.13 and 3.14 for every PR. It surfaced first on #1376, which does not touch this code. The behavior under test is unchanged; only the attribute assertion was version-dependent, and the fix keeps the test correct on both sides of the pin.

**Why cap langchain-openai below 1.6.3 anyway.** The same release also changes endpoint inference under our current `use_responses_api: null` class default: `gpt-6*` models with tools and `gpt-5.6-sol` now auto-route to the Responses API (base.py:695-707 and :1942-1945 at the tag). Both are improvements, but a neuro-san release cut today would behave differently from one cut last week with no change on our side. #1332 makes the Responses API the default deliberately, documents it, and lifts this ceiling.

**Why pin openai to 2.x.** langchain-openai 1.6.x allows `openai<4`, so fresh installs (CI, Docker images) now get openai 3.16.2. openai 3.0 (2026-08-12) switched its HTTP stack to httpx2 and no longer installs httpx. `OpenAILlmPolicy.create_http_client()` and `AzureLlmPolicy` build an `httpx.AsyncClient` (proxy, timeout, connection limits) and pass it as `http_client`; openai 3.x accepts that only through a runtime-only escape hatch that its httpx2 migration guide calls temporary. An offline round trip through `OpenAILlmPolicy` with openai 3.16.2 against the bundled mock LLM server worked, so this is not a known break, but the live path has not been exercised on openai 3.x. Staying on 2.x keeps the HTTP stack we have been running until those clients are migrated to httpx2; that migration lifts the ceiling.

## Verification

- `pytest tests/neuro_san/internals/run_context/langchain/llms/test_openai_llm_policy.py` passes against the environment's langchain-openai 1.6.0 and against 1.6.3 installed with `pip install --no-deps --target /tmp/lco163 langchain-openai==1.6.3` on `PYTHONPATH` (8 passed each). flake8 and pylint are clean.
- `pip install --dry-run --ignore-installed -r requirements.txt -r requirements-build.txt` resolves; it picks langchain-openai 1.6.2, langchain-core 1.6.4, openai 2.54.0 and httpx 0.28.1. langchain-openai 1.6.2 and current master both declare `openai>=2.45.0,<4.0.0`, so the openai pin does not conflict with any released or upcoming langchain-openai.

## Related

#1376 (where the failure surfaced). #1332 rewrites the test module, already carries the same one-line assertion change (e91c791d), and will raise the langchain-openai range to `>=1.4.0,<2.0`, lifting the first ceiling. The openai ceiling is lifted by the httpx2 migration of the OpenAI and Azure policy clients.


tests/neuro_san/internals/run_context/langchain/llms/test_openai_llm_policy.py
```diff
@@ -145,7 +145,11 @@ def test_reasoning_dict_auto_routes_to_responses_api_without_n(self, policies: L
         llm: ChatOpenAI = self._build_llm(policies, {"reasoning": {"effort": "low"}})
         payload: Dict[str, Any] = self._request_payload(llm)
 
-        assert llm.use_responses_api is None
+        # The policy forwards use_responses_api from llm_config unchanged, so it must not have pinned
+        # Chat Completions. Do not assert None here: langchain-openai 1.6.3 added a model validator
+        # that materializes the inferred endpoint on the instance (True once "reasoning" is set),
+        # while earlier releases leave the attribute None.
+        assert llm.use_responses_api is not False
```
Noravee, line 152: Modifying the test to no check that `use_responses_api` is `None` since it will be flipped automatically.

requirements.txt
```diff
@@ -48,7 +48,21 @@ pydantic>=2.9.2,<3.0
 # and all the ensuing transient dependencies for the world.
 # For a list of other LLMs we support out of the box, see requirements-build.txt
 # NOTE: This version should be aligned with the langchain and langchain-core versions above.
-langchain-openai>=1.0.0,<2.0
+# Ceiling <1.6.3 (temporary): langchain-openai 1.6.3 changes endpoint inference under our
+# use_responses_api: null default. gpt-6* models with tools and gpt-5.6-sol now auto-route to
+# the Responses API, and the inferred endpoint is written onto the ChatOpenAI instance at
+# construction. Both are fine, but they must land as a deliberate neuro-san change: #1332 makes
+# the Responses API the default and lifts this ceiling.
+langchain-openai>=1.0.0,<1.6.3
```
Noravee, line 56: New version of langchain-openai `1.6.3` should not have any impacts on general usage of neuro-san. It flips the flag `use_responses_api` to `True` automatically if a set of args is detected in the constructor.

I will do more tests. If everything is fine, the pinned will be remove in #1332.

requirements.txt
```diff
@@ -48,7 +48,21 @@ pydantic>=2.9.2,<3.0
...
+
+# Pinned explicitly (it is otherwise only a transitive dependency of langchain-openai).
+# Floor 2.45.0 matches langchain-openai's own requirement. Ceiling <3 (temporary): openai 3.0
+# (2026-08-12) switched its HTTP stack to httpx2 and no longer installs httpx. OpenAILlmPolicy
+# and AzureLlmPolicy build an httpx.AsyncClient (proxy, timeout, connection limits) and pass it
+# as http_client, which openai 3.x accepts only through a runtime-only escape hatch its migration
+# guide calls temporary. Stay on 2.x until those clients are migrated to httpx2; that migration
+# lifts this ceiling.
+openai>=2.45.0,<3
```
Noravee, line 65: Pin `openai` since `3.0` uses `httpx2` and we don't know when they will remove the http client.

ofrancon (approved):
lgtm

